PR #1, opened 2017-08-02
parse and report async gaps

yjbanov:
Related: https://github.com/flutter/flutter/issues/11479

### DIFF
--- a/lib/src/stack_trace.dart
+++ b/lib/src/stack_trace.dart
@@ -22,9 +22,8 @@ List<Map<String, dynamic>> encodeStackTrace(dynamic stackTrace) {
   final List<Map<String, dynamic>> frames = <Map<String, dynamic>>[];
   for (int t = 0; t < chain.traces.length; t += 1) {
     frames.addAll(chain.traces[t].frames.map(encodeStackTraceFrame));
-    if (t < chain.traces.length - 1) {
+    if (t < chain.traces.length - 1)
       frames.add(asynchronousGapFrameJson);
-    }
   }
   return frames;
 }
--- a/lib/src/stack_trace.dart
+++ b/lib/src/stack_trace.dart
@@ -4,7 +4,32 @@
 
 import 'package:stack_trace/stack_trace.dart';
 
-Map<String, dynamic> stackTraceFrameToJsonFrame(Frame frame) {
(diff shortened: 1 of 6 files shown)


eseidelGoogle:
Isn't crashy able to tell us if this fixes things? :)

yjbanov:
This PR does not fix missing async stack traces. It only fixes missing async gaps, should they appear in the stack. I just confirmed using crashy that release builds are indeed missing async stack traces. Will comment in the original issue.

lib/src/stack_trace.dart
```diff
@@ -4,7 +4,32 @@
...
+  final Chain chain = stackTrace is StackTrace
+      ? new Chain.forTrace(stackTrace)
+      : new Chain.parse(stackTrace);
+
+  final List<Map<String, dynamic>> frames = <Map<String, dynamic>>[];
+  for (int t = 0; t < chain.traces.length; t += 1) {
+    frames.addAll(chain.traces[t].frames.map(encodeStackTraceFrame));
+    if (t < chain.traces.length - 1) {
+      frames.add(asynchronousGapFrameJson);
+    }
```
cbracken: Drop the curly brackets for one-line `if`s.

lib/src/stack_trace.dart
```diff
@@ -4,7 +4,32 @@
...
+///
+/// [stackTrace] must be [String] or [StackTrace].
+List<Map<String, dynamic>> encodeStackTrace(dynamic stackTrace) {
+  assert(stackTrace is String || stackTrace is StackTrace);
+  final Chain chain = stackTrace is StackTrace
+      ? new Chain.forTrace(stackTrace)
+      : new Chain.parse(stackTrace);
+
+  final List<Map<String, dynamic>> frames = <Map<String, dynamic>>[];
+  for (int t = 0; t < chain.traces.length; t += 1) {
```
cbracken, line 23: Why not `t++`?

cbracken (approved):

cbracken:
lgtm

lib/src/stack_trace.dart
yjbanov: That's Flutter style.

lib/src/stack_trace.dart
yjbanov: Done.

PR #11, opened 2018-03-20
release 1.0.0 - the last Dart 1 compatible major version

yjbanov:
I propose that we submit this first. Then https://github.com/flutter/sentry/pull/9 will go next.

/cc @pulyaevskiy @xster


xster:
LG

PR #20, opened 2018-07-05
Support for new DSN format (without secretKey) and remove the quiver dependency.

slightfoot:
I've updated the library to make the secretKey field optional so that it supports the new DSN format from Sentry.

I also noticed that Quiver is only used for Clock to hold a ticking DateTime for reporting. Looking at the use of it I see it can be replaced with a simple function typedef. This I feel is a better way to make the library for flexible and not have the extra dependency on the Quiver library.

### DIFF
--- a/lib/sentry.dart
+++ b/lib/sentry.dart
@@ -159,10 +159,9 @@ class SentryClient {
 
   /// Reports an [event] to Sentry.io.
   Future<SentryResponse> capture({@required Event event}) async {
-
     final DateTime now = _clock();
     String authHeader = 'Sentry sentry_version=6, sentry_client=$sentryClient, '
-      'sentry_timestamp=${now.millisecondsSinceEpoch}, sentry_key=$publicKey';
+        'sentry_timestamp=${now.millisecondsSinceEpoch}, sentry_key=$publicKey';
     if (secretKey != null) {
       authHeader += ', sentry_secret=$secretKey';
     }
--- a/lib/sentry.dart
+++ b/lib/sentry.dart
@@ -68,9 +68,8 @@ class SentryClient {
     uuidGenerator ??= _generateUuidV4WithoutDashes;
     compressPayload ??= true;
 
-    final ClockProvider clockProvider = clock is ClockProvider
-      ? clock
-      : clock.get;
+    final ClockProvider clockProvider =
+        clock is ClockProvider ? clock : clock.get;
 
     final Uri uri = Uri.parse(dsn);
     final List<String> userInfo = uri.userInfo.split(':');
--- a/lib/sentry.dart
+++ b/lib/sentry.dart
@@ -19,7 +19,7 @@ import 'src/version.dart';
 
 export 'src/version.dart';
 
-/// Used to provide time-stamp for logging.
+/// Used to provide timestamp for logging.
 typedef ClockProvider = DateTime Function();
 
 /// Logs crash reports and events to the Sentry.io service.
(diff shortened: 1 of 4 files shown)


slightfoot:
@yjbanov ping!

slightfoot:
@tvolkert any chance of getting this merged in. I've a release to do and would like project referencing the official repo rather than the fork.

yjbanov:
Hi! Sorry, was out on vacation. Looking at this now.

lib/sentry.dart
```diff
@@ -20,6 +19,9 @@ import 'src/version.dart';
 
 export 'src/version.dart';
 
+/// Used to provide time-stamp for logging.
```
yjbanov: nit: [according to Wikipedia](https://en.wikipedia.org/wiki/Timestamp) "timestamp" is the correct spelling :nerd_face:

lib/sentry.dart
```diff
@@ -56,31 +58,27 @@ class SentryClient {
     Event environmentAttributes,
     bool compressPayload,
     Client httpClient,
-    Clock clock,
+    ClockProvider clock,
```
yjbanov: To give people some time to migrate let's make `clock` a `dynamic` parameter. Then in the body of the factory we can do this:

```dart
final ClockProvider clockProvider = clock is ClockProvider
  ? clock
  : clock.get;

// pass clockProvider to SentryClient._
```

This still lets us remove our dependency on `quiver` but won't break existing code. Let's thoroughly document it in the dartdocs though.

tvolkert:
> Sorry, was out on vacation

Me too - @slightfoot you caught the team in vacation season 😄 

(leaving the review to @yjbanov)

lib/sentry.dart
```diff
@@ -161,21 +159,23 @@ class SentryClient {
 
   /// Reports an [event] to Sentry.io.
   Future<SentryResponse> capture({@required Event event}) async {
-    final DateTime now = _clock.now();
+    final DateTime now = _clock();
+    String authHeader = 'Sentry sentry_version=6, sentry_client=$sentryClient, '
+        'sentry_timestamp=${now.millisecondsSinceEpoch}, sentry_key=$publicKey';
+    if (secretKey != null) {
+      authHeader += ', sentry_secret=$secretKey';
```
yjbanov, line 171: Let's add a unit-test for this case.

yjbanov:
Thanks! The fewer dependencies the better!

slightfoot:
Thanks for the review @yjbanov I'll look into the details and answer when I get time to make the suggested changes.

yjbanov:
Adding @tvolkert to the reviewer list as I'll be on vacation again.

lib/sentry.dart
slightfoot: Done.

lib/sentry.dart
slightfoot: Done.

lib/sentry.dart
slightfoot: Added unit test for parsing and http header.

lib/sentry.dart
```diff
@@ -46,7 +48,9 @@ class SentryClient {
   /// make HTTP calls to Sentry.io. This is useful in tests.
   ///
   /// If [clock] is provided, it is used to get time instead of the system
-  /// clock. This is useful in tests.
+  /// clock. This is useful in tests. Should be an implementation of ClockProvider.
```
yjbanov, line 51: References to in-scope identifiers should be wrapped in square brackets so that dartdoc can create links to them, in this case `[ClockProvider]`.

yjbanov (approved):
LGTM with one small nit. Thanks for the contribution!

PR #23, opened 2018-10-08
mergeAttributes: deepcopy RHS Map if LHS isn't Map

dotdoom:


dotdoom:
The problem that this PR solves is that a Map present in `environmentAttributes` gets exposed when [initially merged into `data`](https://github.com/flutter/sentry/blob/5167a5c/lib/sentry.dart#L188-L195). It is then mutated by the following `mergeAttributes` of `event` into `data`.

---

Looks like Travis build fails because it can't resolve some dependencies with Dart 2. I tested this PR by running `echo test/*.dart | xargs -n1 dart`.

yjbanov:
I like this change! Thank you for the fix. I'll have a look at Travis asap (currently AFK).

dotdoom:
Sent https://github.com/flutter/sentry/pull/24 for that, and rebased this PR, too.

yjbanov (approved):

yjbanov:
LGTM

yjbanov:
This is now published in version 2.1.1: https://pub.dartlang.org/packages/sentry

PR #28, opened 2018-10-26
Move sentry.io package to packages

kf6gpe:
Moves sentry.io package to packages, since it's a first-party plugin, so we can continue to decrease our repository footprint. 

Thoughts especially welcome from @yjbanov @filiph, of course.


yjbanov (approved):

yjbanov:
lgtm

PR #30, opened 2018-11-12
Add a stackFrameFilter argument to SentryClient's capture

tom95:
Attempt no 2, description copied:

This change allows filtering sensitive frames on the client or applying custom truncation logic.

I created this in reaction to this discussion: https://forum.sentry.io/t/issue-in-flutter-project-with-stackframe-display-limit-set-to-250/5014/2

Instead of adding complex truncation options that will likely depend on the frameworks you use anyways, I opted to provide a more general callback that could also be used for things like filtering specific frames. In the case of above discussion, we'd simply cut the list passed to us to only send the topmost 250 items.

### DIFF
--- a/lib/sentry.dart
+++ b/lib/sentry.dart
@@ -164,7 +164,10 @@ class SentryClient {
       '${dsnUri.scheme}://${dsnUri.host}/api/$projectId/store/';
 
   /// Reports an [event] to Sentry.io.
-  Future<SentryResponse> capture({@required Event event}) async {
+  Future<SentryResponse> capture({
+    @required Event event,
+    StackFrameFilter stackFrameFilter
+  }) async {
     final DateTime now = _clock();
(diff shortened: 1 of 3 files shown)


tom95:
Hm. Any guidance on how to tackle the CLA bot's disapproval? It does not appear to be possible to inspect CLAs for companies via the link and the troubleshooting section on that topic contains mostly broken links :(

yjbanov (commented):
The CLA bot seems happy, but `dartfmt` failed. Could you please format edited files using `dartfmt`? LGTM with a small comment.

tom95:
Thank you for the review! I added the comment and let dartfmt do its thing.

(I chose to keep this PR over the other because this one contained actual review comments :) )

yjbanov:
Thank you for the contribution and for your patience! :+1:

PR #42, opened 2019-09-02
support dsn that use non-regular port or contains prefix path 

yrom:
This PR will modify the `postUri` value in `SentryClient`.

Added two test cases :
```
const String _testDsnWithPath = 'https://public:secret@sentry.example.com/path/1';
const String _testDsnWithPort = 'https://public:secret@sentry.example.com:8888/1';
```

### DIFF
--- a/lib/sentry.dart
+++ b/lib/sentry.dart
@@ -169,7 +169,7 @@ class SentryClient {
     var pathLength = dsnUri.pathSegments.length;
     var apiPath;
     if (pathLength > 1) {
-      // some paths would present before the projectID in the dsnUri 
+      // some paths would present before the projectID in the dsnUri
       apiPath =
           (dsnUri.pathSegments.sublist(0, pathLength - 1) + ['api']).join('/');
     } else {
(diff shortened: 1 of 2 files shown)


lib/sentry.dart
```diff
@@ -160,8 +160,23 @@ class SentryClient {
...
-  String get postUri =>
-      '${dsnUri.scheme}://${dsnUri.host}/api/$projectId/store/';
+  String get postUri {
+    var port = dsnUri.hasPort &&
+            ((dsnUri.scheme == 'http' && dsnUri.port != 80) ||
+                (dsnUri.scheme == 'https' && dsnUri.port != 443))
+        ? ':${dsnUri.port}'
+        : '';
+    var pathLength = dsnUri.pathSegments.length;
+    var apiPath;
```
yjbanov: To make sure the language doesn't infer `dynamic` by mistake, let's put explicit types on the left-hand side (i.e. `String`, `int`, `String`).

yrom: Updated, thx.

yjbanov (approved):
Great! LGTM.

PR #40, opened 2019-06-15
Add support for Contexts interface

wilkomanger:

PR #43, opened 2019-09-02
Add 'transaction' and 'breadcrumbs' in event

yrom:
ref docs:
transaction: https://docs.sentry.io/development/sdk-dev/event-payloads/#optional-attributes
contexts: https://docs.sentry.io/development/sdk-dev/event-payloads/contexts/
breadcrumbs: https://docs.sentry.io/development/sdk-dev/event-payloads/breadcrumbs/

There is a related pr #40 when i create this pr. In my opinion, the `Event` class should know nothing about what the `contexts` have, so i did not create the wrapper class for the `os`,`device`,etc which it might contains.

### DIFF
--- a/lib/sentry.dart
+++ b/lib/sentry.dart
@@ -347,7 +347,7 @@ class Event {
   final dynamic stackTrace;
 
   /// The name of the transaction which generated this event,
-  /// for e.g., might be the route name: `"/users/<username>/"`.
+  /// for example, the route name: `"/users/<username>/"`.
   final String transaction;
 
   /// How important this event is.
@@ -367,7 +367,8 @@ class Event {
 
   /// List of breadcrumbs for this event.
   ///
-  /// see docs https://docs.sentry.io/enriching-error-data/breadcrumbs/?platform=javascript
+  /// See also:
+  /// * https://docs.sentry.io/enriching-error-data/breadcrumbs/?platform=javascript
   final List<Breadcrumb> breadcrumbs;
 
   /// Information about the current user.
@@ -536,20 +537,25 @@ class Breadcrumb {
   final Map<String, String> data;
   final SeverityLevel level;
 
-  /// default, http, or navigation.
-  /// see docs https://docs.sentry.io/development/sdk-dev/event-payloads/breadcrumbs/#breadcrumb-types
+  /// Describes what type of breadcrumb this is.
+  ///
+  /// Possible values: "default", "http", "navigation".
+  ///
+  /// See also:
+  ///
+  /// * https://docs.sentry.io/development/sdk-dev/event-payloads/breadcrumbs/#breadcrumb-types
   final String type;
 
-  /// Recording time
+  /// The time the breadcrumb was recorded.
   ///
-  /// note: serialized in  seconds
+  /// The value is submitted to Sentry with second precision.
   final DateTime timestamp;
   const Breadcrumb(this.message, this.timestamp,
       {this.category, this.data, this.level = SeverityLevel.info, this.type});
 
   Map<String, dynamic> toJson() {
     var json = <String, dynamic>{
-      'timestamp': timestamp.millisecondsSinceEpoch ~/ 1000,
+      'timestamp': formatDateAsIso8601WithSecondPrecision(timestamp),
     };
     if (message != null) {
       json['message'] = message;
--- a/lib/sentry.dart
+++ b/lib/sentry.dart
@@ -306,6 +306,7 @@ class Event {
     this.release,
     this.environment,
     this.message,
+    this.transaction,
     this.exception,
     this.stackTrace,
     this.level,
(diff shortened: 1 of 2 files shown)


SammyO:
+1 to get this merged

yjbanov (changes requested):
Transaction and breadcrumbs LGTM. Let's leave contexts out of this PR. https://github.com/flutter/sentry/pull/40 is adding a fully schemed version of it.

yrom:
@yjbanov The 'contexts' was discarded.

lib/sentry.dart
```diff
@@ -494,3 +513,59 @@ class User {
...
+/// See also:
+/// * https://docs.sentry.io/development/sdk-dev/event-payloads/breadcrumbs/
+class Breadcrumb {
+  final String message;
+  final String category;
+  final Map<String, String> data;
+  final SeverityLevel level;
+
+  /// default, http, or navigation.
+  /// see docs https://docs.sentry.io/development/sdk-dev/event-payloads/breadcrumbs/#breadcrumb-types
```
yjbanov: Let's reformat this a bit:

```
/// Describes what type of breadcrumb this is.
///
/// Possible values: "default", "http", "navigation".
///
/// See also:
///
/// * https://docs.sentry.io/development/sdk-dev/event-payloads/breadcrumbs/#breadcrumb-types
```

lib/sentry.dart
```diff
@@ -494,3 +513,59 @@ class User {
...
+  /// Recording time
+  ///
+  /// note: serialized in  seconds
+  final DateTime timestamp;
+  const Breadcrumb(this.message, this.timestamp,
+      {this.category, this.data, this.level = SeverityLevel.info, this.type});
+
+  Map<String, dynamic> toJson() {
+    var json = <String, dynamic>{
+      'timestamp': timestamp.millisecondsSinceEpoch ~/ 1000,
```
yjbanov: Let's reuse https://github.com/flutter/sentry/blob/e14f1497d623ae29ca35d4cff5858fb540b83352/lib/src/utils.dart#L29

lib/sentry.dart
```diff
@@ -494,3 +513,59 @@ class User {
...
+  final Map<String, String> data;
+  final SeverityLevel level;
+
+  /// default, http, or navigation.
+  /// see docs https://docs.sentry.io/development/sdk-dev/event-payloads/breadcrumbs/#breadcrumb-types
+  final String type;
+
+  /// Recording time
+  ///
+  /// note: serialized in  seconds
```
yjbanov: Some clean-up:

```
/// The time the breadcrumb was recorded.
///
/// The value is submitted to Sentry with second precision.
```

yjbanov:
Thanks, just a couple more things I noticed.

lib/sentry.dart
```diff
@@ -344,6 +346,10 @@ class Event {
   /// Can be `null`, a [String], or a [StackTrace].
   final dynamic stackTrace;
 
+  /// The name of the transaction which generated this event,
+  /// for e.g., might be the route name: `"/users/<username>/"`.
```
yjbanov: nit: either "for example", or "e.g.", but not both :)
nit: no need for "might be", as we already say "for example"

lib/sentry.dart
```diff
@@ -359,6 +365,11 @@ class Event {
   /// they must be JSON-serializable.
   final Map<String, dynamic> extra;
 
+  /// List of breadcrumbs for this event.
+  ///
+  /// see docs https://docs.sentry.io/enriching-error-data/breadcrumbs/?platform=javascript
```
yjbanov: For consistency, let's format as:

```
/// See also:
/// * https://docs.sentry.io/enriching-error-data/breadcrumbs/?platform=javascript
```

yrom:
@yjbanov Thanks for reviewing. Request changes applied.

yjbanov:
@yrom Curious, why did you decide to rollback the timestamp in https://github.com/flutter/sentry/pull/43/commits/cd6939226ef0da823e49aee9a55f58c1353e2eb9?

yrom:
@yjbanov Looked up the `sentry-java` SDK.
See  https://github.com/getsentry/sentry-java/blob/master/sentry/src/main/java/io/sentry/marshaller/json/JsonMarshaller.java#L299. 
They were using timestamp. 😅

jaumard:
Would love to have this and "Context" merged, as currently reporting is quite light :)

yjbanov:
> @yjbanov Looked up the `sentry-java` SDK.
> See https://github.com/getsentry/sentry-java/blob/master/sentry/src/main/java/io/sentry/marshaller/json/JsonMarshaller.java#L299.
> They were using timestamp.

Would you mind trying `formatDateAsIso8601WithSecondPrecision` instead? This way we'd consistently use the same date-time formatting throughout the client.

yrom:
@yjbanov Sorrry for my late reply. I have updated the commits, again.

yjbanov:
LGTM! Thank you for contributing this feature!

PR #53, opened 2020-04-12
fix: screenDensity is double

### LINKED ISSUE #53521: [sentry] screenDensity has to be a floating point number
[sentry](https://pub.dev/packages/sentry) v3.0.1

Right now when I try to use `Device` context `screenDensity` parameter, I get the following error on sentry.io:

```
There was 1 error encountered while processing this event
contexts.device.screen_density: Discarded invalid value
```

Reason | expected a floating point number
-- | --
Value | 2.625

bruno-garcia:
Resolves flutter/flutter#53521


bruno-garcia:
image

Why isn't the status check updating?

lib/src/base.dart
```diff
@@ -659,7 +659,7 @@ class Device {
   final String screenResolution;
 
   /// A floating point denoting the screen density.
-  final String screenDensity;
+  final double screenDensity;
```
yjbanov, line 662: This is a breaking change. Let's do a major version bump.

bruno-garcia: That's true. Should I bump it as part of this PR?

yjbanov: Yep.

bruno-garcia: Bumped the major and added changelog.

yjbanov (approved):

PR #56, opened 2020-06-04
Expose the "extra" named parameter

ecerroni:
Allow sending to sentry custom extra data that can attach to the event

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,7 @@ main() async {
     await sentry.captureException(
       exception: error,
       stackTrace: stackTrace,
+      extra: additionalCustomData, // Map<String, dynamic> | must be JSON-serializable
     );
   }
 }
@@ -63,6 +64,7 @@ main() async {
         sentry.captureException(
           exception: error,
           stackTrace: stackTrace,
+          extra: additionalCustomData,
         );
         print('Error sent to sentry.io: $error');
       } catch (e) {
@@ -81,6 +83,7 @@ main() async {
       sentry.captureException(
         exception: details.exception,
         stackTrace: details.stack,
+        extra: additionalCustomData,
       );
     } catch (e) {
       print('Sending report to sentry.io failed: $e');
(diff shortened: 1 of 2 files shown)


ecerroni:
@googlebot I signed it!

ecerroni:
CI Build fails. All tests passing, but still, an error from the `"./tool/presubmit.sh"` with no additional details.

image

Any idea?

lib/src/base.dart
```diff
@@ -201,10 +201,12 @@ abstract class SentryClient {
   Future<SentryResponse> captureException({
     @required dynamic exception,
     dynamic stackTrace,
+    dynamic extra,
   }) {
     final event = Event(
       exception: exception,
       stackTrace: stackTrace,
+      extra: extra
```
yjbanov, line 209: nit: add a trailing comma (this will tell `dartfmt` to put each argument on its own line)

yjbanov (commented):
This PR needs unit-tests.

README.md
```diff
@@ -43,6 +43,7 @@ main() async {
     await sentry.captureException(
       exception: error,
       stackTrace: stackTrace,
+      extra: additionalCustomData, // Map<String, dynamic> | must be JSON-serializable
```
bruno-garcia, line 46: I'd try to avoid adding extra as an overload to `captureException` so the API doesn't differ from the other Sentry SDKs ([documented here](https://develop.sentry.dev/sdk/unified-api/)).

Going further we'd like to add [`Scope`s](https://develop.sentry.dev/sdk/unified-api/#scope) and a way to add data to `captureException` would be either by providing a instance of `Scope` or accessing the clone of the current scope via a callback.

We'll be adding such APIs and more in the upcoming months.

bruno-garcia (changes requested):
Thank you for adding the ability to pass `extra` through the `captureException` method but I propose we don't merge this in favor of aligning the API with the other Sentry SDKs.

PR #57, opened 2020-06-11
Use integer type for screenDpi

Mooibroek:
The sentry API expect the value of screenDpi to be an unsigned integer and not a string.

afbeelding

### DIFF
--- a/lib/src/base.dart
+++ b/lib/src/base.dart
@@ -662,7 +662,7 @@ class Device {
   final double screenDensity;
 
   /// A decimal value reflecting the DPI (dots-per-inch) density.
-  final String screenDpi;
+  final int screenDpi;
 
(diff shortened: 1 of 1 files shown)


Mooibroek:
@googlebot I signed it!

bruno-garcia (changes requested):
CI is failing: `  error • The argument type 'String' can't be assigned to the parameter type 'int'. • test/contexts_test.dart:24:9 • argument_type_not_assignable`

Please update the tests and bump a major version since this is a breaking change.

bruno-garcia:
@Mooibroek Thank you for raising this. We need to make a few more changes in order to land this. I've raised #58 which addresses such things based on previous review of similar bug report (see #53), I hope it's OK.

PR #59, opened 2020-06-12
Example submits complete event

bruno-garcia:
To be merged after: #58 

This extends the example to use the whole Event (except culprit which we want to deprecate).

Running the example now reports two events while printing both ids to the console.

image

image

image


yjbanov:
For some reason github is not showing example-related changes, only the screen DPI change. Did you push all local changes to the github branch?

bruno-garcia:
You're right, let me check this.

bruno-garcia:
And the changes are gone.
Sorry, I'll open this again once I sort this out.

PR #60, opened 2020-07-05
Fix file_name to allow sentry detect correct file on flutter app

lfdominguez:


lfdominguez:
@googlebot I signed it!

PR #61, opened 2020-07-14
example complete event

bruno-garcia:
The example now creates two events:

image

One of which uses the complete Sentry event payload (except the exception which is featured in the second example):

image

The goal is to demonstrate how rich an event can be when using an integration which already collects a lot of that data from the device. Integrations will follow.

### DIFF
--- a/example/main.dart
+++ b/example/main.dart
@@ -18,10 +18,12 @@ Future<Null> main(List<String> rawArgs) async {
   final dsn = rawArgs.single;
   final client = SentryClient(dsn: dsn);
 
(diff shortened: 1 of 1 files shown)


example/main.dart
```diff
@@ -38,6 +40,91 @@ Future<Null> main(List<String> rawArgs) async {
   }
 }
 
+Future<Null> captureCompleteExampleEvent(SentryClient client) async {
```
yjbanov, line 44: Maybe leave a doc comment explaining what this example is trying to demonstrate?

yjbanov (approved):

PR #62, opened 2020-07-16
Added missing Contexts to Event.copyWith()

jamieastley:


jamieastley:
@googlebot  I signed it!

bruno-garcia (approved):
It's all shallow copying which could be particularly problematic for `Contexts` given that this data will be made available through [Sentry's Scope](https://docs.sentry.io/enriching-error-data/scopes/) which can be mutated while an event is queued for submission.

Since this is at least keeping it consistent it makes sense to me to merge it but I believe we need to change this approach.

@yjbanov thoughts?

bruno-garcia:
@jamie1192 if you can please add the entry to the changelog, we can get this merged.
More changes will land before 4.0.0 comes out though, likely the `Scope` class included and avoid breaking changes for a while after that.

jamieastley:
no problem @bruno-garcia , all done.

bruno-garcia:
Thanks you @jamie1192

PR #63, opened 2020-07-21
ref: remove default logger

bruno-garcia:
`logger` isn't really a required value. And having a constant value there isn't really very valuable.

This field is usually used when integrating with a logging library, and depending on the logger (aka: category, tag) that creates the record, it takes ends up on that name.

Since this is a "breaking change", fitting in the 5.0 release

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## 5.0.0
 
 - BREAKING CHANGE: Fixed context screenDpi is of type int #58
+- By default no logger it set #63
 
 ## 4.0.0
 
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## 5.0.0
 
 - BREAKING CHANGE: Fixed context screenDpi is of type int #58
+- By default no logger it set #63
```
bruno-garcia: This will go to 4.0.0 if #65 gets merged

lib/src/base.dart
```diff
@@ -155,8 +152,7 @@ abstract class SentryClient {
     final data = <String, dynamic>{
       'project': projectId,
       'event_id': _uuidGenerator(),
-      'timestamp': formatDateAsIso8601WithSecondPrecision(now),
-      'logger': defaultLoggerName,
+      'timestamp': formatDateAsIso8601WithSecondPrecision(now)
```
bruno-garcia:
```suggestion
      'timestamp': formatDateAsIso8601WithSecondPrecision(now),
```

yjbanov (approved):

PR #64, opened 2020-07-28
ref: rename capture to captureEvent

bruno-garcia:
To stay more in line with the other SDKs (Java is also getting renamed now on v2.0):

`capture` gets renamed to `captureEvent`.

We're not overloading these `capture` methods by parameter type but in fact we name them:

`captureException`
`captureEvent`
`captureMessage`

Coming up there'll be `captureTransaction`, `captureSession` etc.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## 5.0.0
 
 - BREAKING CHANGE: Fixed context screenDpi is of type int #58
+- BREAKING CHANGE: Renamed capture method to captureEvent #64
 
 ## 4.0.0
 
--- a/example/main.dart
+++ b/example/main.dart
@@ -112,7 +112,7 @@ Future<Null> captureCompleteExampleEvent(SentryClient client) async {
             timezone: 'America/Toronto',
           )));
 
-  final response = await client.capture(event: event);
+  final response = await client.captureEvent(event: event);
 
   print('\nReporting a complete event example: ');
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## 5.0.0
 
 - BREAKING CHANGE: Fixed context screenDpi is of type int #58
+- BREAKING CHANGE: Renamed capture method to captureEvent #64
```
bruno-garcia: This will go to 4.0.0 if #65 gets merged

example/main.dart
```diff
@@ -112,7 +112,7 @@ Future<Null> captureCompleteExampleEvent(SentryClient client) async {
             timezone: 'America/Toronto',
           )));
 
-  final response = await client.capture(event: event);
+  final response = await client.captureEvent(event: event);
```
bruno-garcia, line 115: I was also thinking of moving to positional arguments here. The call would be:

`captureEvent(event)`; in that case.

With regards to the `StackFrameFilter`, I'd like to remove that as an argument.
It will be provided via the `SentryOptions` object, via the constructor of the client.

I hope these change make sense, I hope to open PRs for that asap if you agree.

CHANGELOG.md
yjbanov: SGTM

yjbanov (approved):

PR #67, opened 2020-08-06
ref: Use Flutter linter rules

bruno-garcia:
Following up with [@yjbanov's suggestion to use Flutter's rules in this repo](https://github.com/flutter/sentry/pull/52#pullrequestreview-405410644):

Bringing the `analysis_options.yaml` from the flutter/flutter repo. The commit hash is at the top of the file for reference. 

The [issue discussed here](https://github.com/flutter/sentry/pull/52#discussion_r407118547) was solved on Dart 2.9.0 so I could remove the `// ignore`.

One modification to the rules was dropping `always_specify_types`, with a comment: `ignored sentry`.

Please let me know if you're not OK with that. I hope we can discuss if the goal is to align 100% with the Flutter rules or another reason to add it.

My take on it, after having the "should we use or not `var`" discussion already back in 2008 (in C#), I've never looked back into declaring types to the left hand side of the assignment operator. But I'm new to the ecosystem and it's possible I'm missing something.

- Adding @zoechi as a reviewer for visibility, since he's helping Sentry out with Flutter. Thank you @zoechi !

### DIFF
--- a/lib/src/base.dart
+++ b/lib/src/base.dart
@@ -43,6 +43,24 @@ abstract class SentryClient {
         compressPayload: compressPayload,
       );
 
+  SentryClient.base({
+    this.httpClient,
+    dynamic clock,
+    UuidGenerator uuidGenerator,
+    String dsn,
+    this.environmentAttributes,
+    String platform,
+    this.origin,
+  })  : _dsn = Dsn.parse(dsn),
+        _uuidGenerator = uuidGenerator ?? generateUuidV4WithoutDashes,
+        _platform = platform ?? sdkPlatform {
+    if (clock == null) {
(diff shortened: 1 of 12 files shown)


lib/src/base.dart
```diff
@@ -753,47 +753,89 @@ class Device {
         break;
     }
 
-    if (name != null) json['name'] = name;
+    if (name != null) {
```
zoechi, line 756: These if null ... look like good candidates for collection if
Not for this PR, but I think it would make the code look nicer

zoechi (approved):

yjbanov (approved):

bruno-garcia:
Thanks!

PR #104, opened 2020-10-09
Use caret and adjust min version

bruno-garcia:
Since we require Dart 2 or higher, we can use the caret notation to define dependency version requirements.

Whenever possible I used `^`: https://dart.dev/tools/pub/dependencies#caret-syntax

Goal: Keep dependencies on the lowest possibly version to allow apps bump if/when they need and not because they've added `sentry`. Dev dependencies can be on the latest version:

Breaking `http` bumped to 0.12 for the nullability requirements

Dev dependencies bumped to the latest version.

`uuid` lowered to 2.0.0 (nothing on the changelog or in API compatibility says we need anything higher).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,10 +5,11 @@
 - BREAKING CHANGE: Fixed context screenDensity is of type double #53
 - BREAKING CHANGE: Fixed context screenDpi is of type int #58
 - BREAKING CHANGE: Renamed capture method to captureEvent #64
+- BREAKING CHANGE: `package:http` min version bumped to 0.12.0 #104
+- BREAKING CHANGE: replace the `package:usage` by `package:uuid` #94
 - By default no logger it set #63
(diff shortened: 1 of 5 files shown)


bruno-garcia:
I'll merge this because the lock-file isnt' matching the pubspec.yaml anyway so `main` is broken:

```
  "runtimeInfo": {
  Unhandled exception:
    "panaVersion": "0.14.4",
  Exception: Running `flutter pub upgrade` failed with the following output:
    "sdkVersion": "2.10.1",
  
  ```
  Bad state: For sentry, the parsed version null did not match the locked version 4.0.0.
  ```
  #0      main (file:///app/bin/main.dart:74)
  #1      _RootZone.runUnary (dart:async/zone.dart:1439)
```

https://github.com/getsentry/sentry-dart/pull/105/checks?check_run_id=1232406921

bruno-garcia:
I'll open a follow up PR to address your feedback

PR #107, opened 2020-10-12
add a Message type to add a logEntry to Event

### LINKED ISSUE #100: Add logEntry to SentryEvent
Let's add the logEntry to the SentryEvent:

https://develop.sentry.dev/sdk/event-payloads/message/

https://github.com/getsentry/sentry-dart/blob/main/dart/lib/src/protocol/event.dart#L49-L52

the current `message` becomes its own class with `formatted`, `message` and `params` fields

rxlabz:

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
Add a new Message class

The Message Interface carries a log message that describes an event or error. Optionally, it can carry a format string and structured parameters. This can help to group similar messages into the same issue. cf. https://develop.sentry.dev/sdk/event-payloads/message/

fix #100 

I updated the existing tests to verify all the message properties are correctly serialized

## :pencil: Checklist

- [x] All tests passing
- [ ] No breaking changes

### DIFF
--- a/dart/lib/src/protocol/message.dart
+++ b/dart/lib/src/protocol/message.dart
@@ -1,3 +1,11 @@
+/// The Message Interface carries a log message that describes an event or error.
+/// Optionally, it can carry a format string and structured parameters. This can help to group similar messages into the same issue.
+/// example of a serialized message : {
+///   "message": {
+///     "message": "My raw message with interpreted strings like %s",
+///     "params": ["this"]
+///   }
+/// }
 class Message {
   /// The fully formatted message. If missing, Sentry will try to interpolate the message.
   final String formatted;
@@ -18,21 +26,4 @@ class Message {
       'params': params,
     };
   }
-
-  @override
-  String toString() {
-    return 'Message{formatted: $formatted, message: $message, params: $params}';
-  }
-
-  @override
-  bool operator ==(Object other) =>
-      identical(this, other) ||
-      other is Message &&
-          runtimeType == other.runtimeType &&
-          formatted == other.formatted &&
-          message == other.message &&
-          params == other.params;
-
-  @override
-  int get hashCode => formatted.hashCode ^ message.hashCode ^ params.hashCode;
 }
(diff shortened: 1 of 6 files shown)


dart/lib/src/protocol/message.dart
```diff
@@ -0,0 +1,38 @@
...
+    };
+  }
+
+  @override
+  String toString() {
+    return 'Message{formatted: $formatted, message: $message, params: $params}';
+  }
+
+  @override
+  bool operator ==(Object other) =>
```
bruno-garcia: We haven't been implementing these jn thr protocol. Is it really worth it?

dart/lib/src/protocol/message.dart
```diff
@@ -0,0 +1,38 @@
+class Message {
```
marandaneto, line 9: can we add a comment with a json example like all the other protocols?
eg https://github.com/getsentry/sentry-dart/blob/main/dart/lib/src/protocol/breadcrumb.dart#L6-L19

dart/lib/src/protocol/message.dart
rxlabz: Not really, I thought maybe it could in some tests, but the existing tests test the serialization equality, so I guess it's not necessary.

dart/lib/src/protocol/message.dart
```diff
@@ -0,0 +1,38 @@
...
+  bool operator ==(Object other) =>
+      identical(this, other) ||
+      other is Message &&
+          runtimeType == other.runtimeType &&
+          formatted == other.formatted &&
+          message == other.message &&
+          params == other.params;
+
+  @override
+  int get hashCode => formatted.hashCode ^ message.hashCode ^ params.hashCode;
```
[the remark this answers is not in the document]
rxlabz: I thought about it, but as it's not done for now in the protocol, I removed them for now.

marandaneto (approved):
LGTM

PR #108, opened 2020-10-13
feature : static SDK main entry withSentry.init(options),...

rxlabz:
…captureEvent(...), .captureException(...)

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring

## :scroll: Description

Add an Sentry class with static methods to init the sdk and capture the events.

Sentry SDK unified API https://develop.sentry.dev/sdk/unified-api/#static-api

By checking if the internal client is correctly used when calling the static methods

## :pencil: Checklist

- [x] All tests passing
- [ ] No breaking changes

## :crystal_ball: Next steps

more static methods


dart/example/main.dart
```diff
@@ -5,7 +5,10 @@
 import 'dart:async';
 import 'dart:io';
 
-import 'package:sentry/sentry.dart';
+import 'package:sentry/src/sentry.dart';
+import 'package:sentry/src/sentry_options.dart';
```
marandaneto: ideally `sentry_options.dart` is already imported by `sentry.dart`

dart/lib/src/sentry.dart
```diff
@@ -0,0 +1,57 @@
...
+import 'stack_trace.dart';
+
+/// Sentry SDK main entry point
+///
+class Sentry {
+  static SentryClient _client;
+
+  Sentry._();
+
+  static void init(SentryOptions options) {
```
marandaneto: instead of taking `SentryOptions` directly, if possible let's do it with a callback configurator, see:

https://github.com/getsentry/sentry-java/blob/main/sentry/src/main/java/io/sentry/Sentry.java#L104-L111

https://github.com/getsentry/sentry-java/blob/main/sentry/src/main/java/io/sentry/Sentry.java#L487-L500

so we have the change to instantiate `SentryOptions` ourselves and set default values, so people just mutate what they really need to afterwards

dart/lib/src/sentry.dart
```diff
@@ -0,0 +1,57 @@
...
+      environmentAttributes: options.environmentAttributes,
+      compressPayload: options.compressPayload,
+      httpClient: options.httpClient,
+      clock: options.clock,
+      uuidGenerator: options.uuidGenerator,
+    );
+  }
+
+  /// Initializes the SDK
+  static void initDns(String dns) => _client = SentryClient(dsn: dns);
```
marandaneto: we don't have this on other SDKs

dart/lib/src/sentry.dart
```diff
@@ -0,0 +1,57 @@
...
+  /// Reports an [event] to Sentry.io.
+  static Future<SentryResponse> captureEvent(
+    Event event, {
+    StackFrameFilter stackFrameFilter,
+  }) async {
+    return _client.captureEvent(
+      event: event,
+      stackFrameFilter: stackFrameFilter,
+    );
+  }
```
marandaneto: our APIs return either a https://github.com/getsentry/sentry-java/blob/main/sentry/src/main/java/io/sentry/protocol/SentryId.java
or an empty `SentryId` which is `00000000-0000-0000-0000-000000000000` I guess.
Also we never take a `stackFrameFilter` as a param, we can work on this stuff later, but this will need to be changed.

dart/lib/src/sentry_options.dart
```diff
@@ -0,0 +1,54 @@
...
+import 'utils.dart';
+
+/// Sentry SDK options
+class SentryOptions {
+  /// Default Log level if not specified Default is DEBUG
+  static final SeverityLevel defaultDiagnosticLevel = SeverityLevel.debug;
+
+  /// The DSN tells the SDK where to send the events to. If this value is not provided, the SDK will
+  ///  just not send any events.
+  final String dsn;
```
marandaneto: if we decide to use the callback: https://github.com/getsentry/sentry-dart/pull/108/files#r503897735

we can't make them `final`

dart/lib/src/sentry_options.dart
```diff
@@ -0,0 +1,54 @@
...
+
+  /// Contains [Event] attributes that are automatically mixed into all events
+  /// captured through this client.
+  ///
+  /// This event is designed to contain static values that do not change from
+  /// event to event, such as local operating system version, the version of
+  /// Dart/Flutter SDK, etc. These attributes have lower precedence than those
+  /// supplied in the even passed to [capture].
+  final Event environmentAttributes;
```
marandaneto: ideally we'd follow already defined names https://github.com/getsentry/sentry-java/blob/main/sentry/src/main/java/io/sentry/SentryOptions.java#L117-L122
if they are about the same attribute

dart/pubspec.yaml
```diff
@@ -6,7 +6,7 @@ description: >
 homepage: https://github.com/getsentry/sentry-dart
 
 environment:
-  sdk: ^2.0.0
+  sdk: ^2.1.0
```
marandaneto: why this change?

dart/test/sentry_test.dart
```diff
@@ -0,0 +1,108 @@
...
+import 'package:test/test.dart';
+
+void main() {
+  group('Sentry.init', () {
+    SentryClient client;
+
+    Exception anException;
+
+    final dns =
+        'https://cb0fad6f5d4e42ebb9c956cb0463edc9@o447951.ingest.sentry.io/5428562';
```
marandaneto: we just need a parseable DSN here, but not a valid one, right?

dart/test/sentry_test.dart
```diff
@@ -0,0 +1,108 @@
...
+    test('should capture the exception', () {
+      Sentry.captureException(anException);
+      verify(client.captureException(exception: anException)).called(1);
+    });
+  });
+}
+
+class MockSentryClient extends Mock implements SentryClient {}
+
+final event = Event(
```
marandaneto, line 39: not possible to reuse the `event_example.dart ` right? just curious if there's some magic to share code across tests/samples

dart/lib/src/sentry_options.dart
rxlabz: renamed to `environment`

dart/test/sentry_test.dart
rxlabz: not really as the test files are not exposed in the library.
We could set a relative path in the test file `import '../example/event_example.dart';` but I'm not sure I'll vote for this, do you ?

dart/lib/src/sentry_options.dart
rxlabz: yes indeed ! the class is now mutable.

dart/test/sentry_test.dart
marandaneto: nah, not really

dart/lib/src/client.dart
```diff
@@ -23,15 +23,15 @@ abstract class SentryClient {
...
+    Event environment,
     bool compressPayload,
     Client httpClient,
     dynamic clock,
     UuidGenerator uuidGenerator,
   }) =>
       createSentryClient(
         dsn: dsn,
-        environmentAttributes: environmentAttributes,
+        environmentAttributes: environment,
```
marandaneto: I'd rename the named field as well

dart/example/main.dart
```diff
@@ -77,3 +64,14 @@ Future<void> bar() async {
 Future<void> baz() async {
   throw StateError('This is a test error');
 }
+
+class Options implements OptionsConfiguration<SentryOptions> {
+  final String dsn;
+
+  Options({this.dsn});
```
marandaneto: why do we have the `dsn` here? I dont get it

dart/lib/src/sentry.dart
rxlabz: I removed it, but now the only way to init Sentry is to define an `OptionsConfiguration<SentryOptions>` which IMHO is not the simplest way for a beginner.

dart/lib/src/sentry.dart
```diff
@@ -52,3 +47,9 @@ class Sentry {
   @visibleForTesting
   static void initClient(SentryClient client) => _client = client;
 }
+
+/// Configuration options callback
+abstract class OptionsConfiguration<T extends SentryOptions> {
```
marandaneto: can't it be just an `interface`?

dart/example/main.dart
```diff
@@ -77,3 +64,14 @@ Future<void> bar() async {
 Future<void> baz() async {
   throw StateError('This is a test error');
 }
+
+class Options implements OptionsConfiguration<SentryOptions> {
```
marandaneto: do we need this? the concept of lambda won't work the same as Java for example?

dart/lib/src/client.dart
rxlabz: 😳

dart/lib/src/sentry.dart
rxlabz: it's now a typedef / function signature

dart/lib/src/browser_client.dart
```diff
@@ -17,15 +17,15 @@ import 'version.dart';
 
 SentryClient createSentryClient({
   @required String dsn,
-  Event environmentAttributes,
+  Event environment,
```
marandaneto: I'm a bit lost here, why an `Event` is called `environment`?

dart/lib/src/sentry.dart
```diff
@@ -0,0 +1,54 @@
+import 'dart:async';
+
+import 'package:meta/meta.dart';
+
+import 'client.dart';
+import 'protocol.dart';
+import 'sentry_options.dart';
+
+/// Configuration options callback
+typedef OptionsConfiguration = void Function(SentryOptions);
```
marandaneto, line 10: I liked that!

dart/example/main.dart
marandaneto: @rxlabz any reason to ignore this one?

dart/lib/src/sentry.dart
marandaneto: our docs state that and all the SDKs should look like this, also the samples, it should not be an issue

dart/pubspec.yaml
marandaneto: @rxlabz mind explaining?

dart/lib/src/browser_client.dart
marandaneto: https://github.com/getsentry/sentry-dart/pull/108#discussion_r503902646

probably they were not exactly the same thing?

rxlabz: from the doc : 
> /// [environment] contain event attributes that do not change over
  /// the course of a program's lifecycle. These attributes will be added to
  /// all events captured via this client. The following attributes often fall
  /// under this category: [Event.serverName], [Event.release], [Event.environment].

marandaneto: ok so then it makes sense to keep as `environmentAttributes` till we split them up to separate fields.

dart/pubspec.yaml
rxlabz: an autofix options error : clicked on 'upgrade sdk' instead of 'import dart:async', and didn't realized

dart/lib/src/sentry_options.dart
rxlabz: reverted to `environmentAttributes`

dart/example/main.dart
rxlabz: no it was updated

dart/lib/src/browser_client.dart
rxlabz: ✅ reverted

dart/lib/src/sentry.dart
rxlabz: should I rename the `SentryResponse` class https://github.com/getsentry/sentry-dart/blob/f63ac278c1de1251a466920e0bd7d5c082477e34/dart/lib/src/client.dart#L234-L256 to `SentryId`?

dart/example/main.dart
rxlabz: ah ok, sorry, I used to let the issue opener close it, but I'll close it when I'm sure it was resolved
.

dart/lib/src/sentry.dart
marandaneto: no, not really, this is the whole response, we'd need to create a `SentryId` class and inside of the client, we just read the `uuid` out of the response and create a `SentryId` with the proper `uuid`, this doesn't need to be in this PR

dart/example/main.dart
marandaneto: no worries, I also close them when I see it, but if there are no comments or reactions, I'd need to look for the code, I think this one should have been closed by me btw haha

PR #109, opened 2020-10-13
feat: add Scope class

marandaneto:

- [ ] Bugfix
- [X] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
add Scope class

we'd like to have Scopes

## :pencil: Checklist

- [X] All tests passing
- [ ] No breaking changes

## :crystal_ball: Next steps
eventProcessors, Contexts, BeforeBreadcrumbCallback, Breadcrumb hint, clone

### DIFF
--- a/dart/lib/src/scope.dart
+++ b/dart/lib/src/scope.dart
@@ -6,14 +6,32 @@ import 'sentry_options.dart';
 /// Scope data to be sent with the event
 class Scope {
   /// How important this event is.
-  SeverityLevel level;
+  SeverityLevel _level;
+
+  SeverityLevel get level => _level;
+
+  set level(SeverityLevel level) {
+    _level = level;
+  }
 
   /// The name of the transaction which generated this event,
   /// for example, the route name: `"/users/<username>/"`.
-  String transaction;
+  String _transaction;
+
+  String get transaction => _transaction;
+
+  set transaction(String transaction) {
+    _transaction = transaction;
+  }
 
   /// Information about the current user.
-  User user;
+  User _user;
+
+  User get user => _user;
+
+  set user(User user) {
+    _user = user;
+  }
 
   /// Used to deduplicate events by grouping ones with the same fingerprint
   /// together.
@@ -22,7 +40,14 @@ class Scope {
   ///
   ///     // A completely custom fingerprint:
   ///     var custom = ['foo', 'bar', 'baz'];
-  List<String> fingerprint;
+  List<String> _fingerprint;
+
+  List<String> get fingerprint =>
+      _fingerprint != null ? List.unmodifiable(_fingerprint) : null;
+
+  set fingerprint(List<String> fingerprint) {
+    _fingerprint = fingerprint;
+  }
 
   /// List of breadcrumbs for this scope.
   ///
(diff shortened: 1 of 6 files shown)


dart/pubspec.yaml
```diff
@@ -15,7 +15,7 @@ dependencies:
   uuid: ^2.0.0
 
 dev_dependencies:
-  mockito: ^4.1.2
+  mockito: ^4.1.1
```
marandaneto, line 18: because of null-safety, this is not working right thru the IDE, downgrading version made it to work

dart/lib/src/scope.dart
```diff
@@ -0,0 +1,80 @@
...
+  ///
+  /// Sentry.io docs do not talk about restrictions on the values, other than
+  /// they must be JSON-serializable.
+  final Map<String, dynamic> extra = {};
+
+  // TODO: eventProcessors, Contexts, BeforeBreadcrumbCallback, Breadcrumb hint, clome
+
+  final SentryOptions _options;
+
+  Scope(this._options) : assert(_options != null, 'SentryOptions is required');
```
marandaneto, line 78: @rxlabz whats the state of art here for required fields? `@required` or `assert` and boom?
afaik `@required` is only used for named params

dart/lib/src/scope.dart
```diff
@@ -0,0 +1,80 @@
...
+
+  /// Adds a breadcrumb to the breadcrumbs queue
+  void addBreadcrumb(Breadcrumb breadcrumb) {
```
marandaneto, line 81: @rxlabz is it possible to do an `assert` on method signature like L49 (ctor)? sounds weird to do inside but I'm fine with it

dart/test/scope_test.dart
```diff
@@ -0,0 +1,238 @@
+import 'package:sentry/sentry.dart';
+import 'package:test/test.dart';
+
+void main() {
+  final fixture = Fixture();
+
+  test('sets $SeverityLevel', () {
```
marandaneto, line 7: @rxlabz I see many people using `groups`, but most of the time a single group, does it even make sense? so I decided to not use it as the class is specific for this class and I can't group things here, maybe the breadcrumbs ring buffer could be grouped though, wdyt?

dart/lib/src/scope.dart
rxlabz: no you can only assert in this way with constructors

rxlabz: we'll remove all this non-null assertion soon ! :)

dart/test/scope_test.dart
rxlabz: groups give `setUp()`& `teardown()`but it's fine to not use them if not needed

rxlabz (approved):
LGMT

dart/lib/src/scope.dart
marandaneto: TIL, thanks for the info :)

dart/lib/src/scope.dart
marandaneto: hopefully when NNBD is GA

dart/test/scope_test.dart
marandaneto: only with groups?
https://github.com/getsentry/sentry-dart/blob/main/flutter/test/sentry_flutter_test.dart
and there are no groups in there

dart/test/scope_test.dart
```diff
@@ -0,0 +1,238 @@
...
+    final sut = fixture.getSut();
+
+    sut.level = SeverityLevel.debug;
+
+    expect(
+      sut.level,
+      SeverityLevel.debug,
```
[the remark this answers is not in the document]
marandaneto: the IDE kinda formats it automatically, will look into it

dart/test/scope_test.dart
rxlabz: you right 😶 !

dart/test/scope_test.dart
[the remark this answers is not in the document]
rxlabz: the Dart formatter breaks in multilines if there is a trailing comma `Severity.debug,`, and keep it in one line else.
Typically if the instruction is less than 80char you'll write `expect(sut.level,SeverityLevel.debug)`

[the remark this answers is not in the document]
marandaneto: yeah, fixed them thx

dart/lib/src/scope.dart
```diff
@@ -0,0 +1,139 @@
...
+/// Scope data to be sent with the event
+class Scope {
+  /// How important this event is.
+  SeverityLevel _level;
+
+  SeverityLevel get level => _level;
+
+  set level(SeverityLevel level) {
+    _level = level;
+  }
```
marandaneto, line 15: @bruno-garcia maybe we should add getters and setters only for List/Maps so they cannot mutate the whole thing?
https://dart.dev/guides/language/effective-dart/usage#dont-wrap-a-field-in-a-getter-and-setter-unnecessarily

bruno-garcia: Makes sense to me!

marandaneto:
merged into #108

PR #110, opened 2020-10-14
fix: dart ci

bruno-garcia:
The coverage action isn't working. Maybe because: https://github.com/stelynx/dart-full-coverage/issues/6


bruno-garcia:
https://github.com/stelynx/dart-full-coverage/issues/6#event-3896595467 was resolved so this could continue (code coverage could)

bruno-garcia:
Oh, forgot that we need to add coverage back in

PR #111, opened 2020-10-14
fix: dart ci no coverage

bruno-garcia:
Disabling coverage until we figure out the hang, ideally with an action such as: https://github.com/marketplace/actions/dart-full-coverage

See #110


bruno-garcia:
Merging this because at least tests fail (anything landing `main` in PRs now hangs forever and I have to manually cancel tons of actions)

PR #113, opened 2020-10-15
Feat : add a Hub class

rxlabz:

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring

## :pencil: Checklist

- [x] All tests passing
- [ ] No breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,7 @@
 - new static API : Sentry.init(), Sentry.captureEvent() #108
 - expect a sdkName based on the test platform #105
 - Added Scope and Breadcrumb ring buffer #109
-- Added Hub ring buffer #113
+- Added Hub to SDK #113
 
 # `package:sentry` changelog
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@
 - new static API : Sentry.init(), Sentry.captureEvent() #108
 - expect a sdkName based on the test platform #105
 - Added Scope and Breadcrumb ring buffer #109
+- Added Hub ring buffer #113
 
 # `package:sentry` changelog
 
--- a/dart/example/main.dart
+++ b/dart/example/main.dart
@@ -25,11 +25,12 @@ Future<void> main(List<String> rawArgs) async {
   // Sends a full Sentry event payload to show the different parts of the UI.
   final response = await Sentry.captureEvent(event);
 
-  if (response.isSuccessful) {
+  print('SentryId : ${response.id}');
+  /*if (response.isSuccessful) {
     print('SUCCESS\nid: ${response.eventId}');
   } else {
     print('FAILURE: ${response.error}');
-  }
+  }*/
 
   try {
     await foo();
@@ -41,11 +42,7 @@ Future<void> main(List<String> rawArgs) async {
       stackTrace: stackTrace,
     );
 
-    if (response.isSuccessful) {
-      print('SUCCESS\nid: ${response.eventId}');
-    } else {
-      print('FAILURE: ${response.error}');
-    }
+    print('SentryId : ${response.id}');
   } finally {
     await Sentry.close();
   }
--- a/dart/example/main.dart
+++ b/dart/example/main.dart
@@ -26,11 +26,6 @@ Future<void> main(List<String> rawArgs) async {
   final response = await Sentry.captureEvent(event);
 
   print('SentryId : ${response.id}');
-  /*if (response.isSuccessful) {
-    print('SUCCESS\nid: ${response.eventId}');
-  } else {
-    print('FAILURE: ${response.error}');
-  }*/
 
   try {
     await foo();
--- a/dart/example/main.dart
+++ b/dart/example/main.dart
@@ -25,7 +25,7 @@ Future<void> main(List<String> rawArgs) async {
   // Sends a full Sentry event payload to show the different parts of the UI.
   final sentryId = await Sentry.captureEvent(event);
 
-  print('SentryId : ${sentryId.id}');
+  print('SentryId : ${sentryId}');
 
   try {
     await foo();
@@ -37,7 +37,7 @@ Future<void> main(List<String> rawArgs) async {
       stackTrace: stackTrace,
     );
 
-    print('SentryId : ${sentryId.id}');
+    print('SentryId : ${sentryId}');
   } finally {
     await Sentry.close();
   }
--- a/dart/example/main.dart
+++ b/dart/example/main.dart
@@ -23,21 +23,21 @@ Future<void> main(List<String> rawArgs) async {
   print('\nReporting a complete event example: ');
 
   // Sends a full Sentry event payload to show the different parts of the UI.
-  final response = await Sentry.captureEvent(event);
+  final sentryId = await Sentry.captureEvent(event);
 
-  print('SentryId : ${response.id}');
+  print('SentryId : ${sentryId.id}');
 
   try {
     await foo();
   } catch (error, stackTrace) {
     print('\nReporting the following stack trace: ');
     print(stackTrace);
-    final response = await Sentry.captureException(
+    final sentryId = await Sentry.captureException(
       error,
       stackTrace: stackTrace,
     );
 
-    print('SentryId : ${response.id}');
+    print('SentryId : ${sentryId.id}');
   } finally {
     await Sentry.close();
   }
--- a/dart/lib/src/client.dart
+++ b/dart/lib/src/client.dart
@@ -182,27 +182,25 @@ abstract class SentryClient {
       body: body,
     );
 
-    if (response.statusCode != 200) {
-      var errorMessage = 'Sentry.io responded with HTTP ${response.statusCode}';
-      if (response.headers['x-sentry-error'] != null) {
-        errorMessage += ': ${response.headers['x-sentry-error']}';
-      }
-      return SentryId.empty();
-    }
-
     final eventId = '${json.decode(response.body)['id']}';
     return SentryId(eventId);
   }
 
-  /// Reports the [exception] and optionally its [stackTrace] to Sentry.io.
-  Future<SentryId> captureException({
-    @required dynamic exception,
-    dynamic stackTrace,
-  }) {
+  /// Reports the [throwable] and optionally its [stackTrace] to Sentry.io.
+  Future<SentryId> captureException(dynamic throwable, {dynamic stackTrace}) {
     final event = Event(
-      exception: exception,
-      stackTrace: stackTrace,
+      exception: throwable,
+      /*stackTrace: stackTrace,*/
     );
+    return captureEvent(event: event);
+  }
+
+  /// Reports the [message]
+  Future<SentryId> captureMessage({
+    @required Message message,
+    SeverityLevel level = SeverityLevel.warning,
+  }) {
+    final event = Event(message: message, level: level);
     return captureEvent(event: event);
   }
 
--- a/dart/lib/src/client.dart
+++ b/dart/lib/src/client.dart
@@ -205,12 +205,17 @@ abstract class SentryClient {
     return captureEvent(event);
   }
 
-  /// Reports the [message]
+  /// Reports the [template]
   Future<SentryId> captureMessage(
-    String message, {
+    String formatted, {
     SeverityLevel level = SeverityLevel.info,
+    String template,
+    List<dynamic> params,
   }) {
-    final event = Event(message: Message(formatted: message), level: level);
+    final event = Event(
+      message: Message(formatted, template: template, params: params),
+      level: level,
+    );
     return captureEvent(event);
   }
 
--- a/dart/lib/src/client.dart
+++ b/dart/lib/src/client.dart
@@ -136,7 +136,7 @@ abstract class SentryClient {
   }
 
   /// Reports an [event] to Sentry.io.
-  Future<SentryResponse> captureEvent({
+  Future<SentryId> captureEvent({
     @required Event event,
     StackFrameFilter stackFrameFilter,
   }) async {
@@ -187,15 +187,15 @@ abstract class SentryClient {
       if (response.headers['x-sentry-error'] != null) {
         errorMessage += ': ${response.headers['x-sentry-error']}';
       }
-      return SentryResponse.failure(errorMessage);
+      return SentryId.empty();
     }
 
     final eventId = '${json.decode(response.body)['id']}';
-    return SentryResponse.success(eventId: eventId);
+    return SentryId(eventId);
   }
 
   /// Reports the [exception] and optionally its [stackTrace] to Sentry.io.
-  Future<SentryResponse> captureException({
+  Future<SentryId> captureException({
     @required dynamic exception,
     dynamic stackTrace,
   }) {
--- a/dart/lib/src/client.dart
+++ b/dart/lib/src/client.dart
@@ -137,9 +137,10 @@ abstract class SentryClient {
   }
 
   /// Reports an [event] to Sentry.io.
-  Future<SentryId> captureEvent({
-    @required Event event,
+  Future<SentryId> captureEvent(
+    Event event, {
     StackFrameFilter stackFrameFilter,
+    Scope scope,
   }) async {
     final now = _clock();
     var authHeader = 'Sentry sentry_version=6, sentry_client=$clientId, '
@@ -201,7 +202,7 @@ abstract class SentryClient {
       exception: throwable,
       stackTrace: stackTrace,
     );
-    return captureEvent(event: event);
+    return captureEvent(event);
   }
 
   /// Reports the [message]
@@ -210,7 +211,7 @@ abstract class SentryClient {
     SeverityLevel level = SeverityLevel.info,
   }) {
     final event = Event(message: message, level: level);
-    return captureEvent(event: event);
+    return captureEvent(event);
   }
 
   Future<void> close() async {
--- a/dart/lib/src/client.dart
+++ b/dart/lib/src/client.dart
@@ -3,6 +3,7 @@ import 'dart:convert';
 
 import 'package:http/http.dart';
 import 'package:meta/meta.dart';
+import 'package:sentry/sentry.dart';
 
 import 'client_stub.dart'
     if (dart.library.html) 'browser_client.dart'
@@ -182,8 +183,16 @@ abstract class SentryClient {
       body: body,
     );
 
-    final eventId = '${json.decode(response.body)['id']}';
-    return SentryId(eventId);
+    if (response.statusCode != 200) {
+      /*var errorMessage = 'Sentry.io responded with HTTP ${response.statusCode}';
+      if (response.headers['x-sentry-error'] != null) {
+        errorMessage += ': ${response.headers['x-sentry-error']}';
+      }*/
+      return SentryId.empty();
+    }
+
+    final eventId = json.decode(response.body)['id'];
+    return eventId != null ? SentryId(eventId) : SentryId.empty();
   }
 
   /// Reports the [throwable] and optionally its [stackTrace] to Sentry.io.
--- a/dart/lib/src/client.dart
+++ b/dart/lib/src/client.dart
@@ -190,7 +190,7 @@ abstract class SentryClient {
   Future<SentryId> captureException(dynamic throwable, {dynamic stackTrace}) {
     final event = Event(
       exception: throwable,
-      /*stackTrace: stackTrace,*/
+      stackTrace: stackTrace,
     );
     return captureEvent(event: event);
   }
--- a/dart/lib/src/client.dart
+++ b/dart/lib/src/client.dart
@@ -207,7 +207,7 @@ abstract class SentryClient {
   /// Reports the [message]
   Future<SentryId> captureMessage({
     @required Message message,
-    SeverityLevel level = SeverityLevel.warning,
+    SeverityLevel level = SeverityLevel.info,
   }) {
     final event = Event(message: message, level: level);
     return captureEvent(event: event);
--- a/dart/lib/src/client.dart
+++ b/dart/lib/src/client.dart
@@ -207,10 +207,10 @@ abstract class SentryClient {
 
   /// Reports the [message]
   Future<SentryId> captureMessage(
-    Message message, {
+    String message, {
     SeverityLevel level = SeverityLevel.info,
   }) {
-    final event = Event(message: message, level: level);
+    final event = Event(message: Message(formatted: message), level: level);
     return captureEvent(event);
   }
 
--- a/dart/lib/src/client.dart
+++ b/dart/lib/src/client.dart
@@ -205,8 +205,8 @@ abstract class SentryClient {
   }
 
   /// Reports the [message]
-  Future<SentryId> captureMessage({
-    @required Message message,
+  Future<SentryId> captureMessage(
+    Message message, {
     SeverityLevel level = SeverityLevel.info,
   }) {
     final event = Event(message: message, level: level);
--- a/dart/lib/src/hub.dart
+++ b/dart/lib/src/hub.dart
@@ -269,23 +269,6 @@ class Hub {
       }
     }
   }
-
-  /// Runs the callback with a new scope which gets dropped at the end
-  void pushScope() {
-    // TODO: implement popScope
-    throw UnimplementedError();
-  }
-
-  void popScope() {
-    // TODO: implement popScope
-    throw UnimplementedError();
-  }
-
-  /// Runs the callback with a new scope which gets dropped at the end
-  void withScope(ScopeCallback callback) {
-    // TODO: implement withScope
-    throw UnimplementedError();
-  }
 }
 
 class _StackItem {
--- a/dart/lib/src/hub.dart
+++ b/dart/lib/src/hub.dart
@@ -146,6 +146,8 @@ class Hub {
   Future<SentryId> captureMessage(
     String message, {
     SeverityLevel level = SeverityLevel.info,
+    String template,
+    List<dynamic> params,
   }) async {
     var sentryId = SentryId.empty();
 
@@ -164,7 +166,12 @@ class Hub {
       if (item != null) {
         try {
           // TODO pass the scope
-          sentryId = await item.client.captureMessage(message, level: level);
+          sentryId = await item.client.captureMessage(
+            message,
+            level: level,
+            template: template,
+            params: params,
+          );
         } catch (err) {
           _options.logger(
             SeverityLevel.error,
--- a/dart/lib/src/hub.dart
+++ b/dart/lib/src/hub.dart
@@ -2,6 +2,7 @@ import 'dart:async';
 import 'dart:collection';
 
 import 'client.dart';
+import 'noop_client.dart';
 import 'protocol.dart';
 import 'scope.dart';
 import 'sentry_options.dart';
@@ -76,7 +77,7 @@ class Hub {
       final item = _stack.last;
       if (item != null) {
         try {
-          sentryId = await item.client.captureEvent(event: event);
+          sentryId = await item.client.captureEvent(event, scope: item.scope);
         } catch (err) {
           /* TODO add Event.id */
           _options.logger(
@@ -116,6 +117,7 @@ class Hub {
       final item = _stack.last;
       if (item != null) {
         try {
+          // TODO pass the scope
           sentryId = await item.client.captureException(
             throwable,
             stackTrace: stackTrace,
@@ -158,10 +160,8 @@ class Hub {
       final item = _stack.last;
       if (item != null) {
         try {
-          sentryId = await item.client.captureMessage(
-            message,
-            level: level,
-          );
+          // TODO pass the scope
+          sentryId = await item.client.captureMessage(message, level: level);
         } catch (err) {
           _options.logger(
             SeverityLevel.error,
@@ -181,7 +181,26 @@ class Hub {
 
   /// Binds a different client to the hub
   void bindClient(SentryClient client) {
-    _stack.add(_StackItem(client, _stack.last.scope));
+    if (!_isEnabled) {
+      _options.logger(SeverityLevel.warning,
+          "Instance is disabled and this 'bindClient' call is a no-op.");
+    } else {
+      final item = _stack.last;
+      if (item != null) {
+        if (client != null) {
+          _options.logger(SeverityLevel.debug, 'New client bound to scope.');
+          item.client = client;
+        } else {
+          _options.logger(SeverityLevel.debug, 'NoOp client bound to scope.');
+          item.client = NoOpSentryClient();
+        }
+      } else {
+        _options.logger(
+          SeverityLevel.fatal,
+          'Stack peek was null when bindClient',
+        );
+      }
+    }
   }
 
   /// Clones the Hub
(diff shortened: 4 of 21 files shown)


rxlabz:
@marandaneto @bruno-garcia Hi, I started this PR, early feedbacks are welcome :)

dart/lib/src/client.dart
```diff
@@ -183,29 +184,35 @@ abstract class SentryClient {
...
+      exception: throwable,
       stackTrace: stackTrace,
     );
     return captureEvent(event: event);
   }
 
+  /// Reports the [message]
+  Future<SentryId> captureMessage({
+    @required Message message,
+    SeverityLevel level = SeverityLevel.warning,
```
marandaneto: default is `info`

dart/lib/src/hub.dart
```diff
@@ -0,0 +1,379 @@
...
+  Hub._(SentryOptions options)
+      : _options = options,
+        _stack = ListQueue() {
+    _stack.add(_StackItem(_getClient(fromOptions: options), Scope(_options)));
+    _isEnabled = true;
+  }
+
+  static void _validateOptions(SentryOptions options) {
+    if (options == null) {
+      throw ArgumentError.notNull('options');
```
marandaneto, line 44: mm I remember asking in other PR about the `assert(condition)` and you said it was the only way, TIL `ArgumentError.notNull`, should we only use this one them?

dart/lib/src/hub.dart
```diff
@@ -0,0 +1,379 @@
...
+        SeverityLevel.warning,
+        'captureEvent called with null parameter.',
+      );
+    } else {
+      final item = _stack.last;
+      if (item != null) {
+        try {
+          sentryId = await item.client.captureEvent(event: event);
+        } catch (err) {
+          /* FIXME(rxlabz) ? Event.id ?*/
```
marandaneto: I suggest avoiding FIXME but instead use TODO, it's easier to find out (IDE helps), also to not use (rxlabz) because I could help on it too

dart/lib/src/hub.dart
```diff
@@ -0,0 +1,379 @@
...
+      }
+    }
+    _lastEventId = sentryId;
+    return sentryId;
+  }
+
+  @override
+  Future<SentryId> captureException({
+    dynamic throwable,
+    dynamic stackTrace,
```
marandaneto: why are we using `dynamic`? should we not use the types?

dart/lib/src/hub.dart
```diff
@@ -0,0 +1,379 @@
...
+  void popScope() {
+    // TODO: implement popScope
+    throw UnimplementedError();
+  }
+
+  @override
+  void pushScope() {
+    // TODO: implement pushScope
+    throw UnimplementedError();
+  }
```
marandaneto: I think we can avoid them, as Dart is single-threaded.
cc @bruno-garcia

dart/lib/src/hub.dart
```diff
@@ -0,0 +1,379 @@
...
+    }
+    _lastEventId = sentryId;
+    return sentryId;
+  }
+
+  @override
+  void addBreadcrumb(Breadcrumb breadCrumb) {
+    // TODO: implement addBreadcrumb
+    throw UnimplementedError();
+  }
```
marandaneto: we can skip it, ideally people will do that thru `configureScope`, the reason is, we've added them on Java to easier migration from Java SDK v1 to v2, but now we are not gonna add them

dart/lib/src/hub.dart
```diff
@@ -0,0 +1,379 @@
...
+  @override
+  void bindClient(SentryClient client) {
+    _stack.add(_StackItem(client, _stack.last.scope));
+  }
+
+  @override
+  void clearBreadcrumbs() {
+    // TODO: implement clearBreadcrumbs
+    throw UnimplementedError();
+  }
```
marandaneto: https://github.com/getsentry/sentry-dart/pull/113/files#r505776960

dart/lib/src/hub.dart
```diff
@@ -0,0 +1,379 @@
...
+  void startSession() {
+    // TODO: implement startSession
+    throw UnimplementedError();
+  }
+
+  @override
+  void stopSession() {
+    // TODO: implement stopSession
+    throw UnimplementedError();
+  }
```
marandaneto: https://github.com/getsentry/sentry-dart/pull/113/files#r505776960

dart/test/test_utils.dart
```diff
@@ -303,12 +303,12 @@ void runTest({Codec<List<int>, List<int>> gzip, bool isWeb = false}) {
     try {
       throw ArgumentError('Test error');
     } catch (error, stackTrace) {
-      final response = await client.captureException(
-          exception: error, stackTrace: stackTrace);
-      expect(response.isSuccessful, false);
-      expect(response.eventId, null);
-      expect(
-          response.error, 'Sentry.io responded with HTTP 401: Invalid api key');
+      final response =
```
marandaneto: we could rename the `response` to sentryId as well

dart/lib/src/hub.dart
```diff
@@ -0,0 +1,379 @@
...
+
+import 'client.dart';
+import 'protocol.dart';
+import 'scope.dart';
+import 'sentry_options.dart';
+
+typedef ScopeCallback = void Function(Scope);
+
+/// SDK API contract which combines a client and scope management
+class Hub implements HubInterface {
```
marandaneto: for the interfaces, is it common to use Name`Interface` in Dart? if theres not really a default pattern, we could do like Java, IHub

rxlabz: Actually I wonder if we need one. It helped me to list the methods but IMO the Dart way would be to not have an interface at all here. The NoOpHub could implements Hub directly ?

dart/lib/src/hub.dart
rxlabz: Good point ! I'm used to code mostly in Flutter where `assert` are executed only in debug mode, but as it's the Dart package so, you right, assert could be used here ( before NNBD ).

dart/lib/src/hub.dart
marandaneto: we use the NoOp pattern because Java, its not null-safety, I believe here we'd need to.
the reason for using interfaces is to be able to test/mock them.

Also, some classes could be overwritten by the customer, if they prefer to have their own implementation, not really needed for Hub and Client I guess, but its common to see for the Transport (which we still dont have for example)

dart/lib/src/hub.dart
rxlabz: there is no type union in Dart, and both Error and Exception are throwable, and not related. In fact you can throw anything in Dart ( but `null` which throws NullThrownError  ).

marandaneto: I see, well making it dynamic works but if it could be literally anything, how we would know which fields to read the info out of the exception, stack trace, messages and so on, we'd need to think this thru, if its the case, we could offer overloads like
captureException(exception type)
captureError(error type)
capture??(any?)

I am just thinking out loud here :)

dart/lib/src/hub.dart
rxlabz: i removed both startSession & stopSession !?

dart/lib/src/hub.dart
rxlabz: There is no common interface for Dart `Errors`
-  a Dart `Error` has no fields https://api.dart.dev/stable/2.10.2/dart-core/Error-class.html.
- `Exception` has an optional argument ( String message ) https://api.dart.dev/stable/2.10.2/dart-core/Exception-class.html
For now we send the `exception.runtimeType` and the `exception.toString()`, whatever the exception is.
```dart
if (exception != null) {
      json['exception'] = [
        <String, dynamic>{
          'type': '${exception.runtimeType}',
          'value': '$exception',
        }
      ];
    }
```

dart/lib/src/hub.dart
marandaneto: the comment was made on lines 229 to 287, so all of them

dart/lib/src/hub.dart
marandaneto: I see but an Error has a stack trace field https://api.dart.dev/stable/2.10.2/dart-core/Error/stackTrace.html

dart/lib/src/hub.dart
rxlabz: Well, finally...
As the package can be used in multiple contexts ( like Flutter or browser ), the asserts will not always be executed, and it can leads to NullError. While NNBD is not landed this is the best option.

dart/lib/src/hub.dart
rxlabz: we could add it like this ?
```dart
if (exception != null) {
      json['exception'] = [
        <String, dynamic>{
          'type': '${exception.runtimeType}',
          'value': '$exception',
           if( exception is Error ) 'stackTrace': '${exception.stackTrace ?? '' }' 
        }
      ];
    }
```

dart/lib/src/hub.dart
rxlabz: For mocking :
```
class MockHub extends Mock implements Hub{}
```

dart/lib/src/client.dart
bruno-garcia: is there an implicit convert between `Message` and `String`?
It should be possible to capture:

`Sentry.captureMessage("string here")` but I'm afraid that with our `Message` class it'll need to be:

`Sentry.captureMessage({message: Message({formatted: "string here"})})` which isn't really straight forward.

dart/lib/src/hub.dart
bruno-garcia: We can have scopes in a single-threaded scenario, it's actually simpler since we don't have the risk of Thread A pushes and Thread B pops and things out of sync. But maybe we leave this out from the first version to simplify things? since we don't really do push and pop scope on Android either, probably better off leaving it out for now and reconsidering this later.

dart/lib/src/client.dart
rxlabz: we could declare 
```dart
Sentry.captureMessage(String message,{ List<dynamic> messageParams = [], SeverityLevel level })
```
 and instanciate a `Message` in the hub with only a `formatted` value if `params.isEmpty` ?

We should use `pkg:sprintf` https://pub.dev/packages/sprintf  to interpolate the formatted string.

marandaneto: as overloading is not supported by Dart, I suggest using all of them as optional named params.
captureMessage({formatted, message, params}) so they choose what to pass, either formatted OR message + params, the docs on top of the method should be clear about it

dart/lib/src/hub.dart
marandaneto: mm I'm lost here now, which one is the best option? `assert` or `ArgumentError.notNull`?

dart/lib/src/hub.dart
rxlabz: or maybe ?
```dart
if (exception != null) {
      json['exception'] = [
        <String, dynamic>{
          'type': '${exception.runtimeType}',
          'value': '$exception',
           'stackTrace': '${exception.stackTrace ?? '' }' 
        }
      ];
      if( exception is Error ){
        json['stacktrace'] = exception.stackTrace ?? '';
     }
    }
```

marandaneto: I'd need to dig a bit more about it, the concept of throwing anything is a bit weird to me right now, can't really tell, but yeah, it seems this conditional `if` does the job.

dart/lib/src/hub.dart
marandaneto: if in Dart a class could `implements` any other class, maybe interfaces are not really necessary, good point

dart/lib/src/hub.dart
rxlabz: sorry, for now the best option is to test null value and throw `ArgumentError.notNull`.

dart/lib/src/scope.dart
```diff
@@ -136,4 +138,28 @@ class Scope {
...
+  @override
+  int get hashCode =>
+      _level.hashCode ^
+      _transaction.hashCode ^
+      _user.hashCode ^
+      _fingerprint.hashCode ^
+      _breadcrumbs.hashCode ^
+      _tags.hashCode ^
+      _extra.hashCode ^
+      _options.hashCode;
```
marandaneto: mm why do we need this?

rxlabz: to test scopes equality in tests.

dart/lib/src/hub.dart
marandaneto: yep, let's go simple/easy for now

dart/lib/src/hub.dart
```diff
@@ -0,0 +1,297 @@
...
+      dsn: fromOptions.dsn,
+      environmentAttributes: fromOptions.environmentAttributes,
+      compressPayload: fromOptions.compressPayload,
+      httpClient: fromOptions.httpClient,
+      clock: fromOptions.clock,
+      uuidGenerator: fromOptions.uuidGenerator,
+    );
+  }
+
+  final ListQueue<_StackItem> _stack;
```
marandaneto, line 25: we use a `Deque` in Java, we'd need to learn the differences if not a Deque available in Dart
push/pop/peek should work in the same way

dart/lib/src/hub.dart
rxlabz: I finally removed it ! unless we don't want expose some Hub methods, it makes no difference.

dart/test/test_utils.dart
rxlabz: ✅

dart/lib/src/hub.dart
marandaneto: then every non exposed method should be private

dart/lib/src/hub.dart
rxlabz: I added the `exception is Error` condition

dart/lib/src/hub.dart
rxlabz: Dart `ListQueue` seems to offer the same features

dart/lib/src/hub.dart
```diff
@@ -0,0 +1,280 @@
...
+        SeverityLevel.warning,
+        "Instance is disabled and this 'captureException' call is a no-op.",
+      );
+    } else if (throwable == null) {
+      _options.logger(
+        SeverityLevel.warning,
+        'captureException called with null parameter.',
+      );
+    } else {
+      final item = _stack.first;
```
bruno-garcia, line 118: Maybe we should just add push and pop scope now, given that all the work to support is is done here

dart/lib/src/hub.dart
```diff
@@ -0,0 +1,280 @@
...
+
+  /// Captures the message.
+  Future<SentryId> captureMessage(
+    Message message, {
+    SeverityLevel level = SeverityLevel.info,
+  }) async {
+    var sentryId = SentryId.empty();
+
+    if (!_isEnabled) {
+      _options.logger(
```
bruno-garcia, line 155: Seems we log and leave the if block, assign the sentryId into the _lastId bit but no event was captured which is misleading.

Maybe we just `return` here?

dart/lib/src/hub.dart
```diff
@@ -0,0 +1,280 @@
...
+        'captureMessage called with null parameter.',
+      );
+    } else {
+      final item = _stack.first;
+      if (item != null) {
+        try {
+          // TODO pass the scope
+          sentryId = await item.client.captureMessage(message, level: level);
+        } catch (err) {
+          _options.logger(
```
bruno-garcia, line 176: Here we clearly failed to capture, so again I wouldn't suggest assigning the eventId. Let's just do that if we at least were able to pass the event to the client?

dart/lib/src/protocol/sentry_id.dart
```diff
@@ -0,0 +1,16 @@
...
+
+const _dashIndexed = [8, 13, 18, 23];
+
+class SentryId {
+  static final String emptyId =
+      List.generate(36, (index) => _dashIndexed.contains(index) ? '-' : '0')
+          .join();
+
+  /// The ID Sentry.io assigned to the submitted event for future reference.
+  final String id;
```
bruno-garcia: Could we have an override on toString instead? Ideally we'd use an instance hereas a string (because it really is)

dart/lib/src/protocol/sentry_id.dart
```diff
@@ -0,0 +1,16 @@
+/// Sentry response id
+
+const _dashIndexed = [8, 13, 18, 23];
+
+class SentryId {
+  static final String emptyId =
```
bruno-garcia: I'd propose hard code the `0000` here instead of running this code. It's a `const`

dart/lib/src/scope.dart
bruno-garcia: Could we move this code to the tests instead? I'd avoid having this here if we don't need to maintain as production code (like putting scopes in as Map keys)

dart/pubspec.yaml
```diff
@@ -13,6 +13,7 @@ dependencies:
   meta: ^1.0.0
   stack_trace: ^1.0.0
   uuid: ^2.0.0
+  collection: ^1.14.13
```
bruno-garcia: We could drop this dependency if we don't have a scope stack?
Would this be a reason to just go with a plain `Scope` field in the `Hub` instead?
The code will be MUCH simpler in that case.

dart/test/hub_test.dart
```diff
@@ -0,0 +1,132 @@
+import 'package:mockito/mockito.dart';
+import 'package:sentry/sentry.dart';
+import 'package:sentry/src/hub.dart';
+import 'package:test/test.dart';
+
+import 'mocks.dart';
+
+void main() {
+  group('Hub instanciation', () {
```
bruno-garcia:
```suggestion
  group('Hub instantiation', () {
```

dart/test/hub_test.dart
```diff
@@ -0,0 +1,132 @@
...
+    test('should not instanciate without a sentryOptions', () {
```
bruno-garcia:
```suggestion
    test('should not instantiate without a sentryOptions', () {
```

bruno-garcia: There are few more occurrences of this typo

bruno-garcia (commented):
I know it's a draft but I left a couple notes, I hope it helps

dart/lib/src/hub.dart
rxlabz: the `_lastEventId = sentryId;` assignment ?

dart/lib/src/client.dart
bruno-garcia: The snippet from Erick would work thouhg, the first arg would be positional, and the following would be named. That would work, right?
I looks good to me since you can:

`Sentry.captureMessage("bla")`;
and

`Sentry.captureMessage("bla", {level: SentryLevel.debug});`

Not sure about params though. The `formatted` stuff is more useful for logging libraries.
I'm unaware of folks using it with Sentry.captureMessage. but maybe it makes sense to take into account here. but in this case the positional argument would be `formatted` (although doesn't have to be named that way), and the named arg could take the template:

```dart
Sentry.captureMessage("message with user id 1", {
level: SentryLevel.debug,
params: ["1"],
template: "message with user id %s"
});
```

Looks pretty horrible though.

dart/lib/src/hub.dart
bruno-garcia: that means adding a dependency, could we reconsider adding this code right now until we are certain we'll add push/pop scope?

CHANGELOG.md
```diff
@@ -20,6 +20,7 @@
 - new static API : Sentry.init(), Sentry.captureEvent() #108
 - expect a sdkName based on the test platform #105
 - Added Scope and Breadcrumb ring buffer #109
+- Added Hub ring buffer #113
```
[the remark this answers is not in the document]
rxlabz: 😳

dart/pubspec.yaml
rxlabz: here, this package is only used for lists and maps "deep comparaisons" in `==` operator overrides

dart/lib/src/scope.dart
rxlabz: I'm not sure I can without changing this kind of tests : 
```dart
test('should capture event', () {
      hub.captureEvent(fakeEvent);
      verify(
        client.captureEvent(fakeEvent, scope: Scope(options), stackFrameFilter: null),
      ).called(1);
    });
```
If we don't override the `==`operator the test will fail as it will only compare scope reference. 
To remove the need of this override, I could change the test to : 
```dart
test('should capture event', () {
      hub.captureEvent(fakeEvent);
      verify(
        client.captureEvent(fakeEvent,scope: anyNamed('scope'), stackFrameFilter: null),
      ).called(1);
    });
```
but it less precise. WDYT?

bruno-garcia: Based on the name of the tests, we're not really asserting anything about the scope, is that right? We just test that an event was captured?

For a test that requires validating equality of the argument, is there a arg validator that takes a callback?
like:

`arg.is(s => s.a == expected.a)` ?

CHANGELOG.md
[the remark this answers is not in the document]
bruno-garcia: Sorry if twas confused. Maybe we could add this entry as:

"Added Hub to SDK" or "added Hub"? 

I just think "ring buffer" is confusing since we have that for breadcrumbs but not here.
This PR includes a stack of scopes though maybe that's the confusion?

dart/pubspec.yaml
bruno-garcia: Maybe we should consider dropping the package and the overrides if there's an alternative.
For context: https://develop.sentry.dev/sdk/philosophy/#dependencies-cost

dart/lib/src/client.dart
rxlabz: I'm not sure I understand how Message will be used. Will it be instantiated with formatted AND template AND params ?

dart/lib/src/scope.dart
rxlabz: this works :
```dart
bool scopeEquals(Scope a, Scope b) {
  return identical(a, b) || a.level == b.level &&
          a.transaction == b.transaction &&
          a.user == b.user &&
          IterableEquality().equals(a.fingerprint, b.fingerprint) &&
          IterableEquality().equals(a.breadcrumbs, b.breadcrumbs) &&
          MapEquality().equals(a.tags, b.tags) &&
          MapEquality().equals(a.extra, b.extra);
}

test(
  'should capture event with correct scope',
  () {
    hub.captureEvent(fakeEvent);
    expect(
        scopeEquals(
          verify(
            client.captureEvent(
              fakeEvent,
              scope: captureAnyNamed('scope'),
              stackFrameFilter: null,
            ),
          ).captured.first,
          Scope(options),
        ),
        true);
  },
);
```
but we still depend on package collection

bruno-garcia: Add the dependency as a test dependency and we're golden

bruno-garcia: that equals funciton could be an extension method on the scope class, defined on the test package?

CHANGELOG.md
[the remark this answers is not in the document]
rxlabz: yes sorry :) I was just shocked by what I wrote (copy/pasted :D ) !

dart/lib/src/client.dart
bruno-garcia: Minimum you'd send is `formatted`, with the interpolated string, which would show in Sentry the right value. Optionally you'd send `message` which would be used in grouping so that:

"user %s logged in" would all group together, instead of "user 1 logged in" and "user 2 logged in" becomes two separate issues in Sentry. But the latter is useful since when debugging you want to know the actual value. That's the difference of formatted and message.

Params is optional, but it can show in the UI the individual arguments, so it's nice to include.

This originally was called "logEntry" since it's designed to be compatible with a logging library, such as you write:

logger.warn("user %s", 1); and under the hood we can replace the %s with 1 and get the 3 things (formatted, message and params).

But for captureMessage, it should be a simple API to send a message. Optional level is nice to have too because by default it's `info`. Having captureMessage take a class, makes it harder to send a basic string (unless we have some implicit operator in Dart). 

I propose we keep what you did here except:

```dart
    @required String message,
    {SeverityLevel level = SeverityLevel.info}) {
final event = Event({
message: Message(formatted: message, level: level)}
captureEvent(event);
}
```

I'd also prose having all these method become extension methods and not part of the actual interface.

Someone implementing a client shoulndn't have to make these choices. The interface could be much simpler, for example: [ISentryClient](https://github.com/getsentry/sentry-dotnet/blob/main/src/Sentry/ISentryClient.cs) and [extensions](https://github.com/getsentry/sentry-dotnet/blob/main/src/Sentry/SentryClientExtensions.cs#L33-L47)"

bruno-garcia: I could suggest this approach in Java because Java 1.8 (or any?) has extension method, but Dart does.

dart/lib/src/client.dart
```diff
@@ -183,27 +185,33 @@ abstract class SentryClient {
...
         errorMessage += ': ${response.headers['x-sentry-error']}';
-      }
-      return SentryResponse.failure(errorMessage);
+      }*/
+      return SentryId.empty();
     }
 
-    final eventId = '${json.decode(response.body)['id']}';
-    return SentryResponse.success(eventId: eventId);
+    final eventId = json.decode(response.body)['id'];
```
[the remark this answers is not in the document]
rxlabz: no it will be == null

dart/lib/src/protocol/sentry_id.dart
rxlabz: ✅

dart/pubspec.yaml
rxlabz: ✅ moved to the devDependencies

dart/example/main.dart
```diff
@@ -23,29 +23,21 @@ Future<void> main(List<String> rawArgs) async {
...
   // Sends a full Sentry event payload to show the different parts of the UI.
-  final response = await Sentry.captureEvent(event);
+  final sentryId = await Sentry.captureEvent(event);
 
-  if (response.isSuccessful) {
-    print('SUCCESS\nid: ${response.eventId}');
-  } else {
-    print('FAILURE: ${response.error}');
-  }
+  print('SentryId : ${sentryId.id}');
```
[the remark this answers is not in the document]
rxlabz: updated : `toString` now returns the uuid without dashes. `id` still return the full uuid ( as a String, as Dart `Uuid`can't be generated from a `String` )

dart/lib/src/hub.dart
rxlabz: its from `dart:collection` (standard lib ) , not `pkg:collection`

dart/lib/src/scope.dart
rxlabz: the "dartiest" way ! 😄

rxlabz: but needs Dart >=2.6 :(

dart/lib/src/client.dart
marandaneto: extension methods in Dart are only available in Dart 2.7, is this realistic enough to add @rxlabz or people still use older versions than that?
got @bruno-garcia point, I thought that when you have `message` and `params`, Sentry would set the `formatted` field automatically, but yeah, `message` and `params` should be named params then

dart/lib/src/scope.dart
marandaneto: I'd also prefer to keep this type of code needed only for tests, on the test packages

dart/lib/src/hub.dart
marandaneto: we could add it later too, let's focus on the missing and required parts for now, of course, if some time left, we'll be back to it even on this sprint

dart/lib/src/client.dart
marandaneto: what does it happen if someone is using Dart < 2.7? does the SDK wont work?

dart/lib/src/hub.dart
marandaneto: `stackTrace` could be read in SentryClient from the `throwable ` if its an error and not taken as params in the Sentry static class/Hub

dart/lib/src/hub.dart
marandaneto: the Java SDK does pretty much the same, the `lastEventId` will be empty if an event failed (eg SDK disabled) to be captured for some reason, I will check other SDKs

dart/lib/src/hub.dart
marandaneto: ok, so if we were able to call captureMessage or captureEvent... we assign the sentryId/lastEventId otherwise it really doesnt make sense, this should be fixedi n Java too

dart/lib/src/hub.dart
marandaneto: https://github.com/getsentry/sentry-dart/pull/113#discussion_r507580774

dart/lib/src/client.dart
rxlabz: yes, to use the library you'll have to install the min sdk

rxlabz: Update : now `hub.captureMessage` and `client.captureMessage` only take a String message required argument, and an optional named `level`.
```dart
Future<SentryId> captureMessage(String message , { SeverityLevel level = SeverityLevel.info});
```

marandaneto: @rxlabz whats about the `params` and `template`? they should also be named params

rxlabz: > I propose we keep what you did here except:

 > Future<SentryId> captureMessage(
    @required String message,
    {SeverityLevel level = SeverityLevel.info}) {
final event = Event({
message: Message(formatted: message, level: level)}
captureEvent(event);
}

I thought that was the final decision. So now we have :
```dart
    String formatted, {
    SeverityLevel level = SeverityLevel.info,
    String template,
    List<dynamic> params,
  }) {
    final event = Event(
      message: Message(formatted, template: template, params: params),
      level: level,
    );
    return captureEvent(event);
  }
```

dart/lib/src/hub.dart
rxlabz: I found the reason why Dart exceptions have no stacktrace : https://groups.google.com/a/dartlang.org/g/misc/c/O1OKnYTUcoo/m/ipIE1jqHD9sJ
I think the current implementation allows to send both : 
- an exception and it's stacktrace if the user catch it and want to send it, 
- and an error with the stacktrace automatically serialized
Is it ok for you @marandaneto ?

dart/lib/src/scope.dart
rxlabz: ✅  moved in the test side, as a simple method for now
We should decide if we update the SDK constraints to use extensions and other "modern dart" features :)

dart/lib/src/client.dart
marandaneto: nice, only `formatted` is mandatory anyway so I'm in for that

dart/lib/src/hub.dart
marandaneto: yeah, sounds good, but then `stacktrace` should not be mandatory right?
also, if the error has a stacktrace and the user still passed a stacktrace, which one wins? we need to think about that, but it doesn't need to be in this PR then

rxlabz: only the throwable is a required argument :
```dart
Future<SentryId> captureException(
    dynamic throwable, {
    dynamic stackTrace,
  })
```
for the if a stacktrace is passed with an error, it will override the error.stacktrace

marandaneto: perfect, we'd need to mention this in the docs later.

dart/lib/src/scope.dart
bruno-garcia: https://github.com/dart-lang/sdk/blob/master/CHANGELOG.md#260---2019-11-05
Completing 1 year:

https://github.com/getsentry/sentry-dart/issues/115

PR #114, opened 2020-10-19
ref: Hub passes the Scope

marandaneto:

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [X] Refactoring

## :scroll: Description
ref: Hub passes the Scope

Client needs to have the Scope in order to prepare the event.
Also a few other improvements I've found on the way

tests

## :pencil: Checklist

- [X] All tests passing
- [ ] No breaking changes


dart/lib/src/hub.dart
```diff
@@ -32,20 +35,18 @@ class Hub {
...
 
   static void _validateOptions(SentryOptions options) {
     if (options == null) {
-      throw ArgumentError.notNull('options');
+      throw ArgumentError.notNull('SentryOptions is required.');
     }
 
-    if (options.dsn == null) {
-      throw ArgumentError.notNull('options.dsn');
+    if (options.dsn == null || options.dsn.isEmpty) {
```
marandaneto: is there a nicer way to test null and empty?

dart/lib/src/noop_hub.dart
```diff
@@ -0,0 +1,51 @@
+import 'dart:async';
+
+import 'package:sentry/src/client.dart';
+import 'package:sentry/src/hub.dart';
+import 'package:sentry/src/protocol/sentry_id.dart';
+import 'package:sentry/src/protocol/level.dart';
+import 'package:sentry/src/protocol/event.dart';
```
marandaneto: IDE does the imports like this but feels weird, whats the proper way?

dart/lib/src/hub.dart
```diff
@@ -74,7 +75,7 @@ class Hub {
         'captureEvent called with null parameter.',
       );
     } else {
-      final item = _stack.first;
+      final item = _peek();
       if (item != null) {
         try {
           sentryId = await item.client.captureEvent(event, scope: item.scope);
```
marandaneto, line 81: @rxlabz I noticed that all the methods are sync now because of `await`

> await: You can use the await keyword to get the completed result of an asynchronous expression. The await keyword only works within an async function.

should we aim for sync calls or keep it async?

@bruno-garcia

dart/lib/src/protocol/sentry_id.dart
```diff
@@ -1,16 +1,16 @@
...
 
 class SentryId {
-  static const String emptyId = '00000000-0000-0000-0000-000000000000';
+  static const String _emptyId = '00000000-0000-0000-0000-000000000000';
 
   /// The ID Sentry.io assigned to the submitted event for future reference.
   final String _id;
 
-  String get id => _id;
+  // TODO: should we generate the new UUID here with an empty ctor?
```
marandaneto, line 9: if so, we'd need to import the `uuid` dependency and make the `_id` of its type

dart/lib/src/hub.dart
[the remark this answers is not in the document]
marandaneto: yeah still feels not so nice, I'd expect something like https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.text/is-null-or-empty.html

dart/lib/src/noop_hub.dart
marandaneto: it didnt complain, thats why

rxlabz: Normally in Dart, relative imports should be preferred https://fuchsia.dev/fuchsia-src/development/languages/dart/style#do_import_all_files_within_a_package_using_relative_paths 
( but I know that the Flutter team doesn't follow all the Dart styleguide :)  )
In IntelliJ, the autocompletion often shows both the relative and `package:` imports.  I don't know if VSCode can be configured is this way.
We could also activate this lint https://dart-lang.github.io/linter/lints/prefer_relative_imports.html which seems to not be activated with `pedantic` to enforce the rule.

dart/lib/src/hub.dart
rxlabz:
```dart
if (options.dsn?.isNotEmpty != true ) { //...}
```
is it nicer ?

dart/lib/src/noop_hub.dart
marandaneto: nice I created a task for it

dart/lib/src/hub.dart
rxlabz: I'm not sure I understand. They are not sync, they return a Future

dart/lib/src/protocol/sentry_id.dart
rxlabz: I maybe missed it, but I didn't found a method to generate a Uuid from String in the uuid package

dart/lib/src/hub.dart
```diff
@@ -162,16 +160,14 @@ class Hub {
...
-            message,
-            level: level,
-            template: template,
-            params: params,
-          );
+          sentryId = await item.client.captureMessage(message,
+              level: level,
+              template: template,
+              params: params,
+              scope: item.scope);
```
rxlabz:
```suggestion
          sentryId = await item.client.captureMessage(message,
              level: level,
              template: template,
              params: params,
              scope: item.scope,
          );
```

here we should typically use a trailing comma

dart/lib/src/noop_hub.dart
```diff
@@ -0,0 +1,51 @@
...
+  @override
+  Future<SentryId> captureException(throwable, {stackTrace}) =>
+      Future.value(SentryId.empty());
+
+  @override
+  Future<SentryId> captureMessage(String message,
+          {SeverityLevel level = SeverityLevel.info,
+          String template,
+          List params}) =>
+      Future.value(SentryId.empty());
```
rxlabz:
```suggestion
  Future<SentryId> captureMessage(
    String message,
    {SeverityLevel level = SeverityLevel.info,
    String template,
    List params,
  }) =>
      Future.value(SentryId.empty());
```

dart/lib/src/sentry.dart
```diff
@@ -6,40 +6,72 @@ import 'client.dart';
...
+  }
+
+  Future<SentryId> captureMessage(
+    String message, {
+    SeverityLevel level,
+    String template,
+    List<dynamic> params,
+  }) async {
+    return currentHub.captureMessage(message,
+        level: level, template: template, params: params);
```
rxlabz: nit : trailing comma

dart/test/hub_test.dart
```diff
@@ -70,13 +70,15 @@ void main() {
...
+      verify(client.captureException(fakeException, scope: anyNamed('scope')))
+          .called(1);
     });
 
     test('should capture message', () {
       hub.captureMessage(fakeMessage.formatted, level: SeverityLevel.info);
       verify(
-        client.captureMessage(fakeMessage.formatted, level: SeverityLevel.info),
+        client.captureMessage(fakeMessage.formatted,
+            level: SeverityLevel.info, scope: anyNamed('scope')),
```
rxlabz: nit: trailing comma

dart/test/sentry_test.dart
```diff
@@ -47,7 +47,8 @@ void main() {
     test('should capture the exception', () async {
       await Sentry.captureException(anException);
       verify(
-        client.captureException(anException, stackTrace: null),
+        client.captureException(anException,
+            stackTrace: null, scope: anyNamed('scope')),
```
rxlabz: nit: trailing comma

dart/lib/src/hub.dart
marandaneto: @rxlabz `item.client.captureEvent` returns a future, but you call this method using `await`
if you use the `await` keyboard, means you the future to be completed, so its a sync call, not async.
or I am not understanding the `await` keyword in Dart, which is also possible

dart/lib/src/protocol/sentry_id.dart
marandaneto: maybe `parse` and `unparse` https://github.com/Daegalus/dart-uuid#uuidparsestring-uuid-list-buffer-null-int-offset-0

dart/lib/src/hub.dart
rxlabz: `await` is just a "syntactic sugar", it doesn't change the nature of the code. Maybe https://dart.dev/codelabs/async-await will be clearer than me.
But I noticed that this method is asynchronous when it calls item.client.captureEvent and synchronous if the hub is not enabled or if the event is null, we probably should change the signature to 
```dart
FutureOr<SentryId> captureEvent(Event event) async {//...}
```

marandaneto: not a big fan because of:

> Note: the FutureOr<T> type is interpreted as dynamic in non strong-mode.

but I got it I guess:
`await item.client.captureEvent` makes a sync call, but the signature of the method
`hub.captureEvent` is also `async` so it's fine.

let's keep as it is till we have a concrete implementation of the transport layer, so we'd know what looks better.

dart/lib/src/protocol/sentry_id.dart
marandaneto: I created a task for it

rxlabz (approved):
LGTM

PR #116, opened 2020-10-20
feat: sentry options

marandaneto:

- [ ] Bugfix
- [X] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
feat: sentry options

users should be able to configure options by mutating it on SDK init.

## :pencil: Checklist

- [X] All tests passing
- [ ] No breaking changes

## :crystal_ball: Next steps
work on the TODOs


dart/lib/src/protocol/level.dart
```diff
@@ -3,6 +3,8 @@ import 'package:meta/meta.dart';
 /// Severity of the logged [Event].
 @immutable
 class SeverityLevel {
+  // TODO: make this as an enum
```
rxlabz: Dart enums are very limited. Here this class is the easiest way to attach a String name, and use it for serialization. If you use an enum you'll need a method with a `switch case`

dart/lib/src/sentry_options.dart
```diff
@@ -47,9 +46,97 @@ class SentryOptions {
...
+  final List<Integration> _integrations = [];
+
+  // TODO: shutdownTimeout, flushTimeoutMillis
+  // https://api.dart.dev/stable/2.10.2/dart-io/HttpClient/close.html doesn't have a timeout param, we'd need to implement manually
+
+  List<Integration> get integrations => List.unmodifiable(_integrations);
+
+  /// Turns debug mode on or off. If debug is enabled SDK will attempt to print out useful debugging
+  /// information if something goes wrong. Default is disabled.
+  bool debug;
```
rxlabz:
```suggestion
  bool debug = false;
```

dart/lib/src/protocol/level.dart
marandaneto: got it, but then I cant do this: https://github.com/getsentry/sentry-java/blob/5ebd49cbfd17eb66175e9247db6d27e4df84a56b/sentry/src/main/java/io/sentry/DiagnosticLogger.java#L37

which is see which level is higher based on the ordinal number

dart/lib/src/sentry_options.dart
marandaneto: but its already `false` by default, right? is it just a convention?

dart/lib/src/sentry_options.dart
```diff
@@ -61,8 +148,53 @@ class SentryOptions {
     Logger logger,
     this.maxBreadcrumbs = 100,
   }) : _logger = logger;
```
rxlabz:
```suggestion
  }) {
    _logger = logger ?? _logger;
  }
```

rxlabz: this and `bool debug = false` fix the tests. Here if no logger is passed `logger` is set to `null`

marandaneto: damn I missed this on the ctor, I removed from the ctor, they are no op by default, thanks

dart/lib/src/protocol/level.dart
rxlabz: we'll need to add an int argument
```dart
int level;

const SeverityLevel._(this.name, this.level);
```

dart/lib/src/sentry_options.dart
rxlabz: no... for now Dart bools are ... null :s

dart/lib/src/protocol/level.dart
marandaneto: yeah sure, if enum doesn't give this for free, it won't be any useful having an enum

marandaneto: looks like an enum has an `index` exactly for these cases.
https://www.tutorialspoint.com/dart_programming/dart_programming_enumeration.htm

@bruno-garcia opinions? I like strongly typed things but I didn't play enough with Dart yet.
I'm also fine with the code snippet above

rxlabz: yes enum have index, but you still will need a switch case for the name. It can be defined as an enum extension... but only with Dart 2.7+
IMHO an enum will need more code and will not be "more typed"

rxlabz (approved):
LGTM

dart/lib/src/protocol/level.dart
marandaneto: an index it is, thx for the feedback

PR #119, opened 2020-10-20
ref: SentryId generates UUID

marandaneto:

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [X] Refactoring

## :scroll: Description
ref: SentryId generates UUID

we don't need to have a reference of the Uuid generator everywhere, and still, it's testable.

## :pencil: Checklist

- [X] All tests passing
- [ ] No breaking changes

### DIFF
--- a/dart/lib/src/protocol/event.dart
+++ b/dart/lib/src/protocol/event.dart
@@ -9,6 +9,7 @@ import '../version.dart';
 class Event {
   /// Creates an event.
   const Event({
+    this.eventId,
     this.loggerName,
     this.serverName,
     this.release,
@@ -33,6 +34,9 @@ class Event {
   /// You do not need to specify this value unless you supplement the default
   /// fingerprint with custom fingerprints.
   static const String defaultFingerprint = '{{ default }}';
+
+  /// The ID Sentry.io assigned to the submitted event for future reference.
+  final SentryId eventId;
 
   /// The logger that logged the event.
   final String loggerName;
@@ -117,6 +121,7 @@ class Event {
   final Sdk sdk;
 
   Event copyWith({
+    SentryId eventId,
     String loggerName,
     String serverName,
     String release,
@@ -136,6 +141,7 @@ class Event {
     Sdk sdk,
   }) =>
       Event(
+        eventId: eventId ?? this.eventId,
         loggerName: loggerName ?? this.loggerName,
         serverName: serverName ?? this.serverName,
         release: release ?? this.release,
@@ -161,6 +167,10 @@ class Event {
     final json = <String, dynamic>{
       'platform': sdkPlatform,
     };
+
+    if (eventId != null) {
+      json['event_id'] = eventId;
+    }
 
     if (loggerName != null) {
       json['logger'] = loggerName;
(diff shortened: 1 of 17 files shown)


dart/lib/src/protocol/event.dart
```diff
@@ -9,6 +9,7 @@ import '../version.dart';
 class Event {
   /// Creates an event.
   const Event({
+    this.eventId,
```
marandaneto: @bruno-garcia @rxlabz the problem here is:

I can't do:
`this.eventId = SentryId.newId(),`
a optional default param must be a constant.

also I cannot do `final SentryId eventId = SentryId.newId();` on the field declaration.

then the class cant be immutable (const ctor) as it is inited with a non-constant value.

I don't want to make it immutable though.

ideas?

rxlabz: I think you can remove the `const` ctor. All the fields are final + `@immutable` are enough. Dart `const` are particular.

> "const" has a meaning that's a bit more complex and subtle in Dart.

from https://news.dartlang.org/2012/06/const-static-final-oh-my.html

marandaneto: not really, I've tried already:

the default value of an optional parameter must be constant.dartnon_constant_default_value

marandaneto: I found this:

```
class ClassB {
  final ClassA classA;
  ClassB({this.classA = const ClassA()}); // <- creation is const
}
```

rxlabz: this should do the job :
```dart
Event({
    SentryId eventId ,
    this.loggerName,
    this.serverName,
    this.release,
    this.environment,
    this.message,
    this.transaction,
    this.exception,
    this.stackTrace,
    this.level,
    this.culprit,
    this.tags,
    this.extra,
    this.fingerprint,
    this.userContext,
    this.contexts,
    this.breadcrumbs,
    this.sdk,
  }):eventId = SentryId.newId();
```

marandaneto: if I do:

`this.eventId = const SentryId.internal(''),`

`const SentryId.internal(this._id);` it'd work but then we'd run the same problem on SentryId class

rxlabz: this should work
```dart
Event({
    SentryId eventId,
    this.loggerName,
    this.serverName,
    this.release,
    this.environment,
    this.message,
    this.transaction,
    this.exception,
    this.stackTrace,
    this.level,
    this.culprit,
    this.tags,
    this.extra,
    this.fingerprint,
    this.userContext,
    this.contexts,
    this.breadcrumbs,
    this.sdk,
  }) : eventId = eventId ?? SentryId.newId();
```

rxlabz: strangely the analyzer prefers this to `: this.eventId = eventId ?? SentryId.newId();` 🤷‍♂️

marandaneto: mmmm I didn't know about the mix of syntactic sugar + named params, nice

bruno-garcia: Would it work to drop `this`, declare the field, add a body to the ctor and add?

`this.eventId = eventId ?? SentryId();`

marandaneto: solved doing this: https://github.com/getsentry/sentry-dart/pull/119#discussion_r508693328

thank you both

PR #121, opened 2020-10-21
ref: Event now is SentryEvent and added GPU

marandaneto:

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [X] Refactoring

## :scroll: Description
ref: Event now is SentryEvent and added GPU

SDKs call an event as SentryEvent, loggerName is now logger, timestamp should be Event's timestamp, added missing GPU

## :pencil: Checklist

- [X] All tests passing
- [ ] No breaking changes

## :crystal_ball: Next steps
TODOs

### DIFF
--- a/dart/lib/src/sentry_options.dart
+++ b/dart/lib/src/sentry_options.dart
@@ -34,12 +34,9 @@ class SentryOptions {
 
   /// If [clock] is provided, it is used to get time instead of the system
   /// clock. This is useful in tests. Should be an implementation of [ClockProvider].
-  /// This parameter is dynamic to maintain backwards compatibility with
-  /// previous use of [Clock](https://pub.dartlang.org/documentation/quiver/latest/quiver.time/Clock-class.html)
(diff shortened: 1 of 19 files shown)


dart/lib/src/sentry_options.dart
```diff
@@ -138,8 +142,14 @@ class SentryOptions {
...
     this.compressPayload,
     this.httpClient,
-    this.clock,
-  });
+    dynamic clock,
+  }) {
+    if (clock == null) {
+      clock = getUtcDateTime;
+    } else {
+      _clock = (clock is ClockProvider ? clock : clock.get) as ClockProvider;
```
marandaneto: @rxlabz can we deprecate one of them or is this due to Dart version?

rxlabz: IMO we totally can type `clock` as ClockProvider.

rxlabz (approved):

PR #122, opened 2020-10-21
feat: before breadcrumb and scope ref

marandaneto:

- [ ] Bugfix
- [X] New feature
- [ ] Enhancement
- [X] Refactoring

## :scroll: Description
feat: before breadcrumb and scope ref

One should be able to filter breadcrumbs on before breadcrumb callback and also add event processors to the scope

## :pencil: Checklist

- [X] All tests passing
- [ ] No breaking changes

## :crystal_ball: Next steps
Contexts is still missing

### DIFF
--- a/dart/lib/src/scope.dart
+++ b/dart/lib/src/scope.dart
@@ -53,7 +53,7 @@ class Scope {
 
   Map<String, dynamic> get extra => Map.unmodifiable(_extra);
 
-  // TODO: EventProcessors, Contexts, BeforeBreadcrumbCallback, Breadcrumb Hint, clone
+  // TODO: Contexts
 
   /// Scope's event processor list
   final List<EventProcessor> _eventProcessors = [];
--- a/dart/lib/src/scope.dart
+++ b/dart/lib/src/scope.dart
@@ -6,32 +6,14 @@ import 'sentry_options.dart';
 /// Scope data to be sent with the event
 class Scope {
   /// How important this event is.
-  SentryLevel _level;
-
-  SentryLevel get level => _level;
-
-  set level(SentryLevel level) {
-    _level = level;
-  }
+  SentryLevel level;
 
   /// The name of the transaction which generated this event,
   /// for example, the route name: `"/users/<username>/"`.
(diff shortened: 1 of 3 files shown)


dart/lib/src/scope.dart
```diff
@@ -153,6 +161,10 @@ class Scope {
       clone.addBreadcrumb(breadcrumb);
     }
 
+    for (final eventProcessor in _eventProcessors) {
```
bruno-garcia, line 164: Since we're in the class, what about a private constructor or factory method that takes all fields as named args?
Then we could just clone all lists etc like `Scope({eventProcessors = [...this.eventprocessors])` and avoid tall these `for`

dart/lib/src/scope.dart
```diff
@@ -100,15 +99,23 @@ class Scope {
...
-    _level = null;
-    _transaction = null;
-    _user = null;
+    level = null;
+    transaction = null;
+    user = null;
     _fingerprint = null;
     _tags.clear();
     _extra.clear();
+    _eventProcessors.clear();
```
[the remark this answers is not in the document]
marandaneto: good catch, I've tested the addEventProcessor only

dart/lib/src/scope.dart
marandaneto: mm yeah but rethinking this right now, maybe it'd make sense to just do a copyWith and follow the convention? tbh I just added the missing field in the existent method, I will check whats the state of the art for cloning objects on Dart.
Factory is public afaik, but a private ctor would work as well

marandaneto: https://pub.dev/documentation/copyable/latest/#examples
using the copyable interface seems to be the standard

marandaneto: I will add this as a task on our project, so I can merge this now

rxlabz: copyWith is probably the "most standard way" ( used a lot in Flutter source code, and in https://pub.dev/packages/freezed, a popular immutable package)

PR #123, opened 2020-10-21
Refacto : add a Transport class

rxlabz:

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring

## :scroll: Description
send the event from a dedicated class

updated existing tests

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/dart/lib/src/browser_client.dart
+++ b/dart/lib/src/browser_client.dart
@@ -9,7 +9,6 @@ import 'package:http/browser_client.dart';
 import 'package:meta/meta.dart';
 
 import 'client.dart';
-import 'protocol.dart';
 import 'sentry_options.dart';
 import 'transport/transport.dart';
 import 'version.dart';
@@ -50,7 +49,7 @@ class SentryBrowserClient extends SentryClient {
           options,
           transport: Transport(
             options: options,
-            sdk: Sdk(name: browserSdkName, version: sdkVersion),
+            sdkIdentifier: '${browserSdkName}/${sdkVersion}',
             platform: platform,
             origin: origin,
           ),
--- a/dart/lib/src/browser_client.dart
+++ b/dart/lib/src/browser_client.dart
@@ -10,7 +10,6 @@ import 'package:meta/meta.dart';
 
 import 'client.dart';
 import 'sentry_options.dart';
-import 'transport/transport.dart';
 import 'version.dart';
 
 SentryClient createSentryClient(SentryOptions options) =>
@@ -32,7 +31,6 @@ class SentryBrowserClient extends SentryClient {
     options.httpClient ??= BrowserClient();
 
     // origin is necessary for sentry to resolve stacktrace
-    origin ??= '${window.location.origin}/';
 
     return SentryBrowserClient._(
       options,
@@ -45,13 +43,5 @@ class SentryBrowserClient extends SentryClient {
     SentryOptions options, {
     String origin,
     @required String platform,
-  }) : super.base(
-          options,
-          transport: Transport(
-            options: options,
-            sdkIdentifier: '${browserSdkName}/${sdkVersion}',
-            platform: platform,
-            origin: origin,
-          ),
-        );
+  }) : super.base(options, origin: '${window.location.origin}/');
 }
--- a/dart/lib/src/browser_client.dart
+++ b/dart/lib/src/browser_client.dart
@@ -42,15 +42,14 @@ class SentryBrowserClient extends SentryClient {
     );
   }
 
-  SentryBrowserClient._(SentryOptions options,
-      {String origin, @required String platform})
-      : super.base(
+  SentryBrowserClient._(
+    SentryOptions options, {
+    String origin,
+    @required String platform,
+  }) : super.base(
           options,
           transport: Transport(
-            dsn: options.dsn,
-            httpClient: options.httpClient,
-            clock: options.clock,
-            compressPayload: false,
+            options: options,
             sdk: Sdk(name: browserSdkName, version: sdkVersion),
             platform: platform,
             origin: origin,
--- a/dart/lib/src/browser_client.dart
+++ b/dart/lib/src/browser_client.dart
@@ -54,7 +54,7 @@ class SentryBrowserClient extends SentryClient {
             sdk: Sdk(name: browserSdkName, version: sdkVersion),
             headersBuilder: SentryClient.buildHeaders,
             platform: platform,
+            origin: origin,
           ),
-          origin: origin,
         );
 }
--- a/dart/lib/src/browser_client.dart
+++ b/dart/lib/src/browser_client.dart
@@ -8,7 +8,9 @@ import 'dart:html' show window;
 import 'package:http/browser_client.dart';
 
 import 'client.dart';
+import 'protocol.dart';
 import 'sentry_options.dart';
+import 'version.dart';
 
 SentryClient createSentryClient(SentryOptions options) =>
     SentryBrowserClient(options);
@@ -25,15 +27,15 @@ class SentryBrowserClient extends SentryClient {
   ///
   /// If [httpClient] is provided, it is used instead of the default client to
   /// make HTTP calls to Sentry.io. This is useful in tests.
-  factory SentryBrowserClient(SentryOptions options, {String origin}) {
+  factory SentryBrowserClient(SentryOptions options) {
     options.httpClient ??= BrowserClient();
 
+    options.sdk ??= Sdk(name: sdkName, version: sdkVersion);
+
     // origin is necessary for sentry to resolve stacktrace
-    return SentryBrowserClient._(options, origin: origin);
+    return SentryBrowserClient._(options);
   }
 
-  SentryBrowserClient._(
-    SentryOptions options, {
-    String origin,
-  }) : super.base(options, origin: '${window.location.origin}/');
+  SentryBrowserClient._(SentryOptions options)
+      : super.base(options, origin: '${window.location.origin}/');
 }
--- a/dart/lib/src/browser_client.dart
+++ b/dart/lib/src/browser_client.dart
@@ -3,14 +3,15 @@
 // found in the LICENSE file.
 
 /// A pure Dart client for Sentry.io crash reporting.
-import 'dart:convert';
 import 'dart:html' show window;
 
 import 'package:http/browser_client.dart';
+import 'package:meta/meta.dart';
 
 import 'client.dart';
 import 'protocol.dart';
 import 'sentry_options.dart';
+import 'transport/transport.dart';
 import 'version.dart';
 
 SentryClient createSentryClient(SentryOptions options) =>
@@ -41,19 +42,19 @@ class SentryBrowserClient extends SentryClient {
     );
   }
 
-  SentryBrowserClient._(SentryOptions options, {String origin, String platform})
+  SentryBrowserClient._(SentryOptions options,
+      {String origin, @required String platform})
       : super.base(
           options,
+          transport: Transport(
+            dsn: options.dsn,
+            httpClient: options.httpClient,
+            clock: options.clock,
+            compressPayload: false,
+            sdk: Sdk(name: browserSdkName, version: sdkVersion),
+            headersBuilder: SentryClient.buildHeaders,
+            platform: platform,
+          ),
           origin: origin,
-          sdk: Sdk(name: browserSdkName, version: sdkVersion),
-          platform: platform,
         );
-
-  @override
-  List<int> bodyEncoder(
-    Map<String, dynamic> data,
-    Map<String, String> headers,
-  ) =>
-      // Gzip compression is implicit on browser
-      utf8.encode(json.encode(data));
 }
--- a/dart/lib/src/browser_client.dart
+++ b/dart/lib/src/browser_client.dart
@@ -6,11 +6,9 @@
 import 'dart:html' show window;
 
 import 'package:http/browser_client.dart';
-import 'package:meta/meta.dart';
 
 import 'client.dart';
 import 'sentry_options.dart';
-import 'version.dart';
 
 SentryClient createSentryClient(SentryOptions options) =>
     SentryBrowserClient(options);
@@ -31,17 +29,11 @@ class SentryBrowserClient extends SentryClient {
     options.httpClient ??= BrowserClient();
 
     // origin is necessary for sentry to resolve stacktrace
-
-    return SentryBrowserClient._(
-      options,
-      origin: origin,
-      platform: browserPlatform,
-    );
+    return SentryBrowserClient._(options, origin: origin);
   }
 
   SentryBrowserClient._(
     SentryOptions options, {
     String origin,
-    @required String platform,
   }) : super.base(options, origin: '${window.location.origin}/');
 }
--- a/dart/lib/src/browser_client.dart
+++ b/dart/lib/src/browser_client.dart
@@ -52,7 +52,6 @@ class SentryBrowserClient extends SentryClient {
             clock: options.clock,
             compressPayload: false,
             sdk: Sdk(name: browserSdkName, version: sdkVersion),
-            headersBuilder: SentryClient.buildHeaders,
             platform: platform,
             origin: origin,
           ),
--- a/dart/lib/src/client.dart
+++ b/dart/lib/src/client.dart
@@ -166,7 +166,4 @@ abstract class SentryClient {
     }
     return event;
   }
-
-  @override
-  String toString() => '$SentryClient("${options.dsn}")';
 }
--- a/dart/lib/src/client.dart
+++ b/dart/lib/src/client.dart
@@ -154,10 +154,14 @@ abstract class SentryClient {
 
       // TODO add tests
       // Merge the scope tags.
-      event = event.copyWith(tags: scope.tags..addAll(event.tags ?? {}));
+      event = event.copyWith(
+          tags: scope.tags.map((key, value) => MapEntry(key, value))
+            ..addAll(event.tags ?? {}));
 
       // Merge the scope extra.
-      event = event.copyWith(extra: scope.extra..addAll(event.extra ?? {}));
+      event = event.copyWith(
+          extra: scope.extra.map((key, value) => MapEntry(key, value))
+            ..addAll(event.extra ?? {}));
 
       // Merge the scope level.
       if (event.level == null) {
--- a/dart/lib/src/client.dart
+++ b/dart/lib/src/client.dart
@@ -48,6 +48,12 @@ abstract class SentryClient {
 
     event = _applyScope(event: event, scope: scope);
 
+    event = event.copyWith(
+      serverName: _options.serverName,
+      environment: _options.environment,
+      release: _options.release,
+    );
+
     return _options.transport.send(event);
   }
 
--- a/dart/lib/src/client.dart
+++ b/dart/lib/src/client.dart
@@ -21,7 +21,6 @@ abstract class SentryClient {
 
   SentryClient.base(
     this.options, {
-    this.origin,
     @required this.transport,
   });
 
@@ -43,9 +42,6 @@ abstract class SentryClient {
   /// * https://docs.sentry.io/learn/context/#capturing-the-user
   User userContext;
 
-  /// Use for browser stacktrace
-  String origin;
-
   /// Reports an [event] to Sentry.io.
   Future<SentryId> captureEvent(
     SentryEvent event, {
@@ -53,6 +49,8 @@ abstract class SentryClient {
     Scope scope,
   }) async {
     event = _processEvent(event, eventProcessors: options.eventProcessors);
+
+    event = _applyScope(event: event, scope: scope);
 
     final data = <String, dynamic>{
       'event_id': event.eventId.toString(),
@@ -62,16 +60,10 @@ abstract class SentryClient {
       mergeAttributes(options.environmentAttributes.toJson(), into: data);
     }
 
-    // Merge the user context.
-    if (userContext != null) {
-      mergeAttributes(<String, dynamic>{'user': userContext.toJson()},
-          into: data);
-    }
-
     mergeAttributes(
       event.toJson(
         stackFrameFilter: stackFrameFilter,
-        origin: origin,
+        origin: transport.origin,
       ),
       into: data,
     );
@@ -124,6 +116,46 @@ abstract class SentryClient {
   }) {
     for (final processor in eventProcessors) {
       event = processor(event, hint);
+    }
+    return event;
+  }
+
+  Event _applyScope({@required Event event, @required Scope scope}) {
+    if (scope != null) {
+      // Merge the scope transaction.
+      if (event.transaction == null) {
+        event = event.copyWith(transaction: scope.transaction);
+      }
+
+      // Merge the user context.
+      if (event.userContext == null) {
+        event = event.copyWith(userContext: scope.user);
+      }
+
+      // Merge the scope fingerprint.
+      if (event.fingerprint == null) {
+        event = event.copyWith(fingerprint: scope.fingerprint);
+      }
+
+      // Merge the scope breadcrumbs.
+      if (event.breadcrumbs == null) {
+        event = event.copyWith(breadcrumbs: scope.breadcrumbs);
+      }
+
+      // Merge the scope tags.
+      if (event.tags == null) {
+        event = event.copyWith(tags: scope.tags);
+      }
+
+      // Merge the scope extra.
+      if (event.extra == null) {
+        event = event.copyWith(extra: scope.extra);
+      }
+
+      // Merge the scope level.
+      if (event.level == null) {
+        event = event.copyWith(level: scope.level);
+      }
     }
     return event;
   }
--- a/dart/lib/src/client.dart
+++ b/dart/lib/src/client.dart
@@ -94,11 +94,7 @@ abstract class SentryClient {
     Scope scope,
   }) {
     final event = SentryEvent(
-      message: Message(
-        formatted,
-        template: template,
-        params: params,
-      ),
+      message: Message(formatted, template: template, params: params),
       level: level,
       timestamp: options.clock(),
     );
@@ -115,7 +111,18 @@ abstract class SentryClient {
     List<EventProcessor> eventProcessors,
   }) {
     for (final processor in eventProcessors) {
-      event = processor(event, hint);
+      try {
+        event = processor(event, hint);
+      } catch (err) {
+        options.logger(
+          SentryLevel.error,
+          'An exception occurred while processing event by a processor : $err',
+        );
+      }
+      if (event == null) {
+        options.logger(SentryLevel.debug, 'Event was dropped by a processor');
+        break;
+      }
     }
     return event;
   }
--- a/dart/lib/src/client.dart
+++ b/dart/lib/src/client.dart
@@ -9,7 +9,6 @@ import 'client_stub.dart'
 import 'protocol.dart';
 import 'stack_trace.dart';
 import 'transport/transport.dart';
-import 'utils.dart';
 
 /// Logs crash reports and events to the Sentry.io service.
 abstract class SentryClient {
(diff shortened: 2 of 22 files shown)


dart/lib/src/browser_client.dart
```diff
@@ -41,19 +42,19 @@ class SentryBrowserClient extends SentryClient {
...
       : super.base(
           options,
-          origin: origin,
-          sdk: Sdk(name: browserSdkName, version: sdkVersion),
-          platform: platform,
+          transport: Transport(
+            dsn: options.dsn,
+            httpClient: options.httpClient,
+            clock: options.clock,
+            compressPayload: false,
```
marandaneto:
```suggestion
            compressPayload: options.compressPayload,
```

dart/lib/src/browser_client.dart
```diff
@@ -41,19 +42,19 @@ class SentryBrowserClient extends SentryClient {
...
+            sdk: Sdk(name: browserSdkName, version: sdkVersion),
```
marandaneto: `options.sdkVersion` already exists, but its probably null, we'd need to find a way to instantiate it based on the platform (dart, flutter or web), but ideally, we just read this field from the options here

dart/lib/src/browser_client.dart
```diff
@@ -41,19 +42,19 @@ class SentryBrowserClient extends SentryClient {
...
+            headersBuilder: SentryClient.buildHeaders,
```
marandaneto: this could be done inside of the Transport layer I guess, why does the client need to know about the headers?

dart/lib/src/browser_client.dart
```diff
@@ -41,19 +42,19 @@ class SentryBrowserClient extends SentryClient {
...
+            platform: platform,
```
marandaneto: wondering if this is the same as `options.sentryClientName`, could you check it?

dart/lib/src/client.dart
```diff
@@ -22,36 +21,14 @@ abstract class SentryClient {
...
-  /// The Sentry.io secret key for the project.
   @visibleForTesting
-  // ignore: invalid_use_of_visible_for_testing_member
-  String get secretKey => _dsn.secretKey;
-
-  /// The ID issued by Sentry.io to your project.
-  ///
-  /// Attached to the event payload.
-  String get projectId => _dsn.projectId;
+  final Transport transport;
```
marandaneto: ideally the transport is read from the `options. transport ` so we don't need to `visibleForTesting`, and we only instantiate a new `transport` here if the current `optins.transport` is a NoOp.
That said, we need a NoOpTransport: https://github.com/getsentry/sentry-java/blob/main/sentry/src/main/java/io/sentry/transport/NoOpTransport.java

let's ignore the `retryAfter` for now (https://github.com/getsentry/sentry-dart/projects/1#card-47870442)

dart/lib/src/browser_client.dart
rxlabz: as the payload compression is not available in browser should we let the choice to the user ?

dart/lib/src/client.dart
```diff
@@ -188,15 +109,63 @@ abstract class SentryClient {
...
 
-  @override
-  String toString() => '$SentryClient("$postUri")';
+  SentryEvent _processEvent(
+    SentryEvent event, {
+    dynamic hint,
+    List<EventProcessor> eventProcessors,
+  }) {
+    for (final processor in eventProcessors) {
+      event = processor(event, hint);
```
marandaneto: if one of the processors returned null, means that this is event is dropped, so we should bail out and log it.
eg https://github.com/getsentry/sentry-java/blob/main/sentry/src/main/java/io/sentry/SentryClient.java#L155-L184

marandaneto: I can help with those things on a different PR though

dart/lib/src/browser_client.dart
marandaneto: ok, I missed that, if the browser client doesn't support that, we should not give an option for sure, but I feel this should be documented in the `compressPayload` field declaration then.

dart/lib/src/client.dart
```diff
@@ -188,15 +109,63 @@ abstract class SentryClient {
...
+      // Merge the scope extra.
+      if (event.extra == null) {
+        event = event.copyWith(extra: scope.extra);
+      }
+
+      // Merge the scope level.
+      if (event.level == null) {
+        event = event.copyWith(level: scope.level);
+      }
+    }
```
marandaneto: for some fields, we actually append items from the Scope to the event, see:
https://github.com/getsentry/sentry-java/blob/main/sentry/src/main/java/io/sentry/SentryClient.java#L281-L335

dart/lib/src/client.dart
```diff
@@ -188,15 +109,63 @@ abstract class SentryClient {
...
+    return event;
+  }
+
+  @override
+  String toString() => '$SentryClient("${options.dsn}")';
```
marandaneto: mm not against it, but I don't see how this is useful :)

dart/lib/src/transport/transport.dart
```diff
@@ -0,0 +1,125 @@
...
+
+  final bool compressPayload;
+
+  /// Used by sentry to differentiate browser from io environment
+  final String platform;
+
+  final ClockProvider _clock;
+
+  /// Use for browser stacktrace
+  final String origin;
```
marandaneto: most of those fields if not all are already in the options, should it not just take options and read from there?

dart/lib/src/browser_client.dart
rxlabz: is `compressPayload` really an option here ? 🤔
Could you choose to send a non compressed payload if you're on a platform which supports `gzip` ?

dart/lib/src/transport/transport.dart
```diff
@@ -0,0 +1,125 @@
...
+    String apiPath;
+    if (pathLength > 1) {
+      // some paths would present before the projectID in the dsnUri
+      apiPath =
+          (dsnUri.pathSegments.sublist(0, pathLength - 1) + ['api']).join('/');
+    } else {
+      apiPath = 'api';
+    }
+    return '${dsnUri.scheme}://${dsnUri.host}$port/$apiPath/$projectId/store/';
+  }
```
marandaneto: does it make sense to move this to the `Dsn` class like https://github.com/getsentry/sentry-java/blob/main/sentry/src/main/java/io/sentry/Dsn.java

ideally, Dsn would offer atributes like `dsn.projectId`, etc...

dart/lib/src/transport/transport.dart
```diff
@@ -0,0 +1,125 @@
...
+
+  Future<SentryId> send(Map<String, dynamic> data) async {
```
marandaneto: our Transport still takes a typed object, `SentryEvent` in the case, and only here in the transport we'd convert to a json-like object
https://github.com/getsentry/sentry-java/blob/main/sentry/src/main/java/io/sentry/transport/ITransport.java#L11

dart/lib/src/browser_client.dart
rxlabz: we could use a conditional import in `transport.dart` to import the bodyEncoder strategy ? and let it be a transport "hidden" implementation details ?

dart/lib/src/transport/transport.dart
```diff
@@ -0,0 +1,125 @@
...
+    final now = _clock();
+    var authHeader = 'Sentry sentry_version=6, sentry_client=$clientId, '
+        'sentry_timestamp=${now.millisecondsSinceEpoch}, sentry_key=$publicKey';
+    if (secretKey != null) {
+      authHeader += ', sentry_secret=$secretKey';
+    }
+
+    mergeAttributes(_getContext(now), into: data);
+
+    final headers = headersBuilder(authHeader);
```
marandaneto: on Java we have an abstraction for this, so we only do it once, and not on every call to `send(...)`, see: https://github.com/getsentry/sentry-java/blob/main/sentry/src/main/java/io/sentry/CredentialsSettingConfigurator.java
could we create such abstraction or at least build it only once on the transport? I think the only moving part here is the `sentry_timestamp`

dart/lib/src/browser_client.dart
rxlabz: I forgot that it's already a conditional import in `transport.dart`. the question is : should we hide `compressPayload ` this from the public API and set it based on the current platform ?

dart/lib/src/transport/transport.dart
```diff
@@ -0,0 +1,125 @@
...
+          (dsnUri.pathSegments.sublist(0, pathLength - 1) + ['api']).join('/');
+    } else {
+      apiPath = 'api';
+    }
+    return '${dsnUri.scheme}://${dsnUri.host}$port/$apiPath/$projectId/store/';
+  }
+
+  Future<SentryId> send(Map<String, dynamic> data) async {
+    final now = _clock();
+    var authHeader = 'Sentry sentry_version=6, sentry_client=$clientId, '
```
marandaneto: latest sentry_version is 7, but we'd need to check if there are differences, I've created a separated issue for that
https://github.com/getsentry/sentry-dart/projects/1#card-47871908

dart/lib/src/browser_client.dart
marandaneto: not really, for flutter, they still can opt in-out of the compressPayload, lets not make this limitation.
lets keep uncompressed for web, and later we fix this too https://github.com/getsentry/sentry-dart/projects/1#card-47870976

dart/lib/src/browser_client.dart
rxlabz: updated  : now it uses a conditional platform in transport.

dart/lib/src/transport/transport.dart
rxlabz: Updated : the transport is now instantiated from a SentryOptions

dart/lib/src/transport/transport.dart
rxlabz: updated : all this logic is now in the `Dsn`class

dart/lib/src/browser_client.dart
rxlabz: it's not an option in Java, is it ? https://github.com/getsentry/sentry-java/blob/67669e1c6d727af618424b7542e35db5c7797c54/sentry/src/main/java/io/sentry/transport/HttpTransport.java#L254-L256 
As Flutter web needs the same behavior than other Dart web projects, I don't really understand why the user could set it ?

marandaneto: its not, but it should, we have this in our Java backlog

dart/lib/src/client.dart
rxlabz: ✅  fixed

dart/lib/src/client.dart
rxlabz: ✅  fixed

dart/lib/src/client.dart
rxlabz: I don't know :) until know the postUri was returned... what kind of infos should we return here ? Do we need it ?

marandaneto: I dont think we need to overwrite the toString for this tbh, makes more sense for data classes only IMO

dart/lib/src/transport/transport.dart
rxlabz: ✅  fixed

dart/lib/src/transport/transport.dart
rxlabz: ✅  I added a CredentialBuilder

dart/lib/src/browser_client.dart
bruno-garcia: Could just be a no-op for Web (Dart or Flutter), since it doesn't apply.
Conditionally having the option will make it harder for users to consume the options since they'll need to conditionally set a value depending on the platform, feels like lots of friction

dart/lib/src/browser_client.dart
bruno-garcia: on a second thought, shouldn't this be just either: `sentry.dart` or `sentry.flutter`? Given that this is what the user installed at the end?

dart/lib/src/browser_client.dart
marandaneto: a condition for importing dart or web http packages already exist, so the only difference here is if they use `web`, and set  `compressPayload=true` won't have any effect till we implement it

dart/lib/src/browser_client.dart
marandaneto: if we need only the identifier of the client, yes, only a String would be fine.

dart/lib/src/client.dart
```diff
@@ -188,48 +86,69 @@ abstract class SentryClient {
...
+
+      // Merge the scope extra.
+      event = event.copyWith(
+          extra: scope.extra.map((key, value) => MapEntry(key, value))
+            ..addAll(event.extra ?? {}));
+
+      // Merge the scope level.
+      if (event.level == null) {
+        event = event.copyWith(level: scope.level);
+      }
```
marandaneto: Scope level has precedence here:
https://github.com/getsentry/sentry-java/blob/main/sentry/src/main/java/io/sentry/SentryClient.java#L327-L330

dart/lib/src/transport/transport.dart
```diff
@@ -0,0 +1,134 @@
...
+      dsn: Dsn.parse(options.dsn),
+      clientId: sdk.identifier,
+    );
+  }
+
+  Future<SentryId> send(SentryEvent event) async {
+    final now = _options.clock();
+
+    var authHeader = _credentialBuilder.build(now.millisecondsSinceEpoch);
+    final headers = buildHeaders(authHeader, sdk: sdk);
```
marandaneto: this could be assigned to a private global field on the ctor body, so we don't need to do on every event

dart/lib/src/transport/transport.dart
```diff
@@ -0,0 +1,134 @@
...
+    };
+
+    if (_options.environmentAttributes != null) {
+      mergeAttributes(_options.environmentAttributes.toJson(), into: data);
+    }
+
+    mergeAttributes(
+      event.toJson(origin: origin),
+      into: data,
+    );
```
marandaneto: this should be the SentryClient's job if I got it right, the transport is only responsible to send it, the event should be fully built and not modified at this point.

marandaneto: I missed the selection for the `_getContext` method here but it's pretty much the same

dart/lib/src/browser_client.dart
bruno-garcia: I'm aware, but there's no gzip on web, the browser will gzip or not and we don't control it

marandaneto: if this is not controlled on the client, totally in for NoOp, I thought this was a limitation because we didn't implement it

dart/lib/src/noop_client.dart
```diff
@@ -21,17 +22,7 @@ class NoOpSentryClient implements SentryClient {
...
-  @override
-  List<int> bodyEncoder(
-    Map<String, dynamic> data,
-    Map<String, String> headers,
-  ) =>
-      [];
-
-  @override
-  Map<String, String> buildHeaders(String authHeader) => {};
+  Transport transport;
```
marandaneto: why is this necessary? one should not care about the `transport` when implementing `SentryClient`, transport comes from `options`

dart/lib/src/client.dart
```diff
@@ -188,48 +86,69 @@ abstract class SentryClient {
...
+        );
+      }
+      if (event == null) {
+        options.logger(SentryLevel.debug, 'Event was dropped by a processor');
+        break;
+      }
     }
-
-    return headers;
+    return event;
```
marandaneto, line 120: still missing the execution of the beforeSendCallback ->
https://github.com/getsentry/sentry-java/blob/5ebd49cbfd17eb66175e9247db6d27e4df84a56b/sentry/src/main/java/io/sentry/SentryClient.java#L97-L103

dart/lib/src/client.dart
marandaneto: also, missing Contexts, created a card for that https://github.com/getsentry/sentry-dart/projects/1#card-47942441
it doesn't need to be on this PR

dart/lib/src/browser_client.dart
marandaneto: we will address this on other PR https://github.com/getsentry/sentry-dart/projects/1#card-47942558

dart/lib/src/client.dart
```diff
@@ -19,38 +17,13 @@ abstract class SentryClient {
...
-    this.options, {
-    String platform,
-    this.origin,
-    Sdk sdk,
-  })  : _dsn = Dsn.parse(options.dsn),
-        _platform = platform ?? sdkPlatform,
-        sdk = sdk ?? Sdk(name: sdkName, version: sdkVersion);
-
-  final Dsn _dsn;
+  SentryClient.base(this.options, {@required this.transport});
```
marandaneto: `options` already have a `transport` field, lets take from the `options` instead of passing as named param.

dart/lib/src/client.dart
marandaneto: we can remove this, even for testing we can set a mock/fake on the options

dart/lib/src/client.dart
```diff
@@ -64,88 +37,22 @@ abstract class SentryClient {
...
-      mergeAttributes(<String, dynamic>{'user': userContext.toJson()},
-          into: data);
-    }
-
-    mergeAttributes(
-      event.toJson(
-        origin: origin,
-      ),
-      into: data,
+    event = event.copyWith(
```
marandaneto, line 48: projectId is not really necessaary, its part of the DSN URL
https://develop.sentry.dev/sdk/event-payloads/

dart/lib/src/io_client.dart
```diff
@@ -2,60 +2,29 @@
...
-    if (options.compressPayload) {
-      headers['Content-Encoding'] = 'gzip';
-      body = gzip.encode(body);
-    }
-    return body;
-  }
+  SentryIOClient._(SentryOptions options, {@required String platform})
+      : super.base(
+          options,
+          transport: Transport(options: options, sdk: sdk, platform: platform),
```
marandaneto: https://github.com/getsentry/sentry-java/blob/5ebd49cbfd17eb66175e9247db6d27e4df84a56b/sentry/src/main/java/io/sentry/SentryClient.java#L43-L47

marandaneto: we set the `transport` to the `options` and then pass the `options` along, so `transport` wont be null

marandaneto: just to remember: we might be able to identify if its web or not on runtime via https://api.flutter.dev/flutter/foundation/kIsWeb-constant.html
so transport wont even need a platform param here, it could be already on sentry options ctor for example

dart/lib/src/transport/header_builder.dart
```diff
@@ -0,0 +1,14 @@
+import '../protocol.dart';
+
+Map<String, String> buildHeaders(/*String authHeader, */ {Sdk sdk}) {
```
marandaneto: it could be a String identifier, instead of the Sdk instance

dart/lib/src/transport/header_builder.dart
```diff
@@ -0,0 +1,14 @@
...
+  final headers = {
+    'Content-Type': 'application/json',
+    // NOTE(lejard_h) overriding user agent on VM and Flutter not sure why
+    // for web it use browser user agent
+    'User-Agent': sdk.identifier
+  };
+
+  headers['User-Agent'] = sdk.identifier;
```
marandaneto: its duplicated

dart/lib/src/transport/transport.dart
```diff
@@ -0,0 +1,131 @@
...
+  @visibleForTesting
+  final Dsn dsn;
+
+  /// Use for browser stacktrace
+  final String origin;
+
+  /// Used by sentry to differentiate browser from io environment
+  final String platform;
+
+  final Sdk sdk;
```
marandaneto: if they are already in the options, they dont need to be public global fields

dart/lib/src/transport/transport.dart
```diff
@@ -0,0 +1,131 @@
...
+  }
+
+  Map<String, dynamic> _getEventData(
+    SentryEvent event, {
+    DateTime timeStamp,
+  }) {
+    final data = event.toJson(origin: origin);
+
+    // TODO add this attributes to event in client
+    if (_options.environmentAttributes != null) {
```
marandaneto: yep, `environmentAttributes` should be even gone, all the values read from `environmentAttributes` should be added on the options, if not there yet

dart/lib/src/browser_client.dart
rxlabz: ✅ fixed

dart/lib/src/browser_client.dart
rxlabz: `platform` is `'dart'` or `'javascript'` , according to the docs ``options.sentryClientName`` is `'sentry.{language}.{platform}/{version}'`

marandaneto: we could do this then?

```
factory SentryIOClient(SentryOptions options) {
    options.platform = sdkPlatform;
    SentryIOClient._(options);
}
```

same for the WebClient, so we move this flag to the options and the `transport` don't need to take it as a param? we do it automatically.

dart/lib/src/client.dart
rxlabz: ✅ fixed

dart/lib/src/transport/header_builder.dart
rxlabz: ✅

dart/lib/src/client.dart
```diff
@@ -64,88 +37,19 @@ abstract class SentryClient {
   /// * https://docs.sentry.io/learn/context/#capturing-the-user
   User userContext;
```
marandaneto: this should be part of the SentryClient only as well, it belongs to the Scope.user

dart/lib/src/client.dart
```diff
@@ -64,88 +37,19 @@ abstract class SentryClient {
...
-      postUri,
-      headers: headers,
-      body: body,
-    );
+    event = _applyScope(event: event, scope: scope);
 
-    if (response.statusCode != 200) {
-      return SentryId.empty();
-    }
+    event = event.copyWith(platform: transport.platform);
```
marandaneto: why do we need to read the `platform` from the `transport`? its already defining on the SentryEvent ctor based on `isWeb`.
I believe `transport.platform` could be done altogether

dart/lib/src/client.dart
```diff
@@ -17,13 +17,19 @@ abstract class SentryClient {
...
+      );
+    }
+  }
 
   @protected
-  SentryOptions options;
-
   @visibleForTesting
-  final Transport transport;
+  SentryOptions options;
```
marandaneto: it could be private I believe

dart/lib/src/client.dart
```diff
@@ -19,38 +18,19 @@ abstract class SentryClient {
...
-  /// The Sentry.io public key for the project.
-  @visibleForTesting
-  // ignore: invalid_use_of_visible_for_testing_member
-  String get publicKey => _dsn.publicKey;
+  SentryClient.base(this.options, {String origin}) {
+    _random = options.sampleRate == null ? null : Random();
+    if (options.transport is NoOpTransport) {
+      options.transport = Transport(
+        options: options,
+        sdkIdentifier: '${sdkName}/${sdkVersion}',
```
marandaneto: could this be replaced by `options.sdkVersion.identifier`? it looks duplicated now

dart/lib/src/browser_client.dart
rxlabz: ✅  platform is now defined at runtime

dart/lib/src/protocol/sentry_event.dart
```diff
@@ -31,10 +32,10 @@ class SentryEvent {
     this.userContext,
     this.contexts,
     this.breadcrumbs,
-    this.sdk,
   })  : eventId = eventId ?? SentryId.newId(),
-        platform = platform ?? sdkPlatform,
-        timestamp = timestamp ?? getUtcDateTime();
+        platform = platform ?? (isWeb ? browserPlatform : sdkPlatform),
```
marandaneto: if this is going to be set here now, why do we need this on the client/transport? I believe this  should be done on the Client, like Java does (via MainEventProcessor)

dart/lib/src/client.dart
rxlabz: ✅

dart/lib/src/sentry_options.dart
```diff
@@ -139,8 +148,9 @@ class SentryOptions {
     this.environmentAttributes,
     this.compressPayload,
     this.httpClient,
+    Transport transport,
     ClockProvider clock = getUtcDateTime,
-  }) {
+  }) : _transport = transport ?? NoOpTransport() {
```
marandaneto: why do we have this here? its already `NoOpTransport` by default, we expose a setter already

dart/lib/src/transport/noop_transport.dart
```diff
@@ -0,0 +1,17 @@
...
+
+class NoOpTransport implements Transport {
+  @override
+  Dsn get dsn => null;
+
+  @override
+  String get origin => null;
+
+  @override
+  String get sdkIdentifier => null;
```
marandaneto: this should not be part of the public API, they should only care about the `send` method

dart/lib/src/version.dart
```diff
@@ -8,19 +8,28 @@
 /// This library contains Sentry.io SDK constants used by this package.
 library version;
 
+import 'utils.dart';
+
 /// The SDK version reported to Sentry.io in the submitted events.
 const String sdkVersion = '4.0.0';
 
+String get sdkName => isWeb ? browserSdkName : ioSdkName;
```
marandaneto: we can make those const private now, if what matters is the `sdkName`

dart/lib/src/client.dart
rxlabz: ✅

dart/lib/src/client.dart
rxlabz: ✅ fixed

dart/lib/src/client.dart
rxlabz: ✅ my bad, I thought `sdkVersion` was just a String, I renamed it `sdk`, and define it in the ioClient & BrowserClient if null

dart/lib/src/sentry_options.dart
rxlabz: we can't access the setter from the constructor, so if we don't want a null transport we needs this.

marandaneto: but the field declaration already has NoOpTransport as a value, it should not belong to the ctor, a setter wont be able to set it to null either, because on the setter we need to check if its null and not allow to set it

dart/lib/src/transport/transport.dart
rxlabz: for now I just let the `dsn` which is used for a lot of tests ( marked as @visibleForTestings )

dart/lib/src/transport/noop_transport.dart
rxlabz: ✅  I just let `dsn` for now as it is used in a lot of tests

dart/lib/src/version.dart
rxlabz: ✅

dart/lib/src/sentry_options.dart
rxlabz: we can't write this : 
```dart
SentryOptions({
    this.dsn,
    this.environmentAttributes,
    this.compressPayload,
    this.httpClient,
    this.transport,
    ClockProvider clock = getUtcDateTime,
  })
```
cause this.transport refers to setter and it can't be accessed from here

we can remove the default value in the declaration of the field, but I think `: _transport = transport ?? NoOpTransport()` is needed

marandaneto: But we dont need a transport in the ctor, as i said, we dont need any ctor params, options is a mutable class and all fields have setters, you could even remove this ctor altogether

rxlabz: ah sorry. in french we say : "I understand fast, but you have to explain me long time" :D

rxlabz: ✅ removed !

dart/lib/src/transport/transport.dart
rxlabz: ✅  removed

dart/lib/src/protocol/sentry_event.dart
rxlabz: ✅ I moved it to the client.

dart/lib/src/client.dart
rxlabz: ✅

dart/lib/src/client.dart
rxlabz: ✅

marandaneto (approved):
nice, thanks for doing this, let's iterate from feature/unified-api

PR #127, opened 2020-10-23
feat: apply sample rate

marandaneto:

- [ ] Bugfix
- [X] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
feat: apply sample rate

one should be able to sample rate events

## :pencil: Checklist

- [ ] All tests passing
- [ ] No breaking changes

### DIFF
--- a/dart/lib/src/client.dart
+++ b/dart/lib/src/client.dart
@@ -181,6 +181,6 @@ abstract class SentryClient {
     if (options.sampleRate != null && _random != null) {
       return (options.sampleRate < _random.nextDouble());
     }
-    return true;
+    return false;
   }
 }
--- a/dart/lib/src/client.dart
+++ b/dart/lib/src/client.dart
@@ -1,4 +1,5 @@
 import 'dart:async';
+import 'dart:math';
 
 import 'package:meta/meta.dart';
 import 'package:sentry/sentry.dart';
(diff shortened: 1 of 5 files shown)


rxlabz (approved):
LGTM

dart/lib/src/client.dart
```diff
@@ -45,6 +51,11 @@ abstract class SentryClient {
   }) async {
     event = _processEvent(event, eventProcessors: options.eventProcessors);
 
+    // dropped by sampling or event processors
+    if (event == null) {
+      return Future.value(SentryId.empty());
```
marandaneto, line 56: @rxlabz should we use `Future.value(SentryId.empty());` or just `SentryId.empty()`?
cus the NoOp Client and Hub return `Future.value(...)` but the real implementation seems to return the String.

bruno-garcia: It returns a Future so it wouldn't compile without this.

Or mark the method `async` but without an `await` in the body that wouldn't make sense

dart/lib/src/client.dart
```diff
@@ -157,4 +176,11 @@ abstract class SentryClient {
     }
     return event;
   }
+
+  bool _sampleRate() {
```
bruno-garcia, line 180: I think the should call `sample`, instead of `sampleRate`.

dart/lib/src/client.dart
marandaneto: It does compile, that's why the question, i guess the "compiler" infers that automatically.

PR #140, opened 2020-10-27
Ref/prepare event & scope.applyToEvent

rxlabz:

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring

## :scroll: Description

- add a dedicated method to prepare the event `SentryClient._prepareEvent(event)`
- add `Scope.applyToEvent(event,hint)`

https://develop.sentry.dev/sdk/unified-api/#scope

## :pencil: Checklist

- [x] All tests passing
- [ ] No breaking changes

### DIFF
--- a/dart/lib/src/client.dart
+++ b/dart/lib/src/client.dart
@@ -43,7 +43,11 @@ abstract class SentryClient {
       return emptyFuture;
     }
 
-    event = scope?.applyToEvent(event, hint) ?? event;
+    if (scope != null) {
+      event = scope.applyToEvent(event, hint);
+    } else {
+      _options.logger(SentryLevel.debug, 'No scope is defined');
+    }
 
     // dropped by scope event processors
     if (event == null) {
--- a/dart/lib/src/client.dart
+++ b/dart/lib/src/client.dart
@@ -1,7 +1,6 @@
 import 'dart:async';
 import 'dart:math';
 
-import 'package:meta/meta.dart';
 import 'package:sentry/sentry.dart';
 import 'package:sentry/src/transport/noop_transport.dart';
 
@@ -44,20 +43,14 @@ abstract class SentryClient {
       return emptyFuture;
     }
 
-    event = _applyScope(event: event, scope: scope);
+    event = scope?.applyToEvent(event, hint) ?? event;
 
     // dropped by scope event processors
     if (event == null) {
       return emptyFuture;
     }
 
-    // TODO create eventProcessors ?
-    event = event.copyWith(
-      serverName: _options.serverName,
-      environment: _options.environment,
-      release: _options.release,
-      platform: event.platform ?? sdkPlatform,
-    );
+    event = _prepareEvent(event);
 
     if (_options.beforeSendCallback != null) {
       try {
@@ -75,6 +68,16 @@ abstract class SentryClient {
     }
 
     return _options.transport.send(event);
+  }
+
+  SentryEvent _prepareEvent(SentryEvent event) {
+    return event.copyWith(
+      serverName: _options.serverName,
+      dist: _options.dist,
+      environment: _options.environment,
+      release: _options.release,
+      platform: event.platform ?? sdkPlatform,
+    );
   }
 
   /// Reports the [throwable] and optionally its [stackTrace] to Sentry.io.
@@ -138,51 +141,6 @@ abstract class SentryClient {
         _options.logger(SentryLevel.debug, 'Event was dropped by a processor');
         break;
       }
-    }
-    return event;
-  }
-
-  SentryEvent _applyScope({
-    @required SentryEvent event,
-    @required Scope scope,
-  }) {
-    if (scope != null) {
-      // Merge the scope transaction.
-      if (event.transaction == null) {
-        event = event.copyWith(transaction: scope.transaction);
-      }
-
-      // Merge the user context.
-      if (event.userContext == null) {
-        event = event.copyWith(userContext: scope.user);
-      }
-
-      // Merge the scope fingerprint.
-      if (event.fingerprint == null) {
-        event = event.copyWith(fingerprint: scope.fingerprint);
-      }
-
-      // Merge the scope breadcrumbs.
-      if (event.breadcrumbs == null) {
-        event = event.copyWith(breadcrumbs: scope.breadcrumbs);
-      }
-
-      // Merge the scope tags.
-      event = event.copyWith(
-          tags: scope.tags.map((key, value) => MapEntry(key, value))
-            ..addAll(event.tags ?? {}));
-
-      // Merge the scope extra.
-      event = event.copyWith(
-          extra: scope.extra.map((key, value) => MapEntry(key, value))
-            ..addAll(event.extra ?? {}));
-
-      // Merge the scope level.
-      if (scope.level != null) {
-        event = event.copyWith(level: scope.level);
-      }
-
-      // TODO: execute scope event processors
     }
     return event;
   }
--- a/dart/lib/src/client.dart
+++ b/dart/lib/src/client.dart
@@ -74,15 +74,13 @@ abstract class SentryClient {
     return _options.transport.send(event);
   }
 
-  SentryEvent _prepareEvent(SentryEvent event) {
-    return event.copyWith(
-      serverName: _options.serverName,
-      dist: _options.dist,
-      environment: _options.environment,
-      release: _options.release,
-      platform: event.platform ?? sdkPlatform,
-    );
-  }
+  SentryEvent _prepareEvent(SentryEvent event) => event.copyWith(
+        serverName: event.serverName ?? _options.serverName,
+        dist: event.dist ?? _options.dist,
+        environment: event.environment ?? _options.environment,
+        release: event.release ?? _options.release,
+        platform: event.platform ?? sdkPlatform,
+      );
 
   /// Reports the [throwable] and optionally its [stackTrace] to Sentry.io.
   Future<SentryId> captureException(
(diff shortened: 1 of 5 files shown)


dart/lib/src/client.dart
```diff
@@ -44,20 +43,14 @@ abstract class SentryClient {
       return emptyFuture;
     }
 
-    event = _applyScope(event: event, scope: scope);
+    event = scope?.applyToEvent(event, hint) ?? event;
```
marandaneto: I prefer to log if `scope` is null, so avoid `scope?` here (otherwise cant log it), also, scope processing could return a null event because of event processors, so ` ?? event` would still keep the event non-null which is wrong

dart/lib/src/client.dart
```diff
@@ -77,6 +70,16 @@ abstract class SentryClient {
...
   }
 
+  SentryEvent _prepareEvent(SentryEvent event) {
+    return event.copyWith(
+      serverName: _options.serverName,
+      dist: _options.dist,
+      environment: _options.environment,
+      release: _options.release,
+      platform: event.platform ?? sdkPlatform,
+    );
```
marandaneto: I'm not sure if this works properly, does this work in the same way of:

`if event. serverName == null ? _options.serverName : event. serverName`?

dart/lib/src/scope.dart
```diff
@@ -142,6 +142,59 @@ class Scope {
   /// Removes an extra from the Scope
   void removeExtra(String key) => _extra.remove(key);
 
+  SentryEvent applyToEvent(SentryEvent event, dynamic hint) {
+    if (event.transaction == null) {
+      event = event.copyWith(transaction: transaction);
+    }
+
+    // Merge the user context.
```
marandaneto: I dont think the wording is `merge` right, its either one or the other

dart/lib/src/scope.dart
```diff
@@ -142,6 +142,59 @@ class Scope {
...
+
+    // Merge the scope tags.
+    event = event.copyWith(
+        tags: tags.map((key, value) => MapEntry(key, value))
+          ..addAll(event.tags ?? {}));
+
+    // Merge the scope extra.
+    event = event.copyWith(
+        extra: extra.map((key, value) => MapEntry(key, value))
+          ..addAll(event.extra ?? {}));
```
marandaneto: if both tags have the same key, which one wins with this syntax? worth a comment, probably the `event` one should be kept.

dart/lib/src/client.dart
rxlabz: no the `copyWith` implementation is `options.serverName != null ? options.serverName : event.serverName`
but we could change its implementation to "fill" only the event's empty properties ?

rxlabz: or here `serverName: event.serverName ?? _options.serverName,`

marandaneto: we should only set a property to the event, read from elsewhere, if the event field is null, that means:

event. serverName == null, it should take the _options.serverName
if event. serverName != null, it should keep event. serverName

dart/lib/src/client.dart
rxlabz: ✅

dart/lib/src/scope.dart
rxlabz: ✅

dart/lib/src/scope.dart
rxlabz: ✅

dart/lib/src/client.dart
rxlabz: ✅

dart/lib/src/scope.dart
```diff
@@ -142,6 +142,65 @@ class Scope {
...
+    }
+
+    // set the scope breadcrumbs if none is set.
+    if (event.breadcrumbs == null) {
+      event = event.copyWith(breadcrumbs: breadcrumbs);
+    }
+
+    // Merge the scope tags
+    // if the scope and the event have tag entries with the same key,
+    // the event tags will be keeped
```
```suggestion
    // the event tags will be kept
```

dart/lib/src/scope.dart
```diff
@@ -142,6 +142,65 @@ class Scope {
   /// Removes an extra from the Scope
   void removeExtra(String key) => _extra.remove(key);
 
+  SentryEvent applyToEvent(SentryEvent event, dynamic hint) {
+    if (event.transaction == null) {
+      event = event.copyWith(transaction: transaction);
+    }
```
marandaneto: what's about testing if the scope values are not null either? so we avoid extra alocation if theres nothing to be copied

eg:
if (event.transaction == null && transaction != null) {
      event = event.copyWith(transaction: transaction);
}

dart/lib/src/scope.dart
```diff
@@ -142,6 +142,65 @@ class Scope {
...
+
+    // set the scope fingerprint if none is set.
+    if (event.fingerprint == null) {
+      event = event.copyWith(fingerprint: fingerprint);
+    }
+
+    // set the scope breadcrumbs if none is set.
+    if (event.breadcrumbs == null) {
+      event = event.copyWith(breadcrumbs: breadcrumbs);
+    }
```
marandaneto: on list and maps we could check if they are not null and empty, so we also save allocation

dart/lib/src/client.dart
```diff
@@ -77,6 +74,14 @@ abstract class SentryClient {
     return _options.transport.send(event);
   }
 
+  SentryEvent _prepareEvent(SentryEvent event) => event.copyWith(
```
marandaneto, line 77: this method should do pretty much the same as `Scope.applyToEvent`, which is apply data if tis null on the event, but they have a different syntax, here is using `field: event.field ?? _options.field`

and on `Scope.applyToEvent`:

```
if (event.field == null) {
      event = event.copyWith(field: field);
    }
```

could we unify the syntax? either using the most friendly one for Dart? the result should be the same though, I believe the 2nd option would be easier to not allocate values if nothing to be copied from the Scope fields

dart/lib/src/scope.dart
rxlabz: should we set the breadcrumbs if the list is empty ?
also what about the same syntax than _prepareEvent : 
```dart
SentryEvent applyToEvent(SentryEvent event, dynamic hint) {
  return event.copyWith(
    transaction: event.transaction ?? transaction,
    user: event.user ?? user,
    // ...
```

dart/lib/src/client.dart
rxlabz: I prefer the first option, it avoids to instanciate a new Event for each field.

```dart
SentryEvent applyToEvent(SentryEvent event, dynamic hint) {
  return event.copyWith(
    transaction: event.transaction ?? transaction,
    user: event.user ?? user,
    // ...
}
```

Is it ok for you ?

marandaneto: sounds good to me

dart/lib/src/client.dart
```diff
@@ -77,6 +74,14 @@ abstract class SentryClient {
...
+        serverName: event.serverName ?? _options.serverName,
+        dist: event.dist ?? _options.dist,
+        environment: event.environment ?? _options.environment,
```
marandaneto: if `_options.environment` is null, it should have taken the default one which is `production`

https://github.com/getsentry/sentry-java/blob/113af8211c07ef622a4e265cc6b3438a6194d9fe/sentry/src/main/java/io/sentry/MainEventProcessor.java#L76-L79

rxlabz: ✅

dart/lib/src/client.dart
rxlabz: ✅

dart/lib/src/scope.dart
rxlabz: ✅

dart/lib/src/client.dart
```diff
@@ -77,6 +74,15 @@ abstract class SentryClient {
     return _options.transport.send(event);
   }
 
+  SentryEvent _prepareEvent(SentryEvent event) => event.copyWith(
```
marandaneto, line 77: can we do the `sdk` field here too?

like https://github.com/getsentry/sentry-java/blob/113af8211c07ef622a4e265cc6b3438a6194d9fe/sentry/src/main/java/io/sentry/MainEventProcessor.java#L86-L88

right now its done on SentryEvent ctor, but we ideally move to here as well:
https://github.com/getsentry/sentry-dart/blob/feature/unified-api/dart/lib/src/protocol/sentry_event.dart#L37

this means, options.sdk should already have this value set

rxlabz: ✅

dart/lib/src/client.dart
```diff
@@ -77,6 +74,16 @@ abstract class SentryClient {
...
+        serverName: event.serverName ?? _options.serverName,
+        dist: event.dist ?? _options.dist,
+        environment:
+            event.environment ?? _options.environment ?? defaultEnvironment,
+        release: event.release ?? _options.release,
+        sdk: event.sdk ?? Sdk(name: sdkName, version: sdkVersion),
```
marandaneto: I meant to read the sdk from the options:

https://github.com/getsentry/sentry-dart/pull/140#discussion_r512756637

marandaneto: `Sdk(name: sdkName, version: sdkVersion)` this could be done on options.sdk field declaration or ctor body so its never null?

rxlabz: in SentryOptions `Sdk sdk = Sdk(name: sdkName, version: sdkVersion)` with the default value ?
should we do the same with defaultEnvironment and initiate options.environment with a default value ?

marandaneto: yes, as default value see here: https://github.com/getsentry/sentry-java/blob/main/sentry/src/main/java/io/sentry/SentryOptions.java#L1107

no about defaultEnvironment, one can overwrite the value with a `null` and we don't want that, unless you add a null checker on the setter, then it should be fine

rxlabz: ✅

marandaneto (approved):
LGTM

PR #148, opened 2020-10-28
unified fixes

rxlabz:

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/dart/lib/src/sentry_browser_client.dart
+++ b/dart/lib/src/sentry_browser_client.dart
@@ -5,10 +5,8 @@
 /// A pure Dart client for Sentry.io crash reporting.
 import 'dart:html' show window;
 
-import 'protocol.dart';
 import 'sentry_client.dart';
 import 'sentry_options.dart';
-import 'version.dart';
 
 SentryClient createSentryClient(SentryOptions options) =>
     SentryBrowserClient(options);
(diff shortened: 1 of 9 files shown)


dart/lib/src/sentry_client.dart
```diff
@@ -22,7 +22,13 @@ abstract class SentryClient {
   SentryClient.base(this._options, {String origin}) {
     _random = _options.sampleRate == null ? null : Random();
     if (_options.transport is NoOpTransport) {
-      _options.transport = HttpTransport(options: _options, origin: origin);
```
rxlabz: @marandaneto  shouldn't we catch the potential Dns.parse exception ? What do you think would be the best strategy or preferred wording here ?

marandaneto: on Java we do on Sentry.init

https://github.com/getsentry/sentry-java/blob/4a9b3cc961e74421b8c56b77cc4808c5ddae2d5b/sentry/src/main/java/io/sentry/Sentry.java#L181-L190

ideally, we try to parse on https://github.com/getsentry/sentry-dart/blob/feature/unified-api/dart/lib/src/sentry.dart#L112-L121

so we crash on Sentry.init and people can fix right away

dart/lib/src/sentry_browser_client.dart
```diff
@@ -25,13 +23,10 @@ class SentryBrowserClient extends SentryClient {
   ///
   /// If [httpClient] is provided, it is used instead of the default client to
   /// make HTTP calls to Sentry.io. This is useful in tests.
-  factory SentryBrowserClient(SentryOptions options) {
-    options.sdk ??= Sdk(name: sdkName, version: sdkVersion);
-
-    // origin is necessary for sentry to resolve stacktrace
-    return SentryBrowserClient._(options);
-  }
+  factory SentryBrowserClient(SentryOptions options) =>
```
marandaneto, line 26: make me wonder if now we need to have a SentryBrowserClient and IOClient only because of the `origin` field.

marandaneto (approved):
LGTM

dart/lib/src/sentry_browser_client.dart
rxlabz: true, we could move the origin settings in transport, and use the same strategy as for gzip compression => it will remove the need for 2 clients

PR #152, opened 2020-10-28
Ref: remove platform specific clients to use SentryClient

rxlabz:

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring

## :scroll: Description

- remove the SentryBrowserClient and SentryIOClient
- declare SentryClient as concrete class

the event origin is the last reason to have a web specific client. By moving the origin to transport layer we provide a single SentryClient class

ran unit tests and examples 

## :pencil: Checklist

- [x] All tests passing
- [ ] No breaking changes


dart/lib/src/sentry_client.dart
```diff
@@ -3,30 +3,31 @@ import 'dart:math';
...
-  factory SentryClient(SentryOptions options) => createSentryClient(options);
-
-  SentryClient.base(this._options, {String origin}) {
+  /// [environmentAttributes] contain event attributes that do not change over
+  /// the course of a program's lifecycle. These attributes will be added to
+  /// all events captured via this client. The following attributes often fall
+  /// under this category: [Event.serverName], [Event.release], [Event.environment].
+  ///
+  /// If [httpClient] is provided, it is used instead of the default client to
+  /// make HTTP calls to Sentry.io. This is useful in tests.
```
marandaneto: I've fixed those comments on the current client, now those docs are not up to date, `environmentAttributes` doesn't exist etc

dart/lib/src/transport/http_transport.dart
```diff
@@ -16,15 +17,15 @@ class HttpTransport implements Transport {
...
 
   CredentialBuilder _credentialBuilder;
 
   final Map<String, String> _headers;
 
-  HttpTransport({@required SentryOptions options, String origin})
+  HttpTransport({@required SentryOptions options /*, String origin*/})
       : _options = options,
-        _origin = origin,
+        /*_origin = origin,*/
```
marandaneto: clean up the comments if they are not relevant

dart/test/sentry_browser_test.dart
```diff
@@ -2,17 +2,16 @@
...
 import 'test_utils.dart';
 
 void main() {
-  group('SentryBrowserClient', () {
+  group('SentryClient on browser', () {
     test('SentryClient constructor build browser client', () {
       final client = SentryClient(SentryOptions(dsn: testDsn));
-      expect(client is SentryBrowserClient, isTrue);
+      expect(client is SentryClient, isTrue);
     });
```
marandaneto: does this test and the flag `isWeb` makes sense now? or still useful?

dart/lib/src/transport/http_transport.dart
rxlabz: ✅

dart/test/sentry_browser_test.dart
rxlabz: ✅  These client tests were useless indeed, but the flag `isWeb` seems necessary to run test_utils both in browser and VM modes.

dart/lib/src/sentry_client.dart

PR #154, opened 2020-10-29
add contexts to scope and apply it

rxlabz:

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring

## :scroll: Description

add a `contexts` property to `Scope`

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes


dart/lib/src/protocol/contexts.dart
```diff
@@ -11,107 +13,170 @@ import 'runtime.dart';
...
-class Contexts {
-  const Contexts({
-    this.device,
-    this.operatingSystem,
-    this.runtimes,
-    this.app,
-    this.browser,
-    this.gpu,
-  });
+class Contexts extends MapView {
```
marandaneto: just out of curiosity, why not `HashMap`? what's the difference between them?

dart/lib/src/protocol/contexts.dart
```diff
@@ -11,107 +13,170 @@ import 'runtime.dart';
...
-              while (json.containsKey(key)) {
-                key = '$key$k';
-                k++;
+    forEach((key, value) {
+      if (value == null) return;
+      switch (key) {
+        case Device.type:
+          Map<String, dynamic> deviceMap;
+          if ((deviceMap = device.toJson()).isNotEmpty) {
+            json['device'] = deviceMap;
```
marandaneto:
```suggestion
            json[Device.type] = deviceMap;
```

wouldd this work too?

dart/lib/src/protocol/contexts.dart
```diff
@@ -11,107 +13,170 @@ import 'runtime.dart';
...
+          break;
+
+        case App.type:
+          Map<String, dynamic> appMap;
+          if ((appMap = app.toJson()).isNotEmpty) {
+            json[App.type] = appMap;
+          }
+          break;
+
+        case 'browser':
```
marandaneto: why do we use the `Class.type ` on the case above and here a plain String?

dart/lib/src/protocol/contexts.dart
```diff
@@ -11,107 +13,170 @@ import 'runtime.dart';
...
+            json['gpu'] = gpuMap;
+          }
+          break;
+
+        case 'runtimes':
+          if (runtimes != null) {
+            if (runtimes.length == 1) {
+              final runtime = runtimes[0];
+              if (runtime != null) {
+                final key = runtime.key ?? 'runtime';
```
marandaneto: Runtime.type?

dart/lib/src/protocol/contexts.dart
```diff
@@ -11,107 +13,170 @@ import 'runtime.dart';
...
+          json[key] = value;
       }
-    }
+    });
 
     return json;
   }
+
+  Contexts clone() {
+    // TODO ? make each class cloneable ?
```
marandaneto: is this related to this? https://github.com/getsentry/sentry-dart/projects/1#card-47865684
if so, we can remove the comment, we already have a card for it

dart/test/scope_test.dart
```diff
@@ -2,6 +2,8 @@ import 'package:collection/collection.dart';
 import 'package:sentry/sentry.dart';
 import 'package:test/test.dart';
 
+// TODO test context apply / clone
```
marandaneto: is this part of this PR?

marandaneto:
this PR is not addressing the SentryClient changes:
which is merging the Event Contexts and the Scope Contexts
https://github.com/getsentry/sentry-java/blob/5ebd49cbfd17eb66175e9247db6d27e4df84a56b/sentry/src/main/java/io/sentry/SentryClient.java#L317-L321

rxlabz:
> this PR is not addressing the SentryClient changes:
> which is merging the Event Contexts and the Scope Contexts

@marandaneto  is this incorrect or incomplete? https://github.com/getsentry/sentry-dart/blob/262305a074a3d0ce497cc0811b221f252a9fcdef/dart/lib/src/scope.dart#L162-L166

dart/test/scope_test.dart
rxlabz: ✅

dart/lib/src/protocol/contexts.dart
rxlabz: ✅

dart/lib/src/protocol/contexts.dart
rxlabz: ✅

dart/lib/src/protocol/contexts.dart
rxlabz: ✅

dart/lib/src/protocol/contexts.dart
rxlabz: ✅

dart/lib/src/protocol/contexts.dart
rxlabz: Extending HashMap requires implementing most of its members. MapView is only a wrapper around an initial Map, which seemed simpler in this case.

marandaneto: I see, thanks.
Should we at least set its type then? `extends MapView<String, dynamic> `

dart/lib/src/scope.dart
```diff
@@ -53,7 +53,13 @@ class Scope {
 
   Map<String, dynamic> get extra => Map.unmodifiable(_extra);
 
-  // TODO: Contexts
+  final Contexts _contexts = Contexts();
+
+  Contexts get contexts => _contexts;
+
+  void setContexts({String key, dynamic value}) {
```
marandaneto: does it need to be named? because the other add methods are not, only this one will be different

marandaneto: missing method `removeContexts`

marandaneto: also, `setContexts` can be `dynamic` as value, but our protocol does not accept non objects here, see what is done on Java, we'd need something similar or `setContexts('a', 'b')` would cause a backend issue

https://github.com/getsentry/sentry-java/blob/main/sentry/src/main/java/io/sentry/Scope.java#L340-L374

dart/test/scope_test.dart
```diff
@@ -222,6 +222,42 @@ void main() {
...
+      ..fingerprint = ['example-dart']
+      ..addBreadcrumb(breadcrumb)
+      ..transaction = '/example/app'
+      ..level = SentryLevel.warning
+      ..setTag('build', '579')
+      ..setExtra('company-name', 'Dart Inc')
+      ..setContexts(key: 'theme', value: 'material')
+      ..addEventProcessor(
+        (event, hint) => event..tags.addAll({'page-locale': 'en-us'}),
+      );
```
marandaneto: please add a few values to the key/value objects to test the precedence over others, eg
a context key on the event cant be replaced by a value of the same key on the scope.

for the level, this is actually the opposite, the scope level has precedence over the event level

marandaneto:
> > this PR is not addressing the SentryClient changes:
> > which is merging the Event Contexts and the Scope Contexts
> 
> @marandaneto is this incorrect or incomplete?
> 
> https://github.com/getsentry/sentry-dart/blob/262305a074a3d0ce497cc0811b221f252a9fcdef/dart/lib/src/scope.dart#L162-L166

more to incomplete than incorrect, I've made a few new comments

dart/lib/src/protocol/runtime.dart
```diff
@@ -1,9 +1,13 @@
+const runtimesType = 'runtimes';
```
marandaneto: just out of curiosity, why the type is not in the class?

dart/lib/src/scope.dart
```diff
@@ -53,7 +53,13 @@ class Scope {
 
   Map<String, dynamic> get extra => Map.unmodifiable(_extra);
 
-  // TODO: Contexts
+  final Contexts _contexts = Contexts();
+
+  Contexts get contexts => _contexts;
```
```suggestion
  Contexts get contexts => Map.unmodifiable(_contexts);
```

dart/lib/src/protocol/contexts.dart
rxlabz: ✅

dart/lib/src/scope.dart
rxlabz: - `setContexts` args are now mandatory 
- added removeContexts
- I added some conditions to the setter ( not null and of type `num`, `bool` or `String` )
 What would be the best strategy in case of invalid value? For now, it's a no_op.

rxlabz: Also, now, the value must be an Object ( with a `toString()` method ) so we could just set the toString() if the value is not a num, a bool, or a String?
```dart
if (key != null && value != null) return;

    _contexts[key] = (value is num || value is bool || value is String) ? value : '$value';
```

dart/lib/src/scope.dart
rxlabz: ✅

dart/lib/src/protocol/runtime.dart
rxlabz: ✅ I moved it in the class and renamed it `Runtime.listType`

dart/lib/src/scope.dart
marandaneto: @rxlabz on Java, we'd not return, but rather wrap into a `value` field if they are primitive types.
This deviates quite a lot of the Java impl. and limit the API quite a lot, I'd suggest doing something similar to Java.

exposing more setContexts methods or increment the current solution.

_contexts[key] = (value is num || value is bool || value is String) ? [{'value': value] : value;

sounds more realistic

dart/test/scope_test.dart
rxlabz: ✅

dart/lib/src/scope.dart
rxlabz: sorry, I missed the java `public void setContexts(final @NotNull String key, final @NotNull Object value)` and misunderstood the Api.

dart/test/scope_test.dart
```diff
@@ -230,12 +230,90 @@ void main() {
...
+      id: '800',
+      username: 'first-user',
+      email: 'first@user.lan',
+      ipAddress: '127.0.0.1',
+      extras: <String, String>{'first-sign-in': '2020-01-01'},
+    );
+
+    final breadcrumb = Breadcrumb(message: 'Authenticated');
+
+    setUp(() {});
```
marandaneto: do we need this?

dart/test/scope_test.dart
```diff
@@ -230,12 +230,90 @@ void main() {
...
+      expect(updatedEvent.breadcrumbs, [breadcrumb]);
+      expect(updatedEvent.level, SentryLevel.warning);
+      expect(updatedEvent.tags,
+          {'etag': '987', 'build': '579', 'page-locale': 'en-us'});
+      expect(
+          updatedEvent.extra, {'e-infos': 'abc', 'company-name': 'Dart Inc'});
+      expect(updatedEvent.contexts['theme'], 'material');
+    });
+
+    test('should not apply context properties when event has already it ', () {
```
marandaneto:
```suggestion
    test('should not apply context properties when event already has it ', () {
```

marandaneto: I'd make this test smaller and more focused on the context field, right now all the user, breadcrumb etc is just noisy for the name of the test, the test `apply context to event` already cover all the other fields, right

marandaneto: I dont see this test testing the `contexts` field though, its not even called

dart/lib/src/protocol/contexts.dart
```diff
@@ -11,107 +13,169 @@ import 'runtime.dart';
...
+    final copy = Contexts(
+      device: device?.clone(),
+      operatingSystem: operatingSystem?.clone(),
+      app: app?.clone(),
+      browser: browser?.clone(),
+      gpu: gpu?.clone(),
+      runtimes: runtimes.map((runtime) => runtime.clone()).toList(),
+    )..addEntries(
+        entries.where((element) => !_defaultFields.contains(element.key)),
+      );
```
marandaneto, line 167: is this case covered by an unit test?

rxlabz: is it enough? https://github.com/getsentry/sentry-dart/blob/99027322f429d6884cf7ec9a85a2053426ffef49/dart/test/contexts_test.dart#L127-L128

dart/test/scope_test.dart
rxlabz: ✅

dart/test/scope_test.dart
rxlabz: ✅

dart/lib/src/scope.dart
rxlabz: ✅

dart/lib/src/scope.dart
```diff
@@ -153,6 +171,17 @@ class Scope {
...
 
+    _contexts.clone().forEach((key, value) {
+      // add the runtime list to the event.contexts.runtimes
+      if (key == Runtime.listType && value is List && value.isNotEmpty) {
+        value.forEach((runtime) => event.contexts.addRuntime(runtime));
+      } else if ((!event.contexts.containsKey(key) ||
+              event.contexts[key] == null) &&
+          value != null) {
+        event.contexts[key] = value;
+      }
```
marandaneto: why do we need to special case here? its not just adding all the keys from a map to another map?

rxlabz: should we override the event.contexts.runtimes ?

rxlabz: In Java, it's not a list so it makes sense to override but with a list, I wasn't sure.

marandaneto: we should probably merge them, yeah.
can you do a method like you do with `_mergeEventExtra` or `_mergeEventExtra`? seems easier to read (to merge runtimes).

also, we dont allow null keys right, so `event.contexts[key] == null` seems unecessary?

rxlabz: ✅

rxlabz: for now the `Contexts` ctor allows null values 
```dart
Contexts({
    Device device,
    OperatingSystem operatingSystem,
    List<Runtime> runtimes,
    App app,
    Browser browser,
    Gpu gpu,
  }) : super({
          Device.type: device,
          OperatingSystem.type: operatingSystem,
          Runtime.listType: runtimes ?? [],
          App.type: app,
          Browser.type: browser,
          Gpu.type: gpu,
        });
```

marandaneto: true, good catch

marandaneto (approved):
LGTM

PR #155, opened 2020-10-30
Feat/more protocol

rxlabz:

- [ ] Bugfix
- [x] New feature
- [x] Enhancement
- [ ] Refactoring

## :scroll: Description

complete the Sentry protocol : 
- add classes : Request, DebugMeta, DebugImage,...
- add SentryException and SentryStacktrace factories
- add inApp excludes and includes logic

mostly added tests to exception serialization, factories and precedence

## :pencil: Checklist

- [x] All tests passing
- [ ] No breaking changes


dart/lib/src/protocol/mechanism.dart
```diff
@@ -0,0 +1,80 @@
...
+  /// The mechanism meta data usually carries error codes reported by
+  /// the runtime or operating system, along with a platform dependent
+  /// interpretation of these codes. SDKs can safely omit code names and
+  /// descriptions for well known error codes, as it will be filled out by
+  /// Sentry. For proprietary or vendor-specific error codes,
+  /// adding these values will give additional information to the user.
+  final Map<String, Object> meta;
+
+  /// Arbitrary extra data that might help the user understand the error thrown by this mechanism
+  final Map<String, Object> data;
```
marandaneto: should it not be `dynamic` instead of `Object`?

dart/lib/src/protocol/mechanism.dart
```diff
@@ -0,0 +1,80 @@
...
+    if (type != null) {
+      json['type'] = type;
+    }
+
+    if (description != null) {
+      json['description'] = description;
+    }
+
+    if (helpLink != null) {
+      json['helpLink'] = helpLink;
```
marandaneto:
```suggestion
      json['help_link'] = helpLink;
```

the json field names should follow the name convention on the docs
https://develop.sentry.dev/sdk/event-payloads/exception/

dart/lib/src/protocol/sentry_stack_frame.dart
```diff
@@ -0,0 +1,103 @@
...
+  String filename;
+  String function;
+  String module;
+  String lineno;
+  String colno;
+  String absPath;
+  String contextLine;
+  bool inApp;
+  String _package;
+  bool _native;
```
marandaneto: are they reserved words like in Java? we dont need the `_` if not

dart/lib/src/protocol/sentry_exception.dart
```diff
@@ -0,0 +1,44 @@
+import 'mechanism.dart';
+import 'sentry_stack_trace.dart';
+
+class SentryException {
```
marandaneto, line 6: missing `threadId` field (if this is even possible on Dart), but we might need because of the integration with the native layers

https://github.com/getsentry/sentry-java/blob/5ebd49cbfd17eb66175e9247db6d27e4df84a56b/sentry/src/main/java/io/sentry/protocol/SentryException.java#L12

dart/lib/src/protocol/sentry_event.dart
```diff
@@ -140,6 +132,8 @@ class SentryEvent {
   /// The SDK Interface describes the Sentry SDK and its configuration used to capture and transmit an event.
   final Sdk sdk;
 
+  final Request request;
```
marandaneto, line 153: add a comment doc please

dart/lib/src/sentry_client.dart
```diff
@@ -87,7 +87,6 @@ class SentryClient {
   }) {
     final event = SentryEvent(
       exception: exception,
-      stackTrace: stackTrace,
```
marandaneto, line 94: this is the stackTrace given by the user, so we'd need to find a way to still pass this along to be used on `SentryExceptionFactory` and `SentryStackTraceFactory`

dart/lib/src/sentry_exception_factory.dart
```diff
@@ -0,0 +1,32 @@
...
+      : _stacktraceFactory = stacktraceFactory;
+
+  SentryException getSentryException(
+    dynamic exception, {
+    StackTrace stackTrace,
+    Mechanism mechanism,
+  }) {
+    if (stackTrace == null && exception is Error) {
+      stackTrace = exception.stackTrace;
+    }
```
marandaneto: consider using an `else if` here, if `stackTrace` is null, get from `StackTrace.current` maybe?

https://github.com/getsentry/sentry-dart/projects/1#card-48533365

dart/lib/src/sentry_exception_factory.dart
```diff
@@ -0,0 +1,32 @@
+import 'protocol/sentry_stack_trace.dart';
+import 'protocol.dart';
+import 'sentry_stack_trace_factory.dart';
+
+/// class to convert Dart Error and exception to SentryException
+class SentryExceptionFactory {
+  final SentryStackTraceFactory _stacktraceFactory;
+
+  SentryExceptionFactory({SentryStackTraceFactory stacktraceFactory})
```
marandaneto: curiosity, why do we use optional param if this is required?
would not be the proper way just `SentryExceptionFactory(SentryStackTraceFactory stacktraceFactory)` ?

dart/lib/src/sentry_stack_trace_factory.dart
```diff
@@ -0,0 +1,57 @@
+import 'package:stack_trace/stack_trace.dart';
+
+import 'protocol/sentry_stack_frame.dart';
+
+class SentryStackTraceFactory {
+  List<SentryStackFrame> getStackFrames(dynamic stackTrace) {
```
marandaneto, line 30: I don't see the `origin` field playing a role here

dart/test/test_utils.dart
```diff
@@ -7,7 +7,7 @@ import 'dart:convert';
 
 import 'package:http/http.dart';
 import 'package:http/testing.dart';
-import 'package:sentry/sentry.dart';
+import 'package:sentry/sentry.dart' hide Request;
```
marandaneto: why this if its just a test?

flutter/example/lib/main.dart
```diff
@@ -37,8 +37,12 @@ Future<void> main() async {
   }, (error, stackTrace) async {
     print('Capture from runZonedGuarded $error');
     final event = SentryEvent(
-      exception: error,
-      stackTrace: stackTrace,
+      exception: SentryException(
+        type: '${error.runtimeType}',
+        value: '$error',
+        stacktrace: SentryStackTrace()
+          ..frames = SentryStackTraceFactory().getStackFrames(stackTrace),
```
marandaneto: I don't think this should be part of the sample, `SentryStackTraceFactory` and `SentryExceptionFactory` are implementation details.

`error` and `stackTrace` should be converted in the event pipeline (SentryClient), using the convert factories

dart/lib/src/sentry_stack_trace_factory.dart
rxlabz: From previous release, in a browser context, absPath should be `'$origin${_absolutePathForCrashReport(frame)}'`

marandaneto: yep but if you didn't take it as a `param` how would you concatenate?
it could be also a ctor param btw as it doesnt change

dart/test/test_utils.dart
rxlabz: If we expose the Request class, we need to avoid the class name conflict with`http.Request`.

dart/lib/src/protocol/mechanism.dart
rxlabz: ✅

dart/lib/src/protocol/mechanism.dart
rxlabz: ✅

dart/lib/src/protocol/sentry_stack_frame.dart
rxlabz: ✅

dart/lib/src/protocol/sentry_exception.dart
rxlabz: ✅ Dart doesn't expose the thread, but I added the field

dart/lib/src/protocol/sentry_event.dart
rxlabz: ✅

dart/lib/sentry.dart
```diff
@@ -8,5 +8,6 @@ export 'src/scope.dart';
 export 'src/sentry.dart';
 export 'src/sentry_client.dart';
 export 'src/sentry_options.dart';
+export 'src/sentry_stack_trace_factory.dart';
```
marandaneto: we should not expose this, this is the implementation detail

dart/lib/src/protocol/sentry_event.dart
```diff
@@ -76,15 +78,20 @@ class SentryEvent {
 
   /// An object that was thrown.
   ///
-  /// It's `runtimeType` and `toString()` are logged. If this behavior is
-  /// undesirable, consider using a custom formatted [message] instead.
-  final dynamic exception;
+  /// It's `runtimeType` and `toString()` are logged.
+  /// If it's an Error, with a stackTrace, the stackTrace is logged.
+  /// If this behavior is undesirable, consider using a custom formatted [message] instead.
+  final dynamic throwable;
```
marandaneto, line 84: `throwable` makes sense in Java, but not in Dart, maybe `error`? wdyt?

dart/lib/src/protocol/sentry_event.dart
```diff
@@ -187,10 +202,12 @@ class SentryEvent {
         contexts: contexts ?? this.contexts,
         breadcrumbs: breadcrumbs ?? this.breadcrumbs,
         sdk: sdk ?? this.sdk,
+        request: request ?? this.request,
       );
 
   /// Serializes this event to JSON.
-  Map<String, dynamic> toJson({String origin}) {
+  Map<String, dynamic> toJson([String origin = '']) {
```
marandaneto: why do we use positional param here?

dart/lib/src/protocol/sentry_event.dart
```diff
@@ -187,10 +202,12 @@ class SentryEvent {
...
+    print('SentryEvent.toJson  $origin');
```
marandaneto: if you think this is important to be logged, please use `options.logger` instead

dart/lib/src/protocol/sentry_exception.dart
```diff
@@ -0,0 +1,59 @@
...
+  String module;
+
+  /// An optional stack trace object
+  SentryStackTrace stacktrace;
+
+  /// An optional object describing the [Mechanism] that created this exception
+  Mechanism mechanism;
+
+  /// Represents a thread id. not available in Dart
+  int threadId;
```
marandaneto: its a Long on Java

dart/lib/src/protocol/sentry_exception.dart
```diff
@@ -0,0 +1,59 @@
+import 'package:meta/meta.dart';
+
+import 'mechanism.dart';
+import 'sentry_stack_trace.dart';
+
+/// The Exception Interface specifies an exception or error that occurred in a program.
+class SentryException {
```
marandaneto, line 6: why do we have final fields in all the data classes but not this one?

dart/lib/src/protocol/sentry_exception.dart
```diff
@@ -0,0 +1,59 @@
...
+
+  SentryException({
+    @required this.type,
+    @required this.value,
+    this.module,
+    this.stacktrace,
+    this.mechanism,
+  });
+
+  Map<String, dynamic> toJson([String origin = '']) {
```
marandaneto: positional param?
I feel that this value is passing by thru all the classes, does it make sense to add this to the SentryEvent as non-serializable?

dart/lib/src/protocol/sentry_stack_frame.dart
```diff
@@ -0,0 +1,106 @@
+class SentryStackFrame {
```
marandaneto, line 1: https://github.com/getsentry/sentry-dart/pull/155/files#r517149327

dart/lib/src/protocol/sentry_stack_frame.dart
```diff
@@ -0,0 +1,106 @@
...
+  static final SentryStackFrame asynchronousGapFrameJson = SentryStackFrame()
+    ..absPath = '<asynchronous suspension>';
+
+  List<String> preContext;
```
marandaneto: missing docs for all the fields

dart/lib/src/protocol/sentry_stack_frame.dart
```diff
@@ -0,0 +1,106 @@
+class SentryStackFrame {
+  static final SentryStackFrame asynchronousGapFrameJson = SentryStackFrame()
```
marandaneto: if this is internal, we can make it private

dart/lib/src/protocol/sentry_stack_frame.dart
```diff
@@ -0,0 +1,106 @@
...
+  String platform;
+  String imageAddr;
+  String symbolAddr;
+  String instructionAddr;
+  String rawFunction;
+
+  Map<String, dynamic> toJson([String origin = '']) {
+    final json = <String, dynamic>{};
+
+    if (preContext?.isNotEmpty ?? false) {
```
marandaneto: most of our NotEmpty checks use:

`field != null && field.isNotEmpty`

guess `field?.isNotEmpty ?? false` is more elegant, should we change in the other places too, I dont like the idea of mixing things up, we could decide the best approach and go for it

dart/lib/src/protocol/sentry_stack_frame.dart
```diff
@@ -0,0 +1,106 @@
+class SentryStackFrame {
+  static final SentryStackFrame asynchronousGapFrameJson = SentryStackFrame()
+    ..absPath = '<asynchronous suspension>';
+
+  List<String> preContext;
+  List<String> postContext;
+  Map<String, String> vars;
+  List<int> framesOmitted;
```
marandaneto: should we add getters to allow mutable list/map?

dart/lib/src/protocol/sentry_stack_trace.dart
```diff
@@ -0,0 +1,33 @@
+import 'sentry_stack_frame.dart';
+
+class SentryStackTrace {
```
marandaneto, line 6: same comments from SentryStackFrame

dart/lib/src/sentry_exception_factory.dart
```diff
@@ -0,0 +1,33 @@
...
+class SentryExceptionFactory {
+  final SentryStackTraceFactory _stacktraceFactory;
+
+  const SentryExceptionFactory({
+    SentryStackTraceFactory stacktraceFactory = const SentryStackTraceFactory(),
+  }) : _stacktraceFactory = stacktraceFactory;
+
+  SentryException getSentryException(
+    dynamic exception, {
+    StackTrace stackTrace,
```
marandaneto: `getStackFrames(dynamic stackTrace)` is a `dynamic`, but here a `StackTrace`, as Dart fully allows a `StackTrace` to be even a String, we should probably use only `dynamic`

dart/lib/src/sentry_stack_trace_factory.dart
```diff
@@ -0,0 +1,60 @@
+import 'package:stack_trace/stack_trace.dart';
+
+import 'protocol/sentry_stack_frame.dart';
+
+class SentryStackTraceFactory {
```
marandaneto, line 10: docs

dart/lib/src/sentry_stack_trace_factory.dart
```diff
@@ -0,0 +1,60 @@
...
+
+      if (t < chain.traces.length - 1) {
+        frames.add(SentryStackFrame.asynchronousGapFrameJson);
+      }
+    }
+
+    return frames.reversed.toList();
+  }
+
+  SentryStackFrame encodeStackTraceFrame(Frame frame) {
```
marandaneto, line 54: should this be private too

dart/lib/src/protocol/sentry_stack_frame.dart
```diff
@@ -0,0 +1,106 @@
...
+  String filename;
```
marandaneto:
```suggestion
  String fileName;
```

dart/lib/src/protocol/sentry_stack_frame.dart
```diff
@@ -0,0 +1,106 @@
...
+  String function;
+  String module;
+  String lineno;
```
marandaneto:
```suggestion
  String lineNo;
```

marandaneto: stick to always `camelCase`, in Java not always is like this because we use reflection and Gson name strategy, but here we do serialization by hands, it's fine to use whatever is easier to read

dart/lib/src/protocol/sentry_stack_frame.dart
```diff
@@ -0,0 +1,106 @@
...
+  String colno;
```
marandaneto: some data types are wrong, `lineno` and `colno` are int, for example, I'd double-check all new additions to this PR otherwise we'll probably see backend errors

dart/test/sentry_event_test.dart
```diff
@@ -109,9 +108,6 @@ void main() {
             'params': ['1', '2']
           },
           'transaction': '/test/1',
-          'exception': [
```
marandaneto, line 112: don't we test this anymore?

dart/test/sentry_event_test.dart
```diff
@@ -134,18 +130,7 @@ void main() {
               },
             ]
           },
-        }..addAll(
```
marandaneto, line 137: don't we test this anymore?

dart/test/test_utils.dart
```diff
@@ -378,16 +381,14 @@ void runTest({Codec<List<int>, List<int>> gzip, bool isWeb = false}) {
 
     try {
       throw ArgumentError('Test error');
-    } catch (error, stackTrace) {
+    } catch (error) {
       final eventWithoutContext = SentryEvent(
         eventId: SentryId.empty(),
-        exception: error,
-        stackTrace: stackTrace,
+        throwable: error,
```
marandaneto, line 377: why are we not testing if caught stack trace?

marandaneto:
still missing DebugMeta on the event level

dart/lib/src/protocol/sentry_stack_frame.dart
rxlabz: ✅

dart/lib/src/protocol/sentry_stack_frame.dart
rxlabz: ✅

dart/lib/src/sentry_stack_trace_factory.dart
```diff
@@ -0,0 +1,60 @@
+import 'package:stack_trace/stack_trace.dart';
+
+import 'protocol/sentry_stack_frame.dart';
+
+class SentryStackTraceFactory {
+  const SentryStackTraceFactory();
```
marandaneto: missing `inAppInclude` and `inAppExclude`

https://github.com/getsentry/sentry-java/blob/113af8211c07ef622a4e265cc6b3438a6194d9fe/sentry/src/main/java/io/sentry/MainEventProcessor.java#L28-L29

https://github.com/getsentry/sentry-java/blob/113af8211c07ef622a4e265cc6b3438a6194d9fe/sentry/src/main/java/io/sentry/SentryStackTraceFactory.java#L67-L94

I'd get the Options as param instead of 2 list as Java

dart/lib/src/sentry_stack_trace_factory.dart
```diff
@@ -0,0 +1,60 @@
...
+    return frames.reversed.toList();
+  }
+
+  SentryStackFrame encodeStackTraceFrame(Frame frame) {
+    final sentryStackFrame = SentryStackFrame()
+      ..absPath = '${_absolutePathForCrashReport(frame)}'
+      ..function = frame.member
+      ..lineNo = frame.line
+      ..colNo = frame.column
+      ..inApp = !frame.isCore;
```
marandaneto: See: https://github.com/getsentry/sentry-dart/pull/155/files#r517188682

we can still use `frame.isCore` in conjunction with inAppExcudes/Includes, but this affects grouping, one might not want to consider 3rd party libs for grouping exceptions together.

dart/test/sentry_event_test.dart
rxlabz: it depends. `SentryEvent` has a `throwable`, an  `exception`, and a `stacktrace` fields.
The "normal" use case is to set a `throwable`, which will be converted to SentryException before transport.
Is there a use case in which the user set directly a SentryException ? in which case we need to add more verifications before setting the `event.exception` in `client.prepareEvent()`

dart/test/sentry_event_test.dart
rxlabz: as the tested `throwable` is an error, it should contain its stacktrace, so I'm not sure it make sense to set the event.stacktrace in this case ?

dart/test/test_utils.dart
rxlabz: It's not really the subject of this test, and as we catch an `Error`, the stacktrace should be in the Error itself.

dart/lib/src/protocol/sentry_exception.dart
rxlabz: `Long` doesn't exist in Dart, but `int` should do the job https://api.dart.dev/stable/dart-core/int-class.html

dart/test/sentry_event_test.dart
marandaneto: yes, a user can set its own `event.exception`, the thing is, on Java, we add an exception if already exist, because its a list, but in Dart is different, we only process the `throwable` if the given `exception` is null, so we dont overwrite a custom exception from the user.

https://github.com/getsentry/sentry-java/blob/113af8211c07ef622a4e265cc6b3438a6194d9fe/sentry/src/main/java/io/sentry/MainEventProcessor.java#L53-L56

this check is missing on Java

dart/test/test_utils.dart
marandaneto: makes sense, I'd just guarantee that another test is testing the `stackTrace` or we'd need to create a new one for this use case

dart/lib/src/protocol/sentry_exception.dart
marandaneto: TIL, thanks

dart/test/sentry_event_test.dart
marandaneto: it doesnt, but lets make sure that this is use case is tested elsewhere (throwable not being an error), or create a new test for it

dart/lib/src/protocol/sentry_event.dart
rxlabz: IMO, as Dart uses both `Error`, and `Exception`, and we already have `SentryEvent.exception`, `error` would be more confusing, `throwable` seems a clearer way to indicate that it can be an `Error` or an `Exception`.

rxlabz: The doc could be more explicit: `/// An Exception or an Error that was thrown.`?

marandaneto: what's about this then? https://github.com/getsentry/sentry-dart/pull/142
we have changed exactly for this reason

marandaneto: well afaik it could be literally anything, it's a `dynamic` field, people can throw a String, so it should be generic enough to work for everything, and for that `throwable` makes sense.
maybe @bruno-garcia ?

rxlabz: We made #142 before having `SentryException`, so it made sense to me. Now having an `error` field, which could be an `Error` or an `Exception`, and an `exception`, which is not an `Exception` but a `SentryException`, seems weird to me.

dart/lib/src/protocol/sentry_exception.dart
rxlabz: In java this class seems mutable and hasn't a `clone()` method, so I translated like this. Should we make it immutable? and add a `clone()` ? or `copyWith()` ?

marandaneto: SentryEvent on Java is mutable and it is immutable on Dart, we'd need to go the way of what makes more sense to the language

dart/lib/src/sentry_exception_factory.dart
rxlabz: ✅  it's now an optional argument with a default value

dart/lib/src/sentry_stack_trace_factory.dart
rxlabz: ✅  the origin is now a SentryStackFrame ctor param

flutter/example/lib/main.dart
rxlabz: ✅  removed

dart/lib/src/protocol/sentry_event.dart
rxlabz: ✅ no more origin param

dart/lib/src/protocol/sentry_event.dart
rxlabz: ✅  removed

dart/lib/src/protocol/sentry_exception.dart
rxlabz: ✅  `SentryException` and `SentryStackFrame` are now immutable.

dart/lib/src/protocol/sentry_exception.dart
rxlabz: ✅  no more `origin` param

dart/lib/src/protocol/sentry_stack_trace.dart
rxlabz: do we need setters here or should we add a `copyWith` method ?

dart/lib/src/sentry_stack_trace_factory.dart
rxlabz: Is `@visibleForTest` acceptable?

dart/lib/src/sentry_client.dart
rxlabz: ✅

dart/lib/src/protocol/mechanism.dart
```diff
@@ -0,0 +1,82 @@
...
+  /// interpretation of these codes. SDKs can safely omit code names and
+  /// descriptions for well known error codes, as it will be filled out by
+  /// Sentry. For proprietary or vendor-specific error codes,
+  /// adding these values will give additional information to the user.
+  final Map<String, dynamic> meta;
+
+  /// Arbitrary extra data that might help the user understand the error thrown by this mechanism
+  final Map<String, dynamic> data;
+
+  final bool synthetic;
```
marandaneto, line 44: missing comment

dart/lib/src/protocol/sentry_stack_trace.dart
marandaneto: we need a way to mutate the fields, afaik `copyWith` is the most used way on Dart and this is already being used on SentryEvent and other classes, I guess that would be the way to go, this has a performance cost though.

@bruno-garcia wdyt?

dart/lib/src/sentry_client.dart
```diff
@@ -68,15 +72,26 @@ class SentryClient {
...
+      serverName: event.serverName ?? _options.serverName,
+      dist: event.dist ?? _options.dist,
+      environment:
+          event.environment ?? _options.environment ?? defaultEnvironment,
+      release: event.release ?? _options.release,
+      sdk: event.sdk ?? _options.sdk,
+      platform: event.platform ?? sdkPlatform,
+    );
+
+    if (event.throwable != null) {
```
marandaneto:
```suggestion
    if (event.throwable != null && event.exception == null) {
```

we should not overwrite an exception set by the client

dart/lib/src/sentry_exception_factory.dart
rxlabz: ✅

dart/lib/src/protocol/sentry_stack_trace.dart
rxlabz: If we want a way to mutate the fields, why do we need to mark them as `final`?

dart/lib/src/sentry_stack_trace_factory.dart
marandaneto: yeah we used in Java as well, but if possible to test `getStackFrames` that forwards the call to `encodeStackTraceFrame`, it'd not be needed.
If the test already exists and it works, let's visibleForTest

dart/lib/src/protocol/sentry_stack_trace.dart
marandaneto: either we make them public and mutable (losing encapsulation) or we make it immutable and a `copyWith` extension, in Dart, the latter sounds the more correct, do we have another option?

dart/lib/src/protocol/sentry_stack_frame.dart
```diff
@@ -0,0 +1,182 @@
...
+    this.package,
+    this.native,
+    this.platform,
+    this.imageAddr,
+    this.symbolAddr,
+    this.instructionAddr,
+    this.rawFunction,
+    this.origin = '',
+    List<int> framesOmitted,
+    List<String> preContrxt,
```
marandaneto:
```suggestion
    List<String> preContext,
```

marandaneto: also missing `_ preContext = preContext;`

dart/lib/src/protocol/mechanism.dart
```diff
@@ -0,0 +1,82 @@
...
+  /// The mechanism meta data usually carries error codes reported by
+  /// the runtime or operating system, along with a platform dependent
+  /// interpretation of these codes. SDKs can safely omit code names and
+  /// descriptions for well known error codes, as it will be filled out by
+  /// Sentry. For proprietary or vendor-specific error codes,
+  /// adding these values will give additional information to the user.
+  final Map<String, dynamic> meta;
+
+  /// Arbitrary extra data that might help the user understand the error thrown by this mechanism
+  final Map<String, dynamic> data;
```
marandaneto: missing a getter with unmodified map

dart/lib/src/protocol/request.dart
```diff
@@ -0,0 +1,84 @@
...
+  /// If a header appears multiple times it,
+  /// needs to be merged according to the HTTP standard for header merging.
+  /// Header names are treated case-insensitively by Sentry.
+  final Map<String, String> headers;
+
+  /// A dictionary containing environment information passed from the server.
+  /// This is where information such as CGI/WSGI/Rack keys go that are not HTTP headers.
+  final Map<String, String> env;
+
+  final Map<String, String> other;
```
marandaneto: missing a getter with unmodified map

dart/lib/src/protocol/request.dart
```diff
@@ -0,0 +1,84 @@
...
+  /// The query string component of the URL.
+  ///
+  /// If the query string is not declared and part of the url parameter,
+  /// Sentry moves it to the query string.
+  final String queryString;
+
+  /// Submitted data in a format that makes the most sense.
+  /// SDKs should discard large bodies by default.
+  /// Can be given as string or structural data of any format.
+  final dynamic data;
```
marandaneto: missing a readonly getter like map or list

dart/lib/src/protocol/sentry_stack_frame.dart
```diff
@@ -0,0 +1,182 @@
...
+  }) : _framesOmitted = framesOmitted;
+
+  /// The absolute path to filename.
+  final String absPath;
+
+  /// A list of source code lines before context_line (in order) – usually [lineno - 5:lineno].
+  final List<String> preContext;
+
+  /// A list of source code lines after context_line (in order) – usually [lineno + 1:lineno + 5].
+  final List<String> postContext;
```
marandaneto: missing a getter with unmodified list

dart/lib/src/sentry_exception_factory.dart
```diff
@@ -0,0 +1,37 @@
...
+    SentryStackTraceFactory stacktraceFactory = const SentryStackTraceFactory(),
+  }) : _stacktraceFactory = stacktraceFactory;
+
+  SentryException getSentryException(
+    dynamic exception, {
+    dynamic stackTrace,
+    Mechanism mechanism,
+  }) {
+    if (stackTrace == null && exception is Error) {
+      stackTrace = exception.stackTrace;
```
marandaneto:
```suggestion
    if (stackTrace == null && exception is Error) {
      stackTrace = exception.stackTrace;
```
```suggestion
    if (exception is Error) {
      stackTrace ??= exception.stackTrace;
```

marandaneto: if a `stackTrace` was given and it's not null, we should keep as it is, I believe this would make more sense, wdyt

dart/lib/src/sentry_stack_trace_factory.dart
```diff
@@ -0,0 +1,64 @@
...
+
+import 'protocol/noop_origin.dart'
+    if (dart.library.html) 'protocol/origin.dart';
+import 'protocol/sentry_stack_frame.dart';
+
+class SentryStackTraceFactory {
+  const SentryStackTraceFactory();
+
+  List<SentryStackFrame> getStackFrames(dynamic stackTrace) {
+    if (stackTrace == null) return [];
```
marandaneto:
```suggestion
    if (stackTrace == null) return null;
```

we should always serialize something to null if its empty

marandaneto: https://github.com/getsentry/sentry-java/blob/1ad9403c9c50396de895b0ead47d5f2e4ba5a0a7/sentry/src/main/java/io/sentry/SentryStackTraceFactory.java#L33

dart/lib/src/transport/http_transport.dart
```diff
@@ -61,7 +60,7 @@ class HttpTransport implements Transport {
     } else {
       _options.logger(
         SentryLevel.debug,
-        'Event ${event.eventId} was sent suceffully.',
+        'Event ${event.eventId} was sent successfully.',
```
marandaneto: merge main into this branch, this is fixed already

marandaneto (commented):
left a few comments

dart/lib/sentry.dart
rxlabz: ✅

dart/lib/src/protocol/sentry_stack_frame.dart
rxlabz: ✅

dart/lib/src/protocol/sentry_stack_frame.dart
rxlabz: ✅

dart/lib/src/protocol/sentry_stack_frame.dart
rxlabz: ✅

dart/lib/src/protocol/sentry_stack_frame.dart
rxlabz: ✅  `field != null && field.isNotEmpty` is probably clearer

dart/lib/src/protocol/sentry_stack_frame.dart
rxlabz: ✅  immutable w/ `copyWith`

dart/lib/src/protocol/sentry_stack_trace.dart
rxlabz: ✅

dart/lib/src/sentry_exception_factory.dart
rxlabz: ✅

dart/lib/src/sentry_stack_trace_factory.dart
rxlabz: ✅

dart/lib/src/sentry_stack_trace_factory.dart
rxlabz: ✅

dart/lib/src/protocol/sentry_stack_frame.dart
rxlabz: ✅

dart/lib/src/protocol/mechanism.dart
rxlabz: ✅

dart/lib/src/sentry_exception_factory.dart
rxlabz: ✅

dart/lib/src/sentry_stack_trace_factory.dart
rxlabz: ✅

dart/lib/src/transport/http_transport.dart
rxlabz: ✅

dart/lib/src/protocol/mechanism.dart
rxlabz: ✅

dart/lib/src/protocol/request.dart
rxlabz: ✅

dart/lib/src/protocol/sentry_stack_frame.dart
rxlabz: ✅

dart/lib/src/protocol/request.dart
rxlabz: ✅

dart/lib/src/protocol/sentry_stack_frame.dart
rxlabz: ✅

dart/lib/src/sentry_client.dart

dart/lib/src/sentry_stack_trace_factory.dart
rxlabz: ✅

dart/lib/src/sentry_stack_trace_factory.dart
rxlabz: ✅

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 ## 4.0.0-alpha.2
 
 - feat: add contexts to scope
+- feat: add missing protocol classes : Request, Mechanism
```
marandaneto: I'd mention the other classes too or remove the specification (Request, Mechanism)

dart/lib/src/protocol/sentry_event.dart
```diff
@@ -238,29 +260,7 @@ class SentryEvent {
...
-    }
-
-    if (stackTrace != null) {
-      json['stacktrace'] = <String, dynamic>{
-        'frames': encodeStackTrace(
-          stackTrace,
-          origin: origin,
-        ),
-      };
+      json['exception'] = [exception.toJson()];
```
marandaneto: the exception doesn't need to be a list because of Dart, theres no nested exceptions, but we need to render in a list way because of serialization on the other native SDKs.
eg:
exception {values [ {exception...}]}
see workaround here https://github.com/getsentry/sentry-dart/pull/158/

dart/lib/src/protocol/sentry_stack_frame.dart
```diff
@@ -0,0 +1,229 @@
...
+
+  /// Signifies whether this frame is related to the execution of the relevant code in this stacktrace.
+  ///
+  /// For example, the frames that might power the framework’s web server of your app are probably not relevant, however calls to the framework’s library once you start handling code likely are.
+  final bool inApp;
+
+  /// The "package" the frame was contained in.
+  final String package;
+
+  final bool native;
```
marandaneto, line 87: is `native` a reserved word in Dart? or just GH making it red?

dart/lib/src/sentry_client.dart
```diff
@@ -22,11 +23,14 @@ class SentryClient {
...
   final Random _random;
 
+  final SentryExceptionFactory _exceptionFactory;
+
   static final _sentryId = Future.value(SentryId.empty());
 
   /// Instantiates a client using [SentryOptions]
   SentryClient._(this._options)
-      : _random = _options.sampleRate == null ? null : Random();
+      : _exceptionFactory = SentryExceptionFactory(options: _options),
```
marandaneto: how would you inject a `SentryExceptionFactory` for testing reasons?

marandaneto: can we consider an optional ctor param with a default value?

dart/lib/src/sentry_stack_trace_factory.dart
```diff
@@ -0,0 +1,112 @@
...
+  /// default.
+  final List<String> _inAppExcludes;
+
+  /// A list of string prefixes of module names that belong to the app. This option takes precedence
+  /// over inAppExcludes.
+  final List<String> _inAppIncludes;
+
+  SentryStackTraceFactory(SentryOptions options)
+      : _inAppExcludes = options?.inAppExcludes,
+        _inAppIncludes = options?.inAppIncludes;
```
marandaneto: options should be mandatory, why do we use `?`

dart/lib/src/sentry_stack_trace_factory.dart
```diff
@@ -0,0 +1,112 @@
...
+      absPath: '$eventOrigin${_absolutePathForCrashReport(frame)}',
+      function: frame.member,
+      lineNo: frame.line,
+      colNo: frame.column,
+      inApp: isInApp(frame),
+      fileName: fileName,
+      package: frame.package,
+      // TODO platform, postContext, preContext, rawFunction,
+      // TODO ? module, native, frame, contextLine, framesOmitted, imageAddr,... ?
+      // module: frame.library,
```
marandaneto: is this a TODO or the language doesn't support those things? I'd change the comment if so

dart/lib/src/protocol/sentry_event.dart
rxlabz: ✅ renamed "throwable" everywhere

dart/lib/src/sentry_stack_trace_factory.dart
```diff
@@ -0,0 +1,112 @@
...
+    }
+    if (_inAppExcludes != null) {
+      for (final exclude in _inAppExcludes) {
+        if (scheme.startsWith(exclude)) {
+          return false;
+        }
+      }
+    }
+
+    if (frame.isCore || frame.uri.scheme == 'package') return false;
```
marandaneto, line 117: don't we consider ` 'dart'` anymore?

dart/lib/src/protocol/sentry_stack_frame.dart
rxlabz: seems it's just GH

dart/lib/src/sentry_stack_trace_factory.dart
```diff
@@ -0,0 +1,112 @@
...
+  bool isInApp(Frame frame) {
+    final scheme = frame.uri.scheme;
+
+    if (scheme == null || scheme.isEmpty) {
+      return true;
+    }
+
+    if (_inAppIncludes != null) {
+      for (final include in _inAppIncludes) {
+        if (scheme.startsWith(include)) {
```
marandaneto: as I understood, every scheme starts with `file:`, `dart:`, `package:`, does it make sense to be `startsWith`? I'd debug this to be sure its the right implementation

dart/lib/src/sentry_client.dart
```dart
SentryClient._(this._options, {SentryExceptionFactory exceptionFactory})
      : _exceptionFactory =
            exceptionFactory ?? SentryExceptionFactory(options: _options),
```

dart/lib/src/sentry_stack_trace_factory.dart
rxlabz: as Dart is not non-nullable for now, options can be passed null

marandaneto: thats why we do `Objects.requireNonNull` on java and throws if null, we'd do the same in Dart, this would mean a bug on the SDK and we should fix it.

dart/lib/src/sentry_stack_trace_factory.dart
rxlabz: it's more a question, in java, the `getStackFrames` doesn't set all this fields, so is it to do ?

dart/lib/src/sentry_stack_trace_factory.dart
rxlabz: `frame.isCore` returns frame.uri.scheme == 'dart'

dart/lib/src/sentry_stack_trace_factory.dart
rxlabz: yes, I need to implement some tests to be sure

CHANGELOG.md
rxlabz: ✅

dart/lib/src/sentry_stack_trace_factory.dart
rxlabz: ✅  I added non nullability verifications

dart/lib/src/protocol/sentry_event.dart

dart/lib/src/scope.dart
```diff
@@ -81,7 +81,11 @@ class Scope {
 
   final SentryOptions _options;
 
-  Scope(this._options) : assert(_options != null, 'SentryOptions is required');
+  Scope(this._options) {
+    if (_options == null) {
+      throw ArgumentError('SentryOptions is required');
```
marandaneto, line 86: u were using `ArgumentError.notNull` but now `ArgumentError(message)`, can we chose one approach and go for it?

rxlabz: ✅ I just replaced the last ArgumentError.notNull(name) by ArgumentError(message)

dart/lib/src/sentry_stack_trace_factory.dart
marandaneto: I see, thanks

dart/lib/src/sentry_stack_trace_factory.dart
marandaneto: only if this exists in the language, the protocol is big enough to support many languages, if its not the case in Dart, we could even remove the comments

rxlabz: ✅  i removed the comment

dart/lib/src/sentry_stack_trace_factory.dart
rxlabz: ✅  finally we test  if `frame.package.startWith...`

dart/lib/src/sentry_stack_trace_factory.dart
```diff
@@ -0,0 +1,121 @@
...
+  bool isInApp(Frame frame) {
+    final scheme = frame.uri.scheme;
+
+    if (scheme == null || scheme.isEmpty) {
+      return true;
+    }
+
+    if (_inAppIncludes != null) {
+      for (final include in _inAppIncludes) {
+        if (frame.package.startsWith(include)) {
```
marandaneto: `package` can be null I guess
https://pub.dev/documentation/stack_trace/latest/stack_trace/Frame/package.html

rxlabz: ✅good catch !

dart/test/test_utils.dart
rxlabz: ✅

dart/test/sentry_event_test.dart
rxlabz: ✅

dart/lib/src/protocol/sentry_event.dart
```diff
@@ -1,9 +1,9 @@
 import 'package:meta/meta.dart';
 
 import '../protocol.dart';
-import '../stack_trace.dart';
 import '../utils.dart';
 import '../version.dart';
+import 'request.dart';
```
marandaneto: should this not be automatically imported by `'../protocol.dart';`?

dart/lib/src/sentry_exception_factory.dart
```diff
@@ -0,0 +1,47 @@
+import 'package:meta/meta.dart';
+
+import 'protocol/sentry_stack_trace.dart';
```
marandaneto: this seems redundant if u imported `'protocol.dart';`

PR #159, opened 2020-11-04
fix: sentry protocol is v7

marandaneto:

- [X] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
fix: sentry protocol is v7

headers were already meeting protocol v7 but it was still sending v6

## :pencil: Checklist

- [X] All tests passing
- [X] No breaking changes

### DIFF
--- a/dart/lib/src/transport/http_transport.dart
+++ b/dart/lib/src/transport/http_transport.dart
@@ -69,7 +69,7 @@ class HttpTransport implements Transport {
     } else {
       _options.logger(
         SentryLevel.debug,
-        'Event ${event.eventId} was sent suceffully.',
+        'Event ${event.eventId} was sent successfully.',
       );
     }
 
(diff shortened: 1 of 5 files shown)


dart/lib/src/utils.dart
```diff
@@ -16,4 +16,4 @@ String formatDateAsIso8601WithSecondPrecision(DateTime date) {
 }
 
 /// helper to detect a browser context
-const isWeb = identical(1.0, 1);
+const isWeb = identical(0, 0.0);
```
marandaneto, line 19: just to be the same as they use in flutter

dart/lib/src/transport/http_transport.dart
```diff
@@ -69,7 +69,7 @@ class HttpTransport implements Transport {
     } else {
       _options.logger(
         SentryLevel.debug,
-        'Event ${event.eventId} was sent suceffully.',
+        'Event ${event.eventId} was sent successfully.',
```
marandaneto, line 72: fixing a typo on the way

dart/lib/src/transport/http_transport.dart
```diff
@@ -103,11 +103,12 @@ class _CredentialBuilder {
       : _authHeader = authHeader,
         _clock = clock;
 
-  factory _CredentialBuilder(Dsn dsn, String clientId, ClockProvider clock) {
+  factory _CredentialBuilder(
```
marandaneto, line 106: renamed to align with the other methods that also takes a `sdkIdentifier` as param

PR #160, opened 2020-11-10
add options.attachStacktrace

rxlabz:

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

- add `SentryOptions.attachStackTrace` to attach stacktrace on captureMessage and captureEvent ( true by default )
- filter 'package:sentry' frames

new tests to verify if stacktrace is correctly sended

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/dart/lib/src/hub.dart
+++ b/dart/lib/src/hub.dart
@@ -54,7 +54,11 @@ class Hub {
   SentryId get lastEventId => _lastEventId;
 
   /// Captures the event.
-  Future<SentryId> captureEvent(SentryEvent event, {dynamic hint}) async {
+  Future<SentryId> captureEvent(
+    SentryEvent event, {
+    dynamic stackTrace,
+    dynamic hint,
+  }) async {
     var sentryId = SentryId.empty();
 
     if (!_isEnabled) {
@@ -73,6 +77,7 @@ class Hub {
         try {
           sentryId = await item.client.captureEvent(
             event,
+            stackTrace: stackTrace,
             scope: item.scope,
             hint: hint,
           );
--- a/dart/lib/src/protocol/sentry_event.dart
+++ b/dart/lib/src/protocol/sentry_event.dart
@@ -1,5 +1,3 @@
-import 'dart:isolate';
-
 import 'package:meta/meta.dart';
 
 import '../protocol.dart';
@@ -263,15 +261,14 @@ class SentryEvent {
       json['exception'] = {
         'values': [exception.toJson()].toList(growable: false)
       };
-    }
-
-    if (stackTrace != null) {
+    } else if (stackTrace != null) {
       json['threads'] = {
         'values': [
           {
-            'id': Isolate.current.debugName,
+            'id': 0,
             'stacktrace': stackTrace.toJson(),
             'crashed': true,
+            'name': 'Current Isolate',
           }
         ]
       };
--- a/dart/lib/src/protocol/sentry_event.dart
+++ b/dart/lib/src/protocol/sentry_event.dart
@@ -83,8 +83,8 @@ class SentryEvent {
   /// If this behavior is undesirable, consider using a custom formatted [message] instead.
   final dynamic throwable;
 
-  /// The stack trace corresponding to the thrown [exception] or attachedStackTrace
-  /// see [SentryOptions.attachStackTrace]
+  /// an optional attached StackTrace
+  /// used when event has no throwable or exception, see [SentryOptions.attachStackTrace]
   final SentryStackTrace stackTrace;
 
   /// an exception or error that occurred in a program
--- a/dart/lib/src/protocol/sentry_event.dart
+++ b/dart/lib/src/protocol/sentry_event.dart
@@ -1,3 +1,5 @@
+import 'dart:isolate';
+
 import 'package:meta/meta.dart';
 
 import '../protocol.dart';
@@ -83,10 +85,9 @@ class SentryEvent {
   /// If this behavior is undesirable, consider using a custom formatted [message] instead.
   final dynamic throwable;
 
-  /// The stack trace corresponding to the thrown [exception].
-  ///
-  /// Can be `null`, a [String], or a [StackTrace].
-  final dynamic stackTrace;
+  /// The stack trace corresponding to the thrown [exception] or attachedStackTrace
+  /// see [SentryOptions.attachStackTrace]
+  final SentryStackTrace stackTrace;
 
   /// an exception or error that occurred in a program
   /// TODO more doc
@@ -262,12 +263,15 @@ class SentryEvent {
       json['exception'] = {
         'values': [exception.toJson()].toList(growable: false)
       };
-    } else if (stackTrace is SentryStackTrace) {
+    }
+
+    if (stackTrace != null) {
       json['threads'] = {
         'values': [
           {
-            'id': 0,
-            'stacktrace': (stackTrace as SentryStackTrace).toJson(),
+            'id': Isolate.current.debugName,
+            'stacktrace': stackTrace.toJson(),
+            'crashed': true,
           }
         ]
       };
--- a/dart/lib/src/protocol/sentry_event.dart
+++ b/dart/lib/src/protocol/sentry_event.dart
@@ -263,7 +263,14 @@ class SentryEvent {
         'values': [exception.toJson()].toList(growable: false)
       };
     } else if (stackTrace is SentryStackTrace) {
-      json['stacktrace'] = (stackTrace as SentryStackTrace).toJson();
+      json['threads'] = {
+        'values': [
+          {
+            'id': 0,
+            'stacktrace': (stackTrace as SentryStackTrace).toJson(),
+          }
+        ]
+      };
     }
 
     if (level != null) {
--- a/dart/lib/src/protocol/sentry_event.dart
+++ b/dart/lib/src/protocol/sentry_event.dart
@@ -262,6 +262,8 @@ class SentryEvent {
       json['exception'] = {
         'values': [exception.toJson()].toList(growable: false)
       };
+    } else if (stackTrace is SentryStackTrace) {
+      json['stacktrace'] = (stackTrace as SentryStackTrace).toJson();
     }
 
     if (level != null) {
--- a/dart/lib/src/sentry_client.dart
+++ b/dart/lib/src/sentry_client.dart
@@ -2,12 +2,12 @@ import 'dart:async';
 import 'dart:math';
 
 import 'package:meta/meta.dart';
-import 'package:sentry/src/sentry_stack_trace_factory.dart';
 
 import 'protocol.dart';
 import 'scope.dart';
 import 'sentry_exception_factory.dart';
 import 'sentry_options.dart';
+import 'sentry_stack_trace_factory.dart';
 import 'transport/http_transport.dart';
 import 'transport/noop_transport.dart';
 import 'version.dart';
--- a/dart/lib/src/sentry_client.dart
+++ b/dart/lib/src/sentry_client.dart
@@ -101,7 +101,9 @@ class SentryClient {
       platform: event.platform ?? sdkPlatform,
     );
 
-    if (stackTrace != null && event.throwable == null) {
+    if (stackTrace != null &&
+        event.throwable == null &&
+        event.exception == null) {
       final frames = _stackTraceFactory.getStackFrames(stackTrace);
 
       if (frames != null && frames.isNotEmpty) {
--- a/dart/lib/src/sentry_client.dart
+++ b/dart/lib/src/sentry_client.dart
@@ -138,14 +138,9 @@ class SentryClient {
       throwable: throwable,
       timestamp: _options.clock(),
     );
-    if (stackTrace != null) {
-      event = event.copyWith(
-        stackTrace: SentryStackTrace(
-          frames: _stackTraceFactory.getStackFrames(stackTrace),
-        ),
-      );
-    }
-    return captureEvent(event, scope: scope, hint: hint);
+
+    return captureEvent(event,
+        stackTrace: stackTrace, scope: scope, hint: hint);
   }
 
   /// Reports the [template]
--- a/dart/lib/src/sentry_client.dart
+++ b/dart/lib/src/sentry_client.dart
@@ -129,7 +129,7 @@ class SentryClient {
     Scope scope,
     dynamic hint,
   }) {
-    var event = SentryEvent(
+    final event = SentryEvent(
       throwable: throwable,
       timestamp: _options.clock(),
     );
--- a/dart/lib/src/sentry_client.dart
+++ b/dart/lib/src/sentry_client.dart
@@ -5,6 +5,7 @@ import 'protocol.dart';
 import 'scope.dart';
 import 'sentry_exception_factory.dart';
 import 'sentry_options.dart';
+import 'sentry_stack_trace_factory.dart';
 import 'transport/http_transport.dart';
 import 'transport/noop_transport.dart';
 import 'version.dart';
@@ -97,6 +98,15 @@ class SentryClient {
           .getSentryException(event.throwable, stackTrace: event.stackTrace);
 
       event = event.copyWith(exception: sentryException);
+    }
+
+    if (_options.attachStacktrace &&
+        event.throwable == null &&
+        event.exception == null) {
+      final frames =
+          SentryStackTraceFactory(_options).getStackFrames(StackTrace.current);
+
+      event = event.copyWith(stackTrace: SentryStackTrace(frames: frames));
     }
 
     return event;
--- a/dart/lib/src/sentry_client.dart
+++ b/dart/lib/src/sentry_client.dart
@@ -114,10 +114,9 @@ class SentryClient {
           .getSentryException(event.throwable, stackTrace: stackTrace);
 
       event = event.copyWith(exception: sentryException);
-    } else if (stackTrace == null &&
-        _options.attachStackTrace &&
-        event.exception == null) {
-      final frames = _stackTraceFactory.getStackFrames(StackTrace.current);
+    } else if (_options.attachStackTrace && event.exception == null) {
+      stackTrace ??= StackTrace.current;
+      final frames = _stackTraceFactory.getStackFrames(stackTrace);
 
       if (frames != null && frames.isNotEmpty) {
         event = event.copyWith(stackTrace: SentryStackTrace(frames: frames));
--- a/dart/lib/src/sentry_client.dart
+++ b/dart/lib/src/sentry_client.dart
@@ -111,9 +111,8 @@ class SentryClient {
           .getSentryException(event.throwable, stackTrace: event.stackTrace);
 
       event = event.copyWith(exception: sentryException);
-    }
-
-    if (_options.attachStackTrace &&
+    } else if (_options.attachStackTrace &&
+        event.stackTrace == null &&
         event.throwable == null &&
         event.exception == null) {
       final frames = _stackTraceFactory.getStackFrames(StackTrace.current);
--- a/dart/lib/src/sentry_client.dart
+++ b/dart/lib/src/sentry_client.dart
@@ -118,7 +118,9 @@ class SentryClient {
         event.exception == null) {
       final frames = _stackTraceFactory.getStackFrames(StackTrace.current);
 
-      event = event.copyWith(stackTrace: SentryStackTrace(frames: frames));
+      if (frames != null && frames.isNotEmpty) {
+        event = event.copyWith(stackTrace: SentryStackTrace(frames: frames));
+      }
     }
 
     return event;
--- a/dart/lib/src/sentry_client.dart
+++ b/dart/lib/src/sentry_client.dart
@@ -5,7 +5,6 @@ import 'protocol.dart';
 import 'scope.dart';
 import 'sentry_exception_factory.dart';
 import 'sentry_options.dart';
-import 'sentry_stack_trace_factory.dart';
 import 'transport/http_transport.dart';
 import 'transport/noop_transport.dart';
 import 'version.dart';
@@ -103,8 +102,8 @@ class SentryClient {
     if (_options.attachStackTrace &&
         event.throwable == null &&
         event.exception == null) {
-      final frames =
-          SentryStackTraceFactory(_options).getStackFrames(StackTrace.current);
+      final frames = _exceptionFactory.stacktraceFactory
+          .getStackFrames(StackTrace.current);
 
       event = event.copyWith(stackTrace: SentryStackTrace(frames: frames));
     }
--- a/dart/lib/src/sentry_client.dart
+++ b/dart/lib/src/sentry_client.dart
@@ -107,7 +107,6 @@ class SentryClient {
       event = event.copyWith(exception: sentryException);
     } else if (_options.attachStackTrace &&
         event.stackTrace == null &&
-        event.throwable == null &&
         event.exception == null) {
       final frames = _stackTraceFactory.getStackFrames(StackTrace.current);
 
--- a/dart/lib/src/sentry_client.dart
+++ b/dart/lib/src/sentry_client.dart
@@ -1,8 +1,6 @@
 import 'dart:async';
 import 'dart:math';
 
-import 'package:meta/meta.dart';
-
 import 'protocol.dart';
 import 'scope.dart';
 import 'sentry_exception_factory.dart';
@@ -18,11 +16,11 @@ class SentryClient {
 
   final Random _random;
 
-  final SentryExceptionFactory _exceptionFactory;
-
-  final SentryStackTraceFactory _stackTraceFactory;
-
   static final _sentryId = Future.value(SentryId.empty());
+
+  SentryExceptionFactory _exceptionFactory;
+
+  SentryStackTraceFactory _stackTraceFactory;
 
   /// Instantiates a client using [SentryOptions]
   factory SentryClient(SentryOptions options) {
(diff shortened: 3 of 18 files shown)


dart/lib/src/sentry_client.dart
```diff
@@ -99,6 +100,15 @@ class SentryClient {
       event = event.copyWith(exception: sentryException);
     }
 
+    if (_options.attachStackTrace &&
```
rxlabz: should we test if event.stackTrace is `null` ?

dart/lib/src/sentry_stack_trace_factory.dart
```diff
@@ -101,14 +103,14 @@ class SentryStackTraceFactory {
 
     if (_inAppIncludes != null) {
       for (final include in _inAppIncludes) {
-        if (frame.package != null && frame.package.startsWith(include)) {
+        if (frame.package != null && frame.package == include) {
```
rxlabz, line 108: as we get a package name I'm not sure that `startWith` is the better option ?

dart/lib/src/protocol/sentry_event.dart
```diff
@@ -262,6 +262,8 @@ class SentryEvent {
       json['exception'] = {
         'values': [exception.toJson()].toList(growable: false)
       };
+    } else if (stackTrace is SentryStackTrace) {
+      json['stacktrace'] = (stackTrace as SentryStackTrace).toJson();
```
marandaneto: why this change? an exception has a stacktrace field already

rxlabz: it's for case we want attach a stacktrace without exception ( for captureMessage and captureEvent )

marandaneto: guess this would be necessary for captureMessage() as theres no exception

rxlabz: ooops I just read that stacktrace should never be a top level property

marandaneto: maybe we add one more field to the SentryEvent, because the docs about the stackTrace:

`Can be `null`, a [String], or a [StackTrace].`

it'd be confusing to add one more type, same as we have `throwable` and `exception`
wdyt?

dart/lib/src/sentry_client.dart
```diff
@@ -99,6 +100,15 @@ class SentryClient {
...
+        event.throwable == null &&
+        event.exception == null) {
+      final frames =
+          SentryStackTraceFactory(_options).getStackFrames(StackTrace.current);
```
marandaneto: please find a way to reuse the instance of `SentryStackTraceFactory`.

`SentryExceptionFactory` already creates an instance of it, it'd be easier to inject/test and also to not allocate memory on each call to captureEvent

dart/lib/src/sentry_stack_trace_factory.dart
marandaneto: that's what I said before, on Java it makes sense because we have packages.like.this and inner packages belong to the parent package

marandaneto:
missing this:

https://github.com/getsentry/sentry-dart/blob/main/dart/lib/src/sentry_exception_factory.dart#L30

this assignment should only be made if attachStacktrace is also enabled

dart/lib/src/protocol/sentry_event.dart
marandaneto: other solution would be to add the threads list

dart/lib/src/sentry_exception_factory.dart
```diff
@@ -8,6 +8,8 @@ import 'sentry_stack_trace_factory.dart';
 class SentryExceptionFactory {
   SentryStackTraceFactory _stacktraceFactory;
 
+  SentryStackTraceFactory get stacktraceFactory => _stacktraceFactory;
```
marandaneto: I'd not go this route, way better if SentryClient creates the instance and pass it along to SentryExceptionFactory, so both hold the reference and dont need to add a public getter, prefer DI sort of solution

dart/lib/src/sentry_client.dart

rxlabz:
> missing this:
> 
> https://github.com/getsentry/sentry-dart/blob/main/dart/lib/src/sentry_exception_factory.dart#L30
> 
> this assignment should only be made if attachStacktrace is also enabled

✅

dart/lib/src/sentry_exception_factory.dart
rxlabz: ✅

dart/lib/src/sentry_client.dart
```diff
@@ -1,6 +1,9 @@
 import 'dart:async';
 import 'dart:math';
 
+import 'package:meta/meta.dart';
+import 'package:sentry/src/sentry_stack_trace_factory.dart';
```
marandaneto: avoid package:

dart/lib/src/protocol/sentry_event.dart
rxlabz: what about a `event.attachedStackTrace` property ? cause a `threads` interface only for a single property could also be confusing

dart/lib/src/sentry_client.dart
```diff
@@ -20,21 +33,22 @@ class SentryClient {
...
-  final SentryOptions _options;
-
-  final Random _random;
-
-  final SentryExceptionFactory _exceptionFactory;
-
-  static final _sentryId = Future.value(SentryId.empty());
+    return SentryClient._(
+      options,
+      stackTraceFactory: SentryStackTraceFactory(options),
```
marandaneto: why dont you do this inside of the private ctor like the `SentryExceptionFactory`? seems that each field is inited in a different place.

dart/lib/src/sentry_client.dart
```diff
@@ -99,6 +113,14 @@ class SentryClient {
       event = event.copyWith(exception: sentryException);
     }
 
+    if (_options.attachStackTrace &&
+        event.throwable == null &&
+        event.exception == null) {
+      final frames = _stackTraceFactory.getStackFrames(StackTrace.current);
+
+      event = event.copyWith(stackTrace: SentryStackTrace(frames: frames));
```
marandaneto: frames can be null or empty, we dont need to copyWIth if its the case

dart/lib/src/sentry_options.dart
```diff
@@ -162,6 +162,14 @@ class SentryOptions {
     _sdk = sdk ?? _sdk;
   }
 
+  bool _attachStackTrace = true;
+
+  bool get attachStackTrace => _attachStackTrace;
```
marandaneto, line 176: docs

dart/lib/src/sentry_client.dart
rxlabz: we need an "existing" instance of SentryStackTraceFactory to instantiate the SentryExceptionFactory

rxlabz: we can't do 
```dart
SentryClient._(this._options)  : _stackTraceFactory = stackTraceFactory,
        _exceptionFactory = SentryExceptionFactory(
          options: _options,
          stacktraceFactory: _stackTraceFactory,
        ),
```

dart/lib/src/protocol/sentry_event.dart
marandaneto: this is just a toJson method, it should not have any business logic here

dart/lib/src/sentry_client.dart

dart/lib/src/sentry_client.dart
marandaneto: can we do both assignments then in the method body? make it non final if necessary, I think it'd be better

dart/lib/src/sentry_options.dart
rxlabz: ✅

dart/lib/src/sentry_client.dart

dart/lib/src/sentry_client.dart
marandaneto: yes only if its null, but actually I'd do this as `else if` with the exceptions check above, its either the exception with a stacktrace or a threads.stacktrace, not both, this way makes possible that we have a exception.stacktrace and still a threads.thread.stacktrace which is wrong.

dart/lib/src/sentry_client.dart

dart/lib/src/protocol/sentry_event.dart
rxlabz: I meant adding a `attachedStackTrace ` field to SentryEvent
 
```dart 
if (exception != null) {
      json['exception'] = {
        'values': [exception.toJson()].toList(growable: false)
      };
} else if (attachedStackTrace != null) {
      json['threads'] = {
        'values': [
          {
            'id': 0,
            'stacktrace': (stackTrace as SentryStackTrace).toJson(),
          }
        ]
      };
    }
```
WDYT?

dart/lib/src/sentry_client.dart
```diff
@@ -20,22 +31,19 @@ class SentryClient {
...
-  final SentryExceptionFactory _exceptionFactory;
-
-  static final _sentryId = Future.value(SentryId.empty());
-
   /// Instantiates a client using [SentryOptions]
-  SentryClient._(this._options, {SentryExceptionFactory exceptionFactory})
-      : _exceptionFactory =
-            exceptionFactory ?? SentryExceptionFactory(options: _options),
-        _random = _options.sampleRate == null ? null : Random();
+  SentryClient._(this._options)
```
marandaneto, line 39: they (SentryStackTraceFactory, SentryExceptionFactory) should still be optional param for testing I guess? but only assign a value if they are null

dart/lib/src/sentry_client.dart
```diff
@@ -97,6 +105,15 @@ class SentryClient {
           .getSentryException(event.throwable, stackTrace: event.stackTrace);
 
       event = event.copyWith(exception: sentryException);
+    } else if (_options.attachStackTrace &&
+        event.stackTrace == null &&
+        event.throwable == null &&
```
marandaneto: at this point, `event.throwable` is not even used anymore, so not worth checking it

dart/lib/src/sentry_exception_factory.dart
```diff
@@ -6,17 +6,21 @@ import 'sentry_stack_trace_factory.dart';
...
     @required SentryOptions options,
-  }) {
-    if (options == null) {
+  }) : _options = options {
+    if (_options == null) {
       throw ArgumentError('SentryOptions is required.');
     }
 
-    _stacktraceFactory = stacktraceFactory ?? SentryStackTraceFactory(options);
+    _stacktraceFactory = stacktraceFactory ?? SentryStackTraceFactory(_options);
```
marandaneto: maybe `SentryStackTraceFactory` should be required too and throw ArgumentError if null, as SentryClient will take care of create a instance of SentryStackTraceFactory

dart/lib/src/sentry_client.dart

dart/lib/src/sentry_exception_factory.dart
rxlabz: ✅

dart/lib/src/sentry_client.dart
rxlabz: sorry but I'm not sure to understand, should we add factories param to SentryClient factory ctor and private ctor to allow users to pass their own factories ?

marandaneto: no, the reasoning is not to allow custom factories, the reason is to make it easily testable, if you can unit test it like this (without fakes or mocks), its fine.

dart/lib/src/protocol/sentry_event.dart
marandaneto: we dont need `attachedStackTrace` to know if we should make the `threads` list or not, this could be done by checking if the exception is null or not.
by big concern here is `stackTrace as SentryStackTrace` as mentioned before.

rxlabz: oh yes sorry I meant 
```dart
if (exception != null) {
      json['exception'] = {
        'values': [exception.toJson()].toList(growable: false)
      };
} else if (attachedStackTrace != null) {
      json['threads'] = {
        'values': [
          {
            'id': 0,
            'stacktrace': attachedStackTrace.toJson(),
          }
        ]
      };
    }
```

with `attachedStackTrace` being a SentryStackTrace.
Do you prefer the threads property options ?

dart/lib/src/sentry_client.dart
rxlabz: seems ok (for current tests)

dart/lib/src/protocol/sentry_event.dart
marandaneto: would be possible to make `stackTrace` strong typed as `SentryStackTrace`? and all the captureEvent, captureException etc takes the given stackTrace (from a try catch) as a melhod param, so we can pass the stackTrace along and we dont really need the stackTrace on the event being dynamic.

marandaneto: so then your implementation would work and we would not need to call it `attachedStackTrace` which is confusing, but only stackTrace

dart/lib/src/protocol/sentry_event.dart
```diff
@@ -262,6 +262,15 @@ class SentryEvent {
       json['exception'] = {
         'values': [exception.toJson()].toList(growable: false)
       };
+    } else if (stackTrace is SentryStackTrace) {
+      json['threads'] = {
+        'values': [
+          {
+            'id': 0,
+            'stacktrace': (stackTrace as SentryStackTrace).toJson(),
```
marandaneto: wondering if its worth to have the thread list or at least a thread object for it, because the threads object has more fields and we could use the `Isolate.current` to populate them, example, thread name with `Isolate.current.debugName`, we can create a card for it, but not really necessary now

marandaneto:
```suggestion
            'stacktrace': (stackTrace as SentryStackTrace).toJson(),
            'crashed': true,
            'name': 'Current Isolate',
```

dart/lib/src/protocol/sentry_event.dart
```diff
@@ -262,12 +263,15 @@ class SentryEvent {
...
-    } else if (stackTrace is SentryStackTrace) {
+    }
+
+    if (stackTrace != null) {
       json['threads'] = {
         'values': [
           {
-            'id': 0,
-            'stacktrace': (stackTrace as SentryStackTrace).toJson(),
+            'id': Isolate.current.debugName,
```
marandaneto: `debugName` is a String, `id` is an Int, why have you changed it?
threads have a `name` field

dart/lib/src/protocol/sentry_event.dart
```diff
@@ -83,10 +85,9 @@ class SentryEvent {
   /// If this behavior is undesirable, consider using a custom formatted [message] instead.
   final dynamic throwable;
 
-  /// The stack trace corresponding to the thrown [exception].
-  ///
-  /// Can be `null`, a [String], or a [StackTrace].
-  final dynamic stackTrace;
+  /// The stack trace corresponding to the thrown [exception] or attachedStackTrace
+  /// see [SentryOptions.attachStackTrace]
```
marandaneto: the docs are off, this is only used if the thrown `throwable` was null, the given `stackTrace` (method params) was also null and `attachStackTrace` is enabled

dart/lib/src/protocol/sentry_event.dart
```diff
@@ -262,12 +263,15 @@ class SentryEvent {
       json['exception'] = {
         'values': [exception.toJson()].toList(growable: false)
       };
-    } else if (stackTrace is SentryStackTrace) {
+    }
+
+    if (stackTrace != null) {
```
marandaneto: it was an `else if` and it was correct, not sure why you have changed it

dart/lib/src/protocol/sentry_event.dart
marandaneto: basically its only used for captureMessage() which doesnt have an exception to set the stackTrace

dart/lib/src/sentry_client.dart
```diff
@@ -97,6 +114,14 @@ class SentryClient {
           .getSentryException(event.throwable, stackTrace: event.stackTrace);
 
       event = event.copyWith(exception: sentryException);
+    } else if (_options.attachStackTrace &&
+        event.stackTrace == null &&
+        event.exception == null) {
+      final frames = _stackTraceFactory.getStackFrames(StackTrace.current);
```
marandaneto: you should use the given `stackTrace` if not null and only if null and `attachStacktrace` is enabled, you use `StackTrace.current`

dart/lib/src/sentry_client.dart
```diff
@@ -56,6 +65,14 @@ class SentryClient {
       _options.logger(SentryLevel.debug, 'No scope is defined');
     }
 
+    if (stackTrace != null) {
+      event = event.copyWith(
+        stackTrace: SentryStackTrace(
+          frames: _stackTraceFactory.getStackFrames(stackTrace),
+        ),
+      );
+    }
```
marandaneto: this looks like being done twice, here and https://github.com/getsentry/sentry-dart/pull/160/files#r521245971

dart/lib/src/sentry_client.dart
```diff
@@ -109,11 +134,17 @@ class SentryClient {
...
-      stackTrace: stackTrace,
       timestamp: _options.clock(),
     );
+    if (stackTrace != null) {
+      event = event.copyWith(
+        stackTrace: SentryStackTrace(
+          frames: _stackTraceFactory.getStackFrames(stackTrace),
+        ),
+      );
+    }
```
marandaneto: why this again here?

dart/lib/src/sentry_exception_factory.dart
```diff
@@ -6,39 +6,44 @@ import 'sentry_stack_trace_factory.dart';
...
-    _stacktraceFactory = stacktraceFactory ?? SentryStackTraceFactory(options);
+    if (_stacktraceFactory == null) {
+      throw ArgumentError('SentryStackTraceFactory is required.');
+    }
   }
 
   SentryException getSentryException(
     dynamic exception, {
-    dynamic stackTrace,
+    SentryStackTrace stackTrace,
```
marandaneto: this should still be `dynamic` I believe, not entirely sure, you assume that `exception.stackTrace` is always populated and its not the case.

`exception.stackTrace` should only be set if theres no `exception/throwable` at all

dart/lib/src/protocol/sentry_event.dart
marandaneto: the docs for `attachStackTrace` is pretty clear about that btw

dart/lib/src/protocol/sentry_event.dart

dart/lib/src/sentry_client.dart

dart/lib/src/sentry_client.dart
```diff
@@ -109,14 +101,18 @@ class SentryClient {
       platform: event.platform ?? sdkPlatform,
     );
 
-    if (event.throwable != null && event.exception == null) {
+    if (stackTrace != null && event.throwable == null) {
```
marandaneto: you should test `event.exception == null` here too

dart/lib/src/sentry_client.dart

dart/lib/src/sentry_client.dart

dart/lib/src/sentry_exception_factory.dart
```diff
@@ -6,34 +6,44 @@ import 'sentry_stack_trace_factory.dart';
...
+    if (stackTrace != null) {
+      sentryStackTrace ??= SentryStackTrace(
+        frames: _stacktraceFactory.getStackFrames(stackTrace),
+      );
+    } else if (exception is Error) {
+      sentryStackTrace ??= SentryStackTrace(
+        frames: _stacktraceFactory.getStackFrames(exception.stackTrace),
+      );
+    } else if (_options.attachStackTrace) {
+      sentryStackTrace ??= SentryStackTrace(
```
marandaneto: why do we use `??=` if the var was just declared and its always null and `else if` branches

dart/lib/src/sentry_client.dart
marandaneto: wrong condition IMO, the `else if` branch was right, but it's missing this:

if (stackTrace == null) stackTrace = StackTrace.current
and:
final frames = _stackTraceFactory.getStackFrames(stackTrace);

dart/lib/src/sentry_client.dart
```diff
@@ -92,11 +101,27 @@ class SentryClient {
...
 
       event = event.copyWith(exception: sentryException);
+    } else if (stackTrace == null &&
+        _options.attachStackTrace &&
+        event.exception == null) {
+      final frames = _stackTraceFactory.getStackFrames(StackTrace.current);
+
+      if (frames != null && frames.isNotEmpty) {
+        event = event.copyWith(stackTrace: SentryStackTrace(frames: frames));
+      }
```
marandaneto: I suggest extracting this to a method on `_stackTraceFactory`, here and `getSentryException` method kinda does the same thing to convern a dynamic stackTrace to a SentryStackTrace and looks like both are different.

what we need is:

if theres an `exception`, do not set `event.stackTrace`, end.

if there's a `throwable`, do not set `event.stackTrace`, because the exception already has most probably an exception, end.

if theres not a `throwable` nor a `exception`, lets set the `event.stackTrace` only if `attachStacktrace` is enabled.

for all the cases above, if the `stackTrace` is not null, use it, if its null, use `stackTrace.current`

dart/lib/src/sentry_exception_factory.dart
rxlabz: ✅

dart/lib/src/sentry_client.dart

dart/lib/src/sentry_client.dart
rxlabz: yeah it was messy. I updated to 
```dart
    if (event.exception != null) return event;

    if (event.throwable != null) {
      final sentryException = _exceptionFactory
          .getSentryException(event.throwable, stackTrace: stackTrace);

      return event.copyWith(exception: sentryException);
    }

    if (stackTrace != null || _options.attachStackTrace) {
      stackTrace ??= StackTrace.current;
      final frames = _stackTraceFactory.getStackFrames(stackTrace);

      if (frames != null && frames.isNotEmpty) {
        event = event.copyWith(stackTrace: SentryStackTrace(frames: frames));
      }
    }

    return event;
```
I'm not sure we really need to create another method. WDYT ?

dart/lib/src/sentry_client.dart
```diff
@@ -101,20 +101,16 @@ class SentryClient {
...
       final sentryException = _exceptionFactory
           .getSentryException(event.throwable, stackTrace: stackTrace);
 
-      event = event.copyWith(exception: sentryException);
-    } else if (_options.attachStackTrace && event.exception == null) {
+      return event.copyWith(exception: sentryException);
+    }
+
+    if (stackTrace != null || _options.attachStackTrace) {
       stackTrace ??= StackTrace.current;
```
marandaneto: this condition seems wrong.

`stackTrace` can be null, and if null, get the StackTrace.current only if attachStackTrace is enabled

dart/lib/src/sentry_client.dart
marandaneto:
```
if (stackTrace != null || _options.attachStackTrace) {
      stackTrace ??= StackTrace.current;
```

still wrong.
the code looks better now, I'd still extract to a method like I said before, DRY

dart/lib/src/sentry_client.dart
rxlabz: sorry I don't understand, here :
- if stacktrace is null AND _options.attachStackTrace is true => we use StackTrace.current
- if stacktrace is not null AND _options.attachStackTrace is false =>  we use stacktrace
- if stacktrace is not null AND _options.attachStackTrace is true =>  we use stacktrace
- if stacktrace is null AND _options.attachStackTrace is false =>  we don't set event.stacktrace
which case is wrong ?

marandaneto: actually its right, I read as `&&` instead of `||`, my bad sorry.
lets just be sure that we cover all the cases via unit tests

rxlabz: ✅ added few tests to be sure

dart/lib/src/sentry_client.dart
```diff
@@ -109,12 +129,17 @@ class SentryClient {
     Scope scope,
     dynamic hint,
   }) {
-    final event = SentryEvent(
+    var event = SentryEvent(
```
marandaneto: does it need to be a `var`?

dart/lib/src/sentry_exception_factory.dart
```diff
@@ -6,34 +6,44 @@ import 'sentry_stack_trace_factory.dart';
...
-    if (exception is Error) {
-      stackTrace ??= exception.stackTrace;
-    } else {
-      stackTrace ??= StackTrace.current;
+    SentryStackTrace sentryStackTrace;
+    if (stackTrace != null) {
+      sentryStackTrace = SentryStackTrace(
+        frames: _stacktraceFactory.getStackFrames(stackTrace),
+      );
+    } else if (exception is Error) {
```
marandaneto: maybe we should check that `exception.stackTrace` is not null as well, can it be null?

marandaneto (approved):
LGTM

dart/lib/src/hub.dart
```diff
@@ -73,6 +77,7 @@ class Hub {
         try {
           sentryId = await item.client.captureEvent(
             event,
+            stackTrace: stackTrace,
```
bruno-garcia, line 80: What about we make the stackTrace part of the hint?
The idea of hint was to have a single .. hint to deal with these extra things without having to expand the public API

marandaneto: the `hint` still could be used for other stuff as we do on Android, but stackTrace on dart is captured separately so it makes sense IMO to have this expanded API.
on Java, a stack trace is always part of the exception, which is not this case.
The other way would be adding a new field (that is also called stackTrace or sentryStackTrace) to the event, which is confusing.

dart/lib/src/sentry_exception_factory.dart
rxlabz: ✅ you right, it could

PR #158, opened 2020-11-03
Feat: Flutter integrations for the Dart SDK

marandaneto:
## :loudspeaker: Type of change

- [ ] Bugfix
- [X] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
Feat: Flutter integrations for the Dart SDK

## :bulb: Motivation and Context
We'd like to Support Flutter Apps ootb.

## :pencil: Checklist

- [X] I reviewed submitted code
- [X] I added tests to verify changes
- [X] All tests passing
- [ ] No breaking changes

PR #161, opened 2020-11-13
flutter native ios 

rxlabz:

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

add the iOS FlutterSentryPlugin with nativeSdkIntegration and loadContexts method channels

handle iOS errors and enrich payload with iOS device infos

added some unit tests + integration tests

## :pencil: Checklist

- [x] All tests passing
- [ ] No breaking changes

## :crystal_ball: Next steps

More Unit tests


flutter/ios/Classes/SwiftSentryFlutterPlugin.swift
```diff
@@ -3,27 +3,118 @@ import Sentry
...
+        registrar.addMethodCallDelegate(instance, channel: channel)
+    }
+
+    public func handle(_ call: FlutterMethodCall, result: @escaping FlutterResult) {
+        if call.method == "initNativeSdk" {
+            initNativeSdk(call, result: result)
+
+        } else if call.method == "captureEnvelope"{
+            captureEnvelope(call, result: result)
+        }
```
marandaneto: a `switch` would make more sense IMO, we'd need more uses cases here

flutter/ios/Classes/SwiftSentryFlutterPlugin.swift
```diff
@@ -3,27 +3,118 @@ import Sentry
...
+            do{
+                try envelope.write(to: filePath, atomically: false, encoding: .utf8)
+                return true
+            } catch {
+                print("writeEnvelope fail")
+                // logger ?
+            }
+        }
+        return false;
+    }
```
marandaneto: should not be 'writeEnvelope', we call `captureEnvelope` on iOS as explained, it's similar to the RN SDK

flutter/ios/Classes/SwiftSentryFlutterPlugin.swift
```diff
@@ -3,27 +3,118 @@ import Sentry
...
+        }
+
+        if (arguments.isEmpty) {
+            result(FlutterError(code: "4", message: "Arguments is null or empty", details: nil) )
+            return
+        }
+
+        SentrySDK.start { options in
+            options.dsn = arguments["dsn"] as? String
+            options.debug = arguments["debug"] as? Bool ?? false
```
marandaneto: instead of adding a default value, we should check if the argument exists.
see the TODO on the code https://github.com/getsentry/sentry-dart/pull/158/files#r523152074

flutter/ios/Classes/SwiftSentryFlutterPlugin.swift
```diff
@@ -3,27 +3,118 @@ import Sentry
...
+
+            /*
+             missing fields : enableAutoNativeBreadcrumbs, diagnostic level,
+             enableNativeCrashHandling,platform, web, packages,anrTimeoutIntervalMillis
+             diagnosticLevel, cacheDirSize
+             */
+
+            options.beforeSend = { event in
+                self.setEventOriginTag(event: event)
+                /// TODO ? addPackages, removeThreadsIfNotAndroid
```
marandaneto: iOS probably doesn't need `removeThreadsIfNotAndroid`

flutter/ios/Classes/SwiftSentryFlutterPlugin.swift
rxlabz: ✅

flutter/ios/Classes/SwiftSentryFlutterPlugin.swift
rxlabz: ✅

flutter/ios/Classes/SwiftSentryFlutterPlugin.swift
rxlabz: ✅

flutter/ios/Classes/SwiftSentryFlutterPlugin.swift
rxlabz: ✅

flutter/ios/Classes/SwiftSentryFlutterPlugin.swift
```diff
@@ -3,27 +3,190 @@ import Sentry
...
+
+        let sdkInfo = SentrySdkInfo(dict: eventSdk)
+        let sentryId = SentryId(uuidString: eventId)
+        let envelopeHeader = SentryEnvelopeHeader.init(id: sentryId, andSdkInfo: sdkInfo)
+
+        let sentryItemHeader = SentryEnvelopeItemHeader(type: itemType, length: itemLength)
+        
+        do {
+            let data = NSKeyedArchiver.archivedData(withRootObject: envelopeParts[2])
+            // TODO Fix Sentry - Error:: Failed to parse envelope item header Error Domain=NSCocoaErrorDomain Code=3840
```
rxlabz: @marandaneto I don't understand why I get this parsing error. Do you have an idea?

flutter/ios/Classes/SwiftSentryFlutterPlugin.swift
```diff
@@ -3,27 +3,190 @@ import Sentry
...
+    private func captureEnvelope(_ call: FlutterMethodCall, result: @escaping FlutterResult){
+        guard let arguments = call.arguments as? [Any],
+              !arguments.isEmpty,
+              let event = arguments.first as? String else {
+            result(FlutterError(code: "2", message: "Envelope is null or empty", details: nil) )
+            return
+        }
+
+
+        let parts = event.split(separator: "\n")
```
marandaneto: maybe instead of passing a plain String from Dart to plugins, I could split up with a Map with 3 items, header, item header, and event, we can discuss this tomorow.

flutter/ios/sentry_flutter.podspec
```diff
@@ -11,11 +11,11 @@ Sentry SDK for Flutter with support to native through sentry-cocoa.
...
   s.source_files = 'Classes/**/*'
-  s.dependency 'Sentry', '~> 6.0.1'
+  s.dependency 'Sentry', '~> 6.0.9'
   s.dependency 'Flutter'
   s.platform = :ios, '9.0'
 
   # Flutter.framework does not contain a i386 slice. Only x86_64 simulators are supported.
   s.pod_target_xcconfig = { 'DEFINES_MODULE' => 'YES', 'VALID_ARCHS[sdk=iphonesimulator*]' => 'x86_64' }
-  s.swift_version = '5.0'
+  s.swift_version = '5.1'
```
marandaneto: is this a necessary change? wondering if this limits somehow people trying to use our SDK and they could not because of version restrictions

flutter/ios/Classes/SwiftSentryFlutterPlugin.swift
marandaneto: googling it I found this:

> Error Domain=NSCocoaErrorDomain Code=3840 "The operation couldn’t be completed. (Cocoa error 3840.)" (JSON text did not start with array or object and option to allow fragments not set.)

can we print each of the envelopeParts[0, 1 and 2] and see if they are valid JSONs or somehow broken?

rxlabz: I've printed all the parts and everything seems fine. The SentryItemHeader is correctly instantiated, but it throws "internally" during the `client.capture(envelop)` call.
The SentryItemHeader is instantiated with json data converted to NsData, then, internally it seems to be re-converted to json, and then throws the parse error https://github.com/getsentry/sentry-cocoa/blob/e93d7de6a70202a6e4b5d405f0d77100220e76aa/Sources/Sentry/SentrySerialization.m#L161.

marandaneto: @philipphofmann could you help us out with that?

marandaneto: @rxlabz would you mind saving the envelope to a .txt file? thanks :)

philipphofmann: I seems like the JSON is not valid, like @marandaneto pointed it out. If you fix this and you get another error please ping me.

philipphofmann:
Can we please fix the swift-lint errors and warnings or add a config file to ignore some of them, if we think they don't make sense. These lint issues make it super hard to take a look at the PR.

flutter/ios/Classes/SwiftSentryFlutterPlugin.swift
rxlabz: <details>
<summary>header / itemHeader / payload</summary>
{"event_id":"5f941d3af7634830a2abc2fa32c557cd","sdk":{"name":"sentry.dart.flutter","version":"4.0.0-alpha.2","packages":[{"name":"pub:sentry","version":"4.0.0-alpha.2"},{"name":"pub:sentry_flutter","version":"4.0.0-alpha.2"}],"integrations":["isolateErrorIntegration","flutterErrorIntegration","nativeSdkIntegration","runZonedGuardedIntegration"]}}

{"content_type":"application/json","type":"event","length":5216}

{"event_id":"5f941d3af7634830a2abc2fa32c557cd","timestamp":"2020-11-17T08:31:18Z","platform":"other","release":"io.sentry.flutter.example@0.1.2+3","dist":"3","environment":"debug","exception":{"values":[{"type":"StateError","value":"Bad state: whats happening here","stacktrace":{"frames":[{"filename":"hooks.dart","function":"_dispatchPointerDataPacket","lineno":96,"colno":31,"abs_path":"dart:ui/hooks.dart","in_app":false},{"filename":"platform_dispatcher.dart","function":"PlatformDispatcher._dispatchPointerDataPacket","lineno":276,"colno":7,"abs_path":"dart:ui/platform_dispatcher.dart","in_app":false},{"filename":"hooks.dart","function":"_invoke1","lineno":176,"colno":10,"abs_path":"dart:ui/hooks.dart","in_app":false},{"filename":"zone.dart","function":"_CustomZone.runUnaryGuarded","lineno":1005,"colno":7,"abs_path":"dart:async/zone.dart","in_app":false},{"filename":"zone.dart","function":"_CustomZone.runUnary","lineno":1100,"colno":19,"abs_path":"dart:async/zone.dart","in_app":false},{"filename":"zone.dart","function":"_rootRunUnary","lineno":1206,"colno":13,"abs_path":"dart:async/zone.dart","in_app":false},{"filename":"binding.dart","package":"flutter","function":"GestureBinding._handlePointerDataPacket","lineno":221,"colno":7,"abs_path":"package:flutter/src/gestures/binding.dart","in_app":false},{"filename":"binding.dart","package":"flutter","function":"GestureBinding._flushPointerEventQueue","lineno":238,"colno":7,"abs_path":"package:flutter/src/gestures/binding.dart","in_app":false},{"filename":"binding.dart","package":"flutter","function":"GestureBinding.handlePointerEvent","lineno":280,"colno":5,"abs_path":"package:flutter/src/gestures/binding.dart","in_app":false},{"filename":"binding.dart","package":"flutter","function":"GestureBinding._handlePointerEventImmediately","lineno":316,"colno":7,"abs_path":"package:flutter/src/gestures/binding.dart","in_app":false},{"filename":"binding.dart","package":"flutter","function":"RendererBinding.dispatchEvent","lineno":279,"colno":11,"abs_path":"package:flutter/src/rendering/binding.dart","in_app":false},{"filename":"binding.dart","package":"flutter","function":"GestureBinding.dispatchEvent","lineno":361,"colno":22,"abs_path":"package:flutter/src/gestures/binding.dart","in_app":false},{"filename":"binding.dart","package":"flutter","function":"GestureBinding.handleEvent","lineno":381,"colno":19,"abs_path":"package:flutter/src/gestures/binding.dart","in_app":false},{"filename":"pointer_router.dart","package":"flutter","function":"PointerRouter.route","lineno":106,"colno":7,"abs_path":"package:flutter/src/gestures/pointer_router.dart","in_app":false},{"filename":"pointer_router.dart","package":"flutter","function":"PointerRouter._dispatchEventToRoutes","lineno":120,"colno":18,"abs_path":"package:flutter/src/gestures/pointer_router.dart","in_app":false},{"filename":"compact_hash.dart","function":"_LinkedHashMapMixin.forEach","lineno":377,"colno":8,"abs_path":"dart:collection-patch/compact_hash.dart","in_app":false},{"filename":"pointer_router.dart","package":"flutter","function":"PointerRouter._dispatchEventToRoutes.<fn>","lineno":122,"colno":9,"abs_path":"package:flutter/src/gestures/pointer_router.dart","in_app":false},{"filename":"pointer_router.dart","package":"flutter","function":"PointerRouter._dispatch","lineno":77,"colno":12,"abs_path":"package:flutter/src/gestures/pointer_router.dart","in_app":false},{"filename":"recognizer.dart","package":"flutter","function":"PrimaryPointerGestureRecognizer.handleEvent","lineno":476,"colno":9,"abs_path":"package:flutter/src/gestures/recognizer.dart","in_app":false},{"filename":"tap.dart","package":"flutter","function":"BaseTapGestureRecognizer.handlePrimaryPointer","lineno":222,"colno":7,"abs_path":"package:flutter/src/gestures/tap.dart","in_app":false},{"filename":"tap.dart","package":"flutter","function":"BaseTapGestureRecognizer._checkUp","lineno":287,"colno":5,"abs_path":"package:flutter/src/gestures/tap.dart","in_app":false},{"filename":"tap.dart","package":"flutter","function":"TapGestureRecognizer.handleTapUp","lineno":598,"colno":11,"abs_path":"package:flutter/src/gestures/tap.dart","in_app":false},{"filename":"recognizer.dart","package":"flutter","function":"GestureRecognizer.invokeCallback","lineno":183,"colno":24,"abs_path":"package:flutter/src/gestures/recognizer.dart","in_app":false},{"filename":"ink_well.dart","package":"flutter","function":"_InkResponseState._handleTap","lineno":990,"colno":20,"abs_path":"package:flutter/src/material/ink_well.dart","in_app":false},{"filename":"main.dart","package":"sentry_flutter_example","function":"_MyAppState.build.<fn>","lineno":81,"colno":32,"abs_path":"package:sentry_flutter_example/main.dart","in_app":true},{"filename":"main.dart","package":"sentry_flutter_example","function":"tryCatch","lineno":179,"colno":5,"abs_path":"package:sentry_flutter_example/main.dart","in_app":true}]}}]},"sdk":{"name":"sentry.dart.flutter","version":"4.0.0-alpha.2","packages":[{"name":"pub:sentry","version":"4.0.0-alpha.2"},{"name":"pub:sentry_flutter","version":"4.0.0-alpha.2"}],"integrations":["isolateErrorIntegration","flutterErrorIntegration","nativeSdkIntegration","runZonedGuardedIntegration"]}}
</details>

Here the code and the printed values :

```swift
// ...
print("itemType : \(itemType) , itemLength : \(itemLength)") // itemType : event , itemLength : 5216
let sentryItemHeader = SentryEnvelopeItemHeader(type: itemType, length: itemLength)
print("sentryItemHeader : \(sentryItemHeader)") // sentryItemHeader : <SentryEnvelopeItemHeader: 0x600002359200>

do {
  let data = NSKeyedArchiver.archivedData(withRootObject: envelopeParts[2])
  let sentryEnvelopeItem = SentryEnvelopeItem( header: sentryItemHeader, data: data)
  print("sentryEnvelopeItem \(sentryEnvelopeItem)") // sentryEnvelopeItem <SentryEnvelopeItem: 0x6000023a4060>
  let envelope = SentryEnvelope.init(header: envelopeHeader, singleItem: sentryEnvelopeItem)
  print("envelope \(envelope)") // envelope <SentryEnvelope: 0x6000023a40c0>

  SentrySDK.currentHub().getClient()?.capture(envelope: envelope)
  result("")
} catch {
  result(FlutterError(code: "2", message: "Cannot serialize event payload", details: nil) )
}
```

marandaneto: JSON looks good @philipphofmann

flutter/ios/Classes/SwiftSentryFlutterPlugin.swift
```diff
@@ -3,27 +3,190 @@ import Sentry
...
+            if let sessionTrackingIntervalMillis = arguments["sessionTrackingIntervalMillis"] as? UInt{
+                options.sessionTrackingIntervalMillis = sessionTrackingIntervalMillis
+            }
+
+            if let dist = arguments["dist"] as? String{
+                options.dist = dist
+            }
+
+            if let integrations = arguments["integrations"] as? [String]{
+                options.integrations = integrations
```
marandaneto: maybe we should append?

flutter/ios/Classes/SwiftSentryFlutterPlugin.swift
```diff
@@ -3,27 +3,190 @@ import Sentry
...
+            }
+            if let maxBreadcrumbs = arguments["maxBreadcrumbs"] as? UInt{
+                options.maxBreadcrumbs = maxBreadcrumbs
+            }
+
+            /*
+             missing fields : enableAutoNativeBreadcrumbs, diagnostic level,
```
marandaneto: https://docs.sentry.io/platforms/apple/guides/ios/usage/#integrations for enableAutoNativeBreadcrumbs

marandaneto: for diagnostic level see what RN does: https://github.com/getsentry/sentry-react-native/blob/master/ios/RNSentry.m#L91-L109

marandaneto: anrTimeoutIntervalMillis is only for Android

flutter/ios/Classes/SwiftSentryFlutterPlugin.swift
```diff
@@ -3,27 +3,190 @@ import Sentry
...
+             enableNativeCrashHandling,platform, web, packages,anrTimeoutIntervalMillis
+             diagnosticLevel, cacheDirSize
```
marandaneto: @philipphofmann do we have a `cacheDirSize` for iOS like Android does? default to 30

marandaneto: `enableNativeCrashHandling` also only for Android

flutter/ios/Classes/SwiftSentryFlutterPlugin.swift
rxlabz: ✅

flutter/ios/Classes/SwiftSentryFlutterPlugin.swift
marandaneto: I'll remove `platform`, its not used

flutter/ios/Classes/SwiftSentryFlutterPlugin.swift
rxlabz: Ok it's fixed, I didn't pass the correct length.

flutter/ios/Classes/SwiftSentryFlutterPlugin.swift
marandaneto: `web` doesnt seem to be used either

flutter/ios/Classes/SwiftSentryFlutterPlugin.swift
marandaneto: packages should be appended on the beforeSend callback as theres no way thru the options object, see where it lives:
https://github.com/getsentry/sentry-cocoa/blob/9f61acd851ad4fde9e853fbd7293c5cd9d7a01bd/Sources/Sentry/Public/SentryEvent.h#L103

flutter/ios/Classes/SwiftSentryFlutterPlugin.swift
marandaneto: those integrations are not the same as the `sdk.integrations`

the gotten `integrations` should be appended on the beforeSend callback,
this also lives in the event.sdk.integrations

flutter/ios/Classes/SwiftSentryFlutterPlugin.swift
```diff
@@ -1,29 +1,195 @@
+// swiftlint:disable all
```
marandaneto: why? :)

philipphofmann: The purpose of linting is to improve the quality of the code. Just disabling all the warnings doesn't make sense to me. Picking out a few to disable is reasonable. For sure the formatting issues made sense to me. We can argue if we want to disable force casting in this file for example because it could be that you have a hard time avoiding this.

flutter/ios/Classes/SwiftSentryFlutterPlugin.swift
marandaneto: `cacheDirSize` only available for Android as well

flutter/ios/Classes/SwiftSentryFlutterPlugin.swift
```diff
@@ -1,29 +1,202 @@
...
+
+            /*
+             missing fields : enableAutoNativeBreadcrumbs, diagnostic level,
+             enableNativeCrashHandling,platform, web, packages,anrTimeoutIntervalMillis
+             diagnosticLevel, cacheDirSize
+             */
+
+            options.beforeSend = { event in
+                self.setEventOriginTag(event: event)
+                /// TODO ? addPackages
```
marandaneto: `event.sdk.packages`, we need to merge them

flutter/ios/Classes/SwiftSentryFlutterPlugin.swift
philipphofmann: Ah okay, great to hear that @rxlabz. Thanks for the update 😀

flutter/ios/Classes/SwiftSentryFlutterPlugin.swift
rxlabz: I'm not sure sentry-cocoa handle `event.sdk.packages` WDYT ? https://github.com/getsentry/sentry-cocoa/blob/a339a4459442238876e502a8830bf6e628169098/Sources/Sentry/Public/SentryEvent.h#L91-L103

marandaneto: @rxlabz not by default because cocoa is a single package, but you can set the packages item to the dictionary yourself.

the idea is:

https://develop.sentry.dev/sdk/event-payloads/sdk/#example

flutter/ios/Classes/SwiftSentryFlutterPlugin.swift
rxlabz: ✅

flutter/ios/Classes/SwiftSentryFlutterPlugin.swift
rxlabz: It would be cleaner

flutter/ios/sentry_flutter.podspec
rxlabz: ✅

flutter/ios/Classes/SwiftSentryFlutterPlugin.swift
rxlabz: ✅

flutter/ios/Classes/SwiftSentryFlutterPlugin.swift
```diff
@@ -3,27 +3,216 @@ import Sentry
...
+
+            if let maxBreadcrumbs = arguments["maxBreadcrumbs"] as? UInt{
+                options.maxBreadcrumbs = maxBreadcrumbs
+            }
+
+            /*
+             missing fields : enableNativeCrashHandling, packages,anrTimeoutIntervalMillis, cacheDirSize
+             */
+
+            options.beforeSend = { event in
```
rxlabz: @marandaneto  beforeSend seems to not be called for now with `captureEnvelope` https://github.com/getsentry/sentry-cocoa/blob/c34b9067584b52e4136e5fddb7050ee6f3ca022a/Sources/Sentry/SentryClient.m#L243-L253

flutter/ios/Classes/SwiftSentryFlutterPlugin.swift
marandaneto: I'll think about it, cus then the parsing would be done on Android, so it's just choosing the side of the hardwork.

flutter/ios/Classes/SwiftSentryFlutterPlugin.swift
marandaneto: in this case, we'd need to merge this into the event before calling `captureEnvelope()`, so the envelope contains this data, I'm confirming this, hold on.

dart/lib/src/protocol/app.dart
```diff
@@ -15,6 +15,16 @@ class App {
     this.deviceAppHash,
   });
 
+  factory App.fromJson(Map<String, dynamic> data) => App(
+        name: data['app_name'],
+        version: data['app_version'],
+        identifier: data['app_identifier'],
+        build: data['app_build'],
+        buildType: data['build_type'],
+        startTime: DateTime.parse(data['app_start_time']),
```
marandaneto: does this work in UTC and has millis precision right? just to be sure about the parsing.

dart/lib/src/protocol/contexts.dart
```diff
@@ -25,6 +25,13 @@ class Contexts extends MapView<String, dynamic> {
           Gpu.type: gpu,
         });
 
+  factory Contexts.fromJson(Map<String, dynamic> data) => Contexts(
```
marandaneto: missing runtimes, browser, and GPU, also we have consts for the contexts keys like `Device.type` instead of hardcoded Srings

flutter/ios/Classes/SwiftSentryFlutterPlugin.swift
```diff
@@ -3,27 +3,257 @@ import Sentry
...
+            result(FlutterMethodNotImplemented)
+        }
+
+
+    }
+
+    private func deviceInfos(result: @escaping FlutterResult){
+        SentrySDK.configureScope{ scope in
+            let serializedScope = scope.serialize()
+            let contexts = serializedScope["context"]
```
marandaneto: `context` or `contexts`?

flutter/ios/Classes/SwiftSentryFlutterPlugin.swift
```diff
@@ -3,27 +3,257 @@ import Sentry
...
+            captureEnvelope(call, result: result)
+            break
+        default:
+            result(FlutterMethodNotImplemented)
+        }
+
+
+    }
+
+    private func deviceInfos(result: @escaping FlutterResult){
```
marandaneto: I'd call this loadContexts or something, as I said, this is not only about the device info

flutter/ios/Classes/SwiftSentryFlutterPlugin.swift
```diff
@@ -3,27 +3,257 @@ import Sentry
...
+
+            var infos = ["contexts":contexts]
+            print("contexts \(String(describing: contexts))")
+            // TODO DEBUG context
+
+            if let integrations = self.sentryOptions?.integrations {
+                infos["integrations"] = integrations
+            }
+
+            // TODO add sdk.packages
```
marandaneto: should this be here? if iOS has no packages, seems irrelevant to return to the Flutter bits

flutter/ios/Classes/SwiftSentryFlutterPlugin.swift
```diff
@@ -3,27 +3,257 @@ import Sentry
...
+                        } else {
+                            event.sdk = ["integrations":integrations]
+                        }
+                    }
+                }
+                return event
+            }*/
+        }
+
+        result("iOS initNativeSdk" )
```
marandaneto: could be as simple as empty

flutter/ios/Classes/SwiftSentryFlutterPlugin.swift
```diff
@@ -3,27 +3,257 @@ import Sentry
...
+        default:
+            result(FlutterMethodNotImplemented)
+        }
+
+
+    }
+
+    private func deviceInfos(result: @escaping FlutterResult){
+        SentrySDK.configureScope{ scope in
+            let serializedScope = scope.serialize()
```
marandaneto: could this fail? wondering what could happen

flutter/ios/Classes/SwiftSentryFlutterPlugin.swift
```diff
@@ -3,27 +3,257 @@ import Sentry
...
+    private func initNativeSdk(_ call: FlutterMethodCall, result: @escaping FlutterResult){
+        guard let arguments = call.arguments as? [String: Any] else {
+            result(FlutterError(code: "4", message: "Arguments is null or empty", details: nil) )
+            return
+        }
+
+        if (arguments.isEmpty) {
+            result(FlutterError(code: "4", message: "Arguments is null or empty", details: nil) )
+            return
+        }
```
marandaneto: set a new code and a specific message as one tests if its null and the other if its empty

flutter/ios/Classes/SwiftSentryFlutterPlugin.swift
```diff
@@ -3,27 +3,257 @@ import Sentry
...
+
+        SentrySDK.start { options in
+            if(arguments["dsn"] != nil ){
+                options.dsn = arguments["dsn"] as? String
+            } else {
+                result(FlutterError(code: "4", message: "A valid Dsn must be provided", details: nil) )
```
marandaneto: remove this, the underlying SDKs already do this

flutter/ios/Classes/SwiftSentryFlutterPlugin.swift
```diff
@@ -3,27 +3,257 @@ import Sentry
...
+                    options.logLevel = .debug
+                    break;
+                case "warning", "info":
+                    options.logLevel = .verbose
+                    break;
+                default:
+                    options.logLevel = .none
+                    break;
+                }
+            }
```
marandaneto: extract this to a method :)

flutter/ios/Classes/SwiftSentryFlutterPlugin.swift
marandaneto: oh btw you need to add the `iOS` package to the Flutter events, so you need to have at least one package made out of the `sdk`, so it makes sense to return

`val package = Packake(sdk.name, sdk.version)`

flutter/ios/Classes/SwiftSentryFlutterPlugin.swift
```diff
@@ -3,27 +3,257 @@ import Sentry
...
+
+            self.sentryOptions = options
+            /*
+             TODO : beforeSend alternative =>
+              - eventOrigin
+             */
+
+            /* for now, in sentry-cocoa, beforeSend is not called before captureEnvelope
+            options.beforeSend = { event in
+                self.setEventOriginTag(event: event)
```
marandaneto: now this is commented but still would require to add the tag for iOS events which run the beforeSend callback.

if an event is captured on Dart/Flutter, we'd need to add before capturing the envelope.
if an event is captured on iOS, we could add just here as before

flutter/ios/Classes/SwiftSentryFlutterPlugin.swift
```diff
@@ -3,27 +3,257 @@ import Sentry
...
+                    if let integrations = arguments["integrations"] as? [String]{
+                        if  var sdkIntegrations = sdk["integrations"] as? [String]{
+                            event.sdk!["integrations"] = sdkIntegrations.append(contentsOf: integrations)
+                        } else {
+                            event.sdk = ["integrations":integrations]
+                        }
+                    }
+                }
+                return event
+            }*/
```
marandaneto: same as https://github.com/getsentry/sentry-dart/pull/161/files#r526670966

if events are raised on Dart/Flutter, we need to append integrations and packages before capturing the envelope.
if events are raised on iOS, we can append them here, on beforeSend

flutter/ios/Classes/SwiftSentryFlutterPlugin.swift
```diff
@@ -3,27 +3,257 @@ import Sentry
...
+        return (sdk["name"] != nil && !(sdk["name"] as! String).isEmpty)
+    }
+
+    func convertStringToDictionary(text: String) -> [String:AnyObject]? {
+        if let data = text.data(using: .utf8) {
+            do {
+                let json = try JSONSerialization.jsonObject(with: data, options: .mutableContainers) as? [String:AnyObject]
+                return json
+            } catch {
+                print("Something went wrong")
```
marandaneto: we can use `SentryLog. logWithMessage` for all loggings here, I believe its static

flutter/ios/Classes/SwiftSentryFlutterPlugin.swift
```diff
@@ -3,27 +3,257 @@ import Sentry
...
+    private func setEventEnvironmentTag(event: Event, origin: String = "ios", environment: String) {
+        event.tags?["event.origin"] = origin
+        event.tags?["event.environment"] = environment
+    }
+
+    private func isValidSdk( sdk: [String: Any]) -> Bool{
+        return (sdk["name"] != nil && !(sdk["name"] as! String).isEmpty)
+    }
+
+    func convertStringToDictionary(text: String) -> [String:AnyObject]? {
```
marandaneto: can it be `String:Any`? instead of `AnyObject`, if I got it right, AnyObjects are only for reference types

flutter/ios/Classes/SwiftSentryFlutterPlugin.swift
```diff
@@ -3,27 +3,257 @@ import Sentry
...
+        if let data = text.data(using: .utf8) {
+            do {
+                let json = try JSONSerialization.jsonObject(with: data, options: .mutableContainers) as? [String:AnyObject]
```
marandaneto: same here `Any`

marandaneto: `mutableContainers` means its a mutable Dictionary, but we are only reading the positions, right?

flutter/ios/Classes/SwiftSentryFlutterPlugin.swift
```diff
@@ -3,27 +3,257 @@ import Sentry
...
+        event.tags?["event.origin"] = origin
+        event.tags?["event.environment"] = environment
+    }
+
+    private func isValidSdk( sdk: [String: Any]) -> Bool{
+        return (sdk["name"] != nil && !(sdk["name"] as! String).isEmpty)
+    }
+
+    func convertStringToDictionary(text: String) -> [String:AnyObject]? {
+        if let data = text.data(using: .utf8) {
```
marandaneto: an `else` and a log would be nice, which means the envelope was not a `jsonObject` or not a `Map`

flutter/ios/Classes/SwiftSentryFlutterPlugin.swift
```diff
@@ -3,27 +3,257 @@ import Sentry
...
+        guard let arguments = call.arguments as? [Any],
+              !arguments.isEmpty,
+              let event = arguments.first as? String else {
+            result(FlutterError(code: "2", message: "Envelope is null or empty", details: nil) )
+            return
+        }
+
+        let parts = event.split(separator: "\n")
+        let envelopeParts: [[String: Any]] = parts.map ({ part in
+            convertStringToDictionary(text: "\(part)")!
```
marandaneto: `"\(part)"` what does it do? I don't get it

flutter/ios/Classes/SwiftSentryFlutterPlugin.swift
```diff
@@ -3,27 +3,257 @@ import Sentry
...
+        })
+
+        let envelopeHeaderDict = envelopeParts[0]
+        let eventSdk = envelopeHeaderDict
+        let itemHeader = envelopeParts[1]
+
+        guard
+            let eventId = envelopeHeaderDict["event_id"] as? String,
+            let itemType = itemHeader["type"] as? String else {
+            result(FlutterError(code: "2", message: "Cannot serialize event", details: nil) )
```
marandaneto: let's `log` before every result(error)

flutter/ios/Classes/SwiftSentryFlutterPlugin.swift
```diff
@@ -3,27 +3,257 @@ import Sentry
...
+            return
+        }
+
+        let parts = event.split(separator: "\n")
+        let envelopeParts: [[String: Any]] = parts.map ({ part in
+            convertStringToDictionary(text: "\(part)")!
+        })
+
+        let envelopeHeaderDict = envelopeParts[0]
+        let eventSdk = envelopeHeaderDict
```
marandaneto: why do we create an extra object which is just declared above?

flutter/ios/Classes/SwiftSentryFlutterPlugin.swift
```diff
@@ -3,27 +3,257 @@ import Sentry
...
+        let envelopeHeader = SentryEnvelopeHeader.init(id: sentryId, andSdkInfo: sdkInfo)
+
+        let payload = envelopeParts[2]
+
+        let data = try! JSONSerialization.data(withJSONObject: payload, options: .init(rawValue: 0))
+        let sentryItemHeader = SentryEnvelopeItemHeader(type: itemType, length: UInt(data.count))
+        let sentryEnvelopeItem = SentryEnvelopeItem( header: sentryItemHeader, data: data)
+
+        let envelope = SentryEnvelope.init(header: envelopeHeader, singleItem: sentryEnvelopeItem)
+        SentrySDK.currentHub().getClient()?.capture(envelope: envelope)
```
marandaneto: the naming of all the vars in this methods are a bit off.

an envelope consists of:

an envelopeHeader
an itemHeader (item is an event in this case)
an item (which is an event in this case)

so `sentryItemHeader` -> `sentry` seems just irrelevant
`sentryEnvelopeItem` -> `envelope` seems irrelevant

flutter/ios/Classes/SwiftSentryFlutterPlugin.swift
```diff
@@ -3,27 +3,257 @@ import Sentry
...
+            return
+        }
+
+        let sdkInfo = SentrySdkInfo(dict: eventSdk)
+        let sentryId = SentryId(uuidString: eventId)
+        let envelopeHeader = SentryEnvelopeHeader.init(id: sentryId, andSdkInfo: sdkInfo)
+
+        let payload = envelopeParts[2]
+
+        let data = try! JSONSerialization.data(withJSONObject: payload, options: .init(rawValue: 0))
```
marandaneto: if this fail, let's log it and return result(error), or what does it happen?

flutter/ios/Classes/SwiftSentryFlutterPlugin.swift
```diff
@@ -3,27 +3,257 @@ import Sentry
...
+            do{
+                try envelope.write(to: filePath, atomically: false, encoding: .utf8)
+                return true
+            } catch {
+                print("writeEnvelope fail")
+                // logger ?
+            }
+        }
+        return false
+    }
```
marandaneto: I believe this is not used anymore, right?

flutter/lib/default_integrations.dart
```diff
@@ -84,6 +84,49 @@ Integration runZonedGuardedIntegration(
...
+      final Map<String, dynamic> infos =
+          Map<String, dynamic>.from(await channel.invokeMethod('deviceInfos'));
+
+      if (infos['contexts'] != null) {
+        options.addEventProcessor((event, dynamic hint) {
+          event = event.copyWith(
+            contexts: Contexts.fromJson(
+              Map<String, dynamic>.from(infos['contexts'] as Map),
+            ),
+          );
```
marandaneto: this would be executed for Android too, so the `Android` and `Web` plugins need to return empty instead of non implemented method

marandaneto: also, the result should be `merged` and not `overwritten`, if the same key already exists in the event, the event has precedence over the gotten contexts

flutter/lib/default_integrations.dart
```diff
@@ -84,6 +84,49 @@ Integration runZonedGuardedIntegration(
...
+          return event;
+        });
+      }
+
+      if (infos['integrations'] != null) {
+        final integrations = List<String>.from(infos['integrations'] as List);
+        options.addEventProcessor((event, dynamic hint) {
```
marandaneto: looks a bit overkill having 2 event processors for the same thing, can we merge them?

flutter/lib/sentry_flutter.dart
```diff
@@ -76,6 +76,7 @@ mixin SentryFlutter {
     }
     options.addIntegration(flutterErrorIntegration);
     options.addIntegration(nativeSdkIntegration(options, _channel));
+    options.addIntegration(deviceInfosIntegration(options, _channel));
```
rxlabz: only if Platform.isIOS

flutter/ios/Classes/SwiftSentryFlutterPlugin.swift
rxlabz: ✅

dart/lib/src/protocol/contexts.dart
rxlabz: ✅

dart/lib/src/protocol/app.dart
rxlabz: ✅ yes

flutter/ios/Classes/SwiftSentryFlutterPlugin.swift
rxlabz: ✅

flutter/ios/Classes/SwiftSentryFlutterPlugin.swift
rxlabz: this info should be read from SentryMeta which is not public. I'll try to find an other way.

flutter/ios/Classes/SwiftSentryFlutterPlugin.swift
rxlabz: ✅

flutter/ios/Classes/SwiftSentryFlutterPlugin.swift
rxlabz: ✅

flutter/ios/Classes/SwiftSentryFlutterPlugin.swift
marandaneto: not sure if I follow the problem, the naming is hardcoded and the version is available thru `options.sdk.version`

flutter/ios/Classes/SwiftSentryFlutterPlugin.swift
rxlabz: ✅

flutter/lib/sentry_flutter.dart
rxlabz: ✅

flutter/ios/Classes/SwiftSentryFlutterPlugin.swift
rxlabz: unfortunately, neither the `sdk` or the `sdkVersion` are available in `options`
sentry-cocoa seems to add the `sdk` to the event with info from the SentryMeta static fields, and it seems not available anywhere else https://github.com/getsentry/sentry-cocoa/blob/c34b9067584b52e4136e5fddb7050ee6f3ca022a/Sources/Sentry/SentryClient.m#L330-L337

rxlabz: ✅  for now I send hardcoded value for both version and name

flutter/ios/Classes/SwiftSentryFlutterPlugin.swift
rxlabz: ✅  from what I see it's not supposed to fail/throw

flutter/lib/default_integrations.dart
rxlabz: ✅

flutter/ios/Classes/SwiftSentryFlutterPlugin.swift
rxlabz: ✅

flutter/ios/Classes/SwiftSentryFlutterPlugin.swift
rxlabz: ✅

flutter/ios/Classes/SwiftSentryFlutterPlugin.swift
rxlabz: ✅ move the two conditions to the `guard`

flutter/ios/Classes/SwiftSentryFlutterPlugin.swift
rxlabz: ✅

dart/lib/src/protocol/contexts.dart
```diff
@@ -25,6 +25,32 @@ class Contexts extends MapView<String, dynamic> {
...
```
marandaneto: if `data` has extra keys that are not "typed" by Contexts, they will be lost, we need to add all the fields.

flutter/ios/Classes/SwiftSentryFlutterPlugin.swift
```diff
@@ -3,27 +3,235 @@ import Sentry
...
-              options.dsn = dsn as? String
-              options.debug = resource["SentryDebug"] as? Bool ?? false
-              options.attachStacktrace = true
-          }
-        }
-      }
-    }
-  }
+
+    var sentryOptions: Options?
```
marandaneto: can we make this private?

flutter/ios/Classes/SwiftSentryFlutterPlugin.swift
marandaneto: method is still called `deviceInfos`

flutter/ios/Classes/SwiftSentryFlutterPlugin.swift
```diff
@@ -3,27 +3,235 @@ import Sentry
...
+
+    }
+
+    private func deviceInfos(result: @escaping FlutterResult){
+        SentrySDK.configureScope{ scope in
+            let serializedScope = scope.serialize()
+            let context = serializedScope["context"]
+
+            var infos = ["contexts":context]
+            // TODO DEBUG context
```
marandaneto: is this going to be worked on?

flutter/ios/Classes/SwiftSentryFlutterPlugin.swift
```diff
@@ -3,27 +3,235 @@ import Sentry
...
+
+            if let integrations = self.sentryOptions?.integrations {
+                infos["integrations"] = integrations
+            }
+
+            // TODO get the sdkVersion from SDK
+            infos["package"] = ["version": "6.0.9" ,"sdk_name":"cocoapods:sentry-cocoa"]
```
marandaneto: @philipphofmann is there a way to get the SDK version? so we don't need to hardcode

marandaneto: @rxlabz as far as I can tell `SentryMeta.versionString` would work?

flutter/ios/Classes/SwiftSentryFlutterPlugin.swift
rxlabz: ✅

flutter/lib/default_integrations.dart
rxlabz: ✅

flutter/ios/Classes/SwiftSentryFlutterPlugin.swift
rxlabz: ✅ SentryLog seems not available outside the library. For now we just `print` a clearer message

flutter/ios/Classes/SwiftSentryFlutterPlugin.swift
rxlabz: ✅ done, needs tests

flutter/ios/Classes/SwiftSentryFlutterPlugin.swift
rxlabz: ✅

flutter/ios/Classes/SwiftSentryFlutterPlugin.swift
rxlabz: ✅ refactored

flutter/ios/Classes/SwiftSentryFlutterPlugin.swift
rxlabz: ✅

flutter/ios/Classes/SwiftSentryFlutterPlugin.swift
rxlabz: ✅

flutter/ios/Classes/SwiftSentryFlutterPlugin.swift
rxlabz: ✅

flutter/ios/Classes/SwiftSentryFlutterPlugin.swift
rxlabz: ✅

dart/lib/src/protocol/contexts.dart
rxlabz: ✅ good catch, thx

flutter/ios/Classes/SwiftSentryFlutterPlugin.swift
rxlabz: ✅ removed

dart/lib/src/protocol/app.dart
marandaneto: out of curiosity for all the setters, if the field doesn't exist in `data` or it's  `null`, what does it happen? I believe it'd break at `DateTime.parse`

flutter/ios/Classes/SwiftSentryFlutterPlugin.swift
```diff
@@ -3,27 +3,244 @@ import Sentry
...
+
+            default:
+                result(FlutterMethodNotImplemented)
+        }
+    }
+
+    private func loadContexts(result: @escaping FlutterResult){
+        /*
+         TODO ? eventOrigin beforeSend alternative
+         */
```
marandaneto: is this going to be addressed?

flutter/lib/default_integrations.dart
```diff
@@ -84,6 +84,65 @@ Integration runZonedGuardedIntegration(
...
 }
 
+Integration loadContextsIntegration(
+  SentryOptions options,
+  MethodChannel channel,
+) {
+  Future<void> integration(Hub hub, SentryOptions options) async {
+    try {
+      final Map<String, dynamic> infos =
+          Map<String, dynamic>.from(await channel.invokeMethod('loadContexts'));
```
marandaneto: this means we'd `loadContexts` only once, on App. start, wondering if this should not be inside of the `eventProcessor`, so we query up to date data on every event, how on in RN?

dart/lib/src/protocol/app.dart
rxlabz: ✅ good catch, thx !

flutter/ios/Classes/SwiftSentryFlutterPlugin.swift
rxlabz: `SentryMeta` doesn't seems public

philipphofmann: Yes, exactly @marandaneto. Craft calls the [VersionBump](https://github.com/getsentry/sentry-cocoa/blob/master/Utils/VersionBump/main.swift) script to update `SentryMeta` when we do a new release.

marandaneto: @philipphofmann yep, we'll get there, but the question here is different, how do we read the Cocoa version on runtime?

philipphofmann: There is no straightforward way to do this. We could easily add a way to retrieve this, but why do we need this? I don't see something like this in react-native: https://github.com/getsentry/sentry-react-native/blob/master/ios/RNSentry.m

marandaneto: @philipphofmann to populate the `sdk.packages`, not all the other SDKs do it, but they should.
so when there's an event, we know which SDKs were involved, eg;

an event captured by Dart SDK, only has the Dart package.
an event captured by the Flutter SDK, has the Dart, Flutter and Native packages (Android or iOS) as they are the ones sending over the network.

philipphofmann: Okay then making this public isn't a problem. Should we just make `SentryMeta` public or should we add methods to `SentrySDK`?

marandaneto: @philipphofmann I like how Android Does, we have a field on options, `SentryOptions.sdkVersion`, on SDK init, we set the values, so it's possible to read thru the `OptionsConfiguration` callback.

```
SentryAndroid.init(options-> {
  // options.sdkVerrsion.name and version
})
```

flutter/lib/default_integrations.dart
rxlabz: ✅ called for each event

flutter/ios/Classes/SwiftSentryFlutterPlugin.swift
rxlabz: ✅ removed

flutter/lib/default_integrations.dart
```diff
@@ -84,6 +84,67 @@ Integration runZonedGuardedIntegration(
   return integration;
 }
 
+Integration loadContextsIntegration(
```
marandaneto, line 118: docs

flutter/lib/default_integrations.dart
```diff
@@ -84,6 +84,67 @@ Integration runZonedGuardedIntegration(
...
+          if (infos['package'] != null) {
+            final package = Map<String, String>.from(infos['package'] as Map);
+            final sdk = event.sdk ?? options.sdk;
+            sdk.addPackage(package['name'], package['version']);
+            event = event.copyWith(sdk: sdk);
+          }
+        } catch (error) {
+          options.logger(
+            SentryLevel.error,
+            'nativeSdkIntegration failed to be installed: $error',
```
marandaneto: looks like this is off (error message), the error here seems to be related to all the querying the data, parsing etc

flutter/lib/sentry_flutter.dart
```diff
@@ -76,6 +77,10 @@ mixin SentryFlutter {
     }
     options.addIntegration(flutterErrorIntegration);
     options.addIntegration(nativeSdkIntegration(options, _channel));
+
+    if (Platform.isIOS) {
```
marandaneto, line 125: we probably need to make it testable, so using this static might need to change, look what I've done with the `PackageLoader` typedef.
We can improve this later, of course.

flutter/lib/default_integrations.dart
rxlabz: ✅

flutter/lib/default_integrations.dart
rxlabz: ✅

flutter/lib/sentry_flutter.dart
rxlabz: added a Todo. for now I should be able to test it with `onPlatform`

flutter/ios/Classes/SwiftSentryFlutterPlugin.swift
philipphofmann: I just opened up a PR for this https://github.com/getsentry/sentry-cocoa/pull/859

flutter/ios/Classes/SwiftSentryFlutterPlugin.swift
marandaneto: RN does `NSLog` but this leaks to the production version, so yeah `print` will do for now.

flutter/example/ios/Podfile.lock
```diff
@@ -1,11 +1,13 @@
 PODS:
   - Flutter (1.0.0)
-  - Sentry (6.0.1):
-    - Sentry/Core (= 6.0.1)
-  - Sentry/Core (6.0.1)
+  - package_info (0.0.1):
```
marandaneto, line 3: @rxlabz this looks fishy, is this 0.0.1?

PR #165, opened 2020-11-17
Fix: Handled mechanism is true and fix samples

marandaneto:

- [X] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

Uncaught exception on Kotlin was not working because the MessageChannel was not returning a value.
Events should be `handled=true`, but let's set the event level to `fatal` at least, as the App never crashes.

## :pencil: Checklist

- [X] All tests passing
- [X] No breaking changes

## :crystal_ball: Next steps
tests, will come on the feature branch

### DIFF
--- a/flutter/lib/default_integrations.dart
+++ b/flutter/lib/default_integrations.dart
@@ -64,6 +64,9 @@ void flutterErrorIntegration(Hub hub, SentryOptions options) {
     if (defaultOnError != null) {
       defaultOnError(errorDetails);
     }
+
+    // we don't call Zone.current.handleUncaughtError because we'd like
+    // to set a specific mechanism for FlutterError.onError.
   };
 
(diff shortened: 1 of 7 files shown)


flutter/lib/default_integrations.dart
```diff
@@ -21,11 +21,16 @@ void isolateErrorIntegration(Hub hub, SentryOptions options) {
         final dynamic throwable = error.first;
         final dynamic stackTrace = error.last;
 
-        const mechanism = Mechanism(type: 'isolateError', handled: false);
+        // TODO: validate if handled should be true or flalse
+        // assuming false as I could not see the App crashing so far.
```
marandaneto:
```suggestion
        // // isolateError doesn't crash the App.
```

rxlabz (approved):
LGTM

PR #179, opened 2020-11-20
Dart pkg: Remove Flutter from the heading

bruno-garcia:
Feedback was that it's confusing why two packages.

Trying to be very clear this is the pure Dart SDK. And point folks to Flutter right at the top (given there are way more Flutter users than other Dart)

### DIFF
--- a/dart/README.md
+++ b/dart/README.md
@@ -5,12 +5,19 @@
   <br />
 </p>
 
-Sentry SDK for Dart and Flutter
+Sentry SDK for Dart
 ===========
 
 | package | build | pub | likes | popularity | pub points |
 | ------- | ------- | ------- | ------- | ------- | ------- |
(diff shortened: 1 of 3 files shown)


dart/README.md
```diff
@@ -5,13 +5,20 @@
   <br />
 </p>
 
-Sentry SDK for Dart and Flutter
+Sentry SDK for Dart
```
bruno-garcia, line 8: So it looks smaller on pub.dev

image

flutter/README.md
```diff
@@ -5,20 +5,15 @@
...
 
-#### Versions
-
-Versions `^4.0.0` are `Prereleases` and are under improvements/testing.
-
-Versions `^4.0.0` integrate our Native SDKs ([Android](https://github.com/getsentry/sentry-java) and [Apple](https://github.com/getsentry/sentry-cocoa)), so you are able to capture errors on Native code as well (Java/Kotlin/C/C++ for Android and Objective-C/Swift for Apple).
-
-The current stable version is the Dart SDK, [3.0.1](https://pub.dev/packages/sentry).
+This package includes support to native crashes through Sentry's native SDKs: ([Android](https://github.com/getsentry/sentry-java) and [iOS](https://github.com/getsentry/sentry-cocoa)).
```
bruno-garcia, line 15: Since pub.dev has the release history already, I thought we could just give a brief description of what's in the box.

PR #181, opened 2020-11-23
loadContextsIntegration tests

rxlabz:

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring

## :scroll: Description

add loadContextsIntegration

new unit tests

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/flutter/test/load_contexts_integrations_test.dart
+++ b/flutter/test/load_contexts_integrations_test.dart
@@ -1,0 +1,67 @@
+import 'package:flutter/services.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:mockito/mockito.dart';
+import 'package:sentry/sentry.dart';
+import 'package:sentry_flutter/sentry_flutter.dart';
+
+import 'mocks.dart';
+
+class MockTransport extends Mock implements Transport {}
+
+void main() {
+  const MethodChannel _channel = MethodChannel('sentry_flutter');
+
+  TestWidgetsFlutterBinding.ensureInitialized();
+
+  bool called = false;
+
+  setUp(() {
+    _channel.setMockMethodCallHandler((MethodCall methodCall) async {
+      called = true;
+      return {
+        'integrations': ['NativeIntegration'],
+        'package': {'sdk_name': 'native-package', 'version': '1.0'},
+        'contexts': {
+          'device': {'name': 'Device1'},
+          'app': {'app_name': 'test-app'},
+          'os': {'name': 'os1'},
+          'gpu': {'name': 'gpu1'},
+          'browser': {'name': 'browser1'},
+          'runtime': {'name': 'RT1'},
+          'theme': 'material',
+        }
+      };
+    });
+  });
+
+  tearDown(() {
+    _channel.setMockMethodCallHandler(null);
+  });
+
+  test('should apply the loadContextsIntegration eventProcessor', () async {
+    final options = SentryOptions()..dsn = fakeDsn;
+    final hub = Hub(options);
+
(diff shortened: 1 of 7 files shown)


flutter/test/load_contexts_integrations_test.dart
```diff
@@ -0,0 +1,146 @@
+import 'package:flutter/services.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:mockito/mockito.dart';
+import 'package:sentry/sentry.dart';
+import 'package:sentry_flutter/sentry_flutter.dart';
+
+import 'mocks.dart';
+
+class MockTransport extends Mock implements Transport {}
```
marandaneto: a file called `mocks.dart` already exists, what's about keeping it in there?

rxlabz: ✅ thanks

flutter/test/sentry_flutter_util.dart
```diff
@@ -6,36 +6,35 @@ import 'package:sentry/sentry.dart';
...
-      options.integrations
-          .where((element) => element == isolateErrorIntegration),
-      isNotEmpty);
+      expect(
+          options.integrations
+              .where((element) => element == isolateErrorIntegration),
+          isNotEmpty);
 
-  expect(4, options.integrations.length);
+      expect(onIOS ? 5 : 4, options.integrations.length);
```
marandaneto: I think this can get complicated to test if its iOS, if its Android, if its Web, how can we improve this, maybe doing an assertion by integration name?

```
if (iOS)
  expect(options.integrations.contains(name), true)
```

etc...

flutter/test/load_contexts_integrations_test.dart
```diff
@@ -0,0 +1,143 @@
+import 'package:flutter/services.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:sentry/sentry.dart';
```
marandaneto: `sentry_flutter.dart` already imports this

flutter/test/sentry_flutter_util.dart
```diff
@@ -6,36 +6,39 @@ import 'package:sentry/sentry.dart';
...
 import 'package:sentry_flutter/src/version.dart';
+
 import 'mocks.dart';
 
-FutureOr<void> configurationTester(
-  SentryOptions options, {
-  bool isWeb = false,
-}) async {
-  options.dsn = fakeDsn;
+FutureOr<void> Function(SentryOptions) getConfigurationTester(bool onIOS) =>
```
marandaneto: `bool isWeb = false` is gone, why? the idea is to test if it's running thru a `Web` integration as well, not done yet, but the flag would be necessary I guess

flutter/test/sentry_flutter_util.dart
rxlabz: ✅ I let the integration number test and added new ones to test "by integration name" https://github.com/getsentry/sentry-dart/blob/1bb5ddf67badfe244a50dcfbe43bbef9134b7118/flutter/test/sentry_flutter_test.dart#L60-L98

flutter/test/load_contexts_integrations_test.dart
rxlabz: ✅

flutter/test/sentry_flutter_util.dart
rxlabz: as the param wasn't used I removed it for now, should I add it ?

rxlabz: ✅ I put it back

PR #185, opened 2020-11-24
fix : stackTrace frames with 'package' uri.scheme are inApp by default

rxlabz:

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring

## :scroll: Description

change the logic behind the stackTraceFactory to not consider all package as !inApp

remove the need to add the application own package to the `inAppIncludes`

new Unit test

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,8 +3,8 @@
 ## 4.0.0-alpha.3 (Next)
 
 - Development.
-- StackTrace factory : package are inApp by default 
-- add loadContextsIntegration tests
+- Fix: StackTrace frames with 'package' uri.scheme are inApp by default #185
+- Enhancement: add loadContextsIntegration tests
 
 ## 4.0.0-alpha.2
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## 4.0.0-alpha.3 (Next)
 
 - Development.
+- StackTrace factory : package are inApp by default 
 - add loadContextsIntegration tests
 
 ## 4.0.0-alpha.2
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,6 @@
 
 ## vNext
 
-- Development.
 - Fix: StackTrace frames with 'package' uri.scheme are inApp by default #185
 - Enhancement: add loadContextsIntegration tests
 - StackTrace factory : package are inApp by default 
--- a/dart/lib/src/sentry_stack_trace_factory.dart
+++ b/dart/lib/src/sentry_stack_trace_factory.dart
@@ -130,7 +130,10 @@ class SentryStackTraceFactory {
       }
     }
 
-    if (frame.isCore) return false;
+    if (frame.isCore ||
+        (frame.uri.scheme == 'package' && frame.package == 'flutter')) {
+      return false;
+    }
 
     return true;
   }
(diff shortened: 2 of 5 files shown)


dart/lib/src/sentry_stack_trace_factory.dart
```diff
@@ -130,7 +130,7 @@ class SentryStackTraceFactory {
       }
     }
 
-    if (frame.isCore || frame.uri.scheme == 'package') return false;
+    if (frame.isCore) return false;
```
rxlabz: @marandaneto we can simply remove the `frame.uri.scheme == 'package'` condition, but I'm not sure that then we still need the `inAppIncludes` loop WDYT ?

marandaneto: so `frame.isCore` means:
  /// Whether this stack frame comes from the Dart core libraries.

is there a way to also check for `frame.isFlutterCore`?

if we remove `frame.uri.scheme == 'package'`, I still want the `Flutter` core packages to be `inApp=false`, if we solve that, it'd be a win already

rxlabz: what about this ?

```dart
if (frame.isCore ||
        (frame.uri.scheme == 'package' && frame.package == 'flutter'))
      return false;
```

marandaneto: looks good, could you try on the samples and see if they are marked `inApp` correctly?

rxlabz: ✅ seems fine, I also added a FlutterError button in the example

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## vNext
 
+- Development.
+- StackTrace factory : package are inApp by default
```
marandaneto, line 7: please align the changelog item + PR's title, a prefix would be nice like Fix: title

flutter/example/lib/main.dart
```diff
@@ -199,6 +210,13 @@ class CocoaExample extends StatelessWidget {
             await channel.invokeMethod<void>('crash');
           },
         ),
+        RaisedButton(
+          child: const Text('Crash swift'),
+          onPressed: () async {
+            const channel = MethodChannel('example.flutter.sentry.io');
+            await channel.invokeMethod<void>('crashSwift');
+          },
+        ),
```
marandaneto: this is not part of this PR right? we don't have this method on the flutter plugin

flutter/example/lib/main.dart
```diff
@@ -16,8 +16,6 @@ Future<void> main() async {
   await SentryFlutter.init(
     (options) {
       options.dsn = _exampleDsn;
-      // Change the 'sentry_flutter_example' below with your own package.
-      options.addInAppInclude('sentry_flutter_example');
```
marandaneto, line 20: we could remove the `addInAppInclude` mention from the READMEs as well

CHANGELOG.md
rxlabz: ✅

flutter/example/lib/main.dart
rxlabz: ✅ ooops, thanks

flutter/example/lib/main.dart
rxlabz: ✅

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## vNext
 
+- Development.
```
```suggestion
```

PR #187, opened 2020-11-25
Feat/dart default integrations

rxlabz:

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

- add `isolateErrorIntegration` and `runZoneGuardedIntegration` to dart default integrations
- add an optional `initialIntegrations` to `Sentry.init` to allow running custom integrations before `runZoneGuardedIntegration`

new unit tests

## :pencil: Checklist

- [x] All tests passing
- [ ] No breaking changes


CHANGELOG.md
```diff
@@ -25,6 +24,7 @@
 - Timestamp has millis precision.
 - For better groupping, add your own package to the `addInAppInclude` list, e.g.  `options.addInAppInclude('sentry_flutter_example');`
 - A few classes of the `Protocol` were renamed.
+- Sentry.init() now needs a callback argument which must run the host app after Sentry initialization.
```
marandaneto: I'd suggest adding a breaking change section on the version above (the one that is gonna be published)

dart/example/main.dart
```diff
@@ -48,7 +52,9 @@ Future<void> main() async {
       ..setTag('build', '579')
       ..setExtra('company-name', 'Dart Inc');
   });
+}
 
+void runApp() async {
```
marandaneto, line 56:
```suggestion
Future<void> runApp() async {
```

CHANGELOG.md
rxlabz: ✅

dart/lib/src/utils.dart
```diff
@@ -20,3 +20,7 @@ String formatDateAsIso8601WithMillisPrecision(DateTime date) {
 
 /// helper to detect a browser context
 const isWeb = identical(0, 0.0);
+
+/// helper to detect if app is in release mode
+const isReleaseMode =
+    bool.fromEnvironment('dart.vm.product', defaultValue: false);
```
marandaneto, line 26: is the other one flutter only?

rxlabz: yes, it's a constant from Flutter foundation

dart/example/main.dart
rxlabz: I'm not sure about this, the function will never return something.

dart/README.md
```diff
@@ -40,12 +40,15 @@ import 'dart:async';
...
-  });
+  await Sentry.init(
+    (options) {
+      options.dsn = 'https://example@sentry.io/add-your-dsn-here';
+    }, 
+    initApp, // Init your App.
+  );
+}
 
+void initApp(){
```
marandaneto:
```suggestion
void initApp() {
```

dart/README.md
```diff
@@ -54,8 +57,10 @@ Future<void> main() async {
       stackTrace: stackTrace,
     );
   }
+  // ...
```
marandaneto: why do we need this?

dart/example/main.dart
rxlabz: and the callback is not awaited
```dart
runZonedGuarded(() {
      callback();
    }, (exception, stackTrace) async {//...}
})
```

dart/lib/src/sentry.dart
```diff
@@ -1,15 +1,22 @@
...
 import 'protocol.dart';
 import 'sentry_client.dart';
 import 'sentry_options.dart';
+import 'utils.dart';
 
 /// Configuration options callback
 typedef OptionsConfiguration = FutureOr<void> Function(SentryOptions);
 
+// run main app callback
+typedef AppRunner = void Function();
```
```suggestion
typedef AppRunner = FutureOr<void> Function();
```

just to not be a limitation of the Function signature

dart/README.md
rxlabz: to illustrate that the app runner will probably contain more code.
What's the better : `// the rest of your code` or nothing ?

dart/lib/src/sentry.dart
```diff
@@ -21,11 +28,19 @@ class Sentry {
...
+  static Future<void> init(
+    OptionsConfiguration optionsConfiguration,
+    AppRunner callback,
+  ) async {
     if (optionsConfiguration == null) {
       throw ArgumentError('OptionsConfiguration is required.');
     }
+    if (callback == null) {
+      throw ArgumentError('AppRunner is required.');
+    }
```
marandaneto: what's about making this an optional parameter for the Dart SDK? not sure if it makes sense to every Dart App.
and you only add the `runZonedGuardedIntegration` if the `callback` is not null

rxlabz: should a user make the decision between using `runZonedGuardedIntegration` or not ?

dart/lib/src/sentry.dart
rxlabz: Why do we need this ? Future<void> and FutureOr<void> should not be used for the signature of a method without `return`

dart/README.md
rxlabz: ✅

dart/example/main.dart
marandaneto: yeah not really needed

dart/lib/src/default_integrations.dart
```diff
@@ -0,0 +1,32 @@
+import 'dart:async';
```
marandaneto, line 1: does it make sense to be `dart_default_integrations.dart ` ? I guess it could conflict to default_integrations.dart  from the flutter package

rxlabz: not really, you can't have file names conflicts, only methods or classes names

dart/lib/src/sentry.dart
```diff
@@ -1,15 +1,22 @@
 import 'dart:async';
 
+import 'default_integrations.dart';
 import 'hub.dart';
 import 'hub_adapter.dart';
 import 'noop_hub.dart';
+import 'noop_isolate_error_integration.dart'
+    if (dart.library.io) 'isolate_error_integration.dart';
```
marandaneto, line 8: does it need to be a NoOp import? did we have a bug on the previous code which added or not the integration? just a question though

dart/lib/src/sentry.dart
marandaneto: if their init methods is async, we must `await` or the SDK won't be fully initialized, I guess?
What I mean is, we'd not force them to just pass a sync method, but could be `async` as well

marandaneto: I'm not sure if its the right thing to do, its food for thought

dart/lib/src/sentry.dart
rxlabz: without this conditional import you can't compile for the web

dart/lib/src/sentry.dart
marandaneto: it's rather a question because I dont know all the use cases.
do you think that all the devs would like the `runZonedGuardedIntegration` or some people prefer to opt out? we can keep like this now, but it'd be nice to have an idea

dart/lib/src/sentry.dart
rxlabz: you don't have to define a `Future<void>` to use `async` and `await`, so it doesn't force you to pass a sync method, it forces only to pass a method which returns nothing.

dart/example_web/web/main.dart
```diff
@@ -79,10 +82,7 @@ void captureException() async {
   } catch (error, stackTrace) {
     print('\nReporting the following stack trace: ');
     print(stackTrace);
-    final sentryId = await Sentry.captureException(
-      error,
-      stackTrace: stackTrace,
-    );
+    final sentryId = await Sentry.captureException(error);
```
marandaneto: is the `stackTrace: stackTrace` removed because `error` is an `Error`?

dart/lib/src/default_integrations.dart
marandaneto: nice

dart/lib/src/sentry.dart
marandaneto: we didn't have this `noop` before and it was compiling the Flutter web project, wondering why it didn't break before

dart/lib/src/sentry.dart
marandaneto: so if the passed function has an `async` method, it'll work as expected? I needed to make the `OptionsConfiguration` a `FutureOr` for this reason.

dart/lib/src/default_integrations.dart
```diff
@@ -0,0 +1,32 @@
...
+
+import 'hub.dart';
+import 'protocol.dart';
+import 'sentry_options.dart';
+import 'throwable_mechanism.dart';
+
+/// integration that capture errors on the runZonedGuarded error handler
+Integration runZonedGuardedIntegration(
+  Function callback,
```
```suggestion
  AppRunner callback,
```

dart/lib/src/sentry.dart
rxlabz: it's because the optionConfiguration is awaited 
-  https://github.com/getsentry/sentry-dart/blob/1e36bd0e66d0a0b22a00cb954a8d90c050183334/flutter/lib/src/sentry_flutter.dart#L31
- https://github.com/getsentry/sentry-dart/blob/1e36bd0e66d0a0b22a00cb954a8d90c050183334/dart/lib/src/sentry.dart#L29

marandaneto: yep, now I ask, does it make sense to `await` the `callback` if we want that callback to be fully run on the `runZonedGuarded`, because that's the way to capture the exceptions in there, I don't know if it'll be caught if an exception happens and there's no `await`

dart/lib/src/default_integrations.dart
rxlabz: ✅

dart/lib/src/sentry.dart
rxlabz: Unfortunately I didn't ran the example with Flutter web for a while. There was a Flutter web compilation error, this PR fixes it.

marandaneto: I see, but CI should have broken this https://github.com/getsentry/sentry-dart/blob/main/.github/workflows/flutter.yml#L62
wondering if this works at all (the CI)

dart/lib/src/sentry.dart
marandaneto: I'm convinced that it should be an optional param for the Dart SDK, as a Console App or any other type of App that don't have an UI won't make much sense to execute a callback, they don't need to run anything inside of a ZonedGuarded

dart/example_web/web/main.dart
rxlabz: It was, but now that I put it back, the captureException fails in release mode.

dart/lib/src/sentry.dart
rxlabz: ✅ I made it optional.

dart/lib/src/sentry.dart
marandaneto: i created a task for it (CI)

dart/example_web/web/main.dart
marandaneto: mm, did you figure this out?

CHANGELOG.md
```diff
@@ -4,9 +4,12 @@
...
 - Fix: StackTrace frames with 'package' uri.scheme are inApp by default #185
 - Enhancement: add loadContextsIntegration tests
-- StackTrace factory : package are inApp by default 
 - Fix: missing app's stack traces for Flutter errors
-- add loadContextsIntegration tests
+- Enhancement: add isolateErrorIntegration and runZonedGuardedIntegration to default integrations in sentry-dart
+
+### Breaking changes
+
+- `Sentry.init` now needs a callback argument which must run the host app after Sentry initialization.
```
```suggestion
- `Sentry.init` has an optional callback argument which runs the host app after Sentry initialization.
```

dart/example/main.dart
```diff
@@ -82,6 +88,8 @@ Future<void> main() async {
   } finally {
     await Sentry.close();
   }
+
+  exit(0);
```
marandaneto: do we need this if we `await` the events to be sent?

dart/lib/src/default_integrations.dart
```diff
@@ -0,0 +1,33 @@
...
+
+import 'hub.dart';
+import 'protocol.dart';
+import 'sentry.dart';
+import 'sentry_options.dart';
+import 'throwable_mechanism.dart';
+
+/// integration that capture errors on the runZonedGuarded error handler
+Integration runZonedGuardedIntegration(
+  AppRunner callback,
```
marandaneto: maybe we should adopt `appRunner` for the callback var name

dart/lib/src/sentry.dart
```diff
@@ -20,11 +27,19 @@ class Sentry {
   static Hub get currentHub => _hub;
 
   /// Initializes the SDK
-  static Future<void> init(OptionsConfiguration optionsConfiguration) async {
+  /// passing a [AppRunner] callback allows to run the app within its own error zone (`runZonedGuarded`)
+  /// https://api.dart.dev/stable/2.10.4/dart-async/runZonedGuarded.html
+  static Future<void> init(
+    OptionsConfiguration optionsConfiguration, [
+    AppRunner callback,
```
```suggestion
    AppRunner appRunner,
```

marandaneto: I like this idea and I think we should bring this into `SentryFlutter` too, as this is a nice add on but most probably opt-in, I already got questions about the `runZonedGuard` and some people prefer to have their own, including when running multiple crash reporting tools.
Can we make this optional on SentryFlutter as well?

flutter/lib/src/sentry_flutter.dart
```diff
@@ -10,34 +9,36 @@ import 'package:sentry/sentry.dart';
...
+import 'web_platform_checker.dart' if (dart.library.io) 'platform_checker.dart';
 
 /// Sentry Flutter SDK main entry point
 mixin SentryFlutter {
   static const _channel = MethodChannel('sentry_flutter');
 
   static Future<void> init(
     OptionsConfiguration optionsConfiguration,
-    Function callback, {
+    AppRunner callback, {
```
```suggestion
    [AppRunner appRunner], {
```

dart/example/main.dart
marandaneto: according to https://dart.dev/codelabs/async-await

`Future<void> main() async`

even a `main` method has a `Future<void>` if its body is `async`, so wondering if we should do the same.

marandaneto: I'm quite convicend that a `FutureOr<void>` def. for `callback` is the safest way, also:
`await callback();`
worst case, it's not used at all as most of the people will just do `void`, but if they wanna do async stuff, we don't need a breaking change again, wdyt?

dart/lib/src/sentry.dart
```diff
@@ -34,6 +49,38 @@ class Sentry {
...
+      // catch any errors that may occur within the entry function, main()
+      // in the ‘root zone’ where all Dart programs start
+      options.addIntegration(isolateErrorIntegration);
+    }
+
+    // finally the runZonedGuarded, catch any errors in Dart code running
+    // ‘outside’ the Flutter framework
+    if (appRunner != null) {
+      options.addIntegration(runZonedGuardedIntegration(appRunner));
+    }
```
marandaneto, line 91: One of the crucial things of this feature was to preserve the `ordering of the execution` of the integrations and this depends on the order they were added to the `options.integrations` list.

I see now that `runZonedGuardedIntegration` is gonna be executed before `flutterErrorIntegration` and this would cause Flutter errors to not be reported on SDK init.

Maybe `Sentry.init` should accept an optional parameter which is the Integration that should be added before `runZonedGuardedIntegration`, or reordering the integrations list on `SentryFlutter.init`, not sure which one is more viable, but the first one seems less flakey.

dart/lib/src/default_integrations.dart
rxlabz: ✅

dart/lib/src/sentry.dart
rxlabz: ✅

flutter/lib/src/sentry_flutter.dart
rxlabz: ✅ I had to add it to optional named arguments

flutter/lib/src/sentry_flutter.dart
```diff
@@ -10,34 +9,36 @@ import 'package:sentry/sentry.dart';
...
 
 /// Sentry Flutter SDK main entry point
 mixin SentryFlutter {
   static const _channel = MethodChannel('sentry_flutter');
 
   static Future<void> init(
-    OptionsConfiguration optionsConfiguration,
-    Function callback, {
+    OptionsConfiguration optionsConfiguration, {
+    AppRunner appRunner,
```
marandaneto: on `Sentry.init` we use: [] arguments, so we don't need to pass the name.
why did we do a different approach here?

I think it'd be less verbose to just doing the same, `[ AppRunner appRunner ]` so a name is not required as this is most probably the most common scenario.

rxlabz: we already had named arguments ( `packageLoader`and `isIOSChecker` ), we can't have both named and anonymous optional arguments. Do you prefer 3 positional optional arguments ?
```dart
static Future<void> init(
    AppRunner appRunner,
    PackageLoader packageLoader = _loadPackageInfo,
    iOSPlatformChecker isIOSChecker = isIOS,
  ]) async {}
```

marandaneto: I see, I didn't know this limitation.

wondering if we could clean up the PackageLoader and iOSPlatformChecker from the `init` method at all then.

on Android, we add all the necessary add ons to the SentryOptions and set default values to them (adding a public setter to overwrite on tests), so we could still mock and test.

My concern is, if we do a different way of initing the SDK, we'll need to change all the documentation to reflect the way of initing it.

We can start with the `[]` params to merge finish this task and refactor later as this won't be a breaking change anyway, those classes ar only used for testing anyway, wdyt?

rxlabz: ✅ I'm agree : as it's mainly for testing purpose, we totally can make them positional for now, and refactoring it later to remove them.

dart/lib/src/sentry.dart
rxlabz: ✅ added a positional `initialIntegrations` to `Sentry.init`

dart/example_web/web/main.dart
rxlabz: so after investigations : 
in browser release mode, capturing the error without adding the stacktrace works, but fails with the optional stacktrace if it is passed as stacktrace, but works if it is passes as String :)

 - this works in web release mode
```dart
final sentryId = await Sentry.captureException(
      error,
      stackTrace: stackTrace.toString(),
    );
```

- this fails in web release mode
```dart
final sentryId = await Sentry.captureException(
      error,
      stackTrace: stackTrace,
    );
```

marandaneto: do you have an idea why this happens?

we could add a condition like:

`stackTrace: (stackTrace != null) ? ((isWeb) ? stackTrace.toString() : stackTrace) : null`

adding a comment why of course.

dart/example/main.dart
rxlabz: yes, without it the dart process seems to never end

dart/example_web/web/main.dart
rxlabz: to be more precise without `toString()`it works in chrome, not working in Safari, but I have no idea why.

dart/example/main.dart
rxlabz: ✅ I changed it to `FutureOr<void>` but I'm not sure of what is the safest way https://github.com/dart-lang/sdk/issues/33218

dart/lib/src/sentry.dart
rxlabz: ✅

dart/example_web/web/main.dart
marandaneto: damn thats problematic, let's fix for Chrome and add a comment for Safari

flutter/lib/src/sentry_flutter.dart
marandaneto: create a task for this in the board, to not be forgotten

dart/example_web/web/main.dart
rxlabz: as it is it works on Chrome, where would be the better place for a comment ? in the readme ? or StacktraceFactory ?

marandaneto: both I'd say, but please create a task for this too, (to research and fix it)

dart/README.md
```diff
@@ -40,10 +40,48 @@ import 'dart:async';
...
+    );
+  }
+}
+
+void aMethodThatMightFail() {
+  throw null;
+}
+```
+
+Or, if you don't want to run your app in its own error zone [runZonedGuarded] :
```
marandaneto:
```suggestion
Or, if you want to run your app in your own error zone [runZonedGuarded] : 
```

dart/README.md
```diff
@@ -40,10 +40,48 @@ import 'dart:async';
...
-  });
+  await Sentry.init(
+    (options) {
+      options.dsn = 'https://example@sentry.io/add-your-dsn-here';
+    }, 
+    initApp, // Init your App.
+  );
+}
+
+void initApp() {
```
marandaneto, line 51:
```suggestion
Future<void> initApp() async {
```

marandaneto: you can't do `await` without `async`

dart/README.md
```diff
@@ -40,10 +40,48 @@ import 'dart:async';
...
+    (options) {
+      options.dsn = 'https://example@sentry.io/add-your-dsn-here';
+    }, 
+  );
+
+  // Init your App.
+  initApp();
+}
 
+void initApp() {
```
marandaneto, line 73: same as https://github.com/getsentry/sentry-dart/pull/187/files#r531580025

dart/README.md
```diff
@@ -40,10 +40,48 @@ import 'dart:async';
...
 
 Future<void> main() async {
-  await Sentry.init((options) {
-    options.dsn = 'https://example@sentry.io/add-your-dsn-here';
-  });
+  await Sentry.init(
+    (options) {
+      options.dsn = 'https://example@sentry.io/add-your-dsn-here';
+    }, 
+    initApp, // Init your App.
```
marandaneto, line 47: I think we should just do the `runApp(MyApp())` here now, having a method that throws and captures is a bit non sense for the appRunner callback

dart/example/main.dart
marandaneto: Dart does not end the process or it gets a deadlock somewhere in our SDK? worth checking

dart/lib/src/sentry.dart
```diff
@@ -1,15 +1,22 @@
...
+    if (dart.library.io) 'isolate_error_integration.dart';
 import 'protocol.dart';
 import 'sentry_client.dart';
 import 'sentry_options.dart';
+import 'utils.dart';
 
 /// Configuration options callback
 typedef OptionsConfiguration = FutureOr<void> Function(SentryOptions);
 
+// run main app callback
```
```suggestion
/// Runs a callback inside of the `runZonedGuarded` method, useful for running your `runApp(MyApp())`
```

dart/test/sentry_test.dart
```diff
@@ -93,17 +102,61 @@ void main() {
...
-      await Sentry.init((options) {
-        options.dsn = fakeDsn;
-        options.addIntegration(integration);
-      });
+      await Sentry.init(
+        (options) {
+          options.dsn = fakeDsn;
+          options.addIntegration(integration);
+        },
+        callback,
```
marandaneto: appRunner

flutter/README.md
```diff
@@ -55,6 +53,24 @@ void aMethodThatMightFail() {
 }
 ```
 
+Or, if you don't want to run your app in its own error zone [runZonedGuarded] :
```
marandaneto:
```suggestion
Or, if you want to run your app in your own error zone [runZonedGuarded] :
```

flutter/lib/src/sentry_flutter.dart
```diff
@@ -10,34 +9,39 @@ import 'package:sentry/sentry.dart';
 import 'default_integrations.dart';
 import 'file_system_transport.dart';
 import 'version.dart';
+import 'web_platform_checker.dart' if (dart.library.io) 'platform_checker.dart';
 
 /// Sentry Flutter SDK main entry point
 mixin SentryFlutter {
   static const _channel = MethodChannel('sentry_flutter');
 
   static Future<void> init(
```
marandaneto, line 22: let's add docs for this method now :)

dart/lib/src/sentry.dart
```diff
@@ -34,6 +50,44 @@ class Sentry {
...
+
+    // if the SENTRY_DSN is set, we read from it.
+    options.dsn = const bool.hasEnvironment('SENTRY_DSN')
+        ? const String.fromEnvironment('SENTRY_DSN')
+        : options.dsn;
+
+    if (initialIntegrations != null && initialIntegrations.isNotEmpty) {
+      initialIntegrations
+          .forEach((integration) => options.addIntegration(integration));
+    }
```
marandaneto, line 78:
```suggestion
    // we need to execute integrations in a specific order sometimes,
    // and this initialIntegrations list makes it possible to inject and add
    // integrations before the runZonedGuardedIntegration gets added
    if (initialIntegrations != null && initialIntegrations.isNotEmpty) {
      initialIntegrations
          .forEach((integration) => options.addIntegration(integration));
    }
```

dart/README.md
rxlabz: there is no `runApp` in Dart. I removed the initApp implementation. Not sure it makes sense to declare an empty method  `Future<void> initApp() async { // your app code }`, so I let `void initApp() { // your app code }` . WDYT ?

dart/lib/src/sentry.dart
rxlabz: ✅

dart/README.md
marandaneto: well this is the README right, the `callback` is mainly used for people that have an UI so `runApp(MyApp())` makes sense, of course on the sample we can't do it otherwise it won't compile

rxlabz: this is the dart/readme , and dart web has no runApp either. It only makes sense in flutter/readme ? Am I missing something ?

flutter/README.md
```diff
@@ -55,6 +53,24 @@ void aMethodThatMightFail() {
 }
```
rxlabz, line 53: not sure it's helpful,  WDYT ?

dart/README.md
marandaneto: so the `appRunner` callback for the Dart SDK is an optional parameter if they don't want to use the Flutter package for some reason. if they decide to use the `appRunner` callback, that means they have a sort of UI to executed on the `runZoned` method, right, it could be a `runApp(MyApp())` or really any other method, this is just an example that they should do inside of this callback.

marandaneto: an empty method + comment // run your app here, also would work btw

dart/example/main.dart
rxlabz: it seems to be caused by the `isolateErrorIntegration`

dart/README.md
rxlabz: ✅

dart/test/sentry_test.dart
rxlabz: ✅

flutter/lib/src/sentry_flutter.dart
rxlabz: ✅

dart/example/main.dart
rxlabz: ✅ I removed the `exit` call

marandaneto: but the problem still exists or what? just to know if you found out or not

dart/example_web/web/main.dart
rxlabz: ✅

dart/example/main.dart
rxlabz: yes, the problem still exists. it's caused by the isolateErrorIntegration, but didn't found how to fix it.

flutter/example/lib/main.dart
```diff
@@ -83,11 +80,6 @@ class _MyAppState extends State<MyApp> {
                 onPressed: () async =>
                     {await compute(loop, 10).catchError(handleError)},
               ),
-              RaisedButton(
-                child: const Text('Dart: Fail in compute'),
-                onPressed: () async =>
-                    {await compute(loop, 10).catchError(handleError)},
-              ),
```
marandaneto, line 91: why this example is gone?

dart/example_web/web/main.dart
marandaneto: task was not created, so I did myself
https://github.com/getsentry/sentry-dart/projects/1#card-50215638

dart/example/main.dart
marandaneto: task was not created, so I did myself
https://github.com/getsentry/sentry-dart/projects/1#card-50215663

flutter/example/lib/main.dart
marandaneto: oh it was duplicated

PR #188, opened 2020-11-25
Ref: add missing docs and move sentry web plugin to the inner src folder

marandaneto:

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [X] Refactoring

## :scroll: Description
Ref: add missing docs and move sentry web plugin to the inner src folder

All the fields should be documented as we get docs for free from pub.dev

## :pencil: Checklist

- [X] All tests passing
- [X] No breaking changes


dart/lib/src/protocol/device.dart
```diff
@@ -1,3 +1,4 @@
+/// If a device is on portrairt or landscape mode
```
marandaneto:
```suggestion
/// If a device is on portrait or landscape mode
```

PR #190, opened 2020-11-26
Fix: Handle event immutable lists and maps

rxlabz:

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

Fix errors when the event contains immutable lists or maps

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes


dart/lib/src/sentry_client.dart
```diff
@@ -181,7 +181,7 @@ class SentryClient {
 
     for (final processor in eventProcessors) {
       try {
-        event = await processor(event, hint);
+        event = await processor(event.copyWith(), hint);
```
marandaneto: do we need this if all the immutability is fixed?

rxlabz: yes, because, as the event members are public, it's the only way to guarantee that nothing is immutable in the event passed to the processor

marandaneto: `copyWith` this should be called to `beforeSend` too then?

marandaneto: and every other callback that could be set by the client

marandaneto: or in the first method of the proessing pipeline, right after captureEvent maybe?

rxlabz: ✅ I finally clone the event in the beginning of the `client.captureEvent`

dart/lib/src/protocol/debug_meta.dart
```diff
@@ -7,10 +7,14 @@ class DebugMeta {
...
-
-  DebugMeta({this.sdk, List<DebugImage> images}) : _images = images;
+  /// The list of debug images contains all dynamic libraries loaded
+  /// into the process and their memory addresses.
+  /// Instruction addresses in the Stack Trace are mapped into the list of debug
+  /// images in order to retrieve debug files for symbolication.
+  List<DebugImage> get images => List.from(_images);
+
+  DebugMeta({this.sdk, List<DebugImage> images})
+      : _images = images != null ? List.unmodifiable(images) : null;
```
marandaneto: its the opposite, the private field should be `modifiable`
the public one should be `unmodifiable`
the `ctor` should copy to be sure that the private field is `modifiable`

dart/lib/src/protocol/sdk_version.dart
```diff
@@ -41,8 +41,9 @@ class SdkVersion {
     List<String> integrations,
     List<SentryPackage> packages,
   })  : assert(name != null || version != null),
-        _integrations = integrations ?? [],
-        _packages = packages ?? [];
+        // List.from prevents from having immutable lists
+        _integrations = integrations != null ? List.from(integrations) : [],
+        _packages = packages != null ? List.from(packages) : [];
```
marandaneto, line 46: now you can fix `SentryFlutter._setSdk` as I needed to do `List.from` manually in there, now the ctor does it automatically

dart/lib/src/protocol/debug_meta.dart
rxlabz: ✅

dart/test/sentry_client_test.dart
```diff
@@ -477,6 +478,47 @@ void main() {
   test("options can't be null", () {
     expect(() => SentryClient(null), throwsArgumentError);
   });
+
+  group('EventProcessors', () {
+    SentryOptions options;
+
+    setUp(() {
+      options = SentryOptions(dsn: fakeDsn);
+      options.addEventProcessor(
```
marandaneto, line 503: can we add a test for the `beforeSend` callback as well? the beforeSend is the most common callback to mutate an event

dart/lib/src/sentry_client.dart
```diff
@@ -52,6 +52,7 @@ class SentryClient {
     dynamic stackTrace,
     dynamic hint,
   }) async {
+    event = event.copyWith(); // clone with mutable list and map fields
```
marandaneto: guess I already asked it, but just to double-check, if we are defending ourselves now and copying all the List/Map references, in which cases do we run into immutable problems? do we still need this?

rxlabz: ✅ true ! I updated the SentryEvent ctor and copyWith to avoid the need for this copyWith call

dart/test/sentry_client_test.dart
rxlabz: ✅

dart/test/hub_test.dart
```diff
@@ -112,21 +112,24 @@ void main() {
...
         scope
-          ..level = SentryLevel.debug
           ..user = fakeUser
+          ..level = SentryLevel.debug
           ..fingerprint = ['1', '2'];
       });
       await hub.captureEvent(fakeEvent);
 
-      await hub.captureEvent(fakeEvent);
+      //await hub.captureEvent(fakeEvent);
```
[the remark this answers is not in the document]
rxlabz: ✅

dart/lib/src/protocol/debug_meta.dart
```diff
@@ -7,10 +7,14 @@ class DebugMeta {
 
   final List<DebugImage> _images;
 
-  /// An immutable list of dynamic libraries loaded into the process (see below).
+  /// The list of debug images contains all dynamic libraries loaded
```
marandaneto: worth adding the `immutable` comment back

dart/test/sentry_client_test.dart
```diff
@@ -470,16 +471,84 @@ void main() {
       final client = SentryClient(options);
       await client.captureEvent(fakeEvent);
 
-      verify(options.transport.send(any)).called(1);
+      //verify(options.transport.send(any)).called(1);
```
marandaneto: could remove the comment

marandaneto (approved):
LGTM, thanks for fixing this

dart/lib/src/protocol/debug_meta.dart
rxlabz: ✅

dart/test/sentry_client_test.dart
rxlabz: ✅

PR #197, opened 2020-11-28
Feature: Flutter NavigationObserver

### LINKED ISSUE #182: [sentry-flutter] Support for navigation Breadcrumbs
I would like to request support for navigation breadcrumbs.

The motivation is, that a developer can better recreate the user flow to track down the reported crash.

This can be implemented using Flutters [NavigatorObserver](https://api.flutter.dev/flutter/widgets/NavigatorObserver-class.html) or [RouteObserver](https://api.flutter.dev/flutter/widgets/RouteObserver-class.html).

A very simplisitic example of this feature could like this:
(Please note that I haven't confirmed if this actually compiles. It is just meant as an illustration of possible solution.)
```dart

Future<void> main() async {
  await SentryFlutter.init(
    (options) {
      // For better groupping, change the 'example' below with your own App's package.
      options.addInAppInclude('sentry_flutter_example');
    },
    () {
        runApp(MyApp())
    },
   );
  }
}

class MyApp extends StatelessWidget {
  @override
  Widget build(BuildContext context) {
    return MaterialApp(
      navigatorObservers: [
        SentryNavigatorObserver(),
      ],
      theme: ThemeData.dark().copyWith(scaffoldBackgroundColor: darkBlue),
      debugShowCheckedModeBanner: false,
      home: Scaffold(
        body: Center(
          child: MyWidget(),
        ),
      ),
    );
  }
}

class SentryNavigatorObserver extends NavigatorObserver {
  void didPush(Route<dynamic> route, Route<dynamic> previousRoute) {
    addBreadcrumb(route.settings.name, "wasPushed");
  }

  void didPop(Route<dynamic> route, Route<dynamic> previousRoute) {
    addBreadcrumb(route.settings.name, "wasPopped");
  }

  void didRemove(Route<dynamic> route, Route<dynamic> previousRoute) {
    addBreadcrumb(route.settings.name, "wasRemoved");
  }

  void didReplace({ Route<dynamic> newRoute, Route<dynamic> oldRoute }) {
    addBreadcrumb(newRoute.settings.name, "didReplace");
  }

  void addBreadCrumb(String routeName, String type) {
    Sentry.addBreadcrumb(Breadcrumb(
      message: routeName,
      category: "navigation",
      type: type,
    ));
  }

  void didStartUserGesture(Route<dynamic> route, Route<dynamic> previousRoute) { 
        // no breadcrumb needed
  }

  void didStopUserGesture() {
    // no breadcrumb needed
  }
}
```

ueman:

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
I've introduced `SentryNavigationObserver` to record navigation events as breadcrumbs.

This adds navigation breadcrumbs for crash reports.
It fixes https://github.com/getsentry/sentry-dart/issues/182

I've written some tests.
However I would need some guidance to make the tests more throughout.

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps
I would like to request some guidance on what tests to write.


flutter/lib/src/navigation/sentry_navigation_observer.dart
```diff
@@ -0,0 +1,71 @@
...
+///
+/// See also:
+///   - https://api.flutter.dev/flutter/widgets/RouteObserver-class.html
+class SentryNavigatorObserver extends RouteObserver<PageRoute<dynamic>> {
+  factory SentryNavigatorObserver(Hub hub) {
+    if (hub != null) {
+      return SentryNavigatorObserver._(hub);
+    }
+    return SentryNavigatorObserver._(Sentry.currentHub);
+  }
```
marandaneto: the idea here would be using an `Integration` (default_integrations.dart) that takes a `Hub` and `SentryOptions`,
also adding a new `bool` to `SentryOptions` so we can turn it off if necessary.
Then adding the `Integration` at `_addDefaultIntegrations` (sentry_flutter.dart)

See how we do on Android:

https://github.com/getsentry/sentry-java/blob/main/sentry-android-core/src/main/java/io/sentry/android/core/ActivityBreadcrumbsIntegration.java

doing this, the Integration will be registered automatically, no need of a single line of code on the client.

flutter/lib/src/navigation/sentry_navigation_observer.dart
```diff
@@ -0,0 +1,71 @@
...
+    String type,
+    RouteSettings from,
+    RouteSettings to,
+  }) {
+    hub.addBreadcrumb(NavigationBreadcrumb(
+      from: from,
+      to: to,
+      navigationType: type,
+    ));
+  }
```
marandaneto: ideally, we use the same values for `Breadcrumbs`, See on Android

https://github.com/getsentry/sentry-java/blob/main/sentry-android-core/src/main/java/io/sentry/android/core/ActivityBreadcrumbsIntegration.java#L63-L67

type= navigation
state= didPop, didReplace...
screen=name of the screen
category=ui.lifecycle
and adding more fields if necessary, so our UI will be ready to display nicely.

flutter/lib/src/navigation/navigation_breadcrumb.dart
```diff
@@ -0,0 +1,59 @@
+import 'package:flutter/material.dart';
+import 'package:sentry/sentry.dart';
+
+/// This class makes it easier to record breadcrumbs for navigation events by
+/// accepting Flutters [RouteSettings](https://api.flutter.dev/flutter/widgets/RouteSettings-class.html).
+///
+/// See also:
+///   - https://flutter.dev/docs/cookbook/navigation/navigate-with-arguments
+class NavigationBreadcrumb extends Breadcrumb {
```
marandaneto: I like the idea, but I think having a class for each type of breadcrumb is a bit overkill,
I'd be in to expose a nice Factory or named constructor in the Breadcrumb class to take parameters instead of setting values manually, See this on Java:

https://github.com/getsentry/sentry-java/blob/3fad970bef0675f63fcfe63cd2ff871b62743ed6/sentry/src/main/java/io/sentry/Breadcrumb.java#L46-L60

marandaneto:
I've removed the WIP title and made it a Draft

marandaneto:
thanks so much @ueman this is going in a nice direction :)
with a little bit of work, this can land on the next version, let me know if you need any help.

flutter/lib/src/navigation/sentry_navigation_observer.dart
marandaneto: or using the `navigation` type: see https://develop.sentry.dev/sdk/event-payloads/breadcrumbs/

```
{
  "type": "navigation",
  "category": "navigation",
  "timestamp": "2016-04-20T20:55:53.845Z",
  "data": {
    "from": "/login",
    "to": "/dashboard"
  }
}
```

flutter/lib/src/navigation/navigation_breadcrumb.dart
ueman: `Breadcrumb` is a class of sentry-dart, so I can't add a constructor or factory with `RouteSettings` parameters without adding a Flutter dependency in sentry-dart. So this was my next best idea.

flutter/lib/src/navigation/sentry_navigation_observer.dart
ueman: Thanks, will do.

flutter/lib/src/navigation/sentry_navigation_observer.dart
ueman: A navigation observer must unfortunately be added manually as they can't be added automatically.
See https://api.flutter.dev/flutter/material/MaterialApp/navigatorObservers.html

```dart
MaterialApp(
   navigatorObservers: [
     SentryNavigatorObserver(),
   ],
   // other parameter ...
)
```

So I guess the `Integration` and the `SentryOptions` option are unnecessary?
I would however add this to the readme so users are aware of it.

ueman: I guess the `ActivityBreadcrumbsIntegration.java` is more equivalent to the [`WidgetsBindingObserver`](https://api.flutter.dev/flutter/widgets/WidgetsBindingObserver-class.html) (this would be a nice addition in another PR) than the `NavigationObserver`.

marandaneto: > I guess the `ActivityBreadcrumbsIntegration.java` is more equivalent to the [`WidgetsBindingObserver`](https://api.flutter.dev/flutter/widgets/WidgetsBindingObserver-class.html) (this would be a nice addition in another PR) than the `NavigationObserver`.

true, I was just trying to give examples on how to use the `Integration` feature

flutter/lib/src/navigation/navigation_breadcrumb.dart
marandaneto: yeah I do get that, but I'd not pass the whole `RouteSettings` but rather the values (as parameters), so if another Router FW is gonna be integrated, the facade is already available thru the `Breadcrumb` class.

flutter/lib/src/navigation/sentry_navigation_observer.dart
```diff
@@ -0,0 +1,84 @@
...
+///
+/// See also:
+///   - https://api.flutter.dev/flutter/widgets/RouteObserver-class.html
+class SentryNavigatorObserver extends RouteObserver<PageRoute<dynamic>> {
+  factory SentryNavigatorObserver({Hub hub}) {
+    if (hub != null) {
+      return SentryNavigatorObserver._(hub);
+    }
+    return SentryNavigatorObserver._(Sentry.currentHub);
+  }
```
marandaneto: instead of using the `Sentry.currentHub` I'd rather use the `HubAdapter` which is basically redirecting to `Sentry.currentHub`, so we don't have a static class reference and still, unit testable.

marandaneto:
```suggestion
  factory SentryNavigatorObserver({Hub hub = HubAdapter()}) {
    return SentryNavigatorObserver._(hub);
  }
```

just an example.

flutter/lib/src/navigation/sentry_navigation_observer.dart
ueman: Ah, I see. So what do you suggest on how to make the `SentryNavigationObserver` available to users?

marandaneto: > A navigation observer must unfortunately be added manually as they can't be added automatically.
> See https://api.flutter.dev/flutter/material/MaterialApp/navigatorObservers.html
> 
> ```dart
> MaterialApp(
>    navigatorObservers: [
>      SentryNavigatorObserver(),
>    ],
>    // other parameter ...
> )
> ```
> 
> So I guess the `Integration` and the `SentryOptions` option are unnecessary?
> I would however add this to the readme so users are aware of it.

Oh I see, that makes sense, yeah if the user needs to add manually, then for sure `Integration` and `SentryOptions` are unnecessary, thanks for pointing this out.

marandaneto: > Ah, I see. So what do you suggest on how to make the `SentryNavigationObserver` available to users?

reading the `NavigatorObserver` docs now I see that it'd not be possible to hook it up automatically unless we create our own `SentryMaterialApp` but I guess this is too much intrusive and if people need to create a new `SentryMaterialApp` anyway, I'd rather having less code and documenting they should do (what u have done on the README), so you are on the right path.

marandaneto: thanks for clarifying btw, I didn't play enough with the Flutter UI yet.

flutter/test/navigation_observer_test.dart
```diff
@@ -0,0 +1,270 @@
...
+      expect(hub.breadcrumbs.length, 1);
+      expect(hub.breadcrumbs.first.data, <String, dynamic>{
+        'state': 'didPush',
+      });
+    });
+  });
+}
+
+/// Used to test if breadcrumbs are really added.
+class _FakeHub implements Hub {
```
marandaneto: some people are not fun of `Mocks` and I get that, but if you use the `MockHub` that already exists and `verify` the passed breadcrumb, you could `assert` the values having less boilerplate (all the `UnimplementedError` methods)

flutter/lib/src/navigation/sentry_navigation_observer.dart
```diff
@@ -0,0 +1,81 @@
...
+
+  final Hub hub;
+
+  @override
+  void didPush(Route<dynamic> route, Route<dynamic> previousRoute) {
+    super.didPush(route, previousRoute);
+    _addBreadcrumb(
+      type: 'didPush',
+      from: previousRoute?.settings,
+      to: route?.settings,
```
marandaneto, line 44: we are not using the null safety feature yet, wondering if this would cause problems, I'd rather just do null checks till the null-safety gets stable and we do a full migration

flutter/test/navigation_observer_test.dart
ueman: Done ✅

flutter/lib/src/navigation/navigation_breadcrumb.dart
ueman: I would like to keep it as it is, because it is pretty Flutter specific and it makes the testing easier.

Also in future versions one could easily add more Flutter specific information. [`PageRoute`](https://api.flutter.dev/flutter/widgets/PageRoute-class.html) and co have much more information, which could potentially be useful. For example the name of the Route itself (eg. MaterialPageRoute, CupertinoPageRoute...) or if the navigation action covers the whole screen or if it is a dialog, bottom sheet or something else.

marandaneto: so the code looks good, but we are coupling a class called `NavigationBreadcrumb` which is a pretty generic name for any navigation FW/3rd party lib (to the RouteObserver lib.)

I'd suggest moving all the code to the `SentryNavigatorObserver` class which is pretty much what is used for, don't get me wrong, the abstraction does make sense, but for an SDK that could be integrated with other 3rd parties not (IMO), as the `NavigationBreadcrumb` takes a `RouteSettings`.

so I see 2 alternatives, either we rename `NavigationBreadcrumb` to something more specific related to the `RouteObserver` or we move this logic to the `_addBreadcrumb` class, really your call, happy with both.

marandaneto:
if you resolve the conflicts and mark it as ready to review, we could see if the CI is happy.

once more, thanks for contributing (L)

flutter/example/lib/main.dart
```diff
@@ -34,12 +34,20 @@ class _MyAppState extends State<MyApp> {
...
         appBar: AppBar(title: const Text('Sentry Flutter Example')),
         body: SingleChildScrollView(
           child: Column(
             children: [
               const Center(child: Text('Trigger an action:\n')),
+              RaisedButton(
+                child: const Text('Open another Scaffold'),
+                onPressed: () =>
+                    SecondaryScaffold.openSecondaryScaffold(context),
+              ),
```
marandaneto: so not sure if there's a bug here or you did intentionally:

> Navigator operation requested with a context that does not include a Navigator.
> The context used to push or pop routes from the Navigator must be that of a widget that is a descendant of a Navigator widget.

marandaneto:
I found a bug on the breadcrumbs parsing and the fix is landing here https://github.com/getsentry/sentry-dart/pull/201

flutter/example/lib/main.dart
ueman: Oh shoot, you're right. That was definitely a mistake on my side. I've fixed it.

marandaneto: cool, waiting for the CI #201 and you can merge main into your branch and GG :)

marandaneto:
Screenshot 2020-11-29 at 10 53 26

marandaneto (approved):
LGTM

marandaneto:
thanks a lot @ueman 
happy to merge when CI is happy.

PR #203, opened 2020-11-29
Feature: WidgetsBindingObserver to track application wide events

ueman:

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
This PR introduces a [WidgetsBindingObserver](https://api.flutter.dev/flutter/widgets/WidgetsBindingObserver-class.html) which can track some application wide changes (like for example lifecycle, accessibility features or display features) as breadcrumbs. It gets added by default.

Currently it only tracks the following events:
- app lifecycle
- display metrics
- changes of light to dark mode and vice versa
- text scale factor
- memory pressure

There are more events which can be tracked, see next steps.

It helps developer to better understand crash reports which result from the events mentioned above.

I've written tests and made sure the example still works as expected.

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps
There is still room for improvement and additional functionality. However I'm not sure how to proceed with those.
Therefore I would like to leave it up to you. I also left documentation and some thoughts on what to do.


CHANGELOG.md
```diff
@@ -12,6 +12,7 @@
 - Fix: NDK integration was being disabled by a typo
 - Fix: missing toList for debug meta #192
 - Enhancement: NavigationObserver to record Breadcrumbs for navigation events #197
+- Enhancement: SentryWidgetsBindingObserver
```
```suggestion
- Enhancement: Add SentryWidgetsBindingObservern an Integration that captures certain window and device events.
```

flutter/lib/src/sentry_flutter.dart
```diff
@@ -119,6 +119,9 @@ mixin SentryFlutter {
     if (isIOS()) {
       options.addIntegration(loadContextsIntegration(options, _channel));
     }
+
+    // track Flutter application events
+    options.addIntegration(widgetsBindingIntegration);
```
marandaneto: does this integration need to be executed before one does `runApp(MyApp)`?
if so, consider adding the integration to this list instead:
https://github.com/getsentry/sentry-dart/blob/main/flutter/lib/src/sentry_flutter.dart#L39-L41

flutter/lib/src/widgets_binding_observer.dart
```diff
@@ -0,0 +1,155 @@
...
+  /// This method records lifecycle events.
+  /// It tries to mimic the behavior of ActivityBreadcrumbsIntegration of Sentry
+  /// Android for lifecycle events.
+  ///
+  /// On Android and iOS this records lifecycle event breadcrumbs which loosely
+  /// correspond to their respectiv platforms lifecycle events.
+  ///
+  /// Does this need to be tracked? The nativ Android an iOS integration
+  /// also tracks this but what about Web, Linux, Windows and MacOS?
+  /// The message could include that this is a Flutter message.
```
marandaneto: that's a good question, right now it's gonna make 'duplicated' breadcrumbs as the Native SDKs (only Android and iOS for now) already track App and Activity lifecycles and we don't officially support the other platforms.
Does the Flutter `WidgetsBindingObserver` API work for all the platforms?

flutter/lib/src/widgets_binding_observer.dart
```diff
@@ -0,0 +1,155 @@
...
+  }
+
+  /// Called when the application's dimensions change. For example,
+  /// when a phone is rotated or an application window is resized.
+  ///
+  /// See also:
+  ///   - [Window.onMetricsChanged](https://api.flutter.dev/flutter/dart-ui/Window/onMetricsChanged.html)
+  @override
+  void didChangeMetrics() {
+    // Should this method push a new scope with a new Device Context?
```
marandaneto: not really, every time one does `captureException`, event or message, we read a fresh state of the device and set to the Contexts automatically (via the Native SDKs), so ideally, this will create only a breadcrumb with the changes (new values).

A bit similar to https://github.com/getsentry/sentry-java/blob/main/sentry-android-core/src/main/java/io/sentry/android/core/AppComponentsBreadcrumbsIntegration.java#L75-L96
but there, we don't track the height and width size, but only that it has changed.

Does this API work for macOS/Linux and Web?

flutter/lib/src/sentry_flutter.dart
ueman: It doesn't necessarily *need* to be executed before `runApp(MyApp)` but it should be executed as soon as possible after `WidgetsBinding.ensureInitialized()`. So yeah, it should be added to the linked list.

flutter/lib/src/widgets_binding_observer.dart
```diff
@@ -0,0 +1,155 @@
...
+  ///   - [Window.onPlatformBrightnessChanged](https://api.flutter.dev/flutter/dart-ui/Window/onPlatformBrightnessChanged.html)
+  @override
+  void didChangePlatformBrightness() {
+    final brightness = WidgetsBinding.instance.window.platformBrightness;
+    final brightnessDescription =
+        brightness == Brightness.dark ? 'dark' : 'light';
+
+    hub.addBreadcrumb(Breadcrumb(
+      message: 'Platform brightness was changed to $brightnessDescription.',
+      category: 'ui.lifecycle',
```
marandaneto: the category here could be something else, platform brightness is not so related to lifecycle, for generic things, I'd do something like:

https://github.com/getsentry/sentry-java/blob/main/sentry-android-core/src/main/java/io/sentry/android/core/PhoneStateBreadcrumbsIntegration.java#L95-L97

so a fixed type and category, but a custom action, so our UI on sentry dashboard could render nicely.

flutter/lib/src/widgets_binding_observer.dart
```diff
@@ -0,0 +1,155 @@
...
+
+  /// See also:
+  ///   - [Window.onTextScaleFactorChanged]https://api.flutter.dev/flutter/dart-ui/Window/onTextScaleFactorChanged.html)
+  @override
+  void didChangeTextScaleFactor() {
+    final newTextScaleFactor = WidgetsBinding.instance.window.textScaleFactor;
+    hub.addBreadcrumb(Breadcrumb(
+      message: 'Text scale factor changed to $newTextScaleFactor.',
+      category: 'ui',
+      level: SentryLevel.warning,
```
marandaneto: same as https://github.com/getsentry/sentry-dart/pull/203/files#r532241561

plus, why is this a `warning` level?

flutter/lib/src/widgets_binding_observer.dart
```diff
@@ -0,0 +1,155 @@
...
+  void didHaveMemoryPressure() {
+    const message =
+        'App had memory pressure. This indicates that the operating system '
+        'would like applications to release caches to free up more memory.';
+    hub.addBreadcrumb(Breadcrumb(
+      message: message,
+      // This is kinda bad. Therefore this gets added as a warning.
+      level: SentryLevel.warning,
+    ));
+  }
```
marandaneto: similar to https://github.com/getsentry/sentry-java/blob/main/sentry-android-core/src/main/java/io/sentry/android/core/AppComponentsBreadcrumbsIntegration.java#L98-L135

flutter/lib/src/widgets_binding_observer.dart
```diff
@@ -0,0 +1,155 @@
...
+        return 'resumed';
+      case AppLifecycleState.inactive:
+        return 'inactive';
+      case AppLifecycleState.paused:
+        return 'paused';
+      case AppLifecycleState.detached:
+        return 'detached';
+    }
+    return '';
+  }
```
marandaneto, line 159: similar to https://github.com/getsentry/sentry-java/blob/main/sentry-android-core/src/main/java/io/sentry/android/core/ActivityBreadcrumbsIntegration.java

flutter/lib/src/sentry_flutter.dart
marandaneto: I see, so yeah adding to the linked list will make more sense, because we'd be getting the events from the very beginning.

marandaneto:
thanks @ueman once more :)

so my main concern here is, a few of those breadcrumbs are already tracked on the Native SDKs (which we get for free from sentry_flutter), so creating duplicated/noisy breadcrumbs is not something we'd like to do, however, if the Flutter API offers this to all the other platforms (macOS, Web, etc...) it'd make sense to turn it off on the Native SDKs and enable it here only.

I could guide you on how to deactivate the Native SDKs, but we'd need to be sure that the Flutter API offers the same feature on the other platforms, do you know if that works? for Web at least?

flutter/lib/src/widgets_binding_observer.dart
ueman: Yes, this API works at least on macOS, Linux and [Windows](https://github.com/flutter/engine/blob/master/shell/platform/windows/win32_flutter_window.cc#L120). I don't know about web but I would guess so.

flutter/lib/src/widgets_binding_observer.dart
ueman: Yes, it does work on all platforms. `WidgetsBindingObserver` is the universal interface to get information about the hosts windows in which Flutter is embedded.

flutter/lib/src/default_integrations.dart
```diff
@@ -140,3 +142,13 @@ Integration nativeSdkIntegration(SentryOptions options, MethodChannel channel) {
...
+
+/// Integration that captures certain window and device events.
+/// See also:
+///   - [WidgetsBindingObserver](https://api.flutter.dev/flutter/widgets/WidgetsBindingObserver-class.html)
+void widgetsBindingIntegration(Hub hub, SentryOptions options) {
+  // We don't need to call `WidgetsFlutterBinding.ensureInitialized()`
+  // because `FlutterSentry.init` already calls it.
+  WidgetsBinding.instance.addObserver(SentryWidgetsBindingObserver(hub: hub));
+  options.sdk.addIntegration('widgetsBindingIntegration');
+}
```
marandaneto: as this is an automatic integration and people may want to turn it off, it'd be nice to expose an option to do so, see: 

https://github.com/getsentry/sentry-dart/blob/main/dart/lib/src/sentry_options.dart#L243-L252

but creating a new bool as this is Flutter's specific, and then before adding the observer to the WidgetsBinding.instance, we check this bool which is enabled by default.

see an example:

https://github.com/getsentry/sentry-java/blob/main/sentry-android-core/src/main/java/io/sentry/android/core/ActivityBreadcrumbsIntegration.java#L45-L48

flutter/lib/src/widgets_binding_observer.dart
marandaneto: thats cool, so I think it'd make sense to align the automatic breadcrumbs we already have on the Native SDKs and only add to this PR the ones that are not tracked by the native SDKs, so we remove the duplication.
The idea is to keep it simple, only adding "new" breadcrumbs.
Once we support all the other platforms, we could do the opposite, turning it off on the Native SDKs and enabling it here, what do you think?

ueman:
So here's what I would like to do:

As Android and iOS already collect some of this data, it would be nice to collect only new events. However because other platforms have no native integration (yet?) it would be nice to record all of them.
Therefore I propose a `WidgetsBindingObserverConfiguration` class in order to configure which events of `WidgetsBindingObserver` should be tracked.

Something like this:
```dart
class WidgetsBindingObserverConfig {
  WidgetsBindingObserverConfig({
    this.trackLifecycle,
    this.trackWindowMetrics,
    this.trackBrightness,
  });

  factory WidgetsBindingObserverConfig.android() {
    return WidgetsBindingObserverConfig(
      trackLifecycle: false,
      trackWindowMetrics: false,
      trackBrightness: false,
    );
  }

  factory WidgetsBindingObserverConfig.web() {
    return WidgetsBindingObserverConfig(
      trackLifecycle: true,
      trackWindowMetrics: true,
      trackBrightness: true,
    );
  }
   // default options for other platforms...

  final bool trackLifecycle;
  final bool trackWindowMetrics;
  final bool trackBrightness;
  // all the other events...
}
```

This way it could already be configured nicely by default. It would also give the user the option to track only Flutters lifecycle events in case the user disables `SentryOptions.enableAutoNativeBreadcrumbs`.

These options doesn't have to be user facing. For them it still could only be the option to enable or disable them and the SDK chooses which to enable per platform. I hope you get what I mean.

What do you think?

marandaneto:
> So here's what I would like to do:
> 
> As Android and iOS already collect some of this data, it would be nice to collect only new events. However because other platforms have no native integration (yet?) it would be nice to record all of them.
> Therefore I propose a `WidgetsBindingObserverConfiguration` class in order to configure which events of `WidgetsBindingObserver` should be tracked.
> 
> Something like this:
> 
> ```dart
> class WidgetsBindingObserverConfig {
>   WidgetsBindingObserverConfig({
>     this.trackLifecycle,
>     this.trackWindowMetrics,
>     this.trackBrightness,
>   });
> 
>   factory WidgetsBindingObserverConfig.android() {
>     return WidgetsBindingObserverConfig(
>       trackLifecycle: false,
>       trackWindowMetrics: false,
>       trackBrightness: false,
>     );
>   }
> 
>   factory WidgetsBindingObserverConfig.web() {
>     return WidgetsBindingObserverConfig(
>       trackLifecycle: true,
>       trackWindowMetrics: true,
>       trackBrightness: true,
>     );
>   }
>    // default options for other platforms...
> 
>   final bool trackLifecycle;
>   final bool trackWindowMetrics;
>   final bool trackBrightness;
>   // all the other events...
> }
> ```
> 
> This way it could already be configured nicely by default. It would also give the user the option to track only Flutters lifecycle events in case the user disables `SentryOptions.enableAutoNativeBreadcrumbs`.
> 
> These options doesn't have to be user facing. For them it still could only be the option to enable or disable them and the SDK chooses which to enable per platform. I hope you get what I mean.
> 
> What do you think?

That sounds a good idea, although I'd not add those options as an extension method, as the `Integration` is created automatically by our SDK, also because the natural/standardized way of enabling things across SDKs is via SentryOptions.

I'm ok with fine grain options to turn things on and off, see on Android, we have pretty much the same.
https://github.com/getsentry/sentry-java/blob/main/sentry-android-core/src/main/java/io/sentry/android/core/SentryAndroidOptions.java#L22-L35

or https://github.com/getsentry/sentry-java/blob/main/sentry-android-core/src/main/java/io/sentry/android/core/SentryAndroidOptions.java#L165-L175

as an easy win for enabling/disabling everything with a one-liner, if we offer a method like this, we could also turn off the `native` breadcrumbs feature automatically.

2 things I'd consider rather important, the breadcrumbs should follow the same `type`, `category`, `action` etc... as the UI renders it nicely based on those settings (See Android).

I'd document that if you `enable` options X, you might want to disable option Y to not get duplicated breadcrumbs.

how does that sound?

ueman:
Yep, sounds good!

ueman:
How am I supposed to add Flutter specific options to `SentryOptions`? 
It doesn't really make sense for them to be on `SentryOptions` but I also can't subclass `SentryOptions` because `SentryFlutter.init` already receives an options instance from `Sentry.init`. So I can't initialize and pass my own subclass.

Should I just add a second options class? Like for example:
```dart
  await SentryFlutter.init(
    (options, flutterOptions) {
      // configure SentryFlutter with flutterOptions 
      // flutterOptions.exampleOption = true;
    },
    // Init your App.
    () => runApp(MyApp()),
  );
```

marandaneto:
> How am I supposed to add Flutter specific options to `SentryOptions`?
> It doesn't really make sense for them to be on `SentryOptions` but I also can't subclass `SentryOptions` because `SentryFlutter.init` already receives an options instance from `Sentry.init`. So I can't initialize and pass my own subclass.
> 
> Should I just add a second options class? Like for example:
> 
> ```dart
>   await SentryFlutter.init(
>     (options, flutterOptions) {
>       options.dsn = 'https://example@sentry.io/add-your-dsn-here';
>       // configure SentryFlutter with flutterOptions 
>       // flutterOptions.exampleOption = true;
>     },
>     // Init your App.
>     () => runApp(MyApp()),
>   );
> ```

so on Android, we had this problem and we created a specific AndroidOptions and we inited via reflection which is suboptimal but solved it.

I believe that at some point even if we create a FlutterOptions we'll end up with custom options that only make sense to a specific platform (Android/iOS/macOS/Linux, etc...).

fwiw I'd not go with a 2nd options (flutterOptions) as you did, this would diverge totally across SDKs, I guess we either find a way to extend SentryOptions having a FlutterOptions, or to just add them all to SentryOptions anyway.

Not sure if it makes sense to go fine-grained here, I'd just add those options to the current class but documenting its usage for X platform, actually I've done this already for a few options but it's not great, I agree.

@bruno-garcia opinions?

ueman:
Dart for Flutter does not support reflection, so that is no option.

bruno-garcia:
No overloading makes it hard because we could simply add an overload to `Sentry.init` that takes `SentryOptions` so we can pass `FlutterSentryOptions` (which would derive from `SentryOptions`) to it.

Last we have is a `apply` method that will apply values from one instance to the other. Still go with the inheritance but in the callback of `init` apply the values to the instance we got from `Sentry.init`. Would that work?

ueman:
> No overloading makes it hard because we could simply add an overload to `Sentry.init` that takes `SentryOptions` so we can pass `FlutterSentryOptions` (which would derive from `SentryOptions`) to it.
> 
> Last we have is a `apply` method that will apply values from one instance to the other. Still go with the inheritance but in the callback of `init` apply the values to the instance we got from `Sentry.init`. Would that work?

One could do
```dart
// sentry.dart
// Option 1: Optional positional parameter
static Future<void> init(
    AppRunner appRunner,
    List<Integration> initialIntegrations,
    SentryOptions sentryOptions,
  ]) async {
    }

    final options = sentryOptions ?? SentryOptions();
    await _initDefaultValues(options, appRunner, initialIntegrations);

    if (options == null) {
    }

    await _init(options);
  }
```

or 

```dart
  // sentry.dart
  // Option 2: Optional named parameter, would be breaking but Sentry v4 is still in alpha 🤷‍♂️
  // The upside here is that one could document parameters a little bit better since they have a user facing name.
  static Future<void> init(
    OptionsConfiguration optionsConfiguration, {
    AppRunner appRunner,
    List<Integration> initialIntegrations,
    SentryOptions sentryOptions,
  }) async {
    }

    final options = sentryOptions ?? SentryOptions();
    await _initDefaultValues(options, appRunner, initialIntegrations);

    if (options == null) {
    }

    await _init(options);
  }
```

Both options are kinda like method overloading.

I would prefer one of the above suggestions over an `apply` method. My reasoning is that this kind of manual mapping is kinda error prone. Additionally the `SentryFlutterOptions` aren't automatically passed to the `Integration` here https://github.com/getsentry/sentry-dart/blob/main/flutter/lib/src/sentry_flutter.dart#L41 if using an `apply` method.

marandaneto:
hey @ueman a bit off-topic but if you'd like to join our discord chat: https://discord.gg/Ww9hbqr
there's a #flutter channel :)

marandaneto:
hey @ueman be aware of #214
we might get a few git conflicts but might be easily fixable, we were needing to change a few things because of other features, so expect a similar solution of yours landing tomorrow :) thanks!

flutter/lib/src/default_integrations.dart
```diff
@@ -153,25 +153,17 @@ class NativeSdkIntegration extends Integration<SentryFlutterOptions> {
...
-    }
+  FutureOr<void> call(Hub hub, SentryFlutterOptions options) {
+    // We don't need to call `WidgetsFlutterBinding.ensureInitialized()`
+    // because `FlutterSentry.init` already calls it.
+    WidgetsBinding.instance.addObserver(SentryWidgetsBindingObserver(
+      hub: hub,
+      options: options,
+    ));
+
+    options.sdk.addIntegration('widgetsBindingIntegration');
```
ueman, line 170: Does it make sense to return a bool from `Integration.call` indicating wether the integration was added
and have another abstract method on `Integration` returning the name? Then one would not have to add `options.sdk.addIntegration('widgetsBindingIntegration');` themself.
(Mybe try to see if `this.runtimeType.toString()` is an 
adequate name?)

flutter/lib/src/sentry_flutter.dart
```diff
@@ -9,51 +9,49 @@ import 'sentry_flutter_options.dart';
...
+typedef FlutterOptionsConfiguration = FutureOr<void> Function(
+    SentryFlutterOptions);
+
 /// Sentry Flutter SDK main entry point
 mixin SentryFlutter {
   static const _channel = MethodChannel('sentry_flutter');
 
   static Future<void> init(
-    OptionsConfiguration optionsConfiguration, {
+    FlutterOptionsConfiguration optionsConfiguration, {
```
ueman: The user should be able to set Flutter specific configuration 😅

flutter/lib/src/default_integrations.dart
marandaneto: `this.runtimeType.toString()` is obfuscated at runtime if `split-debug` feature is enabled, so sadly not, it's not even possible to deobfuscated this yet, it's a limitation of the Dart tooling.

flutter/lib/src/sentry_flutter.dart
marandaneto: mm but this is supposed to work ootb with inheritance, right?

other option would be generics:

typedef FlutterOptionsConfiguration = FutureOr<void> Function<T extends SentryOptions>(
    T);

not sure about the syntax here, but something similar to it.

flutter/lib/src/sentry_flutter_options.dart
```diff
@@ -90,5 +91,157 @@ class SentryFlutterOptions extends SentryOptions {
         : _cacheDirSize;
   }
 
+  /// Consider disabling [SentryOptions.enableAutoNativeBreadcrumbs] if you
+  /// enable this. Otherwise you might record lifecycle events twice.
+  /// Also consider using [enableBreadcrumbTrackingForCurrentPlatform()]
+  /// instead for more sensible defaults.
+  bool get enableLifecycleBreadcrumbs => _enableLifecycleBreadcrumbs ?? false;
```
marandaneto: for all of those options, I'd prefer to set a value by default and not taking any null value, so we'd avoid some NPE checks when passing those values to the Native SDKs and better null-safety compatibility for the future.

flutter/test/sentry_flutter_test.dart
```diff
@@ -71,7 +71,7 @@ void main() {
       final event =
           verify(transport.send(captureAny)).captured.first as SentryEvent;
 
-      expect(event.sdk.integrations.length, 5);
+      expect(event.sdk.integrations.length, 6);
```
marandaneto, line 74: I should have fixed those using `whereType` instead, checking the `length` is suboptimal :(

flutter/lib/src/default_integrations.dart
```diff
@@ -147,6 +149,24 @@ class NativeSdkIntegration extends Integration<SentryFlutterOptions> {
...
+/// Integration that captures certain window and device events.
+/// See also:
+///   - [SentryWidgetsBindingObserver]
+///   - [WidgetsBindingObserver](https://api.flutter.dev/flutter/widgets/WidgetsBindingObserver-class.html)
+class WidgetsBindingIntegration extends Integration<SentryFlutterOptions> {
+  @override
+  FutureOr<void> call(Hub hub, SentryFlutterOptions options) {
+    // We don't need to call `WidgetsFlutterBinding.ensureInitialized()`
+    // because `FlutterSentry.init` already calls it.
+    WidgetsBinding.instance.addObserver(SentryWidgetsBindingObserver(
```
marandaneto: it'd be nice to override the `close` method and do:

`WidgetsBinding.instance.removeObserver`

flutter/lib/src/sentry_flutter.dart
ueman: Not quite. You can do
```dart
typedef OptionsConfiguration<T> = FutureOr<void> Function<T extends SentryOptions>(T options);
```
and use it in Sentry with 
```dart
Sentry.init(
    OptionsConfiguration<SentryOptions> options,
    // ...
}
```
and in SentryFlutter with 
```dart
FlutterSentry.init(
    OptionsConfiguration<SentryFlutterOptions> options,
    // ...
}
```

and then you have to cast `options` [here](https://github.com/getsentry/sentry-dart/blob/main/flutter/lib/src/sentry_flutter.dart#L47) to `SentryFlutterOptions`.

ueman:
I've changed the code based on your feedback. Thanks for the throughout review!

flutter/lib/src/sentry_flutter.dart
marandaneto: if we need to cast it's ok, but if the Client needs to `OptionsConfiguration<T>` is suboptimal, ideally, this is just a lambda, no boilerplate.

I still fail to understand why we need a `FlutterOptionsConfiguration` though, our sample uses the standard `OptionsConfiguration` and still works well with the `SentryFlutterOptions`, so not sure if I am missing something.

Other than that, I guess this PR is pretty good and ready to be landed, so would you mind elaborating on why do we need a specific `FlutterOptionsConfiguration`?

marandaneto: ok I guess I got it, our samples were not using any method only visible to the `SentryFlutterOptions`

flutter/lib/src/default_integrations.dart
```diff
@@ -147,6 +149,31 @@ class NativeSdkIntegration extends Integration<SentryFlutterOptions> {
   }
 }
 
+/// Integration that captures certain window and device events.
+/// See also:
+///   - [SentryWidgetsBindingObserver]
+///   - [WidgetsBindingObserver](https://api.flutter.dev/flutter/widgets/WidgetsBindingObserver-class.html)
+class WidgetsBindingIntegration extends Integration<SentryFlutterOptions> {
+  SentryWidgetsBindingObserver observer;
```
marandaneto: should it be private?

flutter/lib/src/widgets_binding_observer.dart
```diff
@@ -0,0 +1,195 @@
...
+    Hub hub,
+    @required SentryFlutterOptions options,
+  }) {
+    this.hub = hub ?? HubAdapter();
+    assert(options != null);
+    this.options = options;
+  }
+
+  Hub hub;
+  SentryFlutterOptions options;
```
marandaneto: let's make them private too

flutter/lib/src/widgets_binding_observer.dart
```diff
@@ -0,0 +1,195 @@
...
+  /// It tries to mimic the behavior of ActivityBreadcrumbsIntegration of Sentry
+  /// Android for lifecycle events.
+  ///
+  /// On Android and iOS this records lifecycle event breadcrumbs which loosely
+  /// correspond to their respectiv platforms lifecycle events.
+  ///
+  /// See also:
+  ///   - [WidgetsBindingObserver](https://api.flutter.dev/flutter/widgets/WidgetsBindingObserver-class.html)
+  @override
+  void didChangeAppLifecycleState(AppLifecycleState state) {
```
marandaneto, line 43: I'm trying to reproduce this breadcrumb but this method is never been called.

And yes I do call: `options.useFlutterBreadcrumbTracking()`

ideas?

flutter/lib/src/widgets_binding_observer.dart
ueman: Sure

flutter/lib/src/default_integrations.dart
ueman: Yep

flutter/lib/src/widgets_binding_observer.dart
ueman: With the main method of the included example app looking like this:
```dart
Future<void> main() async {
  await SentryFlutter.init(
    (options) {
      options.dsn = _exampleDsn;
      // use breadcrumb tracking of WidgetsBindingObserver
      options.useFlutterBreadcrumbTracking();
      // use breadcrumb tracking of platform Sentry SDKs
      // calling the following disabled the Flutter tracking
      // options.useNativeBreadcrumbTracking();
    },
    // Init your App.
  );
}
```

it gets called in at least the following cases:
- App is open, in foreground and the notification bar gets dragged down (iOS)
- App is brought to the background by pressing the home button (or the swipe action on devices without a button) (Android, iOS)
- App is in foreground and the device gets locked (Android)

If it runs on Flutter for Web I'm note sure if it gets called at all.

Maybe you need to do clean build or a hot restart instead of a hot reload? 
A simple hot reload would not add the integration and observer if the app is already running.

marandaneto: Screenshot 2020-12-04 at 11 32 43

probably a caching issue, yes, thanks!

marandaneto (approved):
@ueman awesome, thanks for this.

LGTM

marandaneto:
@ueman so I upstream the Flutter configurations fix as I wanted to release asap.

Just pull the latest changes and ready to be merged :)

PR #204, opened 2020-11-29
Enha: Enable pedantic to flutter packages and fix issues

marandaneto:

- [ ] Bugfix
- [ ] New feature
- [X] Enhancement
- [ ] Refactoring

## :scroll: Description
Enha: Enable pedantic to flutter packages and fix issues

We'd like to have proper code static analyses

## :pencil: Checklist

- [X] All tests passing
- [X] No breaking changes

### DIFF
--- a/flutter/analysis_options.yaml
+++ b/flutter/analysis_options.yaml
@@ -13,7 +13,6 @@ analyzer:
     deprecated_member_use_from_same_package: warning
   exclude:
     - example/**
-    - example_web/**
 
 linter:
   rules:
--- a/dart/lib/src/hub.dart
+++ b/dart/lib/src/hub.dart
@@ -131,7 +131,7 @@ class Hub {
         } catch (err) {
           _options.logger(
             SentryLevel.error,
-            'Error while capturing exception : ${throwable}',
+            'Error while capturing exception : $throwable',
           );
         } finally {
           _lastEventId = sentryId;
@@ -182,7 +182,7 @@ class Hub {
         } catch (err) {
           _options.logger(
             SentryLevel.error,
-            'Error while capturing message with id: ${message}',
+            'Error while capturing message with id: $message',
           );
         } finally {
           _lastEventId = sentryId;
--- a/dart/lib/src/protocol/sdk_version.dart
+++ b/dart/lib/src/protocol/sdk_version.dart
@@ -61,7 +61,7 @@ class SdkVersion {
   /// An immutable list of packages that compose this SDK.
   List<SentryPackage> get packages => List.unmodifiable(_packages);
 
-  String get identifier => '${name}/${version}';
+  String get identifier => '$name/$version';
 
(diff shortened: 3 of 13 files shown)


dart/analysis_options.yaml
```diff
@@ -11,7 +11,11 @@ analyzer:
...
+  exclude:
+    - example/**
+    - example_web/**
 
 linter:
   rules:
-    - prefer_relative_imports
\ No newline at end of file
+    prefer_relative_imports: true
+    unnecessary_brace_in_string_interps: true
```
marandaneto, line 21: @rxlabz if I got this right, the format should be `name: true or false`, so we preserve the values from pedantic.

https://dart.dev/guides/language/analysis-options#disabling-individual-rules

the way how it is, it basically ignores the pedantic file and only activates the `prefer_relative_imports` one which is not what we want right.

rxlabz: I'm not sure, I think you only need the map syntax if you want to disable some rules

rxlabz (approved):
LGTM

dart/analysis_options.yaml
marandaneto: > If you include an analysis options file such as the one in pedantic, you might want to disable some of the included rules. Disabling individual rules is similar to enabling them, but requires the use of a map rather than a list as the value for the rules

marandaneto: that's pretty clear for me tbh, but not a `pedantic` expert here so :(

rxlabz: > Disabling individual rules is similar to enabling them, but requires the use of a map

not that clear, but seems to me that only "disabling rules... requires the use of a map" :)

rxlabz: ( as changing the syntax don't show new warning or hint, I'm pretty sure that the pedantic rules were applied without the map syntax )

marandaneto: tbh I'm not sure, `prefer_relative_imports` was enabled already and it didn't fail our build before that, see the `sentry_navigation_observer.dart ` file

rxlabz: most of the rules only show hints and warnings, not errors. You need to add them to the analyzer:errors section if you want them to be considered as compilation errors.

flutter/analysis_options.yaml
```diff
@@ -0,0 +1,20 @@
+include: package:pedantic/analysis_options.yaml
+
+analyzer:
+  errors:
+    # treat missing required parameters as a warning (not a hint)
+    missing_required_param: error
+    # treat missing returns as a warning (not a hint)
+    missing_return: error
+    # allow having TODOs in the code
+    todo: ignore
```
bruno-garcia, line 10: TODO: Remove this TODO once we're done with all TODOs.

:trollface:

PR #208, opened 2020-11-30
Fix : the event must never serialize null or empty fields

rxlabz:

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

new checks to be sure to not serialize null or empty fields

unit tests

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/dart/lib/src/protocol/message.dart
+++ b/dart/lib/src/protocol/message.dart
@@ -20,10 +20,20 @@ class Message {
   Message(this.formatted, {this.template, this.params});
 
   Map<String, dynamic> toJson() {
-    return {
-      'formatted': formatted,
-      'message': template,
-      'params': params,
-    };
+    final json = <String, dynamic>{};
+
+    if (formatted != null && formatted.isNotEmpty) {
+      json['formatted'] = formatted;
+    }
+
+    if (template != null && template.isNotEmpty) {
+      json['message'] = template;
+    }
+
+    if (params != null && params.isNotEmpty) {
+      json['params'] = params;
+    }
+
+    return json;
   }
 }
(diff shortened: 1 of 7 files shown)


dart/lib/src/protocol/message.dart
```diff
@@ -20,10 +20,20 @@ class Message {
...
-    };
+    final json = <String, dynamic>{};
+
+    if (formatted != null && formatted.isNotEmpty) {
+      json['formatted'] = formatted;
+    }
+
+    if (template != null && template.isNotEmpty) {
+      json['message'] = template;
+    }
```
marandaneto: we don't need to `isNotEmpty` for `Strings` but only for collections

dart/lib/src/protocol/sentry_event.dart
```diff
@@ -316,18 +320,24 @@ class SentryEvent {
...
-          'version': sdkVersion,
-        };
+    Map<String, dynamic> sdkMap;
+    if (sdk != null && (sdkMap = sdk.toJson()).isNotEmpty) {
+      json['sdk'] = sdkMap;
+    } else {
+      json['sdk'] = <String, String>{
+        'name': sdkName,
+        'version': sdkVersion,
+      };
```
marandaneto: I'd remove this just for the sake of DRY, `_prepareEvent` already sets it.

dart/lib/src/protocol/message.dart
rxlabz:
```suggestion
    if (formatted != null) {
      json['formatted'] = formatted;
    }

    if (template != null) {
      json['message'] = template;
    }
```

dart/lib/src/protocol/sentry_event.dart

PR #209, opened 2020-11-30
Fix: Integrations should be closeable

rxlabz:

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

- add a integration disposer list to `Hub`
- dispose integrations on `Hub.close` 
- update the isolateErrorIntegration to allow closing it's port 

We noticed that the sample Dart process was never terminated due to the isolateErrorIntegration receptor.

unit tests

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes


dart/lib/src/isolate_error_integration.dart
```diff
@@ -5,15 +5,17 @@ import 'protocol.dart';
 import 'sentry_options.dart';
 import 'throwable_mechanism.dart';
 
-/// integration that capture errors on the current Isolate Error handler
-/// which is the main thread.
-void isolateErrorIntegration(Hub hub, SentryOptions options) {
-  final receivePort = _createPort(hub, options);
+Integration initIsolateErrorIntegration(
```
marandaneto: on Android we implement `Closeable` https://github.com/getsentry/sentry-java/blob/main/sentry-android-core/src/main/java/io/sentry/android/core/TempSensorBreadcrumbsIntegration.java#L23
which is the equivalent of this technique for `typedef` I'd say, sounds good.

marandaneto: we don't use `init` as a prefix for the other integrations, I'd rather stick to the normal naming

dart/lib/src/isolate_error_integration.dart
```diff
@@ -5,15 +5,17 @@ import 'protocol.dart';
...
+  void Function(Function) getPortDisposer,
```
marandaneto: what's about a more generic name like `integrationDisposable` so we could reuse for the next integrations, could be a typedef and an optional parameter to the `integration` typedef

dart/lib/src/sentry.dart
```diff
@@ -155,6 +162,8 @@ class Sentry {
     final hub = currentHub;
     _hub = NoOpHub();
     hub.close();
+
+    _disposeIsolate?.call();
```
marandaneto: this should be in the `hub.close()` method, see:

https://github.com/getsentry/sentry-java/blob/2e3679d18ec8be209fb5ad1432fc26c4adf2a268/sentry/src/main/java/io/sentry/Hub.java#L278-L282

ideally, as List, other integrations might have the same problem

rxlabz: ✅

dart/lib/src/isolate_error_integration.dart
rxlabz: ✅

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@
 - Fix: NDK integration was being disabled by a typo
 - Fix: missing toList for debug meta #192
 - Enhancement: NavigationObserver to record Breadcrumbs for navigation events #197
+- Fix: close the isolateErrorIntegration receivePort on Sentry.close()
```
marandaneto: I'd change to something more generic now, which is a system to dispose integrations

dart/lib/src/hub_adapter.dart
```diff
@@ -77,4 +77,7 @@ class HubAdapter implements Hub {
 
   @override
   SentryId get lastEventId => Sentry.lastEventId;
+
+  @override
+  void addIntegrationDisposer(disposer) {}
```
rxlabz: @marandaneto I'm not sure if this method must do something in the HubAdapter ?

dart/lib/src/sentry_options.dart
```diff
@@ -338,7 +338,15 @@ typedef EventProcessor = FutureOr<SentryEvent> Function(
...
-typedef Integration = FutureOr<void> Function(Hub hub, SentryOptions options);
+typedef Integration = FutureOr<void> Function(
+  Hub hub,
+  SentryOptions options, [
+  AddIntegrationDisposer addDisposer,
+]);
+
+typedef IntegrationDisposer = void Function();
+
+typedef AddIntegrationDisposer = void Function(IntegrationDisposer);
```
marandaneto: docs

dart/lib/src/hub.dart
```diff
@@ -312,6 +321,11 @@ class Hub {
       }
     }
   }
+
+  /// add an integration disposer
+  /// disposers are called on [Hub.close]
+  void addIntegrationDisposer(IntegrationDisposer disposer) =>
+      _integrationDisposers.add(disposer);
```
marandaneto: why does it need to be a public method on the Hub, can we get rid of it somehow?

marandaneto: if we can't do it fully automatic, I'd rather prefer to add as named/positional parameter to the  `options.addIntegration` method

dart/lib/src/hub_adapter.dart
marandaneto: I think this should not be part of the Hub at all.
See https://github.com/getsentry/sentry-dart/pull/209#discussion_r533322127

CHANGELOG.md
rxlabz: ✅ `add an integration closing system`?

dart/lib/src/hub_adapter.dart
rxlabz: ✅

bruno-garcia (approved):

PR #210, opened 2020-12-01
Ref: make hints optional

marandaneto:

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [X] Refactoring

## :scroll: Description
Ref: make hints optional

We should use the power of the language instead of forcing users to pass null

## :pencil: Checklist

- [X] All tests passing
- [ ] No breaking changes

### DIFF
--- a/flutter/lib/src/default_integrations.dart
+++ b/flutter/lib/src/default_integrations.dart
@@ -47,7 +47,7 @@ Integration loadContextsIntegration(
 ) {
   Future<void> integration(Hub hub, SentryOptions options) async {
     options.addEventProcessor(
-      (event, dynamic hint) async {
+      (event, {hint}) async {
         try {
           final infos = Map<String, dynamic>.from(
             await channel.invokeMethod('loadContexts'),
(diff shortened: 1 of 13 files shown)


flutter/lib/src/default_integrations.dart
```diff
@@ -47,7 +47,7 @@ Integration loadContextsIntegration(
 ) {
   Future<void> integration(Hub hub, SentryOptions options) async {
     options.addEventProcessor(
-      (event, dynamic hint) async {
+      (event, {hint}) async {
```
rxlabz, line 50:
```suggestion
      (event, {dynamic hint}) async {
```
we should avoid implicit dynamic

flutter/lib/src/default_integrations.dart
```diff
@@ -147,7 +147,7 @@ Integration loadAndroidImageListIntegration(
...
```
rxlabz, line 150: https://github.com/getsentry/sentry-dart/pull/210/files#r533360724

flutter/lib/src/default_integrations.dart
marandaneto: wondering if the `event` should be typed too then?

rxlabz (approved):
LGTM

PR #214, opened 2020-12-02
Ref: create sentry flutter options and use generics

marandaneto:

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :pencil: Checklist

- [ ] All tests passing
- [ ] No breaking changes

### DIFF
--- a/dart/lib/src/integration.dart
+++ b/dart/lib/src/integration.dart
@@ -5,9 +5,9 @@ import 'sentry_options.dart';
 
 /// Code that provides middlewares, bindings or hooks into certain frameworks or environments,
 /// along with code that inserts those bindings and activates them.
-abstract class Integration {
+abstract class Integration<T extends SentryOptions> {
   /// A Callable method for the Integration interface
-  FutureOr<void> call(Hub hub, SentryOptions options);
+  FutureOr<void> call(Hub hub, T options);
 
   /// NoOp by default : only closeable integrations need to override
   void close() {}
--- a/dart/lib/src/sentry.dart
+++ b/dart/lib/src/sentry.dart
@@ -12,8 +12,9 @@ import 'sentry_options.dart';
 import 'utils.dart';
 
 /// Configuration options callback
-typedef OptionsConfiguration = FutureOr<void> Function<T extends SentryOptions>(
-    T);
+typedef OptionsConfiguration = FutureOr<void> Function(SentryOptions);
+// typedef OptionsConfiguration = FutureOr<void> Function<T extends SentryOptions>(
+//     T);
 
 /// Runs a callback inside of the `runZonedGuarded` method, useful for running your `runApp(MyApp())`
 typedef AppRunner = FutureOr<void> Function();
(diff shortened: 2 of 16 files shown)


dart/lib/src/sentry.dart
```diff
@@ -70,19 +68,11 @@ class Sentry {
...
-      initialIntegrations
-          .forEach((integration) => options.addIntegration(integration));
-    }
-
     // Throws when running on the browser
     if (!isWeb) {
       // catch any errors that may occur within the entry function, main()
       // in the ‘root zone’ where all Dart programs start
-      options.addIntegration(IsolateErrorIntegration());
+      options.addIntegrationByIndex(0, IsolateErrorIntegration());
```
marandaneto, line 75: if not adding this method (addIntegrationByIndex), we'd need to refactor more things to set default values via ctor, I still wanna do it but on another PR, because it's not going to be breaking changes anyway

flutter/lib/src/default_integrations.dart
```diff
@@ -221,3 +224,47 @@ class LoadAndroidImageListIntegration extends Integration {
     options.sdk.addIntegration('loadAndroidImageListIntegration');
   }
 }
+
+/// a PackageInfo wrapper to make it testable
+typedef PackageLoader = Future<PackageInfo> Function();
+
+/// an Integration that loads the Release version from Native Apps
+/// or SENTRY_RELEASE and SENTRY_DIST variables
+class LoadReleaseIntegration extends Integration<SentryFlutterOptions> {
```
marandaneto, line 233: not so happy with this being an `Integration` but here is the thing, this should be the last thing to be executed because we want all the previous integrations to be run (so uncaught exception hooks from Flutter/Isolate, etc...) are in place.

flutter/test/sentry_flutter_util.dart
```diff
@@ -12,31 +11,54 @@ import 'mocks.dart';
...
+
       expect(sdkName, options.sdk.name);
       expect(sdkVersion, options.sdk.version);
       expect('pub:sentry_flutter', options.sdk.packages.last.name);
       expect(sdkVersion, options.sdk.packages.last.version);
 
-      expect('packageName@version+buildNumber', options.release);
-      expect('buildNumber', options.dist);
+      // expect('packageName@version+buildNumber', options.release);
+      // expect('buildNumber', options.dist);
```
marandaneto, line 63: I will fix this PR as soon as I get an agreement of https://github.com/getsentry/sentry-dart/pull/214/files#r534165418

dart/lib/src/integration.dart
```diff
@@ -5,9 +5,9 @@ import 'sentry_options.dart';
 
 /// Code that provides middlewares, bindings or hooks into certain frameworks or environments,
 /// along with code that inserts those bindings and activates them.
-abstract class Integration {
+abstract class Integration<T extends SentryOptions> {
   /// A Callable method for the Integration interface
-  FutureOr<void> call(Hub hub, SentryOptions options);
+  FutureOr<void> call(Hub hub, T options);
```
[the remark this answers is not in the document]
bruno-garcia: Where do we need to type check?
At this layer having `SentryOptions` would be enough, so I'm not sure how generics help here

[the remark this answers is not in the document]
bruno-garcia: OH nevermind this is integrations so I see.

Makes sense to me

bruno-garcia (approved):
very nice approach!

Issue #183, opened 2020-11-23
[sentry] Support for HTTP breadcrumbs


ueman:
I would like to request support for HTTP breadcrumbs for Sentry-Dart.

A possible solution could look like this:
```dart
import 'package:http/http.dart' as http;
import 'package:http/http.dart';
import 'package:sentry/sentry.dart';

class SentryHttpClient extends http.BaseClient {
  SentryHttpClient(this._inner);

  final http.Client _inner;

  @override
  Future<StreamedResponse> send(BaseRequest request) async {
    final response = await _inner.send(request);
    Sentry.addBreadcrumbs(
      Breadcrumb(
        message: null,
        type: 'http',
        data: {
          'url': request.url.toString(),
          'method': request.method,
          'status_code': response.statusCode.toString(),
          'reason': response.reasonPhrase,
        },
      ),
    );
    return response;
  }
}

```

marandaneto:
@ueman thanks for this.

what's more common in the Dart/Flutter world, the `http` package or https://api.dart.dev/dev/2.12.0-51.0.dev/dart-io/HttpClient-class.html ?

ueman:
[DIO](https://pub.dev/packages/dio) is another really popular library, just to make your list of HTTP clients complete. I think it uses the [http](https://pub.dev/packages/http) package under the hood.
But I'm not really sure what the most popular solution is.

However as Sentry already depends on the http package this would be a nice addition to the package.

ueman:
This is now possible due to #226

Issue #78, opened 2020-09-25
Performance Monitoring SDK for Flutter


poom201211:
There is very little support for Flutter web at the moment. However, just want to know if its possible for there to be a support for Flutter SDK. The performance monitoring feature is a amazing and I would like to implementing it in my Flutter web project.

bruno-garcia:
The SDK currently supports Flutter Web but given that it's still in the beta channel, it might take a bit longer than adding support for Android and iOS. It's definitely a goal for us but it's likely something for Q1 2021 only.

poom201211:
That's great! Thank you for your reply Bruno.

bruno-garcia:
Let's keep this open until we shit the support. Thanks for raising!

poom201211:
Sure thing, I'm looking forward to the implementation.

mnordine:
I'd like to have this for Dart web.

ueman:
A couple of ideas for performance monitoring:
- Something like [Android's activity tracing](https://docs.sentry.io/platforms/android/performance/included-instrumentation/#androids-activity-instrumentation) but with [SentryNavigatorObserver](https://github.com/getsentry/sentry-dart/blob/dcdbe20f441e21c3fcc626a08dd3f82b09fc16af/flutter/lib/src/navigation/sentry_navigator_observer.dart) and [WidgetsBindingObserver](https://github.com/getsentry/sentry-dart/blob/dcdbe20f441e21c3fcc626a08dd3f82b09fc16af/flutter/lib/src/widgets_binding_observer.dart)
- [MethodChannel](https://api.flutter.dev/flutter/services/MethodChannel-class.html) tracking. For example how long does it take to go from Dart to native to Dart. Or just from Dart to native or vice versa. The [onPlatformMessage()](https://api.flutter.dev/flutter/dart-ui/SingletonFlutterWindow/onPlatformMessage.html) method seems like a good candidate to test.
- A SentryGlobalKey (see [GlobalKey](https://api.flutter.dev/flutter/widgets/GlobalKey-class.html)) to track the lifetime of widgets and get its state in crash reports. I don't actually know if it's possible, but have this idea floating around in my mind.
- SentryHttpClient integration
- Frame duration tracking, see [FrameTiming](https://api.flutter.dev/flutter/dart-ui/FrameTiming-class.html) and [how it should be monitored](https://flutter.dev/docs/perf/metrics)

Lawati97:
[...]

PR #226, opened 2020-12-05
Feature: Sentry HTTP Client

ueman:

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
This PR introduces a [http package](https://pub.dev/documentation/http/latest/http/http-library.html) compatible Client to
track requests as breadcrumbs.

https://github.com/getsentry/sentry-dart/issues/183

I've written tests.

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/dart/lib/src/http_client/sentry_http_client.dart
+++ b/dart/lib/src/http_client/sentry_http_client.dart
@@ -1,0 +1,99 @@
+import 'package:http/http.dart';
+
+import '../../sentry.dart';
+
+/// A [http](https://pub.dev/packages/http)-package compatible HTTP client
+/// which records requests as breadcrumbs.
+///
+/// Remarks:
+/// If this client is used as a wrapper, a call to close does not close the
+/// given client.
+///
+/// The `SentryHttpClient` can be used as a standalone client like this:
+/// ```dart
+/// import 'package:sentry/sentry.dart';
+///
+/// var client = SentryHttpClient();
+/// try {
+///  var uriResponse = await client.post('https://example.com/whatsit/create',
+///      body: {'name': 'doodle', 'color': 'blue'});
+///  print(await client.get(uriResponse.bodyFields['uri']));
+/// } finally {
+///  client.close();
+/// }
+/// ```
+///
+/// The `SentryHttpClient` can also be used as a wrapper for your own
+/// HTTP [Client](https://pub.dev/documentation/http/latest/http/Client-class.html):
+/// ```dart
+/// import 'package:sentry/sentry.dart';
+/// import 'package:http/http.dart' as http;
+///
+/// final myClient = http.Client();
+///
+/// var client = SentryHttpClient(client: myClient);
+/// try {
+///  var uriResponse = await client.post('https://example.com/whatsit/create',
+///      body: {'name': 'doodle', 'color': 'blue'});
+///  print(await client.get(uriResponse.bodyFields['uri']));
+/// } finally {
+///  client.close();
+/// }
+/// ```
+//
+// Possible enhancement:
+// Track the time the HTTP request took.
+// For example with Darts Stopwatch:
+// https://api.dart.dev/stable/2.10.4/dart-core/Stopwatch-class.html
+class SentryHttpClient extends BaseClient {
+  SentryHttpClient({Client client, Hub hub}) {
+    _hub = hub ?? HubAdapter();
+    if (client != null) {
+      _client = client;
+      _userdefinedClient = true;
+    } else {
+      _client == Client();
+    }
+  }
+
+  Client _client;
+  Hub _hub;
+  bool _userdefinedClient = false;
+
+  @override
+  Future<StreamedResponse> send(BaseRequest request) async {
+    final response = await _client.send(request);
+    // See https://develop.sentry.dev/sdk/event-payloads/breadcrumbs/
+    // for how the breadcrumb should look like
+    _hub.addBreadcrumb(
+      Breadcrumb(
+        // TODO: should this always be `http`?
+        // Otherwise use http or https depending on the request.
+        type: request.url.scheme,
+        data: {
+          'url': request.url.toString(),
+          'method': request.method,
+          'status_code': response.statusCode,
+          // reason is optional, therefor only add it in case it is not null
+          if (response.reasonPhrase != null) 'reason': response.reasonPhrase,
+        },
+      ),
+    );
+    return response;
+  }
+
+  @override
+  void close() {
+    // In case the user did not specify a client himself
+    // we should close the client we initialized.
+    //
+    // If the user did specifiy a client himself,
+    // the user is in charge of closing his own client.
+    // Thus we don't close it here.
+    if (!_userdefinedClient) {
+      // TODO: how to test this?
+      // There is no property on a Client which indicates its status.
+      _client.close();
+    }
+  }
+}
--- a/dart/lib/src/http_client/sentry_http_client.dart
+++ b/dart/lib/src/http_client/sentry_http_client.dart
@@ -6,7 +6,7 @@ import '../../sentry.dart';
 /// which records requests as breadcrumbs.
 ///
 /// Remarks:
-/// If this client is used as a wrapper, a call to close does not close the
+/// If this client is used as a wrapper, a call to close also closes the
 /// given client.
 ///
 /// The `SentryHttpClient` can be used as a standalone client like this:
@@ -48,17 +48,11 @@ import '../../sentry.dart';
 class SentryHttpClient extends BaseClient {
   SentryHttpClient({Client client, Hub hub}) {
     _hub = hub ?? HubAdapter();
-    if (client != null) {
-      _client = client;
-      _userdefinedClient = true;
-    } else {
-      _client == Client();
-    }
+    _client = client ?? Client();
   }
 
   Client _client;
   Hub _hub;
-  bool _userdefinedClient = false;
 
   @override
   Future<StreamedResponse> send(BaseRequest request) async {
@@ -67,9 +61,8 @@ class SentryHttpClient extends BaseClient {
     // for how the breadcrumb should look like
     _hub.addBreadcrumb(
       Breadcrumb(
-        // TODO: should this always be `http`?
-        // Otherwise use http or https depending on the request.
-        type: request.url.scheme,
+        type: 'http',
+        category: 'http',
         data: {
           'url': request.url.toString(),
           'method': request.method,
@@ -84,16 +77,7 @@ class SentryHttpClient extends BaseClient {
 
   @override
   void close() {
-    // In case the user did not specify a client himself
-    // we should close the client we initialized.
-    //
-    // If the user did specifiy a client himself,
-    // the user is in charge of closing his own client.
-    // Thus we don't close it here.
-    if (!_userdefinedClient) {
-      // TODO: how to test this?
-      // There is no property on a Client which indicates its status.
-      _client.close();
-    }
+    // See https://github.com/getsentry/sentry-dart/pull/226#discussion_r536984785
+    _client.close();
   }
 }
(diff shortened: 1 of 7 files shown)


dart/lib/src/http_client/sentry_http_client.dart
```diff
@@ -0,0 +1,99 @@
...
+///  print(await client.get(uriResponse.bodyFields['uri']));
+/// } finally {
+///  client.close();
+/// }
+/// ```
+//
+// Possible enhancement:
+// Track the time the HTTP request took.
+// For example with Darts Stopwatch:
+// https://api.dart.dev/stable/2.10.4/dart-core/Stopwatch-class.html
```
marandaneto, line 47: this would be a good addition to #78

dart/lib/src/http_client/sentry_http_client.dart
```diff
@@ -0,0 +1,99 @@
...
+      _client = client;
+      _userdefinedClient = true;
+    } else {
+      _client == Client();
+    }
+  }
+
+  Client _client;
+  Hub _hub;
+  bool _userdefinedClient = false;
```
marandaneto:
```suggestion
  bool _userDefinedClient = false;
```

dart/lib/src/http_client/sentry_http_client.dart
```diff
@@ -0,0 +1,99 @@
...
+    // In case the user did not specify a client himself
+    // we should close the client we initialized.
+    //
+    // If the user did specifiy a client himself,
+    // the user is in charge of closing his own client.
+    // Thus we don't close it here.
+    if (!_userdefinedClient) {
+      // TODO: how to test this?
+      // There is no property on a Client which indicates its status.
+      _client.close();
```
marandaneto: true, I'd just validate if the method was or not called, using a mock maybe, or you create a fake that sets a flag so you know if its called or not.

dart/lib/src/http_client/sentry_http_client.dart
```diff
@@ -0,0 +1,99 @@
...
+  @override
+  Future<StreamedResponse> send(BaseRequest request) async {
+    final response = await _client.send(request);
+    // See https://develop.sentry.dev/sdk/event-payloads/breadcrumbs/
+    // for how the breadcrumb should look like
+    _hub.addBreadcrumb(
+      Breadcrumb(
+        // TODO: should this always be `http`?
+        // Otherwise use http or https depending on the request.
+        type: request.url.scheme,
```
marandaneto: `type` and `category` are hardcoded `http` only, so our UI renders it nicely.

https://github.com/getsentry/sentry-java/blob/main/sentry/src/main/java/io/sentry/Breadcrumb.java#L55-L56

dart/lib/src/http_client/sentry_http_client.dart
```diff
@@ -0,0 +1,99 @@
...
+        data: {
+          'url': request.url.toString(),
+          'method': request.method,
+          'status_code': response.statusCode,
+          // reason is optional, therefor only add it in case it is not null
+          if (response.reasonPhrase != null) 'reason': response.reasonPhrase,
```
marandaneto, line 70: wondering if this would be just noise data like `OK`, `Bad request`, etc... as you could extract the same info with `statusCode`

dart/test/http_client/sentry_http_client_test.dart
```diff
@@ -0,0 +1,236 @@
...
+        expect(e.uri, url);
+      }
+
+      verifyNever(mockHub.addBreadcrumb(captureAny));
+      verifyNever(mockHub.captureException(captureAny));
+    });
+
+    /// SocketException are only a thing on dart:io platforms.
+    /// otherwise this is equal to the test above
+    test('no breadcrumb for SocketException', () async {
```
marandaneto, line 172: is there a way to skip this test to the browser?

using `onPlatform: {'browser': Skip()}` maybe?

dart/lib/src/http_client/sentry_http_client.dart
ueman: Yep, but I only want to call it in case this client was initialized by the constructor so `SentryHttpClient` closes its own client but not the one given by the user. So mocking cannot be used because this code should not be called. `SentryHttpClient` should not interfere with the lifespan of the users http client.

dart/lib/src/http_client/sentry_http_client.dart
ueman: It is, but I've done it like the [breadcrumb](https://develop.sentry.dev/sdk/event-payloads/breadcrumbs/) documentation suggests.

dart/test/http_client/sentry_http_client_test.dart
ueman: Good point, but would't this fail anyway in browser tests because of the `dart:io` import? I'll have to look it up.

dart/lib/src/http_client/sentry_http_client.dart
```diff
@@ -0,0 +1,99 @@
...
+  }
+
+  @override
+  void close() {
+    // In case the user did not specify a client himself
+    // we should close the client we initialized.
+    //
+    // If the user did specifiy a client himself,
+    // the user is in charge of closing his own client.
+    // Thus we don't close it here.
```
marandaneto: tbh I'm wondering why we need this and not simply `_client.close()`

so if the user did not specify a client and call `close()`, obviously we need to close it.

if the user did specify a client, for which reason they would be calling `close` then and not only in their own Client?

Actually, When they are wrapping up their own Client in our Wrapper, they should be using the Wrapper only and their own Client won't be used anymore, so I see the common behavior to actually call `close()` in the wrapper and not to the original Client, unless I am missing something.

dart/test/http_client/sentry_http_client_test.dart
ueman: Hm, the tests all pass on the github web pipelines. So I guess there is no need to skip this test.

dart/lib/src/http_client/sentry_http_client.dart
```diff
@@ -0,0 +1,99 @@
...
+/// final myClient = http.Client();
+///
+/// var client = SentryHttpClient(client: myClient);
+/// try {
+///  var uriResponse = await client.post('https://example.com/whatsit/create',
+///      body: {'name': 'doodle', 'color': 'blue'});
+///  print(await client.get(uriResponse.bodyFields['uri']));
+/// } finally {
+///  client.close();
+/// }
```
marandaneto, line 41: that's a good insight, what's about adding it to the README?

and even better, what's about adding a button to the sample that does an HTTP request and records a breadcrumb automatically? could be as simple as a GET call to Flutter's website.

marandaneto:
left a few comments, but overall looks great, thanks again @ueman, you are killing it :)

dart/lib/src/http_client/sentry_http_client.dart
ueman: I'm fine with that. I'll point this out in the documentation then.

dart/lib/src/http_client/sentry_http_client.dart
marandaneto: oh totally missed it, let's keep it then :D

dart/lib/src/http_client/sentry_http_client.dart
ueman: Out of date because of https://github.com/getsentry/sentry-dart/pull/226#discussion_r536984785

marandaneto: but that's what I meant, using eg `verify` and `verifyNever`

> if the method was or not called

dart/lib/src/http_client/sentry_http_client.dart
marandaneto: I think the limitation would be if they have their own Client being shared across multiple wrappers, and they wanna release the allocation of the Sentry wrapper, that would be a problem, but tbh, I think this should still be fixable by their own code, a Wrapper ideally would mimic a real implementation which is `closing` its Client.

dart/lib/src/http_client/sentry_http_client.dart
ueman: You can only verify if you're using a mock. But in this case I can't use a mock, so no `verify` :(

But well, at least https://github.com/getsentry/sentry-dart/pull/226#discussion_r536984785 saves us from this headache.

marandaneto: yeah, lets spare the headache.

but fwiw, your test called `close does not get called for user defined client` does exactly what I've suggested and you do use a Mock in there.

marandaneto: but yeah, you'd like to test the usecase that the Client is created internally in the wrapper, true, well, solved by not doing it :)

dart/test/http_client/sentry_http_client_test.dart
marandaneto: true, but looks like the CI was hanging and got canceled, mind merging the `main` branch into this one to see if CI gets happy?

dart/lib/src/http_client/sentry_http_client.dart
ueman: Yeah, that was also my original reasoning for implementing it that way. However I haven't done that myself in any of my projects, so I guess it's more of a niche feature.

dart/test/http_client/sentry_http_client_test.dart
ueman: ✅

dart/lib/src/http_client/sentry_http_client.dart
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+  Client _client;
+  Hub _hub;
+
+  @override
+  Future<StreamedResponse> send(BaseRequest request) async {
+    final response = await _client.send(request);
+    // See https://develop.sentry.dev/sdk/event-payloads/breadcrumbs/
+    // for how the breadcrumb should look like
```
marandaneto:
```suggestion
```

marandaneto (approved):
LGTM

dart/test/http_client/sentry_http_client_test.dart
marandaneto: @ueman looks like the CI hangs after running the tests with `--coverage=coverage`, let me research a bit.

marandaneto: not really sure about the reason, maybe related to pausing isolates for JS code, but this should be a reasonable workaround
https://github.com/getsentry/sentry-dart/pull/227

marandaneto: I opened an issue on the coverage repo, so for now, just merge again the main repo into yours, sorry about that, thanks :)

ueman: No problem

marandaneto:
Screenshot 2020-12-06 at 11 18 10

thanks once more @ueman

iducool:
Hi Team,
I am evaluating this SDK for usage in one of my client's project. Tracking HTTP request is one of the mandatory check point for me. so, having following detail would be useful in making decision. 
Is this available to use? or going to be released?
- If available then it would be really helpful if there is any example to show usage. 
- If not released then kindly let us know when this would be available for usage. 

Thanks a lot for making developer's life easier.

ueman:
It's already available. An example can be found here https://github.com/getsentry/sentry-dart/blob/main/dart/README.md#breadcrumbs-for-http-requests

bruno-garcia:
We should put this on docs.sentry.io

ueman:
> We should put this on docs.sentry.io

It is already there:  https://docs.sentry.io/platforms/dart/usage/advanced-usage/

iducool:
Hi Team,

Thanks for quick help.

That implementation looks good however, I am using Dio Package for networking and there is no easy or one linear way of doing it. Dio is rich, popular package and very well adopted by Flutter community. 
Here I am posting my naive implantation for people who are searching for it,

https://gist.github.com/iducool/f80a466f423c34ec65c6230f54571500

Issue #218, opened 2020-12-03
Rename Logger to SentryLogger or similar


kuhnroyal:
Please consider renaming the `Logger` type to something that does not collide with all kinds of logging implementations.

marandaneto:
@kuhnroyal that's a fair point, thanks.

bruno-garcia:
PR's welcomed btw, if you feel like contributing, this could be a good way to get started with the code base

PR #228, opened 2020-12-06
Ref: rename logger to SentryLogger

kuhnroyal:

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring

## :scroll: Description

Renaming `Logger` to `SentryLogger` prevents name/import clashes with different logging packages.

See https://github.com/getsentry/sentry-dart/issues/218

## :pencil: Checklist

- [x] All tests passing
- [ ] No breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - Enhancement: Add SentryWidgetsBindingObserver, an Integration that captures certain window and device events.
 - Ref: Only assign non-null option values in Android native integration in order preserve default values
 - Enhancement: Add 'attachThreads' in options. When enabled, threads are attached to all logged events for Android
+- Ref: Rename typedef `Logger` to `SentryLogger` to prevent name clashes with logging packages
 
 ## 4.0.0-beta.1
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 - Enhancement: Add SentryWidgetsBindingObserver, an Integration that captures certain window and device events.
 - Ref: Only assign non-null option values in Android native integration in order preserve default values
 - Enhancement: Add 'attachThreads' in options. When enabled, threads are attached to all logged events for Android
+- Ref: Rename typedef `Logger` to `SentryLogger` to prevent name clashes with logging packages
```
```suggestion

### Breaking changes

- Ref: Rename typedef `Logger` to `SentryLogger` to prevent name clashes with logging packages
```

marandaneto (approved):
thanks for doing this @kuhnroyal :)
LGTM

kuhnroyal:
@marandaneto I would not consider this a breaking change as it is a typedef and not a class. But I see how this can be a matter of perspective :)

marandaneto:
> @marandaneto I would not consider this a breaking change as it is a typedef and not a class. But I see how this can be a matter of perspective :)

true, I think it's fair that we add a breaking change note, but luckily it does not require any changes on the code, usually, it's better to play safe so if for some reason this breaks something, they will read and fix it quickly.

PR #229, opened 2020-12-06
Fix: package usage as git dependency

kuhnroyal:
Using the development path in the dependency section prevents the package from being used as git dependency. A better approach is to add the custom path to dependency_overrides - this way a consumer can also override the sentry dependency.

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring

## :scroll: Description

By using `dependency_overrides` it is possible to consume the package directly from git(hub) to test latest changes on different branches:
```yaml
dependency_overrides:
  sentry:
    git:
      url: https://github.com/getsentry/sentry-dart.git
      ref: main
      path: dart
  sentry_flutter:
    git:
      url: https://github.com/getsentry/sentry-dart.git
      ref: main
      path: flutter
```
Without using `dependency_overrides` for the local development override this is not possible as the path `../dart` can not be resolved to a local dependency.

Not sure how this ties into build scripts/actions.

See description

I added the snippet to my projects.

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/dart/example_web/pubspec.yaml
+++ b/dart/example_web/pubspec.yaml
@@ -7,10 +7,13 @@ environment:
   sdk: ^2.8.0
 
 dependencies:
-  sentry:
-    path: ../
+  sentry: any
 
 dev_dependencies:
   build_runner: ^1.10.0
   build_web_compilers: ^2.13.0
   pedantic: ^1.9.0
+
+dependency_overrides:
+  sentry:
+    path: ../
--- a/flutter/pubspec.yaml
+++ b/flutter/pubspec.yaml
@@ -13,10 +13,7 @@ dependencies:
     sdk: flutter
   flutter_web_plugins:
     sdk: flutter
-  sentry: #">=4.0.0-beta.2 <4.1.0"
-    # TODO: we need the specific version for publishing it, but the relative path for development.
-    # Figure out how to do it via Craft for releases.
-    path: ../dart
+  sentry: ">=4.0.0-beta.2 <4.1.0"
   package_info: ^0.4.0
 
 dev_dependencies:
@@ -26,6 +23,10 @@ dev_dependencies:
   test: ^1.15.7
   yaml: ^2.2.1 # needed for version match (code and pubspec)
   pedantic: ^1.9.2
+
+dependency_overrides:
+  sentry:
+    path: ../dart
 
 flutter:
   plugin:
--- a/flutter/example/pubspec.yaml
+++ b/flutter/example/pubspec.yaml
@@ -10,11 +10,14 @@ environment:
 dependencies:
   flutter:
     sdk: flutter
+  sentry_flutter: any
+  universal_platform: ^0.1.3
 
+dependency_overrides:
+  sentry:
+    path: ../../dart
   sentry_flutter:
     path: ../
-
-  universal_platform: ^0.1.3
 
 flutter:
   uses-material-design: true


flutter/pubspec.yaml
```diff
@@ -27,6 +24,12 @@ dev_dependencies:
   yaml: ^2.2.1 # needed for version match (code and pubspec)
   pedantic: ^1.9.2
 
+dependency_overrides:
+  sentry:
+    # TODO: we need the specific version for publishing it, but the relative path for development.
+    # Figure out how to do it via Craft for releases.
+    path: ../dart
```
marandaneto:
```suggestion
dependency_overrides:
  sentry:
    path: ../dart
```

marandaneto: looks like this solves the TODO

marandaneto: but CI fails: `Because every version of sentry_flutter from path depends on sentry >=4.0.0-beta.2 <4.1.0 which doesn't match any versions, sentry_flutter from path is forbidden.`

flutter/pubspec.yaml
```diff
@@ -13,10 +13,7 @@ dependencies:
     sdk: flutter
   flutter_web_plugins:
     sdk: flutter
-  sentry: #">=4.0.0-beta.2 <4.1.0"
-    # TODO: we need the specific version for publishing it, but the relative path for development.
-    # Figure out how to do it via Craft for releases.
-    path: ../dart
+  sentry: ">=4.0.0-beta.2 <4.1.0"
```
marandaneto, line 16:
```suggestion
  sentry: ">=4.0.0-beta.1 <4.1.0"
```

flutter/pubspec.yaml
marandaneto: ok to solve this we need to apply this https://github.com/getsentry/sentry-dart/pull/229/files#r537115550

https://dart.dev/tools/pub/dependencies#dependency-overrides
there's a warning note about it.

> Using a dependency override involves some risk. For example, using an override to specify a version outside the range that the package claims to support

I think it's fine, we are bumping the version manually for now, and later we'll automatize this.

kuhnroyal: This is actually just the example failing, I think I can fix this.

flutter/pubspec.yaml
kuhnroyal: This is not needed, it is overridden, the error is not coming from here.

kuhnroyal:
The package builds now. This is a warning/error from pana now.

flutter/pubspec.yaml
marandaneto: oh I see, good catch, thanks

flutter/pubspec.yaml
marandaneto: even better, thanks.

kuhnroyal: Even with this change pana fails locally, doesn't seem to respect pre-release versions.
```
The constraint `>=4.0.0-beta.1 <4.1.0` on sentry does not support the latest published version `3.0.1`
```

marandaneto: is it possible to suppress this error somehow? not a pana expert, otherwise I'll search for it tomorrow.

marandaneto:
the `ktlint` issue has been fixed on `main` btw.

flutter/pubspec.yaml
kuhnroyal: Just looked a bit but there doesn't seem to be a way that would just suppress this and not everything else.

marandaneto: damn, looks like a valid use-case, maybe it's worth opening a new issue on pana's repo.
https://github.com/dart-lang/pana

kuhnroyal: There is this https://github.com/dart-lang/pana/issues/242

kuhnroyal: The overrides are being stripped before analysis so I don't see a way to work around this.

marandaneto: that makes sense I'd say, but the suppress here could just be able to use nonstable versions as alpha, beta, RCs, right.

> The constraint `>=4.0.0-beta.1 <4.1.0` on sentry does not support the latest published version `3.0.1`

if we are able to suppress only this, it'd be fine I guess

kuhnroyal: This is done via the https://github.com/axel-op/dart-package-analyzer action. Could probably exclude the up-to-date dependency score here: https://github.com/getsentry/sentry-dart/blob/main/.github/workflows/flutter.yml#L112

marandaneto: yeah, it's probably running `pub outdated` with `--no-prereleases`.

but it seems that `pub upgrade` fails in the GH action, so I don't see how excluding the up-to-date dependencies would make the GH Action happy, looks like they'd need to offer a way to toggle this on/off.

marandaneto: nvm, I could reproduce it locally, subtracting 10 points would work, let's just do this and add a comment, so I suggest:

```
# 90 due to 'pana' dependencies score when using 'dependency_overrides'
if (( $PERCENTAGE < 90 ))
```

could you do this change and see if CI gets happy? thanks :)

dart/example_web/pubspec.yaml
```diff
@@ -8,9 +8,12 @@ environment:
 
 dependencies:
   sentry:
```
marandaneto: should we do `any` here too or removing from `any` from `sentry_flutter`? just for the sake of equal configuration.

kuhnroyal: Yea can do that to make it more clear

flutter/pubspec.yaml
kuhnroyal: Sure, that was my idea as well, don't see any other way atm.

kuhnroyal:
@marandaneto I would also suggest to have `sentry_flutter` always depend on the specific version of `sentry` and not on a range. The `sentry_flutter` package clearly depends on API and breaking changes of `sentry` of a specific version.

flutter/pubspec.yaml
kuhnroyal: .... It runs `pub outdated --no-dev-dependencies --up-to-date --no-dependency-overrides` so this just completely errors.

marandaneto:
> @marandaneto I would also suggest to have `sentry_flutter` always depend on the specific version of `sentry` and not on a range. The `sentry_flutter` package clearly depends on API and breaking changes of `sentry` of a specific version.

I tried, but it's not allowed, `pub publish` complains, you can't depend on a specific version.

flutter/pubspec.yaml
kuhnroyal: I don't think there currently is a way of running pana without setting everything back to `3.0.1` and only updating the version prior to release.

marandaneto: cool, I think we still need to use `beta.1` here or it fails as it does not resolve the dependency, if we do `beta.1`, it fails because of API incompatibility but this is fine as I'm releasing a beta.2 today :)

kuhnroyal:
> I tried, but it's not allowed, pub publish complains, you can't depend on a specific version.

You could `--force` it but that could also impact the pub.dev score. Not sure but having a range could create conflicts for someone not running `pub upgrade`.

flutter/pubspec.yaml
kuhnroyal: > cool, I think we still need to use beta.1 here or it fails as it does not resolve the dependency, if we do beta.1, it fails because of API incompatibility but this is fine as I'm releasing a beta.2 today :)

It will always fail this way due to the command using `--no-dev-dependencies` and `--no-dependency-overrides`.
I am not sure what to do here. This pana error is not valid as soon as you release it so it should somehow be ignored.

marandaneto: can we try using `beta.1`? if it fails because of API incompatibility, I don't see why this is a problem.

kuhnroyal: It doesn't find `beta.1` due to `--no-dev-dependencies`.
```
The constraint `>=4.0.0-beta.1 <4.1.0` on sentry does not support the latest published version `3.0.1`
```

marandaneto:
> > I tried, but it's not allowed, pub publish complains, you can't depend on a specific version.
> 
> You could `--force` it but that could also impact the pub.dev score. Not sure but having a range could create conflicts for someone not running `pub upgrade`.

I see, yeah I'd not like to do it, but this is only problematic because it's a prerelease, we'll bump the dependency accordingly as soon as the API gets stable, no breaking changes across dependencies.

flutter/pubspec.yaml
marandaneto: mm that's weird, today morning I opened https://github.com/getsentry/sentry-dart/pull/233/files and tried using beta.1 and it worked out, it failed because of breaking changes, but not other reasons.

marandaneto: tbh we are running other CI checks with `format`, `analyze` and `tests`, this is just nice to have, I consider the outcome of this PR more important than the score check.

marandaneto: last case, it'd be fine to deactivate the GH action that runs `pana` for now and enable it back once both packages are stable.

kuhnroyal: pana doesn't seem to take a mono repo with 2 or more dependent packages into account, this will pretty much always fail for dev versions until this is changed.

marandaneto: if it depends on a stable package and API is compatible, it's alright I guess, at least I've seen other packages using this GH Action too.
So here it is #235
Once this is merged, it'd be fine to just fix the conflicts

kuhnroyal:
Rebasing

marandaneto:
thanks @kuhnroyal for the contribution and also for figuring out a solution around this :)

PR #239, opened 2020-12-10
Release: Update READMEs for v4 GA

marandaneto:


CHANGELOG.md
```diff
@@ -1,7 +1,45 @@
...
+
+Improvements:
 
 * Feat: Added a copyWith method to all the protocol classes
 
+Packages were released on [sentry pubdev](https://pub.dev/packages/sentry) and [sentry_flutter pubdev](https://pub.dev/packages/sentry_flutter)
+
+## Sentry Self Hosted Compatibility
+
+* Since version `4.0.0` of the `sentry_flutter`, `Sentry` version >= `v20.6.0` is required. This only applies to on-premise Sentry, if you are using sentry.io no action is needed.
```
bruno-garcia:
```suggestion
* Since version `4.0.0` of the `sentry_flutter`, [`Sentry` server version >= `v20.6.0`](https://github.com/getsentry/onpremise/releases) is required. This only applies to on-premise Sentry, if you are using sentry.io no action is needed.
```

flutter/pubspec.yaml
```diff
@@ -13,7 +13,7 @@ dependencies:
     sdk: flutter
   flutter_web_plugins:
     sdk: flutter
-  sentry: ">=4.0.0-beta.3 <4.1.0"
+  sentry: ">=4.0.0 <4.1.0"
```
bruno-garcia: We really want to limit to 4.1.0?
Should this be open, or up to 5.0?

bruno-garcia (approved):
The changelog is beautiful to read.
I added one note and a suggestion

flutter/pubspec.yaml
marandaneto: @bruno-garcia I'm afraid of compatibility as we can't limit to a specific version, at least tighten like this, we know that it's not gonna be broken, for example, we could do `minor` semver changes and it's not gonna be broken unless they upgrade the `sentry_flutter` dependency first, wdyt?

bruno-garcia: yeah honestly it won't matter because we version them together so they'll always get the exact version

marandaneto: ok gonna do < 5.0.0

Issue #240, opened 2020-12-11
[sentry_flutter] Lifecycle breadcrumb is kinda wrong


ueman:
[This breadcrumb](https://github.com/getsentry/sentry-dart/blob/main/flutter/lib/src/widgets_binding_observer.dart#L43-L57) should be analog to [this Android breadcrumb](https://github.com/getsentry/sentry-java/blob/main/sentry-android-core/src/main/java/io/sentry/android/core/LifecycleWatcher.java#L119-L128).
Currently it is more like an Activity-Lifecycle event despite being an application wide lifecycle event.
I've stumbled across this, while researching [this](https://github.com/getsentry/develop/issues/213#issuecomment-743026394).

marandaneto:
@ueman good catch, it's more about App's lifecycle (background/foreground), maybe it makes sense to just toggle the Android option `enableAppLifecycleBreadcrumbs` if using the WidgetsBinding then.
Would you like to contribute? :)

We probably need to fix here the naming of the options (Which is checked by our WidgetsBinding wrapper), enabling/disabling the correct fields on the Native SDKs and also docs.

ueman:
Yeah, but it could take some time until I come around to it.

marandaneto:
on Android, we could disable only this one instead https://github.com/getsentry/sentry-java/blob/main/sentry-android-core/src/main/java/io/sentry/android/core/SentryAndroidOptions.java#L27-L28

If using the WidgetsBinding wrapper.

PR #241, opened 2020-12-12
Refactoring app lifecycle breadcrumbs

ueman:

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring

## :scroll: Description
I've changed the breadcrumbs according to the notes in https://github.com/getsentry/sentry-dart/issues/240 .
I've also improved the documentation a little bit.

https://github.com/getsentry/sentry-dart/issues/240

I've changed existing tests accordingly.

## :pencil: Checklist

- [x] All tests passing
- [ ] No breaking changes

## :crystal_ball: Next steps
https://github.com/getsentry/sentry-dart/issues/240 suggests to disable/enable the native integration accordingly and to refactor the options of `FlutterSentryOptions`. I would suggest to first agree on a naming scheme for the options.

Maybe something like `enableFlutterX` (or just `enableX`) and `enableNativeX` so that it is more clear what and where something gets enabled. It would also help to know which options must be enable/disabled accordingly in the native part.

### DIFF
--- a/flutter/lib/src/sentry_flutter_options.dart
+++ b/flutter/lib/src/sentry_flutter_options.dart
@@ -1,7 +1,9 @@
 import 'package:flutter/foundation.dart' as foundation;
 import 'package:sentry/sentry.dart';
 
-/// This class add options which are only availble in a Flutter environment.
+/// This class adds options which are only availble in a Flutter environment.
+/// Note that some of these options require native Sentry integration, which is
+/// not available on all platforms.
 class SentryFlutterOptions extends SentryOptions {
   SentryFlutterOptions() : super();
 
@@ -17,7 +19,7 @@ class SentryFlutterOptions extends SentryOptions {
 
   bool _enableNativeCrashHandling = true;
 
-  /// Enable or Disable the Crash handling on the Native SDKs (Android/iOS)
+  /// Enable or disable the Crash handling on the Native SDKs (Android/iOS)
   bool get enableNativeCrashHandling => _enableNativeCrashHandling;
 
   set enableNativeCrashHandling(bool nativeCrashHandling) {
@@ -27,9 +29,9 @@ class SentryFlutterOptions extends SentryOptions {
 
   int _autoSessionTrackingIntervalMillis = 30000;
 
-  /// The session tracking interval in millis. This is the interval to end a session if the App goes
-  /// to the background.
-  /// See: enableAutoSessionTracking
+  /// The session tracking interval in millis. This is the interval to end a
+  /// session if the App goes to the background.
+  /// See: [enableAutoSessionTracking]
   int get autoSessionTrackingIntervalMillis =>
       _autoSessionTrackingIntervalMillis;
 
@@ -58,7 +60,7 @@ class SentryFlutterOptions extends SentryOptions {
 
   /// ANR Timeout internal in Millis Default is 5000 = 5s Used by AnrIntegration.
   /// Available only for Android.
-  /// See: anrEnabled
+  /// See: [anrEnabled]
   int get anrTimeoutIntervalMillis => _anrTimeoutIntervalMillis;
 
   set anrTimeoutIntervalMillis(int anrTimeoutIntervalMillis) {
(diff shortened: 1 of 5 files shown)


flutter/lib/src/sentry_flutter_options.dart
```diff
@@ -91,18 +96,37 @@ class SentryFlutterOptions extends SentryOptions {
         : _cacheDirSize;
   }
 
+  @Deprecated('Use enableAppLifecycleBreadcrumbs instead')
+  bool get enableLifecycleBreadcrumbs => _enableAppLifecycleBreadcrumbs;
+
+  @Deprecated('Use enableAppLifecycleBreadcrumbs instead')
```
marandaneto: I liked that, so it's just forwarding the options, but not breaking it.

flutter/lib/src/widgets_binding_observer.dart
```diff
@@ -41,14 +41,14 @@ class SentryWidgetsBindingObserver with WidgetsBindingObserver {
...
   void didChangeAppLifecycleState(AppLifecycleState state) {
-    if (!_options.enableLifecycleBreadcrumbs) {
+    if (!_options.enableAppLifecycleBreadcrumbs) {
       return;
     }
-    // According to
+    // References:
     // https://develop.sentry.dev/sdk/event-payloads/breadcrumbs/
-    // this is of the type navigation.
+    // https://github.com/getsentry/sentry-java/blob/main/sentry-android-core/src/main/java/io/sentry/android/core/LifecycleWatcher.java#L119-L128
```
marandaneto: a tip when using links to other repos is using a permalink (that includes a SHA) so if we ref. the code and the lines change, it'd still work.

marandaneto:
```suggestion
    // https://github.com/getsentry/sentry-java/blob/de00462e3499fa9a21a7992317503f1ccda7d226/sentry-android-core/src/main/java/io/sentry/android/core/LifecycleWatcher.java#L119-L128
```

marandaneto:
about the `Next steps`, makes sense, `enableX` and `enableNativeX` is a good idea, I'll discuss this internally beforehand as it seems a good idea also across SDKs, like React-Native, soon Capacitor, etc, thanks :)

marandaneto (approved):
thanks @ueman 
LGTM

flutter/lib/src/sentry_flutter_options.dart
ueman: yep, would be bad to have a breaking change just after a major release.
It would be nice if it includes when it gets removed. So that users get nudged into using the superseding method.
`Use enableAppLifecycleBreadcrumbs instead. This option gets removed in Sentry 5.0.0` or something like that. What do you think?

marandaneto: if they are using `pedantic` or something similar, they'd be warned by the linter (because of `Deprecated`), right? I guess this would be already good enough.
but not sure if I fully understand your suggestion, what do you mean by:

> It would be nice if it includes when it gets removed

ueman: I think it would be nice to let the user know when this options get removed. Just like the example above:
`Use enableAppLifecycleBreadcrumbs instead. This option gets removed in Sentry 5.0.0`

marandaneto: oh I see, so just a better message you mean, sounds good to me.
On Java, we have a specific annotation for this: https://github.com/JetBrains/java-annotations/blob/master/common/src/main/java/org/jetbrains/annotations/ApiStatus.java#L75-L92

if we don't have something similar on the `meta` pkg, a message is good enough :)

ueman: Neither [dart:core](https://api.dart.dev/stable/2.10.4/dart-core/Deprecated-class.html) nor [meta](https://pub.dev/documentation/meta/latest/meta/meta-library.html) has something like that.

Is `5.0.0` a good target?

marandaneto: yep, sounds good, thanks for that!

flutter/lib/src/sentry_flutter_options.dart
```diff
@@ -91,18 +96,37 @@ class SentryFlutterOptions extends SentryOptions {
         : _cacheDirSize;
   }
 
+  @Deprecated('Use enableAppLifecycleBreadcrumbs instead')
```
marandaneto:
```suggestion
  @Deprecated('Use enableAppLifecycleBreadcrumbs instead. This option gets removed in Sentry 5.0.0')
```

flutter/lib/src/sentry_flutter_options.dart
marandaneto:
```suggestion
  @Deprecated('Use enableAppLifecycleBreadcrumbs instead. This option gets removed in Sentry 5.0.0')
```

marandaneto:
@ueman `Formatting failed: 1`, would you mind to fix this one? so I can merge & release :) thanks!

Issue #242, opened 2020-12-15
Error while sending more throughout FlutterErrorDetails


ueman:
1. Please describe the issue and how to reproduce it, or link to a small reproducible application:
I'm trying to get a better and more details error report but I get the following exception
```
E/FileObserver(13655): Unhandled exception in FileObserver io.sentry.android.core.EnvelopeFileObserver@94251ab
E/FileObserver(13655): java.lang.IllegalArgumentException: Invalid envelope. Item at index '1'. has no header delimiter.
E/FileObserver(13655): 	at io.sentry.EnvelopeReader.read(EnvelopeReader.java:80)
E/FileObserver(13655): 	at io.sentry.OutboxSender.processFile(OutboxSender.java:60)
E/FileObserver(13655): 	at io.sentry.OutboxSender.processEnvelopeFile(OutboxSender.java:100)
E/FileObserver(13655): 	at io.sentry.android.core.EnvelopeFileObserver.onEvent(EnvelopeFileObserver.java:58)
E/FileObserver(13655): 	at android.os.FileObserver$ObserverThread.onEvent(FileObserver.java:163)
E/FileObserver(13655): 	at android.os.FileObserver$ObserverThread.observe(Native Method)
E/FileObserver(13655): 	at android.os.FileObserver$ObserverThread.run(FileObserver.java:113)

```
To reproduce the issue checkout this [branch](https://github.com/ueman/sentry-dart/tree/feater/error-report)  and press the `Open Scaffold with UI error` button.

It only happens on Android, iOS works just fine.

2. Include the output of the command `flutter doctor` below:
```
Doctor summary (to see all details, run flutter doctor -v):
[✓] Flutter (Channel stable, 1.22.5, on macOS 11.0.1 20B29 darwin-x64, locale de-DE)
 
[✓] Android toolchain - develop for Android devices (Android SDK version 30.0.2)
[✓] Xcode - develop for iOS and macOS (Xcode 12.2)
[!] Android Studio (version 4.1)
    ✗ Flutter plugin not installed; this adds Flutter specific functionality.
    ✗ Dart plugin not installed; this adds Dart specific functionality.
[✓] VS Code (version 1.51.1)
[✓] Connected device (1 available)

! Doctor found issues in 1 category.
```

# Motivation

This change in the linked branch improve the error message from 
```
FlutterError
A RenderFlex overflowed by 1175 pixels on the right.
```
to
```

════════ Exception caught by rendering library ═════════════════════════════════
[...]

PR #245, opened 2020-12-16
Fix: Envelope length should be based on the UTF8 array instead of String length

marandaneto:

- [X] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
Fix: Envelope length should be based on the UTF8 array instead of String length

#242

## :pencil: Checklist

- [X] All tests passing
- [X] No breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # vNext
 
 * Ref: Changed category of Flutter lifecycle tracking [#240](https://github.com/getsentry/sentry-dart/issues/240)
+* Fix: Envelope length should be based on the UTF8 array instead of String length
 
 # 4.0.0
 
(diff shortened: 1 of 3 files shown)


ueman (approved):
LGTM.

ueman:
But dang, that is nasty. If I understand correctly this prevents every event which contains unicode chars from being send, right?

marandaneto:
> But dang, that is nasty. If I understand correctly this prevents every event which contains unicode chars from being send, right?

not really, a subset of it and usually they are not part of "normal" events, but it's rather critical of course, I'm cutting a release pretty soon.

marandaneto:
`pana` is failing but nothing has changed (works locally though), created https://github.com/axel-op/dart-package-analyzer/issues/12 and will merge it

Issue #243, opened 2020-12-15
FlutterErrorIntegration does not respect FlutterErrorDetails.silent flag


ueman:
1. Please describe the issue and how to reproduce it, or link to a small reproducible application:

The FlutterErrorIntegration does not respect the [silent](https://api.flutter.dev/flutter/foundation/FlutterErrorDetails/silent.html) flag of `FlutterErrorDetails`. There is probably a good reason for that, so maybe Sentry should respect that?

marandaneto:
@ueman good point, there's no reason, the flag was unknown to me.

so just to confirm, what we could do is:

```
if (details.silent != true) {
  // sentry captures the event
} else {
  // bail out
}
```

is this the suggested behavior?

ueman:
The docs state 
```
This is used by certain exception handlers that catch errors that could be triggered by environmental conditions (as opposed to logic errors). For example, the HTTP library sets this flag so as to not report every 404 error to the console on end-user devices, while still allowing a custom error handler to see the errors even in release builds.
```
So I would probably do it like you suggested but also provide the option to report silent errors.

marandaneto:
that's a good idea, although Sentry offers the `beforeSend` callback to filter out events that you don't want to send over the wire, ideally, you'd not even capture them, so a flag would be ideal.

 I'd say we could go for respecting the `silent != true` OR `options.reportSilentFlutterErrors == true`, captures the event.

`options.reportSilentFlutterErrors` would be `false` by default btw, the naming is just an example, maybe we could find something better, wdyt?

ueman:
The callback would also not know about that silent flag :( Otherwise the option sounds good to me.

marandaneto:
@ueman yep, I guess not raising an event if the `silent` flag is `true` makes sense, but as you said, some people might still want it, adding an option (reportSilentFlutterErrors) would solve both use-cases, thank you for raising this.

marandaneto:
Also, consider the improvements discussed here #242

Sending the content of the `FlutterErrorDetails` as a `message` rather than only the `stack` makes sense for some type of errors.

marandaneto:
@ueman very true about some `FlutterError` errors.

sometimes there are no stack traces and `FlutterErrorDetails.toString()` has the offending file.
sometimes there's a stack trace, but the offending file is not in there, but it appears in the `FlutterErrorDetails.toString()`
I don't see a way around it other than always logging the `FlutterErrorDetails.toString()` as part of the exception.

[...]

PR #251, opened 2020-12-19
Refactoring: Init Flutter first + documentation

ueman:

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring

## :scroll: Description
`WidgetsFlutterBinding.ensureInitialized();` should be the first thing, the SDK calls. This prevents all kinds of mistakes which could result from default integrations assuming it was already caught. (For example PlatoformChannels and so on).
Previously it was called after the default integrations were added.

I've also improved the documentation of the `FlutterErrorIntegration` regarding UI and layout errors.
This should also set the users expections regarding catching [these](https://flutter.dev/docs/testing/common-errors) errors
 in release mode.

See 
- https://github.com/getsentry/sentry-dart/issues/243
- https://github.com/flutter/flutter/issues/72598

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/flutter/lib/src/sentry_flutter.dart
+++ b/flutter/lib/src/sentry_flutter.dart
@@ -34,6 +34,11 @@ mixin SentryFlutter {
     if (optionsConfiguration == null) {
       throw ArgumentError('OptionsConfiguration is required.');
     }
+
+    // It is necessary to initialize Flutter method channels so that
+    // our plugin can call into the native code.
+    WidgetsFlutterBinding.ensureInitialized();
+
     final flutterOptions = SentryFlutterOptions();
     // first step is to install the native integration and set default values,
     // so we are able to capture future errors.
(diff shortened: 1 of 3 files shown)


flutter/lib/src/sentry_flutter.dart
```diff
@@ -34,6 +34,11 @@ mixin SentryFlutter {
     if (optionsConfiguration == null) {
       throw ArgumentError('OptionsConfiguration is required.');
     }
+
+    // It is necessary to initialize Flutter method channels so that
+    // our plugin can call into the native code.
+    WidgetsFlutterBinding.ensureInitialized();
```
marandaneto, line 40: I'm not against the change, but I also don't see why, there's a difference between adding the default integrations to the list and executing them.

https://github.com/getsentry/sentry-dart/blob/main/dart/lib/src/sentry.dart#L111-L114

the integrations only get the chance to run after the SDK is fully initialized, which is the last step, at this point, the SDK is enabled and `WidgetsFlutterBinding.ensureInitialized` has been called for quite a bit.

ueman: Ah I see. I was not aware of that.  Still, I would argue for `WidgetsFlutterBinding.ensureInitialized()` because it seems to me this is a better practice. However I'm fine with either solution.

marandaneto: let's merge as it is, thanks :)

PR #266, opened 2021-01-07
Fix: NPE check before initing the SentryWidgetsBindingObserver

marandaneto:
## :scroll: Description
Fix: NPE check before initing the SentryWidgetsBindingObserver

If `WidgetsFlutterBinding.ensureInitialized()` fails somehow, this would prevent the SDK to keep initing.

## :pencil: Checklist

- [X] All tests passing
- [X] No breaking changes

### DIFF
--- a/flutter/lib/src/default_integrations.dart
+++ b/flutter/lib/src/default_integrations.dart
@@ -199,11 +199,21 @@ class WidgetsBindingIntegration extends Integration<SentryFlutterOptions> {
     if (instance != null) {
       instance.addObserver(_observer);
       options.sdk.addIntegration('widgetsBindingIntegration');
+    } else {
+      options.logger(
+        SentryLevel.error,
+        'widgetsBindingIntegration failed to be installed',
+      );
     }
   }
 
   @override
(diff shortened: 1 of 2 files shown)


flutter/lib/src/default_integrations.dart
```diff
@@ -194,9 +194,12 @@ class WidgetsBindingIntegration extends Integration<SentryFlutterOptions> {
 
     // We don't need to call `WidgetsFlutterBinding.ensureInitialized()`
     // because `FlutterSentry.init` already calls it.
-    WidgetsBinding.instance.addObserver(_observer);
-
-    options.sdk.addIntegration('widgetsBindingIntegration');
+    // If the instance is not created, we skip it to keep going.
+    final instance = WidgetsBinding.instance;
+    if (instance != null) {
+      instance.addObserver(_observer);
```
ueman, line 200: Shouldn't that be logged to the console?

marandaneto: good catch, added an else branch that calls `options.logger`

ueman (approved):

PR #269, opened 2021-01-18
fix: Pass auto session tracking interval to iOS

philipphofmann:
## :scroll: Description
In SwiftSentryFlutterPlugin we look for sessionTrackingIntervalMillis in
the arguments. Instead, it must be autoSessionTrackingIntervalMillis, see https://github.com/getsentry/sentry-dart/blob/8406d676d4839606b89665ed65a99dcdb3910a98/flutter/lib/src/sentry_flutter_options.dart#L35.

We want to be able to change the `autoSessionTrackingIntervalMillis ` in iOS.

With an emulator.

## :pencil: Checklist

- [x] I reviewed the submitted code
- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # vNext
 
+* Fix: Pass auto session tracking interval to iOS
 * Fix: Deprecated binaryMessenger (MethodChannel member) for Flutter Web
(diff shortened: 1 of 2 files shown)


marandaneto (approved):
good catch, probably a ctrl c/ctrl v typo, thanks

PR #283, opened 2021-01-26
Readme: Fix custom runZonedGuarded

marandaneto:
## :scroll: Description
Readme: Fix custom runZonedGuarded

Sentry should be inited within custom runZonedGuarded


bruno-garcia (commented):
I thought the whole point of `SentryFlutter.init` was to avoid having to do this.

Is this a must? Is it because of the bug fix that just went out?

marandaneto:
> I thought the whole point of `SentryFlutter.init` was to avoid having to do this.
> 
> Is this a must? Is it because of the bug fix that just went out?

the changed code snippet is under this section:

> Or, if you want to run your app in your own error zone [runZonedGuarded]

this section didn't change:

```

Future<void> main() async {
  await SentryFlutter.init(
    (options) {
    },
    // Init your App.
  );
}
```

if you pass the `appRunner` you are good to go

PR #295, opened 2021-02-09
Fix: Pana Flutter upper bound deprecation

marandaneto:
## :scroll: Description
Pana Flutter upper bound deprecation

Package validation found the following potential issue:
* The Flutter constraint should not have an upper bound.
  In your pubspec.yaml the constraint is currently `>=1.17.0 <2.0.0`.

  You can replace that with just the lower bound: `>=1.17.0`.

  See https://dart.dev/go/flutter-upper-bound-deprecation

## :pencil: Checklist

- [X] All tests passing
- [X] No breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # vNext
 
 * Bump: sentry-android to v4.0.0
+* Fix: Pana Flutter upper bound deprecation
 
 # 4.0.4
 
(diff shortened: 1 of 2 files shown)


bruno-garcia (approved):
LGTM but we're still at 90 points scope, not sure if this change affected that

marandaneto:
> LGTM but we're still at 90 points scope, not sure if this change affected that

the score is something else, a new version from the `http` package, the docker image from the static analyzer is outdated probably, I will need to force a `dart pub upgrade` to download newer versions, I guess.

Issue #247, opened 2020-12-16
Add support to null safety


bruno-garcia:
Flutter is gearing up towards null-safety.

This issue is to track progress adding support to it.

- [X] Support on `sentry` #298
- [X] Support on `sentry_flutter` #337
- [ ] CI uses new Action #334
- [ ] Rename conflicting classes #250

Related to #99

The migration steps are here: https://dart.dev/null-safety/migration-guide

Important to note we should start by migrating our dependencies to the null safety version.

- [X] `uuid` support tracking: https://github.com/Daegalus/dart-uuid/issues/50
- [X] `http` support tracking: https://github.com/dart-lang/http/issues/501
- [X] `device_info` [seems to be almost there](https://github.com/flutter/plugins/pull/3324/files#diff-d2b030e7532ec572e6e20d53bf4e20d867156087f65f38efef4b8e5e16cdd296)

marandaneto:
`package_info` not null-safety ready, but easy to get rid of.
`uuid` not null-safety ready, not hard to get rid of but it'd be a breaking change.
`http` not null-safety ready, hard to get rid of and it'd be a breaking change.

marandaneto:
@bruno-garcia for me, this is a dup.
what's the motivation for creating a new one? see #99

I'd suggest either closing this one or the other.

ueman:
[http](https://pub.dev/packages/http/versions/0.13.0-nullsafety.0), [uuid](https://pub.dev/packages/uuid/versions/3.0.0-nullsafety.0) and [device_info](https://pub.dev/packages/device_info/versions/2.0.0-nullsafety.2) are now migrated to nullsafety.

marandaneto:
we need `package_info` though.
https://pub.dev/packages/package_info/versions
that's the easiest one to get rid of anyway.

ueman:
[package_info](https://pub.dev/packages/package_info/versions/0.5.0-nullsafety/changelog) has also joined the null safety club. So the migration is now unblocked.

marandaneto:
https://pub.dev/packages/http/versions/0.13.0 even went null-safety GA already, so we'll lose 10 pub points https://pub.dev/packages/sentry/score

ueman:
https://github.com/getsentry/sentry-dart/issues/250 and https://github.com/getsentry/sentry-dart/pull/311 are missing for a null safe release.

fzyzcjy:
Today null safety becomes **stable**! Is there any updates?

[...]

PR #298, opened 2021-02-13
Refactoring: Dart Nullsafety

ueman:
## :scroll: Description

This PR migrates the Dart part of this Sentry SDK to null safety.

Some notes:
- Darts migrate tool is really helpful
- mockito makes a lot of problems with non-nullable fields. It's basically useless for them.
- Some tests failed because mockito can't handle non-nullable fields. They were rewritten
- Some mocks were written manually. They're kinda inspired by the [bloc library tests for nullsafety](https://github.com/felangel/bloc/blob/null_safety/packages/bloc/test/mocks/mock_bloc_observer.dart)
- CI builds with Dart & Flutter beta
- `dynamic` can hold null values by default, even without a `?`
- https://dart.dev/null-safety/migration-guide

A lot of test were touched so they should be reviewed carefully.
Most of the user facing APIs accept null values. However events, messages or exceptions are getting dropped if they are null.
So the question is should the API allow null values? I would guess so because it makes the API a little bit more user friendly but I'm kinda torn on that. For now null values are allowed.

https://github.com/getsentry/sentry-dart/issues/247

- Tests are passing
- Manual tests with Dart example
- Manual tests with Flutter example

## :pencil: Checklist

- [x] All tests passing
- [ ] No breaking changes

## :crystal_ball: Next steps
- Migrate the Sentry Flutter SDK to nullsafety
- Version number should be changed because this is a breaking change
    - should be something like `5.0.0-nullsafety.0` according to https://dart.dev/null-safety/migration-guide
- Decide how mocks should be written in the future
- Re-enable CI on stable once Dart and Flutter releases their null safe versions.

### DIFF
--- a/dart/lib/src/hub.dart
+++ b/dart/lib/src/hub.dart
@@ -129,7 +129,7 @@ class Hub {
   /// Captures the message.
   Future<SentryId> captureMessage(
     String? message, {
-    SentryLevel? level = SentryLevel.info,
+    SentryLevel? level,
     String? template,
     List<dynamic>? params,
     dynamic hint,
--- a/dart/lib/src/hub.dart
+++ b/dart/lib/src/hub.dart
@@ -129,7 +129,7 @@ class Hub {
   /// Captures the message.
   Future<SentryId> captureMessage(
     String? message, {
-    SentryLevel level,
+    SentryLevel level = SentryLevel.info,
     String? template,
     List<dynamic>? params,
     dynamic hint,
--- a/dart/lib/src/hub.dart
+++ b/dart/lib/src/hub.dart
@@ -17,8 +17,9 @@ class Hub {
 
   final ListQueue<_StackItem> _stack = ListQueue();
 
-  /// if stack is empty, it throws IterableElementError.noElement()
-  _StackItem? _peek() => _stack.isNotEmpty ? _stack.first : null;
+  // peek can never return null since Stack can be created only with an item and
+  // pop does not drop the last item.
+  _StackItem _peek() => _stack.first;
 
   final SentryOptions _options;
 
@@ -64,27 +65,21 @@ class Hub {
       );
     } else {
       final item = _peek();
-      if (item != null) {
-        try {
-          sentryId = await item.client.captureEvent(
-            event,
-            stackTrace: stackTrace,
-            scope: item.scope,
-            hint: hint,
-          );
-        } catch (err) {
-          _options.logger(
-            SentryLevel.error,
-            'Error while capturing event with id: ${event.eventId.toString()}',
-          );
-        } finally {
-          _lastEventId = sentryId;
-        }
-      } else {
-        _options.logger(
-          SentryLevel.fatal,
-          'Stack peek was null when captureEvent',
+
+      try {
+        sentryId = await item.client.captureEvent(
+          event,
+          stackTrace: stackTrace,
+          scope: item.scope,
+          hint: hint,
         );
+      } catch (err) {
+        _options.logger(
+          SentryLevel.error,
+          'Error while capturing event with id: ${event.eventId.toString()}',
+        );
+      } finally {
+        _lastEventId = sentryId;
       }
     }
     return sentryId;
@@ -110,27 +105,21 @@ class Hub {
       );
     } else {
       final item = _peek();
-      if (item != null) {
-        try {
-          sentryId = await item.client.captureException(
-            throwable,
-            stackTrace: stackTrace,
-            scope: item.scope,
-            hint: hint,
-          );
-        } catch (err) {
-          _options.logger(
-            SentryLevel.error,
-            'Error while capturing exception : $throwable',
-          );
-        } finally {
-          _lastEventId = sentryId;
-        }
-      } else {
-        _options.logger(
-          SentryLevel.fatal,
-          'Stack peek was null when captureException',
+
+      try {
+        sentryId = await item.client.captureException(
+          throwable,
+          stackTrace: stackTrace,
+          scope: item.scope,
+          hint: hint,
         );
+      } catch (err) {
+        _options.logger(
+          SentryLevel.error,
+          'Error while capturing exception : $throwable',
+        );
+      } finally {
+        _lastEventId = sentryId;
       }
     }
 
@@ -159,29 +148,23 @@ class Hub {
       );
     } else {
       final item = _peek();
-      if (item != null) {
-        try {
-          sentryId = await item.client.captureMessage(
-            message,
-            level: level,
-            template: template,
-            params: params,
-            scope: item.scope,
-            hint: hint,
-          );
-        } catch (err) {
-          _options.logger(
-            SentryLevel.error,
-            'Error while capturing message with id: $message',
-          );
-        } finally {
-          _lastEventId = sentryId;
-        }
-      } else {
-        _options.logger(
-          SentryLevel.fatal,
-          'Stack peek was null when captureMessage',
+
+      try {
+        sentryId = await item.client.captureMessage(
+          message,
+          level: level,
+          template: template,
+          params: params,
+          scope: item.scope,
+          hint: hint,
         );
+      } catch (err) {
+        _options.logger(
+          SentryLevel.error,
+          'Error while capturing message with id: $message',
+        );
+      } finally {
+        _lastEventId = sentryId;
       }
     }
     return sentryId;
@@ -196,14 +179,7 @@ class Hub {
       );
     } else {
       final item = _peek();
-      if (item != null) {
-        item.scope.addBreadcrumb(crumb, hint: hint);
-      } else {
-        _options.logger(
-          SentryLevel.fatal,
-          'Stack peek was null when addBreadcrumb',
-        );
-      }
+      item.scope.addBreadcrumb(crumb, hint: hint);
     }
   }
 
@@ -214,15 +190,8 @@ class Hub {
           "Instance is disabled and this 'bindClient' call is a no-op.");
     } else {
       final item = _peek();
-      if (item != null) {
-        _options.logger(SentryLevel.debug, 'New client bound to scope.');
-        item.client = client;
-      } else {
-        _options.logger(
-          SentryLevel.fatal,
-          'Stack peek was null when bindClient',
-        );
-      }
+      _options.logger(SentryLevel.debug, 'New client bound to scope.');
+      item.client = client;
     }
   }
 
@@ -252,21 +221,15 @@ class Hub {
       }
 
       final item = _peek();
-      if (item != null) {
-        try {
-          item.client.close();
-        } catch (err) {
-          _options.logger(
-            SentryLevel.error,
-            'Error while closing the Hub.',
-          );
-        }
-      } else {
+      try {
+        item.client.close();
+      } catch (err) {
         _options.logger(
-          SentryLevel.fatal,
-          'Stack peek was NULL when closing Hub',
+          SentryLevel.error,
+          'Error while closing the Hub.',
         );
       }
+
       _isEnabled = false;
     }
   }
@@ -280,19 +243,13 @@ class Hub {
       );
     } else {
       final item = _peek();
-      if (item != null) {
-        try {
-          callback(item.scope);
-        } catch (err) {
-          _options.logger(
-            SentryLevel.error,
-            "Error in the 'configureScope' callback.",
-          );
-        }
-      } else {
+
+      try {
+        callback(item.scope);
+      } catch (err) {
         _options.logger(
-          SentryLevel.fatal,
-          'Stack peek was NULL when configureScope',
+          SentryLevel.error,
+          "Error in the 'configureScope' callback.",
         );
       }
     }
--- a/dart/lib/src/hub.dart
+++ b/dart/lib/src/hub.dart
@@ -24,6 +24,7 @@ class Hub {
   final SentryOptions _options;
 
   factory Hub(SentryOptions options) {
+    assert(options != null, 'SentryOptions is required.');
     _validateOptions(options);
 
     return Hub._(options);
@@ -35,10 +36,6 @@ class Hub {
   }
 
   static void _validateOptions(SentryOptions options) {
-    if (options == null) {
-      throw ArgumentError('SentryOptions is required.');
-    }
-
     if (options.dsn?.isNotEmpty != true) {
       throw ArgumentError('DSN is required.');
     }
--- a/dart/lib/src/hub.dart
+++ b/dart/lib/src/hub.dart
@@ -219,7 +219,7 @@ class Hub {
   }
 
   /// Binds a different client to the hub
-  void bindClient(SentryClient client) {
+  void bindClient(SentryClient? client) {
     if (!_isEnabled) {
       _options.logger(SentryLevel.warning,
           "Instance is disabled and this 'bindClient' call is a no-op.");
--- a/dart/lib/src/hub.dart
+++ b/dart/lib/src/hub.dart
@@ -98,9 +98,9 @@ class Hub {
 
   /// Captures the exception
   Future<SentryId> captureException(
-    dynamic /*?*/ throwable, {
-    dynamic /*?*/ stackTrace,
-    dynamic /*?*/ hint,
+    dynamic throwable, {
+    dynamic stackTrace,
+    dynamic hint,
   }) async {
     var sentryId = SentryId.empty();
 
--- a/dart/lib/src/hub.dart
+++ b/dart/lib/src/hub.dart
@@ -35,7 +35,7 @@ class Hub {
   }
 
   static void _validateOptions(SentryOptions options) {
-    if (options.dsn.isEmpty) {
+    if (options.dsn == null) {
       throw ArgumentError('DSN is required.');
     }
   }
@@ -52,7 +52,7 @@ class Hub {
 
   /// Captures the event.
   Future<SentryId> captureEvent(
-    SentryEvent? event, {
+    SentryEvent event, {
     dynamic stackTrace,
     dynamic hint,
   }) async {
@@ -62,11 +62,6 @@ class Hub {
       _options.logger(
         SentryLevel.warning,
         "Instance is disabled and this 'captureEvent' call is a no-op.",
-      );
-    } else if (event == null) {
-      _options.logger(
-        SentryLevel.warning,
-        'captureEvent called with null parameter.',
       );
     } else {
       final item = _peek();
@@ -194,16 +189,11 @@ class Hub {
   }
 
   /// Adds a breacrumb to the current Scope
-  void addBreadcrumb(Breadcrumb? crumb, {dynamic hint}) {
+  void addBreadcrumb(Breadcrumb crumb, {dynamic hint}) {
     if (!_isEnabled) {
       _options.logger(
         SentryLevel.warning,
         "Instance is disabled and this 'addBreadcrumb' call is a no-op.",
-      );
-    } else if (crumb == null) {
-      _options.logger(
-        SentryLevel.warning,
-        'addBreadcrumb called with null parameter.',
       );
     } else {
       final item = _peek();
@@ -219,20 +209,15 @@ class Hub {
   }
 
   /// Binds a different client to the hub
-  void bindClient(SentryClient? client) {
+  void bindClient(SentryClient client) {
     if (!_isEnabled) {
       _options.logger(SentryLevel.warning,
           "Instance is disabled and this 'bindClient' call is a no-op.");
     } else {
       final item = _peek();
       if (item != null) {
-        if (client != null) {
-          _options.logger(SentryLevel.debug, 'New client bound to scope.');
-          item.client = client;
-        } else {
-          _options.logger(SentryLevel.debug, 'NoOp client bound to scope.');
-          item.client = NoOpSentryClient();
-        }
+        _options.logger(SentryLevel.debug, 'New client bound to scope.');
+        item.client = client;
       } else {
         _options.logger(
           SentryLevel.fatal,
--- a/dart/lib/src/hub.dart
+++ b/dart/lib/src/hub.dart
@@ -53,7 +53,7 @@ class Hub {
 
   /// Captures the event.
   Future<SentryId> captureEvent(
-    SentryEvent event, {
+    SentryEvent /*?*/ event, {
     dynamic stackTrace,
     dynamic hint,
   }) async {
@@ -99,9 +99,9 @@ class Hub {
 
   /// Captures the exception
   Future<SentryId> captureException(
-    dynamic throwable, {
-    dynamic stackTrace,
-    dynamic hint,
+    dynamic /*?*/ throwable, {
+    dynamic /*?*/ stackTrace,
+    dynamic /*?*/ hint,
   }) async {
     var sentryId = SentryId.empty();
 
@@ -146,7 +146,7 @@ class Hub {
 
   /// Captures the message.
   Future<SentryId> captureMessage(
-    String message, {
+    String /*?*/ message, {
     SentryLevel level = SentryLevel.info,
     String template,
     List<dynamic> params,
@@ -195,7 +195,7 @@ class Hub {
   }
 
   /// Adds a breacrumb to the current Scope
-  void addBreadcrumb(Breadcrumb crumb, {dynamic hint}) {
+  void addBreadcrumb(Breadcrumb /*?*/ crumb, {dynamic hint}) {
     if (!_isEnabled) {
       _options.logger(
         SentryLevel.warning,
--- a/dart/lib/src/hub.dart
+++ b/dart/lib/src/hub.dart
@@ -19,12 +19,11 @@ class Hub {
   final ListQueue<_StackItem> _stack = ListQueue();
 
   /// if stack is empty, it throws IterableElementError.noElement()
-  _StackItem _peek() => _stack.isNotEmpty ? _stack.first : null;
+  _StackItem? _peek() => _stack.isNotEmpty ? _stack.first : null;
 
   final SentryOptions _options;
 
   factory Hub(SentryOptions options) {
-    assert(options != null, 'SentryOptions is required.');
     _validateOptions(options);
 
     return Hub._(options);
@@ -36,7 +35,7 @@ class Hub {
   }
 
   static void _validateOptions(SentryOptions options) {
-    if (options.dsn?.isNotEmpty != true) {
+    if (options.dsn.isEmpty) {
       throw ArgumentError('DSN is required.');
     }
   }
@@ -53,7 +52,7 @@ class Hub {
 
   /// Captures the event.
   Future<SentryId> captureEvent(
-    SentryEvent /*?*/ event, {
+    SentryEvent? event, {
     dynamic stackTrace,
     dynamic hint,
   }) async {
@@ -146,10 +145,10 @@ class Hub {
 
   /// Captures the message.
   Future<SentryId> captureMessage(
-    String /*?*/ message, {
-    SentryLevel level = SentryLevel.info,
-    String template,
-    List<dynamic> params,
+    String? message, {
+    SentryLevel? level = SentryLevel.info,
+    String? template,
+    List<dynamic>? params,
     dynamic hint,
   }) async {
     var sentryId = SentryId.empty();
@@ -195,7 +194,7 @@ class Hub {
   }
 
   /// Adds a breacrumb to the current Scope
-  void addBreadcrumb(Breadcrumb /*?*/ crumb, {dynamic hint}) {
+  void addBreadcrumb(Breadcrumb? crumb, {dynamic hint}) {
     if (!_isEnabled) {
       _options.logger(
         SentryLevel.warning,
--- a/dart/lib/src/hub.dart
+++ b/dart/lib/src/hub.dart
@@ -129,7 +129,7 @@ class Hub {
   /// Captures the message.
   Future<SentryId> captureMessage(
     String? message, {
-    SentryLevel level = SentryLevel.info,
+    SentryLevel? level = SentryLevel.info,
     String? template,
     List<dynamic>? params,
     dynamic hint,
--- a/dart/lib/src/hub.dart
+++ b/dart/lib/src/hub.dart
@@ -1,7 +1,6 @@
 import 'dart:async';
 import 'dart:collection';
 
-import 'noop_sentry_client.dart';
 import 'protocol.dart';
 import 'scope.dart';
 import 'sentry_client.dart';
--- a/dart/lib/src/hub.dart
+++ b/dart/lib/src/hub.dart
@@ -146,7 +146,7 @@ class Hub {
   /// Captures the message.
   Future<SentryId> captureMessage(
     String? message, {
-    SentryLevel? level = SentryLevel.info,
+    SentryLevel level = SentryLevel.info,
     String? template,
     List<dynamic>? params,
     dynamic hint,
--- a/dart/lib/src/hub_adapter.dart
+++ b/dart/lib/src/hub_adapter.dart
@@ -49,7 +49,7 @@ class HubAdapter implements Hub {
   @override
   Future<SentryId> captureMessage(
     String? message, {
-    SentryLevel? level = SentryLevel.info,
+    SentryLevel? level,
     String? template,
     List? params,
     dynamic hint,
--- a/dart/lib/src/hub_adapter.dart
+++ b/dart/lib/src/hub_adapter.dart
@@ -16,7 +16,7 @@ class HubAdapter implements Hub {
   }
 
   @override
-  void addBreadcrumb(Breadcrumb crumb, {dynamic hint}) =>
+  void addBreadcrumb(Breadcrumb? crumb, {dynamic hint}) =>
       Sentry.addBreadcrumb(crumb, hint: hint);
 
   @override
@@ -24,7 +24,7 @@ class HubAdapter implements Hub {
 
   @override
   Future<SentryId> captureEvent(
-    SentryEvent event, {
+    SentryEvent? event, {
     dynamic stackTrace,
     dynamic hint,
   }) =>
@@ -48,10 +48,10 @@ class HubAdapter implements Hub {
 
   @override
   Future<SentryId> captureMessage(
-    String message, {
-    SentryLevel level = SentryLevel.info,
-    String template,
-    List params,
+    String? message, {
+    SentryLevel? level = SentryLevel.info,
+    String? template,
+    List? params,
     dynamic hint,
   }) =>
       Sentry.captureMessage(
--- a/dart/lib/src/hub_adapter.dart
+++ b/dart/lib/src/hub_adapter.dart
@@ -49,7 +49,7 @@ class HubAdapter implements Hub {
   @override
   Future<SentryId> captureMessage(
     String? message, {
-    SentryLevel level = SentryLevel.info,
+    SentryLevel? level = SentryLevel.info,
     String? template,
     List? params,
     dynamic hint,
--- a/dart/lib/src/hub_adapter.dart
+++ b/dart/lib/src/hub_adapter.dart
@@ -49,7 +49,7 @@ class HubAdapter implements Hub {
   @override
   Future<SentryId> captureMessage(
     String? message, {
-    SentryLevel level,
+    SentryLevel level = SentryLevel.info,
     String? template,
     List? params,
     dynamic hint,
--- a/dart/lib/src/hub_adapter.dart
+++ b/dart/lib/src/hub_adapter.dart
@@ -20,7 +20,7 @@ class HubAdapter implements Hub {
       Sentry.addBreadcrumb(crumb, hint: hint);
 
   @override
-  void bindClient(SentryClient client) => Sentry.bindClient(client);
+  void bindClient(SentryClient? client) => Sentry.bindClient(client);
 
   @override
   Future<SentryId> captureEvent(
--- a/dart/lib/src/hub_adapter.dart
+++ b/dart/lib/src/hub_adapter.dart
@@ -49,7 +49,7 @@ class HubAdapter implements Hub {
   @override
   Future<SentryId> captureMessage(
     String? message, {
-    SentryLevel? level = SentryLevel.info,
+    SentryLevel level = SentryLevel.info,
     String? template,
     List? params,
     dynamic hint,
--- a/dart/lib/src/hub_adapter.dart
+++ b/dart/lib/src/hub_adapter.dart
@@ -16,15 +16,15 @@ class HubAdapter implements Hub {
   }
 
   @override
-  void addBreadcrumb(Breadcrumb? crumb, {dynamic hint}) =>
+  void addBreadcrumb(Breadcrumb crumb, {dynamic hint}) =>
       Sentry.addBreadcrumb(crumb, hint: hint);
 
   @override
-  void bindClient(SentryClient? client) => Sentry.bindClient(client);
+  void bindClient(SentryClient client) => Sentry.bindClient(client);
 
   @override
   Future<SentryId> captureEvent(
-    SentryEvent? event, {
+    SentryEvent event, {
     dynamic stackTrace,
     dynamic hint,
   }) =>
--- a/dart/lib/src/isolate_error_integration.dart
+++ b/dart/lib/src/isolate_error_integration.dart
@@ -37,7 +37,6 @@ RawReceivePort _createPort(Hub hub, SentryOptions options) {
 }
 
 /// Parse and raise an event out of the Isolate error.
-@internal
 @visibleForTesting
 Future<void> handleIsolateError(
   Hub hub,
--- a/dart/lib/src/isolate_error_integration.dart
+++ b/dart/lib/src/isolate_error_integration.dart
@@ -37,6 +37,7 @@ RawReceivePort _createPort(Hub hub, SentryOptions options) {
 }
 
 /// Parse and raise an event out of the Isolate error.
+@internal
 @visibleForTesting
 Future<void> handleIsolateError(
   Hub hub,
--- a/dart/lib/src/isolate_error_integration.dart
+++ b/dart/lib/src/isolate_error_integration.dart
@@ -10,7 +10,7 @@ import 'sentry_options.dart';
 import 'throwable_mechanism.dart';
 
 class IsolateErrorIntegration extends Integration {
-  RawReceivePort _receivePort;
+  late RawReceivePort _receivePort;
 
   @override
   FutureOr<void> call(Hub hub, SentryOptions options) async {
--- a/dart/lib/src/noop_hub.dart
+++ b/dart/lib/src/noop_hub.dart
@@ -18,7 +18,7 @@ class NoOpHub implements Hub {
 
   @override
   Future<SentryId> captureEvent(
-    SentryEvent event, {
+    SentryEvent? event, {
     dynamic stackTrace,
     dynamic hint,
   }) =>
@@ -34,10 +34,10 @@ class NoOpHub implements Hub {
 
   @override
   Future<SentryId> captureMessage(
-    String message, {
-    SentryLevel level = SentryLevel.info,
-    String template,
-    List params,
+    String? message, {
+    SentryLevel? level = SentryLevel.info,
+    String? template,
+    List? params,
     dynamic hint,
   }) =>
       Future.value(SentryId.empty());
@@ -58,5 +58,5 @@ class NoOpHub implements Hub {
   SentryId get lastEventId => SentryId.empty();
 
   @override
-  void addBreadcrumb(Breadcrumb crumb, {dynamic hint}) {}
+  void addBreadcrumb(Breadcrumb? crumb, {dynamic hint}) {}
 }
--- a/dart/lib/src/noop_hub.dart
+++ b/dart/lib/src/noop_hub.dart
@@ -14,7 +14,7 @@ class NoOpHub implements Hub {
   }
 
   @override
-  void bindClient(SentryClient client) {}
+  void bindClient(SentryClient? client) {}
 
   @override
   Future<SentryId> captureEvent(
--- a/dart/lib/src/noop_hub.dart
+++ b/dart/lib/src/noop_hub.dart
@@ -58,5 +58,5 @@ class NoOpHub implements Hub {
   SentryId get lastEventId => SentryId.empty();
 
   @override
-  void addBreadcrumb(Breadcrumb? crumb, {dynamic hint}) {}
+  void addBreadcrumb(Breadcrumb crumb, {dynamic hint}) {}
 }
--- a/dart/lib/src/noop_hub.dart
+++ b/dart/lib/src/noop_hub.dart
@@ -14,11 +14,11 @@ class NoOpHub implements Hub {
   }
 
   @override
-  void bindClient(SentryClient? client) {}
+  void bindClient(SentryClient client) {}
 
   @override
   Future<SentryId> captureEvent(
-    SentryEvent? event, {
+    SentryEvent event, {
     dynamic stackTrace,
     dynamic hint,
   }) =>
--- a/dart/lib/src/protocol/app.dart
+++ b/dart/lib/src/protocol/app.dart
@@ -53,26 +53,26 @@ class App {
 
   /// Produces a [Map] that can be serialized to JSON.
   Map<String, dynamic> toJson() {
-    final json = <String, String?>{};
+    final json = <String, String>{};
 
     if (name != null) {
-      json['app_name'] = name;
+      json['app_name'] = name!;
     }
 
     if (version != null) {
-      json['app_version'] = version;
+      json['app_version'] = version!;
     }
 
     if (identifier != null) {
-      json['app_identifier'] = identifier;
+      json['app_identifier'] = identifier!;
     }
 
     if (build != null) {
-      json['app_build'] = build;
+      json['app_build'] = build!;
     }
 
     if (buildType != null) {
-      json['build_type'] = buildType;
+      json['build_type'] = buildType!;
     }
 
     if (startTime != null) {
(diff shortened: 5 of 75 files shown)


dart/lib/src/protocol/debug_image.dart
```diff
@@ -11,32 +11,32 @@ import 'package:meta/meta.dart';
 /// more details : https://develop.sentry.dev/sdk/event-payloads/debugmeta/
 @immutable
 class DebugImage {
```
ueman, line 13: DebugImage has a couple fields which are required as per their comments. Should I make them non-nullable?

dart/lib/src/protocol/mechanism.dart
```diff
@@ -15,15 +15,15 @@ class Mechanism {
...
-  final String description;
+  final String? description;
 
   /// Optional fully qualified URL to an online help resource, possible interpolated with error parameters
-  final String helpLink;
+  final String? helpLink;
 
   /// Optional flag indicating whether the exception has been handled by the user (e.g. via try..catch)
-  final bool handled;
+  final bool? handled;
```
ueman, line 24: Should this be non-nullable?

dart/lib/src/protocol/sdk_version.dart
```diff
@@ -36,11 +36,11 @@ import 'sentry_package.dart';
 class SdkVersion {
   /// Creates an [SdkVersion] object which represents the SDK that created an [Event].
   SdkVersion({
-    @required this.name,
-    @required this.version,
-    List<String> integrations,
-    List<SentryPackage> packages,
-  })  : assert(name != null || version != null),
+    required this.name,
```
ueman, line 39: `SdkVersion` required name and version which make it unlikely that anyone just passed on of them like the assert suggests.
Currently they are both non-nullable.

dart/lib/src/protocol/debug_image.dart
marandaneto: I think only `type` should be required as the requirements change based on the type
what we could do is to offer named ctors so we could use non-nullable field based on the ctor (maybe lets not do it now).
all the fields should be nullable though, to support each type of image (based on type)

dart/lib/src/protocol/mechanism.dart
marandaneto: nullable, https://develop.sentry.dev/sdk/event-payloads/exception/#attributes-1
it's optional and false by default, no need to serialize a value that is false-by-default

dart/lib/src/protocol/sdk_version.dart
marandaneto: name and version are required, non-nullable https://develop.sentry.dev/sdk/event-payloads/sdk/
the thing is, we set this automatically via the SDK, probably no one is overwriting this.
`integrations` and `packages`, nullable

dart/lib/src/protocol/mechanism.dart
```diff
@@ -77,9 +77,7 @@ class Mechanism {
   Map<String, dynamic> toJson() {
     final json = <String, dynamic>{};
 
-    if (type != null) {
-      json['type'] = type;
-    }
+    json['type'] = type;
```
ueman, line 80: Only add if everything else is present

dart/lib/src/protocol/sentry_event.dart
```diff
@@ -285,52 +280,52 @@ class SentryEvent {
     }
 
     if (level != null) {
-      json['level'] = level.name;
+      json['level'] = level!.name;
```
ueman, line 283: Hm, level should probably be non-nullable with a default value

dart/lib/src/protocol/sentry_stack_trace.dart
```diff
@@ -5,11 +5,11 @@ import 'sentry_stack_frame.dart';
...
 class SentryStackTrace {
-  const SentryStackTrace({
-    @required List<SentryStackFrame> frames,
-    Map<String, String> registers,
+  SentryStackTrace({
+    required List<SentryStackFrame> frames,
+    Map<String, String>? registers,
   })  : _frames = frames,
-        _registers = registers;
+        _registers = Map.from(registers ?? {});
```
ueman, line 12: Check if I can use `_registers = Map.from(registers ?? const {});`
Also probably enable the following lints in another PR:

- https://dart-lang.github.io/linter/lints/prefer_const_constructors.html
- https://dart-lang.github.io/linter/lints/prefer_const_declarations.html
- https://dart-lang.github.io/linter/lints/prefer_const_literals_to_create_immutables.html

Const ist good because https://stackoverflow.com/a/50431087/12002462

dart/lib/src/sentry_options.dart
```diff
@@ -20,45 +20,32 @@ class SentryOptions {
...
   /// If [compressPayload] is `true` the outgoing HTTP payloads are compressed
   /// using gzip. Otherwise, the payloads are sent in plain UTF8-encoded JSON
   /// text. If not specified, the compression is enabled by default.
-  bool get compressPayload => _compressPayload;
-
-  set compressPayload(bool compressPayload) =>
-      _compressPayload = compressPayload ?? _compressPayload;
-
-  Client _httpClient = NoOpClient();
+  bool compressPayload = true;
```
ueman, line 30: Edit docs, because it is now always specified.

dart/lib/src/sentry_options.dart
```diff
@@ -20,45 +20,32 @@ class SentryOptions {
...
 
   /// If [httpClient] is provided, it is used instead of the default client to
   /// make HTTP calls to Sentry.io. This is useful in tests.
-  Client get httpClient => _httpClient;
-
-  set httpClient(Client httpClient) => _httpClient = httpClient ?? _httpClient;
-
-  ClockProvider _clock = getUtcDateTime;
+  /// If you don't need to send event, use [NoOpClient].
```
ueman:
```suggestion
  /// If you don't need to send events, use [NoOpClient].
```

dart/lib/src/sentry_stack_trace_factory.dart
```diff
@@ -7,23 +7,16 @@ import 'sentry_options.dart';
...
-    if (options == null) {
-      throw ArgumentError('SentryOptions is required.');
-    }
-    _options = options;
-  }
+  SentryStackTraceFactory(this.options);
 
   /// returns the [SentryStackFrame] list from a stackTrace ([StackTrace] or [String])
   List<SentryStackFrame> getStackFrames(dynamic stackTrace) {
-    if (stackTrace == null) return null;
```
ueman, line 25: This method should probably be changed to `getStackFramesFromString(String)` and `getStackFramesFromString(StackTrace)` because all other types and null are not allowed.

dart/lib/src/transport/http_transport.dart
```diff
@@ -54,7 +50,7 @@ class HttpTransport implements Transport {
 
     final response = await _options.httpClient.post(
       _dsn.postUri,
-      headers: _credentialBuilder.configure(_headers),
+      headers: _credentialBuilder.configure(_headers) as Map<String, String>?,
```
ueman: double check this

dart/pubspec.yaml
```diff
@@ -7,18 +7,18 @@ homepage: https://docs.sentry.io/platforms/dart/
 repository: https://github.com/getsentry/sentry-dart
 
 environment:
-  sdk: ">=2.8.0 <3.0.0"
+  sdk: '>=2.12.0-259.8.beta <3.0.0'
```
ueman: This should be `sdk: '>=2.12.0-0 <3.0.0'`

ueman: Update: As per https://medium.com/dartlang/preparing-the-dart-and-flutter-ecosystem-for-null-safety-e550ce72c010 this should now be `>=2.12.0–259.9.beta < 3.0.0`

ueman:
```suggestion
  sdk: '>=2.12.0-259.9.beta <3.0.0'
```

dart/lib/src/hub.dart
```diff
@@ -56,7 +52,7 @@ class Hub {
 
   /// Captures the event.
   Future<SentryId> captureEvent(
-    SentryEvent event, {
+    SentryEvent? event, {
```
```suggestion
    SentryEvent event, {
```

it'd not make sense to accept a null SentryEvent, we'd drop it anyway

dart/lib/src/hub.dart
```diff
@@ -149,10 +145,10 @@ class Hub {
 
   /// Captures the message.
   Future<SentryId> captureMessage(
-    String message, {
+    String? message, {
```
marandaneto, line 131: here it does make sense to be nullable because values come at runtime from external sources

dart/lib/src/hub.dart
```diff
@@ -198,7 +194,7 @@ class Hub {
   }
 
   /// Adds a breacrumb to the current Scope
-  void addBreadcrumb(Breadcrumb crumb, {dynamic hint}) {
+  void addBreadcrumb(Breadcrumb? crumb, {dynamic hint}) {
```
```suggestion
  void addBreadcrumb(Breadcrumb crumb, {dynamic hint}) {
```

same case as SentryEvent

dart/lib/src/hub.dart
```diff
@@ -223,7 +219,7 @@ class Hub {
   }
 
   /// Binds a different client to the hub
-  void bindClient(SentryClient client) {
+  void bindClient(SentryClient? client) {
```
```suggestion
  void bindClient(SentryClient client) {
```

dart/lib/src/isolate_error_integration.dart
```diff
@@ -37,6 +37,7 @@ RawReceivePort _createPort(Hub hub, SentryOptions options) {
 }
 
 /// Parse and raise an event out of the Isolate error.
+@internal
 @visibleForTesting
```
marandaneto: is it common to use both of them together?

dart/lib/src/protocol/app.dart
```diff
@@ -31,29 +31,29 @@ class App {
...
+  final DateTime? startTime;
 
   /// Application specific device identifier.
-  final String deviceAppHash;
+  final String? deviceAppHash;
 
   /// Produces a [Map] that can be serialized to JSON.
   Map<String, dynamic> toJson() {
-    final json = <String, String>{};
+    final json = <String, String?>{};
```
marandaneto: gues the value is non-null, we check for nullability each field I guess

dart/lib/src/protocol/app.dart
```diff
@@ -76,7 +76,7 @@ class App {
     }
 
     if (startTime != null) {
-      json['app_start_time'] = startTime.toIso8601String();
+      json['app_start_time'] = startTime!.toIso8601String();
```
marandaneto, line 79: is this required because its non final? so it could mutate from one thread to another

dart/lib/src/protocol/breadcrumb.dart
```diff
@@ -50,7 +50,7 @@ class Breadcrumb {
   /// See also:
   ///
   /// * https://docs.sentry.io/development/sdk-dev/event-payloads/breadcrumbs/#breadcrumb-types
-  final Map<String, dynamic> data;
+  final Map<String, dynamic>? data;
```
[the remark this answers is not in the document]
marandaneto: that's a possibility, the downside is memory footprint, those maps/lists are created but maybe never used

dart/lib/src/protocol/breadcrumb.dart
```diff
@@ -87,25 +87,25 @@ class Breadcrumb {
...
     }
-    if (data != null && data.isNotEmpty) {
+    if (data?.isNotEmpty ?? false) {
       json['data'] = data;
     }
-    if (level != null) {
-      json['level'] = level.name;
-    }
+
+    json['level'] = level.name;
```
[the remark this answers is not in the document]
marandaneto: why so? https://develop.sentry.dev/sdk/event-payloads/breadcrumbs/ everything is optional

dart/lib/src/protocol/debug_image.dart
```diff
@@ -57,9 +57,7 @@ class DebugImage {
       json['uuid'] = uuid;
     }
 
-    if (type != null) {
-      json['type'] = type;
-    }
+    json['type'] = type;
```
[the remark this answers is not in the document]
marandaneto: type is always required https://develop.sentry.dev/sdk/event-payloads/debugmeta#debug-images
I guess we dont need to check the other fields, the server will just ignore the debug image if its incomplete, better than checking on each SDK, and this might change over time

dart/lib/src/protocol/debug_meta.dart
```diff
@@ -16,18 +16,18 @@ class DebugMeta {
...
-      : _images = images != null ? List.from(images) : null;
+  DebugMeta({this.sdk, List<DebugImage>? images})
+      : _images = List.from(images ?? []);
 
   Map<String, dynamic> toJson() {
     final json = <String, dynamic>{};
 
     Map<String, dynamic> sdkInfo;
-    if (sdk != null && (sdkInfo = sdk.toJson()).isNotEmpty) {
+    if (sdk != null && (sdkInfo = sdk!.toJson()).isNotEmpty) {
```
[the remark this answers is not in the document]
marandaneto: makes sense, but ideally we'd not call always `toJson` twice

dart/lib/src/protocol/device.dart
```diff
@@ -162,6 +162,9 @@ class Device {
       case Orientation.landscape:
         orientation = 'landscape';
         break;
+      case null:
```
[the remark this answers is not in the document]
marandaneto: damn, can we use a `default` then? makes more sense IMO

dart/lib/src/protocol/dsn.dart
```diff
@@ -14,34 +14,36 @@ class Dsn {
...
   /// The DSN URI.
-  final Uri uri;
+  final Uri? uri;
 
-  String get postUri {
-    final port = uri.hasPort &&
-            ((uri.scheme == 'http' && uri.port != 80) ||
-                (uri.scheme == 'https' && uri.port != 443))
-        ? ':${uri.port}'
+  Uri get postUri {
```
marandaneto, line 27: the signature change was required or just an improvement?

dart/lib/src/protocol/dsn.dart
```diff
@@ -14,34 +14,36 @@ class Dsn {
...
+    final port = uri!.hasPort &&
```
marandaneto: can we create a new local field like `final uriCopy = uri!` and using only `uriCopy`?
calling `uri!` every time affects readability

dart/lib/src/protocol/mechanism.dart
```diff
@@ -42,27 +42,27 @@ class Mechanism {
...
     this.handled,
     this.synthetic,
-    Map<String, dynamic> meta,
-    Map<String, dynamic> data,
-  })  : _meta = meta != null ? Map.from(meta) : null,
-        _data = data != null ? Map.from(data) : null;
+    Map<String, dynamic>? meta,
+    Map<String, dynamic>? data,
+  })  : _meta = Map.from(meta ?? {}),
+        _data = Map.from(data ?? {});
```
marandaneto: the downside of doing this is allocating a Map when its not necessary, if values are nullable, we dont need to do it right

dart/lib/src/protocol/mechanism.dart
marandaneto: similar to a comment made above

dart/lib/src/protocol/message.dart
```diff
@@ -11,14 +11,14 @@ import 'package:meta/meta.dart';
 @immutable
 class Message {
   /// The fully formatted message. If missing, Sentry will try to interpolate the message.
-  final String formatted;
+  final String? formatted;
```
marandaneto: this is required https://develop.sentry.dev/sdk/event-payloads/message/

dart/lib/src/protocol/sdk_version.dart
```diff
@@ -36,14 +36,14 @@ import 'sentry_package.dart';
...
+    required this.version,
+    List<String>? integrations,
+    List<SentryPackage>? packages,
+  })  :
         // List.from prevents from having immutable lists
-        _integrations = integrations != null ? List.from(integrations) : [],
-        _packages = packages != null ? List.from(packages) : [];
+        _integrations = List.from(integrations ?? []),
+        _packages = List.from(packages ?? []);
```
marandaneto, line 46: similar to comment above

dart/lib/src/protocol/sentry_event.dart
```diff
@@ -30,17 +30,17 @@ class SentryEvent {
...
         contexts = contexts ?? Contexts(),
         modules = modules != null ? Map.from(modules) : null,
-        tags = tags != null ? Map.from(tags) : null,
-        extra = extra != null ? Map.from(extra) : null,
-        fingerprint = fingerprint != null ? List.from(fingerprint) : null,
-        breadcrumbs = breadcrumbs != null ? List.from(breadcrumbs) : null;
+        tags = Map.from(tags ?? {}),
+        extra = Map.from(extra ?? {}),
+        fingerprint = List.from(fingerprint ?? []),
+        breadcrumbs = List.from(breadcrumbs ?? []);
```
marandaneto: similar to comment above, well I wont comment on the next ones, if we decide doing like this, we could make maps and lists non nullable and only add to json if non empty

dart/lib/src/protocol/sentry_event.dart
marandaneto: level is inferred on the server if not available, we don't need to set it then.

dart/lib/src/scope.dart
```diff
@@ -24,15 +24,14 @@ class Scope {
...
 
-  set fingerprint(List<String> fingerprint) {
-    _fingerprint = (fingerprint != null ? List.from(fingerprint) : fingerprint);
+  set fingerprint(List<String>? fingerprint) {
+    _fingerprint = List.from(fingerprint ?? []);
   }
 
   /// List of breadcrumbs for this scope.
-  final Queue<Breadcrumb> _breadcrumbs = Queue();
+  final Queue<Breadcrumb?> _breadcrumbs = Queue();
```
```suggestion
  final Queue<Breadcrumb> _breadcrumbs = Queue();
```

dart/lib/src/scope.dart
```diff
@@ -6,16 +6,16 @@ import 'sentry_options.dart';
...
   /// for example, the route name: `"/users/<username>/"`.
-  String transaction;
+  String? transaction;
 
   /// Information about the current user.
-  User user;
+  User? user;
 
-  List<String> _fingerprint;
+  List<String> _fingerprint = [];
```
marandaneto, line 18: it'd be fine to be nullable too

dart/lib/src/scope.dart
```diff
@@ -60,7 +59,7 @@ class Scope {
   Map<String, dynamic> get contexts => Map.unmodifiable(_contexts);
 
   /// add an entry to the Scope's contexts
-  void setContexts(String key, dynamic value) {
+  void setContexts(String? key, dynamic value) {
```
```suggestion
  void setContexts(String key, dynamic value) {
```

`if (key == null || value == null) return;` could be gone too

dart/lib/src/scope.dart
```diff
@@ -83,24 +82,18 @@ class Scope {
...
-      throw ArgumentError('SentryOptions is required');
-    }
-  }
+  Scope(this._options);
 
   /// Adds a breadcrumb to the breadcrumbs queue
-  void addBreadcrumb(Breadcrumb breadcrumb, {dynamic hint}) {
-    assert(breadcrumb != null, "Breadcrumb can't be null");
-
+  void addBreadcrumb(Breadcrumb? breadcrumb, {dynamic hint}) {
```
```suggestion
  void addBreadcrumb(Breadcrumb breadcrumb, {dynamic hint}) {
```

dart/lib/src/scope.dart
```diff
@@ -157,23 +145,21 @@ class Scope {
...
-    assert(value != null, "Value can't be null");
-
     _extra[key] = value;
   }
 
   /// Removes an extra from the Scope
   void removeExtra(String key) => _extra.remove(key);
 
-  Future<SentryEvent> applyToEvent(SentryEvent event, dynamic hint) async {
+  Future<SentryEvent?> applyToEvent(SentryEvent event, dynamic hint) async {
```
marandaneto, line 156:
```suggestion
  Future<SentryEvent?> applyToEvent(SentryEvent event, dynamic hint?) async {
```

is `dynamic` special case? because every `hint` should be nullable

dart/lib/src/sentry.dart
```diff
@@ -36,27 +36,19 @@ class Sentry {
...
-    SentryOptions options,
+    AppRunner? appRunner,
+    SentryOptions? options,
   }) async {
-    if (optionsConfiguration == null) {
-      throw ArgumentError('OptionsConfiguration is required.');
-    }
-
-    final sentryOptions = options ?? SentryOptions();
+    final sentryOptions = options ?? SentryOptions(dsn: '');
```
```suggestion
    final sentryOptions = options ?? SentryOptions();
```

no DSN is != than empty DSN

dart/lib/src/sentry.dart
```diff
@@ -132,7 +124,7 @@ class Sentry {
 
   /// Reports an [event] to Sentry.io.
   static Future<SentryId> captureEvent(
-    SentryEvent event, {
+    SentryEvent? event, {
```
marandaneto: I guess Ive made the comments for this class already in the Hub class, which is pretty much the same

dart/lib/src/sentry_exception_factory.dart
```diff
@@ -14,8 +12,8 @@ class SentryExceptionFactory {
   SentryExceptionFactory._(this._options, this._stacktraceFactory);
 
   factory SentryExceptionFactory({
-    @required SentryOptions options,
-    @required SentryStackTraceFactory stacktraceFactory,
+    required SentryOptions? options,
+    required SentryStackTraceFactory? stacktraceFactory,
```
marandaneto:
```suggestion
    required SentryOptions options,
    required SentryStackTraceFactory stacktraceFactory,
```

dart/lib/src/sentry_options.dart
```diff
@@ -20,45 +20,32 @@ class SentryOptions {
...
   static final SentryLevel _defaultDiagnosticLevel = SentryLevel.debug;
 
-  /// The DSN tells the SDK where to send the events to. If this value is not provided, the SDK will
-  ///  just not send any events.
-  String dsn;
-
-  bool _compressPayload = true;
+  /// The DSN tells the SDK where to send the events to. If an empty string is
+  /// used, the SDK will not send any events.
+  String dsn = '';
```
marandaneto: empty DSN means SDK is disabled, no DSN at all, means, you have a problem, I'd make it nullable

dart/lib/src/scope.dart
ueman: `dynamic` is null by default, even without a `?`. I put that also in the notes of the PR itself

dart/lib/src/sentry_options.dart
```diff
@@ -180,35 +143,20 @@ class SentryOptions {
...
-    _attachThreads = attachThreads ?? _attachThreads;
-  }
+  bool attachThreads = false;
 
   // TODO: Scope observers, enableScopeSync
 
   // TODO: sendDefaultPii
 
-  SentryOptions({this.dsn}) {
+  SentryOptions({required this.dsn}) {
```
marandaneto: this should still be optional IMO, its a nullable field

dart/lib/src/sentry_stack_trace_factory.dart
```diff
@@ -7,23 +7,16 @@ import 'sentry_options.dart';
 
 /// converts [StackTrace] to [SentryStackFrames]
 class SentryStackTraceFactory {
-  SentryOptions _options;
+  SentryOptions options;
```
marandaneto: it should be private

dart/lib/src/sentry_stack_trace_factory.dart
marandaneto: yep, but the docs mention that and we check its types in the method body, its an internal class, I'd say we are fine, nobody should use it directly

dart/lib/src/transport/http_transport.dart
```diff
@@ -118,9 +114,9 @@ class _CredentialBuilder {
   }
 
   static String _buildAuthHeader({
-    String publicKey,
-    String secretKey,
-    String sdkIdentifier,
+    String? publicKey,
+    String? secretKey,
+    String? sdkIdentifier,
```
marandaneto: wondering if it makes sense to be nullable, a broken header wont be any helpful

dart/lib/src/protocol/mechanism.dart
ueman: Yeah, that's why I added comments about the `const` lints. That way it is not a real performance penalty and non-nullable lists are nicer to work with.
See also https://stackoverflow.com/a/50431087/12002462

dart/lib/src/protocol/dsn.dart
ueman: This is because `http` now requires an Uri instead of String or Uri

dart/lib/src/isolate_error_integration.dart
ueman: Let me double check this

dart/lib/src/protocol/breadcrumb.dart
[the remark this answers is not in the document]
ueman: Well if we make all those empty lists const, every empty list is the same list. So there is no real memory footprint increase. I found working with non-nullable list is a little bit easier than working with nullable lists, that's why I made most of these changes.

dart/lib/src/scope.dart
marandaneto: ops missed that. thanks

dart/lib/src/protocol/breadcrumb.dart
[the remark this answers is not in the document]
ueman: Because previously level had a default value, that's why the migration tool made it non-nullable.
I guess it is also a little bit nicer to work with because you don't have to check if level is null in https://pub.dev/documentation/sentry/latest/sentry/SentryOptions/beforeBreadcrumb.html 

I could imagine that dropping Breadcrumbs because of the level is quite common.

However I'm fine with either way.

dart/lib/src/protocol/debug_meta.dart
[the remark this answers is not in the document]
ueman: good point

dart/lib/src/protocol/device.dart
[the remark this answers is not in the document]
ueman: I like exhaustive cases more than default, because the compiler complains if another value gets added to the enum.
Also I could image people wondering why it has a default even though it looks exhaustive. Your call however

marandaneto:
@ueman an option for:

> dynamic can hold null values by default, even without a ?

would be using `Object` and `Object?` everywhere, seems the right approach when using null-safety, wondering if it's too many changes or a String replace would do, if the change makes sense, of course.

dart/lib/src/protocol/device.dart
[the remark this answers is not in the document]
marandaneto: I agree with your point of view when looking at readability

dart/lib/src/protocol/breadcrumb.dart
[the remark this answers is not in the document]
marandaneto: `timestamp` is never null, because the ctor sets it automatically, so I'd prefer to keep the same behavior, we could make it nullable (level), but setting a default value `info`, its like this across other SDKs too.
because of the `timestamp`, `json.isNotEmpty` would always return false anyway

dart/lib/src/protocol/mechanism.dart
marandaneto: great, thanks

ueman:
> @ueman an option for:
> 
> > dynamic can hold null values by default, even without a ?
> 
> would be using `Object` and `Object?` everywhere, seems the right approach when using null-safety, wondering if it's too many changes or a String replace would do, if the change makes sense, of course.

Good question. I think it's not much of a change because dynamic isn't used that much.
However most dynamic usages would be replaced by `Object?` so I don't see much value in it. 
`dynamic` can be anything and `Object?` can also be anything. 
All instances where `dynamic` shouldn't be null however should probably be replaced by `Object`.

I'll look into it when I finish this up early next week.

marandaneto:
> > @ueman an option for:
> > > dynamic can hold null values by default, even without a ?
> > 
> > 
> > would be using `Object` and `Object?` everywhere, seems the right approach when using null-safety, wondering if it's too many changes or a String replace would do, if the change makes sense, of course.
> 
> Good question. I think it's not much of a change because dynamic isn't used that much.
> However most dynamic usages would be replaced by `Object?` so I don't see much value in it.
> `dynamic` can be anything and `Object?` can also be anything.
> All instances where `dynamic` shouldn't be null however should probably be replaced by `Object`.
> 
> I'll look into it when I finish this up early next week.

yes, that's the difference, you can make it nullable or not by using Object or Object?, I found a few tweets about it and also people on the flutter team are moving away from `dynamic` in favor of `Object/Object?` exactly for this reason

bruno-garcia:
We could release `4.0.0-nullsafety.0` from this branch once it's ready to get merged?

Null safety will GA in 2 weeks, at that point we could probably merge it and publish from `main`

dart/lib/src/protocol/message.dart
ueman: I would leave it as optional because it gets filled from nullable fields, such as `SentryClient.captureMessage(...)` and the message could be null.

dart/lib/src/transport/http_transport.dart
ueman: secret seems to be nullable, the others are not

dart/lib/src/protocol/app.dart
ueman: The compiler complains otherwise. I'm not quite sure tho why. It's final and also the compiler should be able to infer that it's not null.

ueman:
Okay, so I hope I implemented all the feedback and haven't missed something.

Regarding changing `dynamic` to `Object` or `Object?`:
That should definitely be done in another PR because it is not a small change. It also can't be done completely with find and replace because it is implicitly dynamic at a few places. The linter can be made more strict to catch this.

Regarding nullable lists/maps:
I've changed it back to nullables. Const lists and maps are also unmodifiable so it didn't work out as I hoped :(

Other than that it should now be good to go.

marandaneto:
> We could release `4.0.0-nullsafety.0` from this branch once it's ready to get merged?

yes but not from this branch which is a PR from a remote repo, I'll merge it into a feat/null_safety probably

> Null safety will GA in 2 weeks, at that point we could probably merge it and publish from `main`

there's no deadline for null safety GA yet, so we'll likely need to port severe bug fixes on `main` branch if any.

dart/lib/src/hub.dart
```diff
@@ -19,7 +18,7 @@ class Hub {
   final ListQueue<_StackItem> _stack = ListQueue();
 
   /// if stack is empty, it throws IterableElementError.noElement()
-  _StackItem _peek() => _stack.isNotEmpty ? _stack.first : null;
+  _StackItem? _peek() => _stack.isNotEmpty ? _stack.first : null;
```
marandaneto: ideally this would be non null, we'd need to adjust the code similar to https://github.com/getsentry/sentry-java/blob/main/sentry/src/main/java/io/sentry/Stack.java#L69-L74

dart/lib/src/hub.dart
marandaneto: it's resolved but still nullable, why would you think it'd be better nullable?

marandaneto: maybe a reason would be that people can captureMessage at runtime but not checking if its a null value?

dart/lib/src/noop_hub.dart
```diff
@@ -14,11 +14,11 @@ class NoOpHub implements Hub {
   }
 
   @override
-  void bindClient(SentryClient client) {}
+  void bindClient(SentryClient? client) {}
```
marandaneto: on `hub` this is non nullable

dart/lib/src/noop_hub.dart
```diff
@@ -14,11 +14,11 @@ class NoOpHub implements Hub {
...
 
   @override
   Future<SentryId> captureEvent(
-    SentryEvent event, {
+    SentryEvent? event, {
```
marandaneto: I guess this class is not properly checked, it differs from the real implementation

dart/lib/src/protocol/device.dart
```diff
@@ -35,7 +35,8 @@ class Device {
     this.bootTime,
     this.timezone,
   }) : assert(
-            batteryLevel == null || (batteryLevel >= 0 && batteryLevel <= 100));
+          batteryLevel == null || (batteryLevel >= 0 && batteryLevel <= 100),
```
marandaneto, line 38: `batteryLevel == null` does it still make sense this check?

dart/lib/src/protocol/message.dart
marandaneto: SentryClient.captureMessage(null) should not raise an event at all, it's not going to be useful.
if we create a Message without a `formatted` field, it'll cause a backend error during ingestion.

dart/lib/src/protocol/sentry_stack_trace.dart
marandaneto: why don't we use `const` here?

dart/lib/src/scope.dart
```diff
@@ -24,11 +24,10 @@ class Scope {
...
   ///     // A completely custom fingerprint:
   ///     var custom = ['foo', 'bar', 'baz'];
-  List<String> get fingerprint =>
-      _fingerprint != null ? List.unmodifiable(_fingerprint) : null;
+  List<String> get fingerprint => List.unmodifiable(_fingerprint);
 
-  set fingerprint(List<String> fingerprint) {
-    _fingerprint = (fingerprint != null ? List.from(fingerprint) : fingerprint);
+  set fingerprint(List<String>? fingerprint) {
+    _fingerprint = List.from(fingerprint ?? []);
```
marandaneto: const?
if the field is not nullable, I'd make this param also not nullable, or the other way around

dart/lib/src/scope.dart
```diff
@@ -83,39 +80,42 @@ class Scope {
...
     // run before breadcrumb callback if set
-    if (_options.beforeBreadcrumb != null) {
-      breadcrumb = _options.beforeBreadcrumb(breadcrumb, hint: hint);
-
-      if (breadcrumb == null) {
-        _options.logger(
-            SentryLevel.info, 'Breadcrumb was dropped by beforeBreadcrumb');
-        return;
+    if (_options.beforeBreadcrumb == null) {
+      // remove first item if list if full
```
```suggestion
      // remove first item if list is full
```

dart/lib/src/scope.dart
```diff
@@ -83,39 +80,42 @@ class Scope {
...
+      if (_breadcrumbs.length >= _options.maxBreadcrumbs &&
+          _breadcrumbs.isNotEmpty) {
+        _breadcrumbs.removeFirst();
```
marandaneto: we duplicated this logic and previously we did it only once, how this change make it better?

dart/lib/src/scope.dart
```diff
@@ -190,22 +184,23 @@ class Scope {
       }
     });
 
+    SentryEvent? processedEvent = event;
```
marandaneto, line 182: why are we copying instead of using the same instance? we are mutating it anyway

ueman: Because the passed instance is non-nullable but because of the processing it can now be nullable. Thus we need to declare a new variable which is nullable.

dart/lib/src/sentry_client.dart
```diff
@@ -60,37 +56,39 @@ class SentryClient {
       return _sentryId;
     }
 
-    event = _prepareEvent(event, stackTrace: stackTrace);
+    SentryEvent? preparedEvent = _prepareEvent(event, stackTrace: stackTrace);
```
marandaneto, line 59: probably the same question as above, copying the reference instead of using it directly

dart/lib/src/noop_hub.dart
ueman: Dang, why didn't the compiler complain about that? :(

dart/lib/src/sentry_client.dart
ueman: Also same answer as above. The parameter is non-nullable, but it can be null after processing.

dart/lib/src/sentry_stack_trace_factory.dart
```diff
@@ -115,7 +108,7 @@ class SentryStackTraceFactory {
       // unparsed      #04 abs 000000723d4b8c3b virt 0000000000071c3b _kDartIsolateSnapshotInstructions+0x66c3b
 
       // we are only interested on the #01, 02... items which contains the 'abs' addresses.
-      final matches = _absRegex.allMatches(member);
+      final matches = _absRegex.allMatches(member!);
```
marandaneto: if `member` is nullable, we cant force `!` but instead checking it

dart/lib/src/protocol/device.dart
ueman: Yep, batteryLevel is null if the device has no battery. Eg. Android TV boxes, Apple TV...

marandaneto:
@ueman thanks once more :)
I did another pass and looks good, 2 points I'd like to address:

sometimes we are not using `const` for [] or {}, is there a reason or must forgot about them?
2nd is sometimes we're copying the reference when mutating the event, previously not, is there a reason?

dart/lib/src/scope.dart
marandaneto: makes sense, missed that, thanks

ueman:
> sometimes we are not using `const` for [] or {}, is there a reason or must forgot about them?

`const []` makes the list immutable, so there's no way to add anything. 

> 2nd is sometimes we're copying the reference when mutating the event, previously not, is there a reason?

Yep, I've added the explanation to your review. But for all other which don't want to search:
Because the passed instance is non-nullable but because of the processing it can now be nullable. Thus we need to declare a new variable which is nullable.

dart/lib/src/scope.dart
ueman: It didn't, I'll change it

marandaneto:
> > sometimes we are not using `const` for [] or {}, is there a reason or must forgot about them?
> 
> `const []` makes the list immutable, so there's no way to add anything.
> 
> > 2nd is sometimes we're copying the reference when mutating the event, previously not, is there a reason?
> 
> Yep, I've added the explanation to your review. But for all other which don't want to search:
> Because the passed instance is non-nullable but because of the processing it can now be nullable. Thus we need to declare a new variable which is nullable.

got it, thanks, hard to get these things during code review if not trying out :)

dart/lib/src/sentry_stack_trace_factory.dart
ueman: Good catch

marandaneto:
@ueman I've created the `feat/null-safety` branch and now we have a few conflicts because of a bug fix :(
after working on them, we could merge and cut a release :)

dart/lib/src/hub.dart
```diff
@@ -149,10 +128,10 @@ class Hub {
 
   /// Captures the message.
   Future<SentryId> captureMessage(
-    String message, {
-    SentryLevel level,
-    String template,
-    List<dynamic> params,
+    String? message, {
+    SentryLevel level = SentryLevel.info,
```
```suggestion
    SentryLevel level,
```

marandaneto: this is done o nthe SentryClient now

dart/lib/src/hub_adapter.dart
```diff
@@ -48,10 +48,10 @@ class HubAdapter implements Hub {
 
   @override
   Future<SentryId> captureMessage(
-    String message, {
-    SentryLevel level,
-    String template,
-    List params,
+    String? message, {
+    SentryLevel level = SentryLevel.info,
```
marandaneto:
```suggestion
    SentryLevel level,
```

dart/lib/src/noop_hub.dart
```diff
@@ -58,5 +58,5 @@ class NoOpHub implements Hub {
   SentryId get lastEventId => SentryId.empty();
 
   @override
-  void addBreadcrumb(Breadcrumb crumb, {dynamic hint}) {}
+  void addBreadcrumb(Breadcrumb? crumb, {dynamic hint}) {}
```
marandaneto:
```suggestion
  void addBreadcrumb(Breadcrumb crumb, {dynamic hint}) {}
```

it does not make sense to be nullable, people would need to create a `Breadcrumb` at runtime or using the addBreadcrumb(String)

dart/lib/src/protocol/message.dart
marandaneto: just reforcing this, captureMessage(String?) is fine, but we'll bail out before creating a Message, so `formatted` should be non nulable to not generate a backend error

dart/lib/src/sentry.dart
```diff
@@ -151,10 +147,10 @@ class Sentry {
       );
 
   static Future<SentryId> captureMessage(
-    String message, {
-    SentryLevel level,
-    String template,
-    List<dynamic> params,
+    String? message, {
+    SentryLevel level = SentryLevel.info,
```
```suggestion
    SentryLevel level,
```

dart/lib/src/sentry_client.dart
```diff
@@ -155,49 +153,50 @@ class SentryClient {
 
   /// Reports the [template]
   Future<SentryId> captureMessage(
-    String formatted, {
-    SentryLevel level,
-    String template,
-    List<dynamic> params,
-    Scope scope,
+    String? formatted, {
+    SentryLevel level = SentryLevel.info,
```
marandaneto:
```suggestion
    SentryLevel level,
```

please double check all the `SentryLevel level = SentryLevel.info`

dart/lib/src/hub.dart
ueman: I know but then SentryLevel is a required field (because it is not nullable) which would mean that anyone who is using the Hub directly would always need to specify a SentryLevel.

dart/test/mocks/mock_hub.dart
```diff
@@ -0,0 +1,121 @@
...
+  }
+
+  @override
+  // TODO: implement isEnabled
+  bool get isEnabled => throw UnimplementedError();
+
+  @override
+  // TODO: implement lastEventId
+  SentryId get lastEventId => throw UnimplementedError();
+}
```
marandaneto, line 78: we had tests for this before, not anymore, I think we should work on them or the chance to add a regression is higher

dart/lib/src/sentry_client.dart
ueman: SentryLevel is a required field (because it is not nullable) If it doesn't have a default value  which would mean that anyone who is using the Hub directly would always need to specify a SentryLevel. Should I still make the change?

flutter/pubspec.yaml
```diff
@@ -13,20 +13,21 @@ dependencies:
     sdk: flutter
   flutter_web_plugins:
     sdk: flutter
-  sentry: ">=4.0.0 <5.0.0"
-  package_info: ^0.4.0
+  # this should point to pub
+  sentry:
+    path: ../dart
```
marandaneto: should we revert this before merging?

flutter/pubspec.yaml
```diff
@@ -13,20 +13,21 @@ dependencies:
...
+  mockito: ^5.0.0-nullsafety.7
+  yaml: ^3.0.0 # needed for version match (code and pubspec)
+  pedantic: ^1.10.0
 
-dependency_overrides:
-  sentry:
-    path: ../dart
+# dependency_overrides:
+#  sentry:
+#    path: ../dart
```
marandaneto: this helps developing and to debug, I'd keep as it is till we get a full null safety migration

dart/test/mocks/mock_hub.dart
ueman: I am not quite sure which test you refer to. All of those unimplemented things were also previously not verified with the mocks through mockito. Or am I overlooking something?
The only test which does something similar is https://github.com/getsentry/sentry-dart/blob/e8fbf20c673e3e9f85b697c3151d00c2b9a24a1c/dart/test/hub_test.dart#L90 but it is for the SentryClient

dart/lib/src/hub.dart
marandaneto: we could simply `SentryLevel? level` then?

dart/lib/src/sentry_client.dart
marandaneto: I will resolve this one in favor of https://github.com/getsentry/sentry-dart/pull/298#discussion_r580836701
so depending on what we decide, we change for all of them

dart/lib/src/hub.dart
ueman: Yes, so we would just keep the `Sentry` static API non-nullable and all the other nullable?

dart/test/mocks/mock_hub.dart
marandaneto: `hub_test` has tests with `expect(hub.isEnabled...`

`hub_test` has tests with `hub.configureScope...`

and the one you've mentioned, `should save the lastEventId`

dart/lib/src/hub.dart
marandaneto: SentryEvent.level itself is nullable, so all of them should be nullable, the difference is, captureMessage defaults to `info` on `SentryClient` if none is set.

ueman: Ah I see, I guess that's where most of my confusion comes from. Do all others default to `info`, too? Or just `captureMessage`?

dart/test/mocks/mock_hub.dart
ueman: The one I mentioned is for the `SentryClient` tho. 

The following tests 

https://github.com/getsentry/sentry-dart/blob/e8fbf20c673e3e9f85b697c3151d00c2b9a24a1c/dart/test/hub_test.dart#L177 

https://github.com/getsentry/sentry-dart/blob/e8fbf20c673e3e9f85b697c3151d00c2b9a24a1c/dart/test/hub_test.dart#L34

https://github.com/getsentry/sentry-dart/blob/e8fbf20c673e3e9f85b697c3151d00c2b9a24a1c/dart/test/hub_test.dart#L113 

https://github.com/getsentry/sentry-dart/blob/e8fbf20c673e3e9f85b697c3151d00c2b9a24a1c/dart/test/hub_test.dart#L141 

aren't using the mock.

dart/lib/src/hub.dart
marandaneto: only `captureMessage`.
everything else is `error` by default, but nullable on our side, the backend infers that automatically if no level is available.

ueman: got it

marandaneto: there's still plenty of `SentryLevel? level = SentryLevel.info`
they all should be `SentryLevel? level` only, the reason is here:
https://github.com/getsentry/sentry-dart/blob/main/dart/lib/src/sentry_client.dart#L167

if we decide to change the level, we don't need to change in ~6 methods, but only one place.

ueman: Yeah, I just pushed a commit which removes them.

dart/lib/src/protocol/message.dart
```diff
@@ -33,17 +33,17 @@ class Message {
...
     }
 
     return json;
   }
 
   Message copyWith({
-    String formatted,
-    String template,
-    List<dynamic> params,
+    String? formatted,
```
marandaneto, line 42: same as https://github.com/getsentry/sentry-dart/pull/298/files#r580839287

dart/test/mocks/mock_hub.dart
marandaneto: that's true :) my bad

flutter/pubspec.yaml
```diff
@@ -13,16 +13,16 @@ dependencies:
     sdk: flutter
   flutter_web_plugins:
     sdk: flutter
-  sentry: ">=4.0.0 <5.0.0"
-  package_info: ^0.4.0
+  # this should point to pub
+  sentry: ^4.0.5
```
marandaneto, line 17: would not be better to keep between compatible versions?

dart/lib/src/protocol/message.dart
bruno-garcia: I think captureMessage shouldn't take null

The code can gracefully handle a null (programming error) but the annotation should be that a message is required.

For all the public API when we need to value to do something meaningful

ueman:
We're now at a point where the discussion is more about the API and if it should take nulls or not than the underlying null safety changes. Does it make sense to merge this and do these changes yourself? I'm not quite sure I am the right person to decide on it 😅

bruno-garcia:
Sounds good to me but it's up to @marandaneto. 
I didn't review the PR though just added a comment.

dart/lib/src/protocol/message.dart
marandaneto: this would give a lot of boilerplate or force an error on the users code, think about?

final str = getthisfromApi();
// str is nullable
Sentry.captureMessage(str!); // could cause an error on app's code.

// doing this would be annoying
if (str != null) {
  Sentry.captureMessage(str!);
}

marandaneto: as a `Message` is created only after capturing, hence the field `formatted` should not be nullable

marandaneto:
there are only 3 comments open 2 of them are the same, about Message.formatted that should not be nullable.
the other one is a question, after that, LGTM

flutter/pubspec.yaml
ueman: It has to be changed anyway before you release this even if I change it to `">=4.0.0 <5.0.0"`. As Sentry should probably be `5.0.0-prerelease.0` and at that time there is no sentry with a `5.0.0` release. There's also no reason to release a `sentry_flutter` because it's not null safe yet.

ueman: Or am I overlooking something?

marandaneto: we can discuss versioning in another PR, lets merge it, my question was more about, why changing from `>= ... <` to `^...`

marandaneto (approved):
@ueman thanks a lot for this, a lot of back and forth but its also a change in the whole code base so yeah.
happy to merge :)

ueman:
Also thanks alot for the review.

PR #308, opened 2021-02-22
Fix: captureMessage defaults SentryLevel to info

### LINKED ISSUE #305: `captureMessage` defaults to level `error`.
I'm using `sentry.dart` 4.0.5 on a server app. I tried a `captureMessage` but the `level comes up as `error`:

`Sentry.captureMessage('test from pub client with crumbs');`

As a result the last crumb, which becomes the message itself, also has level `error`:

image

[Sample on sentry.io](https://sentry.io/organizations/bruno-garcia/issues/2229790948/?project=1857674&query=is%3Aunresolved)

Also, note `was active: false`. Is this the correct behavior?

marandaneto:
## :scroll: Description
Fix: captureMessage defaults SentryLevel to info

Fix #305

## :pencil: Checklist

- [X] All tests passing
- [X] No breaking changes

### DIFF
--- a/dart/test/sentry_client_test.dart
+++ b/dart/test/sentry_client_test.dart
@@ -118,8 +118,7 @@ void main() {
       expect(capturedEvent.message.formatted, 'simple message 1');
       expect(capturedEvent.message.template, 'simple message %d');
       expect(capturedEvent.message.params, [1]);
-
-      expect(capturedEvent.stackTrace is SentryStackTrace, true);
+      expect(capturedEvent.level, SentryLevel.error);
     });
 
     test('capture message defaults to info level', () async {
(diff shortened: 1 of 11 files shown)


dart/test/sentry_client_test.dart
```diff
@@ -122,6 +122,19 @@ void main() {
       expect(capturedEvent.stackTrace is SentryStackTrace, true);
     });
 
+    test('capture message defaults to info level', () async {
```
ueman, line 124: Does it make sense to create a test like this for the static Sentry methods, too? To make sure that it gets passed down correctly. Or does a test like that already exist?

marandaneto: `should capture message` test on SentryClient pass `SentryLevel.error` and asserts it, but yeah one coming from the Sentry static class makes sense, will do it, thanks

ueman (approved):

Issue #304, opened 2021-02-21
beforeSend filtering based on exception type does not work 


mnordine:
_Platform:_
- [ ]  Dart
- [x] Dart Web
- [ ]  Flutter Android or iOS
- [ ]  Flutter Web

_Platform installed with:_
- [x] pub.dev
- [ ] GitHub

The version of the SDK (See pubspec.lock):
**4.0.5**

---
I have the following issue:

The following example straight from your documentation at https://docs.sentry.io/platforms/dart/configuration/filtering/ does not work:

```dart
import 'package:sentry/sentry.dart';

SentryEvent beforeSend(SentryEvent event, {dynamic hint}) {
  if (event.exception is DatabaseException) {
    event = event.copyWith(fingerprint: ['database-connection-error']);
  }
  return event;
}

Future<void> main() async {
  await Sentry.init((options) => options.beforeSend = beforeSend);
}
```

First of all, `event.exception is DatabaseException` will not work on any Dart platform, since `event.exception` is a `SentryException`. So this line will always be `false`. 

I have adapted the above code so we won't report any of our `ServerException`s.:

```dart
SentryEvent beforeSend(SentryEvent event, {hint}) {
  if (event.exception is ServerException) return null;
  return event;
}
```

As stated above, this will never work, on any Dart platform. You may say, "oh, `SentryException` has a `type` field, just compare that, like so:

```dart
SentryEvent beforeSend(SentryEvent event, {hint}) {
  if (event.exception.type == 'ServerException') return null;
  return event;
}
```

This will work on Dart VM, and perhaps AOT, but crucially, **will not** work using `dart2js`. It will minify the type, so it could be something like `minified:ht`, and the comparison will always return `false`. Perhaps the original exception should be stored on the `SentryException`? Then, this would work everywhere:

```dart
SentryEvent beforeSend(SentryEvent event, {hint}) {
  if (event.exception.original is ServerException) return null;
  return event;
}
```

marandaneto:
@mnordine thanks for reporting, indeed it should be:

`if (event.throwable is ServerException)`

as `event.throwable` is the original exception, this code snippet is about using `fingerprint` and not dropping the event.
to drop the event, you need to return null, this is said in this sentence: https://docs.sentry.io/platforms/dart/configuration/filtering/#using-beforesend

> so you can use that to modify the event’s data or drop it completely (by returning null) based on custom logic and the data available on the event

we'll fix the docs :)

mnordine:
[...]

PR #311, opened 2021-02-22
Fix: SentryEvent.throwable returns the unwrapped throwable instead of the throwableMechanism

marandaneto:
## :scroll: Description
Fix: SentryEvent.throwable returns the unwrapped throwable instead of the throwableMechanism

#304

## :pencil: Checklist

- [X] All tests passing
- [X] No breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # vNext
 
 * Fix: captureMessage defaults SentryLevel to info
+* Fix: Expose SentryEvent.originThrowable to get access to the origin Throwable
 
 # 4.0.5
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # vNext
 
 * Fix: captureMessage defaults SentryLevel to info
-* Fix: Expose SentryEvent.originThrowable to get access to the origin Throwable
+* Fix: SentryEvent.throwable returns the unwrapped throwable instead of the throwableMechanism
 
 # 4.0.5
 
--- a/dart/lib/src/protocol/sentry_event.dart
+++ b/dart/lib/src/protocol/sentry_event.dart
@@ -25,7 +25,7 @@ class SentryEvent {
     this.environment,
     this.message,
     this.transaction,
-    this.throwable,
+    dynamic throwable,
     this.stackTrace,
     this.exception,
     this.level,
@@ -41,7 +41,8 @@ class SentryEvent {
         tags = tags != null ? Map.from(tags) : null,
         extra = extra != null ? Map.from(extra) : null,
         fingerprint = fingerprint != null ? List.from(fingerprint) : null,
-        breadcrumbs = breadcrumbs != null ? List.from(breadcrumbs) : null;
+        breadcrumbs = breadcrumbs != null ? List.from(breadcrumbs) : null,
+        _throwable = throwable;
 
   /// Refers to the default fingerprinting algorithm.
   ///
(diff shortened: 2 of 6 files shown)


bruno-garcia (commented):
@marandaneto not possible to take the same approach as we did in Java and on the getter unwrap our throwable?

having two fields to get a throwable isn't ideal

marandaneto:
docs https://github.com/getsentry/sentry-docs/pull/3138

marandaneto:
> @marandaneto not possible to take the same approach as we did in Java and on the getter unwrap our throwable?
> 
> having two fields to get a throwable isn't ideal

I did exactly the same approach :) our internal SDK needs access to the mechanism but clients would need the unwrapped version, that's why 2 getters

bruno-garcia:
> > @marandaneto not possible to take the same approach as we did in Java and on the getter unwrap our throwable?
> > having two fields to get a throwable isn't ideal
> 
> I did exactly the same approach :) our internal SDK needs access to the mechanism but clients would need the unwrapped version, that's why 2 getters

Shouldn't we reconsider this?
For our internal API we can use an internal API instead of adding two APIs one of which is useless to users?

marandaneto:
we'd break literally every `beforeSend` and that was the reason adding the `originThrowable` at that time, why would we change now?

it's gonna be public anyway because of being used by another packages

bruno-garcia:
> we'd break literally every `beforeSend` and that was the reason adding the `originThrowable` at that time, why would we change now?
> 
> it's gonna be public anyway because of being used by another packages

What would break, I don't follow?
We should no longer returns our wrapping throwable, but return the original one.

For the internal API we can have another getter (package internal) that we can use to get the wrapper exception

How would we even document:

```
getThrowable - the throwable (maybe wrapped by some Sentry throwable so we can add more data)
originThrowable - The throwable that was captured (maybe the same as getThorwable, sometimes)

Please use only originThrowable.
```

marandaneto:
> > we'd break literally every `beforeSend` and that was the reason adding the `originThrowable` at that time, why would we change now?
> > it's gonna be public anyway because of being used by another packages
> 
> What would break, I don't follow?
> We should no longer returns our wrapping throwable, but return the original one.
> 
> For the internal API we can have another getter (package internal) that we can use to get the wrapper exception
> 
> How would we even document:
> 
> ```
> getThrowable - the throwable (maybe wrapped by some Sentry throwable so we can add more data)
> originThrowable - The throwable that was captured (maybe the same as getThorwable, sometimes)
> 
> Please use only originThrowable.
> ```

https://github.com/getsentry/sentry-java/issues/1201
during stand up we'll have agreed to use `getOriginThrowable`, the points you've mentioned right now were already on the table before, we just avoided a breaking change.

bruno-garcia:
Not sure it's worth increasing the public API surface in the expense of potential more confusion just to avoid a breaking change without at least putting some thought into it.

Could we please consider simply returning the unwrapped Throwable and having an internal API instead for the internal behavior? Should we be using `hint`s like the other SDKs to add the mechanism information instead? Maybe we should do things differently here altogether, I added this mechanism stuff back on Java 1.7, there might be a better way to go about this now.

We can bump a major if that's needed.

Regarding the name: why 'originThrowable"? Is it the "original throwable"?

What about "sourceThrowable" I'm not sure any of these names are much better given we already have `getThrowable()`.

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # vNext
 
 * Fix: captureMessage defaults SentryLevel to info
+* Fix: Expose SentryEvent.originThrowable to get access to the origin Throwable
```
```suggestion
* Fix: SentryEvent.throwable returns the unwrapped throwable instead of the throwableMechanism
```

marandaneto:
> Not sure it's worth increasing the public API surface in the expense of potential more confusion just to avoid a breaking change without at least putting some thought into it.
> 
> Could we please consider simply returning the unwrapped Throwable and having an internal API instead for the internal behavior? Should we be using `hint`s like the other SDKs to add the mechanism information instead? Maybe we should do things differently here altogether, I added this mechanism stuff back on Java 1.7, there might be a better way to go about this now.
> 
> We can bump a major if that's needed.
> 
> Regarding the name: why 'originThrowable"? Is it the "original throwable"?
> 
> What about "sourceThrowable" I'm not sure any of these names are much better given we already have `getThrowable()`.

let's keep the discussion in a single PR, right now it's spread across a few others.
I've changed this PR to unwrap the `Throwable` for `Event.throwable` and added a `Event.throwableMechanism` that returns the wrapped `ThrowableMechanism`, as this was not released yet, it was in time to fix it, on Java, we can discuss later as it'd be a breaking change.

marandaneto:
btw looks like JS use hints for this https://docs.sentry.io/platforms/javascript/configuration/filtering/#using-hints similar to https://forum.sentry.io/t/capture-sentry-error/12811

ueman (approved):

bruno-garcia:
> btw looks like JS use hints for this [docs.sentry.io/platforms/javascript/configuration/filtering/#using-hints](https://docs.sentry.io/platforms/javascript/configuration/filtering/#using-hints) similar to [forum.sentry.io/t/capture-sentry-error/12811](https://forum.sentry.io/t/capture-sentry-error/12811)

Yes on other SDKs is all about hints.

bruno-garcia (commented):
Nice! LGTM

PR #316, opened 2021-02-23
chore: enable package analysis

bruno-garcia:
#skip-changelog

### DIFF
--- a/.github/workflows/flutter.yml
+++ b/.github/workflows/flutter.yml
@@ -110,7 +110,7 @@ jobs:
           TOTAL_MAX: ${{ steps.analysis.outputs.total_max }}
         run: |
           PERCENTAGE=$(( $TOTAL * 100 / $TOTAL_MAX ))
-          if (( $PERCENTAGE < 100 ))
+          if (( $PERCENTAGE < 90 ))
           then
             echo Score too low!
             exit 1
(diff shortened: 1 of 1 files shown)


bruno-garcia:
Odd that it says we got 90 points, but on `pub.dev` says we have 100:

image

.github/workflows/flutter.yml
```diff
@@ -95,29 +95,26 @@ jobs:
...
+        with:
+          githubToken: ${{ secrets.GITHUB_TOKEN }}
+          relativePath: flutter/
+      - name: Check scores
+        env:
+          TOTAL: ${{ steps.analysis.outputs.total }}
+          TOTAL_MAX: ${{ steps.analysis.outputs.total_max }}
+        run: |
+          PERCENTAGE=$(( $TOTAL * 100 / $TOTAL_MAX ))
+          if (( $PERCENTAGE < 100 ))
```
marandaneto:
```suggestion
          if (( $PERCENTAGE < 90 ))
```

marandaneto:
> Odd that it says we got 90 points, but on `pub.dev` says we have 100:
> 
> image

it already happened a few times, it depends on the flutter channel, pana version and also your dependency versions, the combination of all of them affects the scoring, I'll just lower it down to 90 so we make CI happy

Issue #317, opened 2021-02-24
Setting enableNativeCrashHandling to false does not disable native reports


Dave0921:
_Platform:_
- [ ]  Dart
- [x]  Flutter Android or iOS
- [ ]  Flutter Web

_IDE:_
- [ ]  VSCode
- [x]  IntelliJ/AS
- [ ]  XCode
- [ ]  Other, which one?

_split-debug-info and obfuscate (Flutter Android or iOS) or CanvasKit (Flutter Web):_
- [ ]  Enabled
- [x]  Disabled

_Platform installed with:_
- [x] pub.dev
- [ ] GitHub

Output of the command `flutter doctor -v` below:

```
[✓] Flutter (Channel stable, 1.22.5, on Mac OS X 10.15.6 19G2021 darwin-x64, locale en-CA)
    • Flutter version 1.22.5 at -----------------
    • Framework revision 7891006299 (3 months ago), 2020-12-10 11:54:40 -0800
    • Engine revision ae90085a84
    • Dart version 2.10.4

 
[✓] Android toolchain - develop for Android devices (Android SDK version 30.0.2)
    • Android SDK at -----------------
    • Platform android-30, build-tools 30.0.2
    • Java binary at: /Applications/Android Studio.app/Contents/jre/jdk/Contents/Home/bin/java
    • Java version OpenJDK Runtime Environment (build 1.8.0_242-release-1644-b3-6222593)
    • All Android licenses accepted.

[✓] Xcode - develop for iOS and macOS (Xcode 12.3)
    • Xcode at /Applications/Xcode.app/Contents/Developer
    • Xcode 12.3, Build version 12C33
    • CocoaPods version 1.10.0

[✓] Android Studio (version 4.1)
    • Android Studio at /Applications/Android Studio.app/Contents
    • Flutter plugin installed
    • Dart plugin version 201.9245
    • Java version OpenJDK Runtime Environment (build 1.8.0_242-release-1644-b3-6222593)

[✓] VS Code (version 1.53.0)
    • VS Code at /Applications/Visual Studio Code.app/Contents
    • Flutter extension version 3.19.0

[✓] Connected device (1 available)
    • iPhone 12 (mobile) • BEFBD9BF-8751-497F-A1F5-7465631F3641 • ios • com.apple.CoreSimulator.SimRuntime.iOS-14-3
      (simulator)

• No issues found!
```

The version of the SDK (See pubspec.lock):
**4.0.5**

---
I have the following issue:

[...]

PR #323, opened 2021-02-26
Feat: Support enableNativeCrashHandling on iOS

marandaneto:
## :scroll: Description
Feat: Support enableNativeCrashHandling on iOS

#317

## :pencil: Checklist

- [X] All tests passing
- [X] No breaking changes

### DIFF
--- a/flutter/example/lib/main.dart
+++ b/flutter/example/lib/main.dart
@@ -18,6 +18,7 @@ Future<void> main() async {
       // options.useFlutterBreadcrumbTracking();
       // use breadcrumb tracking of platform Sentry SDKs
       options.useNativeBreadcrumbTracking();
+      options.enableNativeCrashHandling = false;
     },
     // Init your App.
     appRunner: () => runApp(MyApp()),
--- a/flutter/example/lib/main.dart
+++ b/flutter/example/lib/main.dart
@@ -18,7 +18,6 @@ Future<void> main() async {
       // options.useFlutterBreadcrumbTracking();
       // use breadcrumb tracking of platform Sentry SDKs
       options.useNativeBreadcrumbTracking();
-      options.enableNativeCrashHandling = false;
     },
     // Init your App.
     appRunner: () => runApp(MyApp()),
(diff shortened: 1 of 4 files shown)


flutter/example/lib/main.dart
```diff
@@ -18,6 +18,7 @@ Future<void> main() async {
       // options.useFlutterBreadcrumbTracking();
       // use breadcrumb tracking of platform Sentry SDKs
       options.useNativeBreadcrumbTracking();
+      options.enableNativeCrashHandling = false;
```
marandaneto: TODO: remove it

flutter/lib/src/sentry_flutter_options.dart
```diff
@@ -22,7 +22,12 @@ class SentryFlutterOptions extends SentryOptions {
   /// Enable or disable the Crash handling on the Native SDKs, e.g.,
   /// UncaughtExceptionHandler and [anrEnabled] for Android.
   ///
-  /// Available only for Android.
+  /// SentryCrashIntegration (KSCrash) for iOS, turning this feature off on iOS
+  /// has a side effect which is missing the Device's context, e.g.,
+  /// App, Device and Operation System.
```
marandaneto, line 27: @philipphofmann is there a workaround for this or not possible right now? could not find it

flutter/lib/src/sentry_flutter_options.dart
```diff
@@ -22,7 +22,12 @@ class SentryFlutterOptions extends SentryOptions {
...
+  ///
+  /// Disabling this feature would affect the [enableAutoSessionTracking]
```
philipphofmann: `l`:
```suggestion
  /// Disabling this feature affects the [enableAutoSessionTracking]
```

philipphofmann (approved):
Nice, LGTM.

flutter/lib/src/sentry_flutter_options.dart
philipphofmann: Not an easy one. We could extract some of the device context from SentryCrashIntegration, because it doesn't depend on SentryCrash. For most of the data, we need SentryCrash to be running. We would need to figure out if we can initialize SentryCrash without all the signal handlers and crash monitors. I was not able to figure this out after a short research.

marandaneto: thanks, let's go this way for now and if this becomes an issue again, we see if we get this prioritized.

PR #340, opened 2021-03-07
Publish to TestFlight

bruno-garcia:
Still need to figure out some things like today it's heavily based on fastlane including building it.
We need to adjust to build with `flutter build ios` instead

### DIFF
--- a/.github/workflows/testflight.yml
+++ b/.github/workflows/testflight.yml
@@ -1,0 +1,85 @@
+name: Upload to TestFlight
+on:
+  push:
+    branches:
+      - main
+      - feat/sample-test-flight
+
+env:
+  FASTLANE_KEYCHAIN_NAME: fastlane.keychain
+
+jobs:
+
+  upload_to_testflight:
+    name: Build and Upload Sample App to TestFlight
+    runs-on: macos-latest
+    working-directory: ./flutter/example
+
+    steps:
+      - uses: actions/checkout@v2
+      - run: sudo xcode-select -s /Applications/Xcode_12.4.app/Contents/Developer
+
+      # The Github Actions hang when Xcode is trying to sign the app because
+      # a modal window pops up to access the signing key's keychain. This modal
+      # lets the action hang. A workaround is to create a local keychain and
+      # store the signing keys in there.
+      # https://github.com/actions/virtual-environments/issues/1820#issuecomment-719549887
+      
+      # Sometimes the keychain already exists
+      - name: Delete local keychain
+        continue-on-error: true
+        run: security delete-keychain $FASTLANE_KEYCHAIN_NAME
+          
+      - name: Create local keychain
+        run: |
+          security create-keychain -p ${{ secrets.FASTLANE_KEYCHAIN_PASSWORD }} $FASTLANE_KEYCHAIN_NAME
+          security set-keychain-settings -lut 21600 $FASTLANE_KEYCHAIN_NAME
+          security unlock-keychain -p ${{ secrets.FASTLANE_KEYCHAIN_PASSWORD }} $FASTLANE_KEYCHAIN_NAME
+          security list-keychain -d user -s $FASTLANE_KEYCHAIN_NAME
+  
+      - uses: actions/setup-ruby@v1
+        with:
+          ruby-version: '2.7'
+
+      - name: Install Sentry CLI
+        run: brew install getsentry/tools/sentry-cli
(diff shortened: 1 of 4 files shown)


bruno-garcia:
/cc @philipphofmann this was my experiment we can discuss tomorrow.
Would also add to https://github.com/getsentry/sentry-cordova https://github.com/getsentry/sentry-xamarin and eventually to https://github.com/getsentry/sentry-capacitor

.github/workflows/testflight.yml
```diff
@@ -0,0 +1,85 @@
+name: Upload to TestFlight
+on:
+  push:
+    branches:
+      - main
+      - feat/sample-test-flight
```
philipphofmann, line 6: `l`: Don't forget to remove before merging 😉

.github/workflows/testflight.yml
```diff
@@ -0,0 +1,85 @@
...
+      - name: Delete local keychain
+        continue-on-error: true
+        run: security delete-keychain $FASTLANE_KEYCHAIN_NAME
+          
+      - name: Create local keychain
+        run: |
+          security create-keychain -p ${{ secrets.FASTLANE_KEYCHAIN_PASSWORD }} $FASTLANE_KEYCHAIN_NAME
+          security set-keychain-settings -lut 21600 $FASTLANE_KEYCHAIN_NAME
+          security unlock-keychain -p ${{ secrets.FASTLANE_KEYCHAIN_PASSWORD }} $FASTLANE_KEYCHAIN_NAME
+          security list-keychain -d user -s $FASTLANE_KEYCHAIN_NAME
```
philipphofmann, line 38: `l`: I moved this to Fastlane in Cocoa to make it easier for local debugging. See https://github.com/getsentry/sentry-cocoa/blob/ab0326835655fbe7cbfec32a80dd1efb9876f128/fastlane/Fastfile#L46-L59

flutter/example/fastlane/Fastfile
```diff
@@ -0,0 +1,94 @@
+default_platform(:ios)
+
+# Download dSYMs does not work with app_store_connect_api_key yet. Therefore
+# We use username and password to authenticate. This feature is blocked by the
+# App Store Connect API team, see https://github.com/fastlane/fastlane/discussions/17485#discussioncomment-125619
+# When this is fixed we can remove SPACESHIP_SKIP_2FA_UPGRADE, FASTLANE_USER and FASTLANE_PASSWORD
+ENV["SPACESHIP_SKIP_2FA_UPGRADE"] = "1"
```
philipphofmann, line 7: This is not going to work anymore sadly. We need to set `FASTLANE_SESSION`. You can generate one with [spaceauth](https://docs.fastlane.tools/best-practices/continuous-integration/#spaceauth). Checkout our internal notion page for Cocoa for more details.

flutter/example/fastlane/Fastfile
```diff
@@ -0,0 +1,94 @@
...
+  end
+
+  desc "Upload Release-iphoneos to TestFlight and symbols to Sentry"
+  lane :ios_swift_to_testflight do
+    sync_code_signing(
+      type: "appstore", 
+      keychain_name: ENV["FASTLANE_KEYCHAIN_NAME"],
+      keychain_password: ENV["FASTLANE_KEYCHAIN_PASSWORD"],
+      readonly: true
+    )
```
philipphofmann, line 57: This is going to fail because we pass readonly true. I think you have to run match locally to generate the certificates and put them in our private git repo first, see https://docs.fastlane.tools/actions/match/#run

philipphofmann (commented):
Awesome that we add this now to more projects. This is going to make our lives easier. 🥇

flutter/example/fastlane/Fastfile
bruno-garcia: This is a good reminder that this setup should be in a state where we can easily maintain across repos. gittree is one option. other thoughts?

It's a lot of different bits and pieces to copy paste across 5 or more repos

philipphofmann: I agree. As already discussed it would be nice to get it up and running here and then try to share the script. We could use `import_from_git` from Fastlane https://docs.fastlane.tools/actions/import_from_git/

bruno-garcia:
@philipphofmann I'll close this since a lot changed anyway. Will create an issue so we can prioritze later

PR #339, opened 2021-03-05
Upgrade dependencies (`http 0.13.0` etc) for Flutter 2 upgrade

zeusbaba:
## :scroll: Description
Upgrade dependencies (`http 0.13.0` etc) as it is widely used after Flutter 2 release!  

## :bulb: Motivation and Context
After Flutter2 release, `http 0.13.0` is the preferred version used by many developers. `Sentry` package dependencies required to be up2date to meet this demand. 

## :green_heart: How did you test it?
My apps which have the latest and greatest package versions began failing caused by the issue as described in #338 . The updates in this PR resolves it. 

## :pencil: Checklist

- [ x ] I reviewed submitted code
- [ x ] I added tests to verify changes
- [ x ] I updated the docs if needed
- [ x ] All tests passing
- [ x ] No breaking changes

## :crystal_ball: Next steps
Review and publish a new release after merge

PR #341, opened 2021-03-07
upgrade dependencies

vanfresh:
## :scroll: Description
Update dependencies


ThexXTURBOXx:
As already stated in #339 the null-safety update is in progress and if you need a temporary fix until it is released, use a `dependency_override` :)

marandaneto:
@vanfresh thanks for the contribution, but this is already migrated, see #247

Issue #314, opened 2021-02-23
Annoying plugin registrant warning in flutter web


gummicq:
_Platform:_
- [ ]  Dart
- [ ]  Flutter Android or iOS
- [x]  Flutter Web

_IDE:_
- [x]  VSCode
- [ ]  IntelliJ/AS
- [ ]  XCode
- [ ]  Other, which one?

_split-debug-info and obfuscate (Flutter Android or iOS) or CanvasKit (Flutter Web):_
- [ ]  Enabled
- [x]  Disabled

_Platform installed with:_
- [x] pub.dev
- [ ] GitHub

Output of the command `flutter doctor -v` below:

[✓] Flutter (Channel beta, 1.26.0-17.6.pre, on macOS 11.2 20D64 darwin-x64, locale en-GB)
    • Flutter version 1.26.0-17.6.pre at /Users/gummi/development/flutter
    • Framework revision a29104a69b (7 days ago), 2021-02-16 09:26:56 -0800
    • Engine revision 21fa8bb99e
    • Dart version 2.12.0 (build 2.12.0-259.12.beta)

[✓] Android toolchain - develop for Android devices (Android SDK version 30.0.1)
    • Android SDK at /Users/gummi/Library/Android/sdk
    • Platform android-30, build-tools 30.0.1
    • Java binary at: /Applications/Android Studio.app/Contents/jre/jdk/Contents/Home/bin/java
    • Java version OpenJDK Runtime Environment (build 1.8.0_242-release-1644-b3-6222593)
    • All Android licenses accepted.

[✓] Xcode - develop for iOS and macOS
    • Xcode at /Applications/Xcode.app/Contents/Developer
    • Xcode 12.4, Build version 12D4e
    • CocoaPods version 1.10.1

[✓] Chrome - develop for the web
    • Chrome at /Applications/Google Chrome.app/Contents/MacOS/Google Chrome

[✓] Android Studio (version 4.0)
    • Android Studio at /Applications/Android Studio.app/Contents
    • Flutter plugin version 47.1.2
    • Dart plugin version 193.7361
    • Java version OpenJDK Runtime Environment (build 1.8.0_242-release-1644-b3-6222593)

[✓] VS Code (version 1.53.2)
    • VS Code at /Applications/Visual Studio Code.app/Contents
    • Flutter extension version 3.19.0

[✓] Connected device (1 available)
    • Chrome (web) • chrome • web-javascript • Google Chrome 88.0.4324.182

• No issues found!

[...]

PR #357, opened 2021-03-13
Fix: Rename some protocol classes

ueman:
## :scroll: Description

This is a _breaking change_.

This PR renames a couple classes: 
- App -> SentryApp
- Browser -> SentryBrowser
- Device -> SentryDevice
- Message -> SentryMessage
- Request -> SentryRequest
- User -> SentryUser
- Gpu -> SentryGpu
- OperatingSystem -> SentryOperatingSystem
- Integration -> SentryIntegration

I wonder if we should rename the following classes too:
- Breadcrumb
- Contexts

The following classes are probably fine:
- DebugImage
- DebugMeta
- Dsn
- Mechanism
- SdkInfo
- SdkVersion 

## :bulb: Motivation and Context
See #250 

## :green_heart: How did you test it?
No functional changes, tests are still green

## :pencil: Checklist
- [x] I reviewed submitted code
- [ ] I added tests to verify changes
- [ ] I updated the docs if needed
- [x] All tests passing
- [ ] No breaking changes

## :crystal_ball: Next steps
Update documentation:
- https://github.com/getsentry/sentry-docs/pull/3244

PR #358, opened 2021-03-15
Fix: Plugin registrant warning

ueman:
## :scroll: Description
I moved the class `SentryFlutterWeb` from `src/sentry_flutter_web.dart` to `sentry_flutter_web.dart`.
I also enabled [this](https://dart-lang.github.io/linter/lints/implementation_imports.html) lint to prevent this kind of errors in
the future.

I'm not sure if this is considered a breaking change or not. Technically it is, because the imports changed I guess.

See #314

I've enabled the lint and it didn't show any warnings.

## :pencil: Checklist

- [x] All tests passing
- [ ] No breaking changes


flutter/analysis_options.yaml
```diff
@@ -24,3 +24,4 @@ linter:
   rules:
     prefer_relative_imports: true
     unnecessary_brace_in_string_interps: true
+    implementation_imports: true
```
marandaneto, line 27: currently, we're using `package:pedantic`, is there a more aggressive solution? often I see us enabling new rules that should be avoided in the 1st place :D (not for this PR though)

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 * Fix: Multiple FlutterError.onError calls in FlutterErrorIntegration (#345)
 * Fix: Pass hint to EventProcessors (#356)
 * Fix: EventProcessors were not dropping events when returning null (#353)
+* Fix: Plugin Registrant class moved to barrel file (#314)
```
marandaneto: by Danger, requires PRs id

marandaneto:
> I'm not sure if this is considered a breaking change or not. Technically it is, because the imports changed I guess.

theoretically, no as the import as code generated right, but technically yes, mmm, we could take the same approach as https://github.com/getsentry/sentry-dart/pull/357/files#r594101972
adding a break changes section and explaining what to do shortly.
we are in a pre-release stage as well, so it'd be fine.

flutter/analysis_options.yaml
ueman: As per https://dart-lang.github.io/linter/lints/ there are three official rule sets: pedantic, effective_dart and flutter. 
There's also https://pub.dev/packages/lint which describes itself as very strict.

marandaneto (approved):
LGTM

flutter/analysis_options.yaml
marandaneto: thanks, I'll have a look at this one (lint) later this week.

PR #278, opened 2021-01-22
Bump: Sentry-cocoa to 6.1.4

marandaneto:
## :scroll: Description
Bump: Sentry-cocoa to 6.1.4

## :bulb: Motivation and Context
#271

## :green_heart: How did you test it?
running on simulator

## :pencil: Checklist

- [X] I reviewed submitted code
- [ ] I added tests to verify changes
- [X] All tests passing
- [X] No breaking changes

PR #360, opened 2021-03-15
Feat: Send default pii

ueman:
## :scroll: Description
This PR enables the user to set `SentryOptions.sendDefaultPii`.
I tried to mimic the Java SDKs approach [here](https://github.com/getsentry/sentry-java/blob/d8efafea2a26eaa6f2d45e706c674254f27421e5/sentry/src/main/java/io/sentry/MainEventProcessor.java#L142-L150).

When I send an event with `sendDefaultPii == true` I get this as a result.
Bildschirmfoto 2021-03-15 um 11 46 19
I'm not quite sure why it has this `unknown` field. I would guess however, it's because the `User` adds null values to the JSON, see https://github.com/getsentry/sentry-dart/blob/00417e10f05f2ae6166be0ab5ae0ad10b2662424/dart/lib/src/protocol/user.dart#L58-L64

See #275 

I've written new tests. 
I also tested it manually with the flutter example app.

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps
https://github.com/getsentry/sentry-docs/pull/3243


CHANGELOG.md
```diff
@@ -4,7 +4,7 @@
 * Fix: Multiple FlutterError.onError calls in FlutterErrorIntegration (#345)
 * Fix: Pass hint to EventProcessors (#356)
 * Fix: EventProcessors were not dropping events when returning null (#353)
-
+* Feat: Send default Pii options
```
```suggestion
* Feat: Send default Pii options (#360)
```

marandaneto:
we should sync this option to `NativeSdkIntegration` class and Native SDKs (eg SentryFlutterPlugin.kt and SwiftSentryFlutterPlugin.swift)

dart/lib/src/event_processors/user_event_processor.dart
```diff
@@ -0,0 +1,31 @@
...
+const _defaultIpAddress = '{{auto}}';
+
+/// This adds an [EventProcessor] which modifies the users IP address according
+/// to [SentryOptions.sendDefaultPii].
+void addUserEventProcessor(SentryOptions options) {
+  options.addEventProcessor(
+      (event, {hint}) => userEventProcessor(options, event, hint: hint));
+}
+
+FutureOr<SentryEvent?> userEventProcessor(
```
marandaneto: some folks are against using event processors when not needed, I guess that would be a use case to not use it as the code is Pure Dart.

this could be done for example -> `SentryClient._prepareEvent`

Java does use an event processor so it did make sense for you to just mimic it, let's try to fix this approach from now on, when it's possible, obviously :)

ueman:
I've removed the EventProcessor and moved the code to SentryClient.
The setting is now also synced to native.

Example events:
Android: https://sentry.io/organizations/sentry-sdks/issues/2149575550/events/5c937f468d4045da9167bb5c2d33e93e/?project=5428562
iOS: https://sentry.io/organizations/sentry-sdks/issues/2149575550/events/25492f60983d422e88105ebd84ce16c5/?project=5428562

I've also fixed the User, so that the null fields aren't send.

To sync sendDefaultPii to iOS I had to bump Sentry-Cocoa to 6.2.1.

dart/lib/src/sentry_client.dart
```diff
@@ -80,9 +85,10 @@ class SentryClient {
       return _sentryId;
     }
 
-    if (_options.beforeSend != null) {
+    final beforeSend = _options.beforeSend;
```
ueman, line 88: This way the dart compiler can infer that beforeSend is not null and the `!` isn't needed anymore.

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 * Fix: Multiple FlutterError.onError calls in FlutterErrorIntegration (#345)
 * Fix: Pass hint to EventProcessors (#356)
 * Fix: EventProcessors were not dropping events when returning null (#353)
+* Feat: Send default Pii options (#360)
```
ueman: Should I add the Sentry Cocoa bump to the changelog?

flutter/ios/sentry_flutter.podspec
```diff
@@ -11,7 +11,7 @@ Sentry SDK for Flutter with support to native through sentry-cocoa.
   s.source           = { :git => "https://github.com/getsentry/sentry-dart.git",
                          :tag => s.version.to_s }
   s.source_files = 'Classes/**/*'
-  s.dependency 'Sentry', '~> 6.1.4'
+  s.dependency 'Sentry', '~> 6.2.1'
```
ueman, line 14: 6.1.4 didn't have `sendDefaultPii`

flutter/ios/Classes/SwiftSentryFlutterPlugin.swift
```diff
@@ -176,6 +176,11 @@ public class SwiftSentryFlutterPlugin: NSObject, FlutterPlugin {
         if let maxBreadcrumbs = arguments["maxBreadcrumbs"] as? UInt {
             options.maxBreadcrumbs = maxBreadcrumbs
         }
+        if let sendDefaultPii = arguments["sendDefaultPii"] as? Bool {
+            options.sendDefaultPii = sendDefaultPii
+        }
+
```
marandaneto: https://realm.github.io/SwiftLint/trailing_whitespace.html

CHANGELOG.md
```suggestion
* Feat: Send default PII options (#360)
* Bump: sentry-cocoa to v6.2.1
```

flutter/ios/sentry_flutter.podspec
marandaneto: update our samples too please

marandaneto: eg https://github.com/getsentry/sentry-dart/pull/278/files
or we remove the lockfile from the repo, so it updates every time, wdyt?

dart/lib/src/protocol/sentry_user.dart
```diff
@@ -56,11 +56,11 @@ class SentryUser {
...
-      'id': id,
-      'username': username,
-      'email': email,
-      'ip_address': ipAddress,
-      'extras': extras,
+      if (id != null) 'id': id,
+      if (username != null) 'username': username,
+      if (email != null) 'email': email,
+      if (ipAddress != null) 'ip_address': ipAddress,
+      if (extras != null) 'extras': extras,
```
marandaneto: yeah looks like `toJson` for the User protocol was not doing the checks.
I did a quick search and this was the only one.

marandaneto:
```suggestion
      if (id != null) 'id': id,
      if (username != null) 'username': username,
      if (email != null) 'email': email,
      if (ipAddress != null) 'ip_address': ipAddress,
      if (extras?.isNotEmpty ?? false) 'extras': extras,
```

marandaneto: lets also not add an empty map

flutter/ios/sentry_flutter.podspec
ueman: The Podfile.lock is already gitignored: https://github.com/getsentry/sentry-dart/blob/main/.gitignore
I would remove it. Seems easier to manage.

dart/lib/src/protocol/sentry_user.dart
ueman: ah good catch

dart/test/sentry_client_test.dart
```diff
@@ -396,6 +396,71 @@ void main() {
     });
   });
 
+  group('SentryClient : apply default pii', () {
+    test('sendDefaultPii is disabled', () async {
+      var options = SentryOptions(dsn: fakeDsn);
+      options.sendDefaultPii = false;
+      final transport = MockTransport();
+      options.transport = transport;
+      final client = SentryClient(options);
```
marandaneto: I'd prefer to use a Fixture pattern, so we can avoid all this code duplication within the tests, which might change only the value of `sendDefaultPii`, an example would be: `file_system_transport_test.dart` or the Java source code eg https://github.com/getsentry/sentry-java/blob/main/sentry/src/test/java/io/sentry/DiagnosticLoggerTest.kt

flutter/ios/sentry_flutter.podspec
marandaneto: sounds good to me, I guess I've removed a few other lock files and this one was missing, we just need to be sure that the `flutter run` works without the lock file with a clean repo (and generates a new one automatically)

ueman: It's already removed https://github.com/getsentry/sentry-dart/commit/b1a68173134c9e2335786c37590ff4aa64a9a546

dart/test/sentry_client_test.dart
ueman: Makes sense 👍

flutter/ios/sentry_flutter.podspec
ueman: It works with a clean repo. Just set it up from zero yesterday and I didn't have any problems.

marandaneto: @ueman but you cloned the repo that had the lock file, that's what I meant, but that's what CI does pretty much, so we are good

dart/test/sentry_client_test.dart
```diff
@@ -551,3 +602,17 @@ SentryEvent beforeSendCallback(SentryEvent event, {dynamic hint}) {
...
+    var options = SentryOptions(dsn: fakeDsn);
+    options.sendDefaultPii = sendDefaultPii;
+    transport = MockTransport();
+    options.transport = transport;
+    client = SentryClient(options);
+  }
+
+  late MockTransport transport;
+  late SentryClient client;
+}
```
marandaneto:
```suggestion
class Fixture {
  
  SentryClient getSut(bool sendDefaultPii = true) {
    var options = SentryOptions(dsn: fakeDsn);
    transport = MockTransport();
    options.transport = transport;
    return SentryClient(options);
  }

  late MockTransport transport;
  late SentryClient client;
}
```

marandaneto: thats pseudocode, but just to give a hint on how it could be, having a generic name helps for copypasta to other tests, also, getSut() creates a new instance of SentryClient so on each tests, you have a clean state, you don't actually need to create a Fixture per test, but a Sut per test

ueman: Okay, I hope I got it right now :D

marandaneto (approved):
LGTM

dart/test/sentry_client_test.dart
```diff
@@ -551,3 +611,13 @@ SentryEvent beforeSendCallback(SentryEvent event, {dynamic hint}) {
 SentryEvent? eventProcessorDropEvent(SentryEvent event, {dynamic hint}) {
   return null;
 }
+
+class Fixture {
+  /// Test Fixture for tests with [SentryOptions.sendDefaultPii]
+  SentryClient getSut(bool sendDefaultPii, Transport transport) {
```
marandaneto, line 617: It's approved already so feel free to ignore :)
but `Transport transport` could still be a global field of Fixture and accessed via `fixture.transport` as it does nothing but only mocked.

ueman: Yeah, I wondered what would be the better way of doing that. Let's keep it as is for now.

Issue #306, opened 2021-02-22
Read `SENTRY_RELEASE` and other env vars in the `dart` package


bruno-garcia:
Discovering release via env var is also useful in the core `dart` package to be used by Dart VM projects.

marandaneto:
@ueman this issue is about https://github.com/getsentry/sentry-dart/blob/fecf4599fa2eb97d03fc4ef9fee765f3ecb23278/flutter/lib/src/default_integrations.dart#L362-L369
ideally, for non Android/iOS, this should work for the Dart package too, maybe moving this around

https://github.com/getsentry/sentry-dart/blob/main/dart/lib/src/sentry.dart#L64-L72

ueman:
Thanks, I already figured that out, though.

One question I have is the order of precedence.
Is it `environment variables` over `SentryOptions` over `values read from package_info or device`?

marandaneto:
see here https://docs.sentry.io/platforms/dart/configuration/options/

it depends on the property tbh, eg
options.dsn has precedence over SENTRY_DSN

SENTRY_ENVIRONMENT has precedence over options.environment

package_info has precedence over SENTRY_RELEASE, SENTRY_RELEASE has precedence over options.release (this is tricky as package_info is an integration and executed only later)

`dist` isn't documented but I guess it'd be similar to `environment`

PR #375, opened 2021-03-22
Fix: Read all environment variables in sentry 

ueman:
## :scroll: Description

Environment variables are now read in sentry.
This also includes https://github.com/getsentry/sentry-dart/pull/376

https://github.com/getsentry/sentry-dart/issues/306

I've written new tests.

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 * Feat: Send default PII options (#360)
 * Bump: sentry-cocoa to v6.2.1 (#360)
 * Fix: missing event.origin and event.environment tags for non-native events on iOS (#369)
+* Fix: Set `SentryOptions.debug` in `sentry` (#376)
 * Fix: Read all environment variables in `sentry` (#375)
 
 ## Breaking Changes:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 * Feat: Send default PII options (#360)
 * Bump: sentry-cocoa to v6.2.1 (#360)
 * Fix: missing event.origin and event.environment tags for non-native events on iOS (#369)
-* Fix: Read all environment variables in sentry_dart
+* Fix: Read all environment variables in sentry_dart (#375)
 
 ## Breaking Changes:
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 * Feat: Send default PII options (#360)
 * Bump: sentry-cocoa to v6.2.1 (#360)
 * Fix: missing event.origin and event.environment tags for non-native events on iOS (#369)
-* Fix: Read all environment variables in sentry_dart (#375)
+* Fix: Read all environment variables in `sentry` (#375)
 
 ## Breaking Changes:
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 * Feat: Send default PII options (#360)
 * Bump: sentry-cocoa to v6.2.1 (#360)
 * Fix: missing event.origin and event.environment tags for non-native events on iOS (#369)
+* Fix: Read all environment variables in sentry_dart
 
 ## Breaking Changes:
 
--- a/dart/lib/src/environment_variables.dart
+++ b/dart/lib/src/environment_variables.dart
@@ -1,3 +1,5 @@
+import 'platform_checker.dart';
+
 /// Reads environment variables from the system.
 /// In an Flutter environment these can be set via
 /// `flutter build --dart-define=VARIABLE_NAME=VARIABLE_VALUE`.
@@ -30,4 +32,19 @@ class EnvironmentVariables {
   String? get dist => const bool.hasEnvironment(_sentryDist)
       ? const String.fromEnvironment(_sentryDist)
       : null;
+
+  /// Returns an environment based on the compilation mode of Dart or Flutter.
+  /// This can be set as [SentryOptions.environment]
+  String environmentForMode(PlatformChecker checker) {
+    // We infer the enviroment based on the release/non-release and profile
+    // constants.
+
+    if (checker.isReleaseMode()) {
+      return 'production';
+    }
+    if (checker.isProfileMode()) {
+      return 'profile';
+    }
+    return 'debug';
+  }
 }
--- a/dart/lib/src/environment_variables.dart
+++ b/dart/lib/src/environment_variables.dart
@@ -4,10 +4,6 @@ import 'sentry_options.dart';
 
 /// This method reads available environment variables and uses them
 /// accordingly.
-///
-/// Please note, that environment variables are overriden by the values of
-/// [SentryOptions].
-///
 /// To see which environment variables are available, see [EnvironmentVariables]
 ///
 /// The precendence of these options are tricky,
--- a/dart/lib/src/environment_variables.dart
+++ b/dart/lib/src/environment_variables.dart
@@ -41,5 +41,7 @@ class EnvironmentVariables {
   String? get dist => _readString('SENTRY_DIST');
 
   String? _readString(String key) =>
-      bool.hasEnvironment(key) ? String.fromEnvironment(key) : null;
+      bool.fromEnvironment(key, defaultValue: false)
+          ? String.fromEnvironment(key)
+          : null;
 }
--- a/dart/lib/src/environment_variables.dart
+++ b/dart/lib/src/environment_variables.dart
@@ -1,0 +1,49 @@
+import 'package:meta/meta.dart';
+
+import 'sentry_options.dart';
+
+/// This method reads available environment variables and uses them
+/// accordingly.
+///
+/// Please note, that environment variables are overriden by the values of
+/// [SentryOptions].
+///
+/// To see which environment variables are available, see [EnvironmentVariables]
+///
+/// The precendence of these options are tricky,
+/// see https://docs.sentry.io/platforms/dart/configuration/options/
+/// and https://github.com/getsentry/sentry-dart/issues/306
+@internal
+void setEnvironmentVariables(SentryOptions options, EnvironmentVariables vars) {
+  // options has precendence over vars
+  options.dsn = options.dsn ?? vars.dsn;
+
+  // vars has precedence over options
+  options.environment = vars.environment ?? options.environment;
+  options.release = vars.release ?? options.release;
+  options.dist = vars.dist ?? options.dist;
+}
+
+/// Reads environment variables from the system.
+/// In an Flutter environment these can be set via
+/// `flutter build --dart-define=VARIABLE_NAME=VARIABLE_VALUE`.
+class EnvironmentVariables {
+  /// `SENTRY_ENVIRONMENT`
+  /// See [SentryOptions.environment]
+  String? get environment => _readString('SENTRY_ENVIRONMENT');
+
+  /// `SENTRY_DSN`
+  /// See [SentryOptions.dsn]
+  String? get dsn => _readString('SENTRY_DSN');
+
+  // `SENTRY_RELEASE`
+  /// See [SentryOptions.release]
+  String? get release => _readString('SENTRY_RELEASE');
+
+  /// `SENTRY_DIST`
+  /// See [SentryOptions.dist]
+  String? get dist => _readString('SENTRY_DIST');
+
+  String? _readString(String key) =>
+      bool.hasEnvironment(key) ? String.fromEnvironment(key) : null;
+}
--- a/dart/lib/src/environment_variables.dart
+++ b/dart/lib/src/environment_variables.dart
@@ -1,30 +1,3 @@
-import 'package:meta/meta.dart';
-
-import 'sentry_options.dart';
-
-/// This method reads available environment variables and uses them
-/// accordingly.
-/// To see which environment variables are available, see [EnvironmentVariables]
-///
-/// The precendence of these options are tricky,
-/// see https://docs.sentry.io/platforms/dart/configuration/options/
-/// and https://github.com/getsentry/sentry-dart/issues/306
-@internal
-void setEnvironmentVariables(SentryOptions options, EnvironmentVariables vars) {
-  // options has precendence over vars
-  options.dsn = options.dsn ?? vars.dsn;
-
-  var environment = options.platformChecker.environment;
-  options.environment = options.environment ?? environment;
-
-  // vars has precedence over options
-  options.environment = vars.environment ?? options.environment;
-
-  // vars has precedence over options
-  options.release = vars.release ?? options.release;
-  options.dist = vars.dist ?? options.dist;
-}
-
 /// Reads environment variables from the system.
 /// In an Flutter environment these can be set via
 /// `flutter build --dart-define=VARIABLE_NAME=VARIABLE_VALUE`.
--- a/dart/lib/src/environment_variables.dart
+++ b/dart/lib/src/environment_variables.dart
@@ -14,8 +14,13 @@ void setEnvironmentVariables(SentryOptions options, EnvironmentVariables vars) {
   // options has precendence over vars
   options.dsn = options.dsn ?? vars.dsn;
 
+  var environment = options.platformChecker.environment;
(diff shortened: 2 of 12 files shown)


marandaneto:
@ueman is PR title a typo? `sentry_flutter`, we should read from `sentry` package, so it'd work for non mobile apps

ueman:
> @ueman is PR title a typo? `sentry_flutter`, we should read from `sentry` package, so it'd work for non mobile apps

Yep, that was a typo

dart/lib/src/environment_variables.dart
```diff
@@ -0,0 +1,45 @@
...
+
+/// This method reads available environment variables and uses them
+/// accordingly.
+/// To see which environment variables are available, see [EnvironmentVariables]
+///
+/// The precendence of these options are tricky,
+/// see https://docs.sentry.io/platforms/dart/configuration/options/
+/// and https://github.com/getsentry/sentry-dart/issues/306
+@internal
+void setEnvironmentVariables(SentryOptions options, EnvironmentVariables vars) {
```
marandaneto: what's the inspiration for being a non-instance method? could be just a private method on the Sentry static class

dart/lib/src/sentry.dart
```diff
@@ -53,23 +54,10 @@ class Sentry {
...
-        : environment;
-
-    // if the SENTRY_DSN is set, we read from it.
-    options.dsn = const bool.hasEnvironment('SENTRY_DSN')
-        ? const String.fromEnvironment('SENTRY_DSN')
-        : options.dsn;
+    var environment = options.platformChecker.environment;
+    options.environment = options.environment ?? environment;
+
+    setEnvironmentVariables(options, EnvironmentVariables());
```
marandaneto: yo uset `options.environment` outside of the method `setEnvironmentVariables`, ideally, we'd have a single method that does everything, pass `environment` as a param if its the case

dart/lib/src/sentry_options.dart
```diff
@@ -16,6 +16,9 @@ import 'platform_checker.dart';
 const defaultEnvironment = 'production';
 
 /// Sentry SDK options
+// TODO: Scope observers, enableScopeSync
+// TODO: shutdownTimeout, flushTimeoutMillis
+// https://api.dart.dev/stable/2.10.2/dart-io/HttpClient/close.html doesn't have a timeout param, we'd need to implement manually
```
marandaneto, line 18: this reads weird, should not be part of the class docs (its an internal comment), I'd move this above `/// Sentry SDK options` and have a blank line in between

dart/lib/src/platform_checker.dart
```diff
@@ -1,3 +1,5 @@
+import '../sentry.dart';
```
marandaneto: import could be a bit more specific here, instead everything

dart/lib/src/sentry.dart
ueman: Makes sense

flutter/lib/src/default_integrations.dart
```diff
@@ -350,6 +350,7 @@ class LoadReleaseIntegration extends Integration<SentryFlutterOptions> {
   @override
   FutureOr<void> call(Hub hub, SentryFlutterOptions options) async {
     try {
+      // For web we read the environment variables in sentry_dart
```
marandaneto: not only web anymore, but non-Android/iOS

dart/lib/src/environment_variables.dart
ueman: That way I can test it more easily. If it is a private method on `Sentry` I can't easily pass `MockEnvironmentVariables` to it. Well I could by adding it as parameter to `Sentry.init` but that doesn't seem that good.

What would you suggest instead?

ueman: Another solution is that `EnvironmentVariables` could be property on SentryOptions just like `PlatformChecker`

ueman:
I've opened an issue for failing Flutter beta builds here: https://github.com/getsentry/sentry-dart/issues/377

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 * Feat: Send default PII options (#360)
 * Bump: sentry-cocoa to v6.2.1 (#360)
 * Fix: missing event.origin and event.environment tags for non-native events on iOS (#369)
+* Fix: Read all environment variables in sentry_dart (#375)
```
```suggestion
* Fix: Read all environment variables in `sentry` (#375)
```

marandaneto: PRs title too

dart/lib/src/environment_variables.dart
marandaneto: > Another solution is that EnvironmentVariables could be property on SentryOptions just like PlatformChecker

I think this one makes more sense

flutter/lib/src/default_integrations.dart
```diff
@@ -350,6 +349,7 @@ class LoadReleaseIntegration extends Integration<SentryFlutterOptions> {
   @override
   FutureOr<void> call(Hub hub, SentryFlutterOptions options) async {
     try {
+      // For non Flutter apps, we read the environment variables in sentry_dart
```
marandaneto: this comment is off IMO, `kIsWeb` maybe be true for a non flutter app, I'd prefer `non Android or iOS` or something more specific

```suggestion
      // For non Flutter apps, we read the environment variables in sentry
```

dart/lib/src/environment_variables.dart
```diff
@@ -0,0 +1,60 @@
...
+
+  var environment = options.platformChecker.environment;
+  options.environment = options.environment ?? environment;
+
+  // vars has precedence over options
+  options.environment = vars.environment ?? options.environment;
+
+  // vars has precedence over options
+  options.release = vars.release ?? options.release;
+  options.dist = vars.dist ?? options.dist;
```
marandaneto: after reading the docs again I guess I've misunderstood it and we should fix it, please help me if I am wrong.

options should always have precedence over vars, the user always has the chance to overwrite any value so it must take precedence.

* use options.dsn and if not, use vars.dsn
* use options.environment and if not, use vars.environment and if not, use options.platformChecker.environment
* use options.release and if not, vars.release and if not, package_info
* use options.dist and if not, vars.dist

`OptionsConfiguration` always run last so the user always should have the chance to overwrite any value, the gotcha here is for `release` as the integration reads the package_info after that

sorry for the confusion.

flutter/lib/src/default_integrations.dart
```diff
@@ -358,15 +358,6 @@ class LoadReleaseIntegration extends Integration<SentryFlutterOptions> {
 
         options.release = release;
         options.dist = packageInfo.buildNumber;
```
```suggestion
        options.release = options.release ?? release;
        options.dist = options.dist ?? packageInfo.buildNumber;
```

I believe here we should check if there was already a value, if the user has overwritten with the `OptionsConfiguration` callback, we should honor it

CHANGELOG.md
```diff
@@ -6,6 +6,8 @@
 * Feat: Send default PII options (#360)
 * Bump: sentry-cocoa to v6.2.1 (#360)
 * Fix: missing event.origin and event.environment tags for non-native events on iOS (#369)
+* Fix: Set `SentryOptions.debug` in `sentry` (#376)
```
marandaneto, line 10: this PR now mixes up 2 PRs, as we do squash and merge, it's suboptimal, let's be sure in the next PRs to not do this anymore.
We could still use target branches, at the end we should merge them into the main directly, so we guarantee that we have the proper history.
for now, lets update this PRs title and description that it also reflects the `SentryOptions.debug` changes

ueman: Ah sorry. I didn't think of that. I had set this branch as a target in the other PR so that I can avoid merge conflicts because both PRs touch the same files.

flutter/lib/src/sentry_flutter.dart
```diff
@@ -60,8 +60,6 @@ mixin SentryFlutter {
     SentryFlutterOptions options,
     MethodChannel channel,
   ) async {
-    options.debug = kDebugMode;
```
marandaneto, line 63: I believe removing this would introduce a bug now, after deep thinking, we should do this in both packages first thing.

otherwise all the logging being done before this flag is set in the sentry package will be still `false` when it should be enabled.
actually, this bug even already exists I guess, here we should also do this:

https://github.com/getsentry/sentry-dart/blob/main/dart/lib/src/sentry.dart#L199-L202

otherwise even if `debug=true`, there's no logging at all (its a hunch)

dart/lib/src/environment_variables.dart
ueman: So I could pass `EnvironmentVariables` as a constructor parameter to `SentryOptions` and set them.
The user is then free to overwrite them in the `OptionsConfiguration`. Then `EnvironmentVariables` does not have to be a property of `SentryOptions`.

CHANGELOG.md
marandaneto: no worries, I did this many times :D

dart/lib/src/environment_variables.dart
marandaneto: I dont think so, users never creates their own `SentryOptions`, but always mutate the SDKs `SentryOptions`, the reason is, we always have the chance to set default values before the user has the chance to mutate it, so I'd be in favor of being a property to `SentryOptions` that we read from, like all the other internal properties, users dont have to interact with `EnvironmentVariables`

flutter/lib/src/sentry_flutter.dart
ueman: In Sentry it is already the first thing we do. But yeah in sentry_flutter it makes sense to do as first thing.
I'm not sure though how I would go about testing this.

dart/test/environment_test.dart
```diff
@@ -0,0 +1,54 @@
+import 'package:sentry/sentry.dart';
+import 'package:test/test.dart';
+
+import 'mocks.dart';
+import 'mocks/mock_environment_variables.dart';
+
+void main() {
+  // See https://docs.sentry.io/platforms/dart/configuration/options/
+  // and https://github.com/getsentry/sentry-dart/issues/306
+  group('Environment Variables', () {
```
marandaneto, line 10: we need to implement `tearDown` here and call Sentry.close() because Sentry is a static class and has state

dart/lib/src/sentry.dart
```diff
@@ -52,24 +53,12 @@ class Sentry {
...
-        : environment;
-
-    // if the SENTRY_DSN is set, we read from it.
-    options.dsn = const bool.hasEnvironment('SENTRY_DSN')
-        ? const String.fromEnvironment('SENTRY_DSN')
-        : options.dsn;
+    SentryOptions options,
+    AppRunner? appRunner,
+  ) async {
+    options.debug = options.platformChecker.isDebugMode();
```
bruno-garcia, line 59: Is this running before or after the user callback? jsut checking because if it's after we would be overriding the user-defined value. If it's before it just means we default to debug=true and users would be forced to turn it off if they don't want to see Sentry's debugging messages while developing (likely after the initial setup).

marandaneto: literally nothing is run after the user conf. callback except integrations

dart/lib/src/environment_variables.dart
```diff
@@ -0,0 +1,33 @@
...
+/// `flutter build --dart-define=VARIABLE_NAME=VARIABLE_VALUE`.
+class EnvironmentVariables {
+  static const _sentryEnvironment = 'SENTRY_ENVIRONMENT';
+  static const _sentryDsn = 'SENTRY_DSN';
+  static const _sentryRelease = 'SENTRY_RELEASE';
+  static const _sentryDist = 'SENTRY_DSN';
+
+  /// `SENTRY_ENVIRONMENT`
+  /// See [SentryOptions.environment]
+  String? get environment => const bool.hasEnvironment(_sentryEnvironment)
```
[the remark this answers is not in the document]
ueman: That's a named constructor so it can't return null.
So it wouldn't throw but it would return an empty string (`''`) if the environment variable does not exist.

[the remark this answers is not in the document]
marandaneto: maybe `fromEnvironment` passing `defaultValue=null` if possible would do then, but I guess it does not work either, so the way how it is, its the correct way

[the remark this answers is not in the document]
ueman: You can't pass null because it only accepts non-null strings (because null safety), so yeah it is already correct as is.

bruno-garcia:
Removed the #376 from the title

flutter/lib/src/sentry_flutter.dart
ueman: I created an issue for this here: https://github.com/getsentry/sentry-dart/issues/384

ueman:
@marandaneto I fixed the last missing bits. Please let me know if it is now good to go.

dart/lib/src/platform_checker.dart
```diff
@@ -1,3 +1,5 @@
+import 'sentry_options.dart';
```
marandaneto: maybe we move `defaultEnvironment` to `EnvironmentVariables` instead

dart/lib/src/platform_checker.dart
```diff
@@ -17,4 +19,18 @@ class PlatformChecker {
   bool isProfileMode() {
     return const bool.fromEnvironment('dart.vm.profile', defaultValue: false);
   }
+
+  /// This can be set as [SentryOptions.environment]
+  String get environment {
```
marandaneto: should this be a method instead? quite a few conditions to be just a property, maybe we move hardcoded values to `EnvironmentVariables` instead

dart/lib/src/sentry.dart
```diff
@@ -79,6 +68,25 @@ class Sentry {
...
+  /// The precendence of these options are also described on
+  /// https://docs.sentry.io/platforms/dart/configuration/options/
+  static void _setEnvironmentVariables(SentryOptions options) {
+    final vars = options.environmentVariables;
+    options.dsn = options.dsn ?? vars.dsn;
+
+    if (options.environment == null) {
+      var environment = options.platformChecker.environment;
+      options.environment = vars.environment ?? environment;
+    }
```
marandaneto: this wouldnt be the same as:

`options.environment = options.environment ?? options.platformChecker.environment;` or it needs to be this way?

dart/test/environment_test.dart
```diff
@@ -0,0 +1,58 @@
...
+        options: options,
+      );
+
+      expect(options.dsn, fakeDsn);
+      expect(options.environment, 'prod');
+      expect(options.release, 'release-1.2.3');
+      expect(options.dist, 'foo');
+    });
+
+    test('No options are set', () async {
```
marandaneto: maybe let's improve the test name, something like, SentryOptions are overridden by environment

dart/lib/src/sentry.dart
ueman: We do have three values which we need to consider:
- `SentryOptions.environment`
- Environment variable
- The value we set based on Darts release, profile and debug mode.

So I first check if the user has set something and then look for the other two.
Does that clear it up?

flutter/test/default_integrations_test.dart
```diff
@@ -3,10 +3,12 @@ import 'package:flutter/services.dart';
 import 'package:flutter/widgets.dart';
 import 'package:flutter_test/flutter_test.dart';
 import 'package:mockito/mockito.dart';
+import 'package:package_info_plus/package_info_plus.dart';
```
marandaneto, line 6: merge main into this branch to not show the diff of non related stuff

marandaneto:
> @marandaneto I fixed the last missing bits. Please let me know if it is now good to go.

made a few small suggestions but other than that LGTM

dart/lib/src/sentry.dart
marandaneto: oh yeah that makes sense, it'd be possible still with this `ternary` operator but it'd be a bit complex to read, lets keep as it is

marandaneto (approved):
LGTM

PR #378, opened 2021-03-22
Fix: Flutter Beta CI

ueman:
## :scroll: Description
CI fails on Flutter beta channel, this PR fixes that

## :bulb: Motivation and Context
https://github.com/getsentry/sentry-dart/issues/377

## :pencil: Checklist

- [ ] I reviewed submitted code
- [ ] I added tests to verify changes
- [ ] I updated the docs if needed
- [ ] All tests passing
- [ ] No breaking changes

PR #382, opened 2021-03-23
Fix: Disable Flutter beta builds on CI

ueman:
## :scroll: Description
This change disables Flutter beta builds.
It also removes a fix which is not longer needed and fixes the LICENSE path in the `sentry_flutter.podspec` file.

_#skip-changelog_

Currently Flutter beta fails for Android and iOS.

I tried to fix it in https://github.com/getsentry/sentry-dart/pull/378 but I didn't succeed.
Also see https://github.com/flutter/flutter/issues/78785

## :pencil: Checklist

- [ ] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps
Fix Android beta builds

### DIFF
--- a/.github/workflows/flutter.yml
+++ b/.github/workflows/flutter.yml
@@ -17,7 +17,7 @@ jobs:
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
         target: ['ios', 'android', 'web']
-        channel: ['stable', 'beta']
+        channel: ['stable']
         exclude:
           - os: ubuntu-latest
             target: ios
--- a/.github/workflows/flutter.yml
+++ b/.github/workflows/flutter.yml
@@ -17,6 +17,7 @@ jobs:
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
         target: ['ios', 'android', 'web']
+        # 'beta' channel is not working https://github.com/flutter/flutter/issues/78785
         channel: ['stable']
         exclude:
           - os: ubuntu-latest
--- a/.github/workflows/flutter.yml
+++ b/.github/workflows/flutter.yml
@@ -42,11 +42,6 @@ jobs:
 
         - run: flutter upgrade
 
-        # https://github.com/flutter/flutter/issues/59522#issuecomment-646208247
-        - name: Fix iOS build bug
-          if: matrix.target == 'ios'
-          run: rm flutter/example/ios/Podfile
-
         - name: Test
           run: |
             cd flutter
--- a/flutter/ios/sentry_flutter.podspec
+++ b/flutter/ios/sentry_flutter.podspec
@@ -6,7 +6,7 @@ Pod::Spec.new do |s|
 Sentry SDK for Flutter with support to native through sentry-cocoa.
                        DESC
   s.homepage         = 'https://sentry.io'
-  s.license          = { :type => 'MIT', :file => '../../LICENSE' }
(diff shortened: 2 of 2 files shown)


ueman:
Do I need to add a changelog entry for this?

.github/workflows/flutter.yml
```diff
@@ -17,7 +17,7 @@ jobs:
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
         target: ['ios', 'android', 'web']
-        channel: ['stable', 'beta']
+        channel: ['stable']
```
marandaneto, line 21: instead of removing the `channel`, lets exclude the channel `beta` for iOS only, see the `exclude` on this file, also add a comment about the iOS issue

marandaneto:
> Do I need to add a changelog entry for this?

no, changelog is only for user's facing features/changes, but add the skip changelog to description, so Danger gets happy

flutter/ios/sentry_flutter.podspec
```diff
@@ -6,7 +6,7 @@ Pod::Spec.new do |s|
 Sentry SDK for Flutter with support to native through sentry-cocoa.
                        DESC
   s.homepage         = 'https://sentry.io'
-  s.license          = { :type => 'MIT', :file => '../../LICENSE' }
+  s.license          = { :type => 'MIT', :file => '../LICENSE' }
```
marandaneto, line 9: the change makes sense, but `../../LICENSE` also exists, what was the warning?

ueman: If you run `pod install` in `flutter/example/ios` this warning happens: 
```
[!] A license was specified in podspec `sentry_flutter` but the file does not exist - /Users/jonas/Documents/projects/sentry-dart/flutter/example/ios/.symlinks/plugins/LICENSE

[!] Unable to read the license file `../../LICENSE` for the spec `sentry_flutter (0.0.1)`

[!] Unable to read the license file `../../LICENSE` for the spec `sentry_flutter (0.0.1)`

[!] CocoaPods did not set the base configuration of your project because your project already has a custom config set. In order for CocoaPods integration to work at all, please either set the base configurations of the target `Runner` to `Target Support Files/Pods-Runner/Pods-Runner.profile.xcconfig` or include the `Target Support Files/Pods-Runner/Pods-Runner.profile.xcconfig` in your build configuration (`Flutter/Release.xcconfig`).

[!] Your project does not explicitly specify the CocoaPods master specs repo. Since CDN is now used as the default, you may safely remove it from your repos directory via `pod repo remove master`. To suppress this warning please add `warn_for_unused_master_specs_repo => false` to your Podfile.
```

marandaneto: I see, looks like that this could depend on where your App is placed (in our case, flutter/example/ios).
wondering if a fresh project with our SDK also throws this warning or not.

ueman: Yep, it does.

.github/workflows/flutter.yml
marandaneto: maybe lets keep like this as its broken for Android too but lets write a comment why we've commented out the beta channel and link to the GH issue

marandaneto, line 21:
```suggestion
        # 'beta' channel is not working https://github.com/flutter/flutter/issues/78785
        channel: ['stable']
```

Issue #374, opened 2021-03-18
CI should compile examples for Dart/Dart Web and Flutter


marandaneto:
I guess this is a regression, we need to check it out

ueman:
Could be done together with https://github.com/getsentry/sentry-dart/issues/334

marandaneto:
I guess when I've added the `exclude` section to the `analysis_options`file, this kinda bypassed the build of the examples folder, its something we could check out, maybe just fixing the linting for the examples and removing that would already work, or, last case, we add a new Job to the CI that compiles the examples

Issue #334, opened 2021-03-04
CI uses new dart action


marandaneto:
https://github.com/dart-lang/setup-dart

We need to use that because our current Action that setups Dart does not have the latest stable yet and the Action got deprecated, so we'd need to migrate to make the CI happy

PR #385, opened 2021-03-25
Fix: Use new Dart GH Action

ueman:
## :scroll: Description
Use new Dart GitHub Action and fix all Linter warnings in the examples

_#skip-changelog_

#334 and #374

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/.github/workflows/dart.yml
+++ b/.github/workflows/dart.yml
@@ -17,9 +17,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        # TODO: cedx/setup-dart@v2 doesn't work on Windows (doesn't add pub to the PATH?)
-        # os: [ubuntu-latest, windows-latest, macos-latest]
-        os: [ubuntu-latest, macos-latest]
+        os: [ubuntu-latest, windows-latest, macos-latest]
         sdk: [stable, beta, dev]
         exclude:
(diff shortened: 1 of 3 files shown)


ueman:
#skip-changelog

marandaneto:
@ueman it should be in the description, not PR's title, or as a comment, otherwise it does not work

ueman:
> @ueman it should be in the description, not PR's title, or as a comment, otherwise it does not work

Dang it :D Thanks!

.github/workflows/dart.yml
```diff
@@ -35,7 +33,7 @@ jobs:
             sdk: dev
 
     steps:
-        - uses: cedx/setup-dart@v2
+        - uses: dart-lang/setup-dart@v1
```
marandaneto, line 36: its still using `release-channel` to config the channel, its different for this action

marandaneto:
we still need to address https://github.com/getsentry/sentry-dart/issues/374 for the Flutter package, could be a different PR though, but I guess it already works for the Flutter package.

marandaneto (approved):

ueman:
> we still need to address #374 for the Flutter package, could be a different PR though, but I guess it already works for the Flutter package.

We already build the example so I would say they're good as is:

https://github.com/getsentry/sentry-dart/blob/main/.github/workflows/flutter.yml#L62-L75

Issue #352, opened 2021-03-12
Knowing when an event has sent successfully


fzyzcjy:
_Platform:_
- [ ]  Dart
- [x]  Flutter Android or iOS
- [ ]  Flutter Web

_IDE:_
- [ ]  VSCode
- [x]  IntelliJ/AS
- [ ]  XCode
- [ ]  Other, which one?

_split-debug-info and obfuscate (Flutter Android or iOS) or CanvasKit (Flutter Web):_
- [ ]  Enabled
- [x]  Disabled

_Platform installed with:_
- [x] pub.dev
- [ ] GitHub

Output of the command `flutter doctor -v` below:
```
[✓] Flutter (Channel stable, 2.0.1, on Mac OS X 10.15.7 19H524 darwin-x64, locale zh-Hans-CN)
    • Flutter version 2.0.1 at /Users/tom/opt/flutter
    • Framework revision c5a4b4029c (7 天前), 2021-03-04 09:47:48 -0800
    • Engine revision 40441def69
    • Dart version 2.12.0
    • Pub download mirror https://pub.flutter-io.cn
    • Flutter download mirror https://storage.flutter-io.cn

[✓] Android toolchain - develop for Android devices (Android SDK version 29.0.3)
    • Android SDK at /Users/tom/Library/Android/sdk
    • Platform android-30, build-tools 29.0.3
    • ANDROID_SDK_ROOT = /Users/tom/Library/Android/sdk
    • Java binary at: /Users/tom/Library/Application
      Support/JetBrains/Toolbox/apps/AndroidStudio/ch-0/201.7042882/Android
      Studio.app/Contents/jre/jdk/Contents/Home/bin/java
    • Java version OpenJDK Runtime Environment (build 1.8.0_242-release-1644-b3-6915495)
    • All Android licenses accepted.

[✓] Xcode - develop for iOS and macOS
    • Xcode at /Applications/Xcode.app/Contents/Developer
    • Xcode 12.4, Build version 12D4e
    • CocoaPods version 1.10.0

[✓] Chrome - develop for the web
    • Chrome at /Applications/Google Chrome.app/Contents/MacOS/Google Chrome

[✓] Android Studio (version 4.1)
    • Android Studio at /Users/tom/Library/Application
      Support/JetBrains/Toolbox/apps/AndroidStudio/ch-0/201.7042882/Android Studio.app/Contents
    • Flutter plugin can be installed from:
[...]

Issue #291, opened 2021-02-03
`close` should to disable completely the SDK


bruno-garcia:
Pass the call to Android and iOS to ensure all layers get closed.

Depends on: 
https://github.com/getsentry/sentry-java/issues/1225
https://github.com/getsentry/sentry-cocoa/issues/934

The behavior should be that calling `Sentry.close()` will block the caller until all events queued up to the point of `close` being called are flushed out. And then the SDK can be closed and the method returns.

If it's simpler to implement it in a way that it'll keep waiting for events to be flushed even if those were queued after `close` was called, that's fine but essentially that would mean that if you have `capture` in a tight loop, it'll always stop only upon timeout.

The behavior above is describing `flush` which is an optional API here but is part of the behavior of `close`. 

Example of such implementation: https://github.com/getsentry/sentry-dotnet/blob/2337261333d5743996d6d96ca3aa289d34a9d3ec/src/Sentry/Internal/BackgroundWorker.cs#L182-L261

This will unblock the issue #352 since the user can call `Sentry.close()` to make sure events are not lost, before exiting the app.

marandaneto:
@ueman maybe let's create a new one which is only about using the latest Cocoa once is GA, so we could remove the enabled boolean with the proper method call, otherwise, we might forget :)

the issue title could be like Bump Cocoa SDK to v7 and then specifying what else we have to do, calling `close` instead fiddling with the `enabled` flag

PR #388, opened 2021-03-29
Fix: Close native SDKs

ueman:
## :scroll: Description
This change closes native SDKs if one calls `Sentry.close()`.
Currently only Android because https://github.com/getsentry/sentry-cocoa/issues/934 is still open and I didn't find a method to do this.

@philipphofmann Am I overlooking something on iOS?

From the issue:
> The behavior should be that calling Sentry.close() will block the caller until all events queued up to the point of close being called are flushed out. And then the SDK can be closed and the method returns.

I don't think we need this because every send event can be awaited.
`Sentry.captureEvent` returns a `Future<SentryId>`, so if you awaited it, you can be sure that it was sent to Sentry.

#291

I added tests and a button in the example to verify manually.

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps
If we need need a `Sentry.flush()` method, I would prefer to do it in another PR, see https://github.com/getsentry/sentry-dart/issues/352 for the relevant issue.


flutter/ios/Classes/SwiftSentryFlutterPlugin.swift
```diff
@@ -120,6 +123,11 @@ public class SwiftSentryFlutterPlugin: NSObject, FlutterPlugin {
 
         result("")
     }
+    
+    private func closeNativeSdk(_ call: FlutterMethodCall, result: @escaping FlutterResult) {
+        // TODO: Call SentrySDK when available
```
ueman, line 129:
```suggestion
        // TODO: Call SentrySDK when available
        // https://github.com/getsentry/sentry-cocoa/issues/934
```

flutter/lib/src/default_integrations.dart
```diff
@@ -221,6 +221,15 @@ class NativeSdkIntegration extends Integration<SentryFlutterOptions> {
       );
     }
   }
+
+  @override
+  FutureOr<void> close() async {
+    try {
+      await _channel.invokeMethod<void>('closeNativeSdk');
+    } catch (error) {
+      // TODO what to do?
```
ueman: @marandaneto Currently we can't log something in the `Integration.close` emthods because `SentryOptions` are passed to `Integration.call`. They aren't an instance property.

Do we need to log this error? If yes I would add SentryOptions as an instance property.

CHANGELOG.md
```diff
@@ -7,9 +7,11 @@
 * Bump: sentry-cocoa to v6.2.1 (#360)
 * Feat: Migration from `package_info` to `package_info_plus` plugin (#370)
 * Fix: missing event.origin and event.environment tags for non-native events on iOS (#369)
+* Fix: `Sentry.close()` closes native SDK integrations
```
```suggestion
* Fix: `Sentry.close()` closes native SDK integrations (#388)
```

flutter/lib/src/default_integrations.dart
marandaneto: you mean `options.logger` does not work? maybe it's something we should fix across the SDK then, if it's the case, let's raise a new issue for that, we don't need to tackle within this PR.

marandaneto:
let's keep it as a Draft till we get the native SDKs to offer an API for that (eg iOS) or (Android calls sentry-native shutdown), so we need to bump the Native SDKs too

flutter/lib/src/default_integrations.dart
ueman: It does work, but I don't have access to it in the close method. So my question is should we add `SentryOptions` as a parameter to the `Integration.close` method. Or just pass `SentryOptions` as a constructor parameter.

ueman:
> let's keep it as a Draft till we get the native SDKs to offer an API for that (eg iOS) or (Android calls sentry-native shutdown), so we need to bump the Native SDKs too

Android already works though, except for the sentry-native part. But yeah, makes sense.

flutter/lib/src/default_integrations.dart
marandaneto: you get the `SentryFlutterOptions` instance when the method `call` is called, save it to a global field (of the integration), or take it as a ctor param, both would work

marandaneto:
PR for Android closing sentry-native -> https://github.com/getsentry/sentry-java/pull/1358

flutter/ios/Classes/SwiftSentryFlutterPlugin.swift
```diff
@@ -121,6 +124,13 @@ public class SwiftSentryFlutterPlugin: NSObject, FlutterPlugin {
         result("")
     }
 
+    private func closeNativeSdk(_ call: FlutterMethodCall, result: @escaping FlutterResult) {
+        // swiftlint:disable:next todo
+        // TODO: Call SentrySDK when available
```
bruno-garcia, line 129: We can merge this and keep sentry-cocoa as no-op for now.

flutter/android/src/main/kotlin/io/sentry/flutter/SentryFlutterPlugin.kt
```diff
@@ -164,6 +166,11 @@ class SentryFlutterPlugin : FlutterPlugin, MethodCallHandler {
     result.success(newDebugImages)
   }
 
+  private fun closeNativeSdk(call: MethodCall, result: Result) {
```
```suggestion
  private fun closeNativeSdk(result: Result) {
```

apparently we don't need it

flutter/ios/Classes/SwiftSentryFlutterPlugin.swift
marandaneto: @philipphofmann can u tell us how to swap the `enabled` flag? could be a workaround for now

is it possible to mutate this https://github.com/getsentry/sentry-cocoa/blob/master/Sources/Sentry/SentryOptions.m#L22 after SDK is init?

philipphofmann: Yes, you can swap the enabled flag. The client is then just dropping everything. See https://github.com/getsentry/sentry-cocoa/blob/789d52b29386cb0267be6b9696ce29df038f40dc/Sources/Sentry/SentryClient.m#L448-L451.

marandaneto: @philipphofmann yep, we'd like to know `how` to swap it, I only see a getter but not a setter across SentrySDK/Hub and Client

philipphofmann: We made the [SentrySDK currentHub] private in 7.x.x. For Hybrid SDKs, it is fine to create a category to make this method visible, like https://github.com/getsentry/sentry-cocoa/blob/master/Sources/Sentry/include/SentrySDK+Private.h.  Then you should be able to do something like `SentrySDK.currentHub().getClient()?.options.enabled = false`. It's not nice, but it should work.

marandaneto: @philipphofmann we'd do the 1st iteration without v7, so we could do this directly right now, thanks.

`SentrySDK.currentHub().getClient()?.options.enabled = false`

Once v7 is GA, we'll probably have a proper SentrySDK.close() method and could get rid of it altogether.

.github/workflows/web-example-ghpages.yml
```diff
@@ -11,6 +11,6 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - uses: subosito/flutter-action@v1
-      - uses: erickzanardo/flutter-gh-pages@v2
+      - uses: erickzanardo/flutter-gh-pages@v3
```
marandaneto: merge main into this branch, please

CHANGELOG.md
```diff
@@ -1,17 +1,22 @@
...
 * Feat: Migration from `package_info` to `package_info_plus` plugin (#370)
-* Fix: missing event.origin and event.environment tags for non-native events on iOS (#369)
 * Fix: Set `SentryOptions.debug` in `sentry` (#376)
 * Fix: Read all environment variables in `sentry` (#375)
+* Fix: `Sentry.close()` closes native SDK integrations (#388)
 
 ## Breaking Changes:
 
+* Return type of `Sentry.close()` changed from `void` to `Future<void>` and `Integration.close()` changed from `void` to `FutureOr<void>` (#395)
+* Return type of `Sentry.close()` changed from `void` to `Future<void>`
```
marandaneto: this is dup

CHANGELOG.md
```diff
@@ -1,17 +1,22 @@
...
+* Fix: event.origin and event.environment tags have wrong value for iOS (#365) and (#369)
 * Fix: Fix deprecated `registrar.messenger` call in `SentryFlutterWeb` (#364)
 * Fix: Enable breadcrumb recording mechanism based on platform (#366)
 * Feat: Send default PII options (#360)
 * Bump: sentry-cocoa to v6.2.1 (#360)
 * Feat: Migration from `package_info` to `package_info_plus` plugin (#370)
-* Fix: missing event.origin and event.environment tags for non-native events on iOS (#369)
 * Fix: Set `SentryOptions.debug` in `sentry` (#376)
 * Fix: Read all environment variables in `sentry` (#375)
+* Fix: `Sentry.close()` closes native SDK integrations (#388)
```
marandaneto: this should be under `Unreleased`

flutter/test/default_integrations_test.dart
```diff
@@ -177,6 +179,21 @@ void main() {
         false);
   });
 
+  test('nativeSdkIntegration closes native SDK', () async {
+    final completer = Completer<bool>();
```
marandaneto: this could be as simple as a boolean var, no need of a `Completer`, just a nit though, feel free to ignore it

ueman: Nah, gotta fix it.

marandaneto (approved):
LGTM

Issue #377, opened 2021-03-22
Flutter Beta builds are broken on CI


ueman:
_Platform:_
- [ ]  Dart
- [x]  Flutter Android or iOS
- [ ]  Flutter Web

See the following PRs: 
- https://github.com/getsentry/sentry-dart/pull/376
- https://github.com/getsentry/sentry-dart/pull/375
- https://github.com/getsentry/sentry-dart/pull/370

iOS complains about a missing Podfile on beta.
Android has a problem with references/imports in the MainActivity.kt

 flutter/flutter#78785

marandaneto:
@ueman please prioritize this then, let me know if you need any help

ueman:
I'm already on it

ueman:
I think Flutter beta is broken.
On Android I also get errors similar to iOS.
```
/sentry-dart/flutter/example/android/app/src/main/java/io/flutter/plugins/GeneratedPluginRegistrant.java:17: error: package io.sentry.flutter does not exist
    flutterEngine.getPlugins().add(new io.sentry.flutter.SentryFlutterPlugin());
/sentry-dart/flutter/example/android/app/src/main/java/io/flutter/plugins/GeneratedPluginRegistrant.java:16: error: package io.flutter.plugins.packageinfo does not exist
    flutterEngine.getPlugins().add(new io.flutter.plugins.packageinfo.PackageInfoPlugin());
```

which is also from `GeneratedPluginRegistrant`.

But I have no idea why :(

marandaneto:
> I think Flutter beta is broken.
> On Android I also get errors similar to iOS.
> 
> ```
> /sentry-dart/flutter/example/android/app/src/main/java/io/flutter/plugins/GeneratedPluginRegistrant.java:17: error: package io.sentry.flutter does not exist
>     flutterEngine.getPlugins().add(new io.sentry.flutter.SentryFlutterPlugin());
> /sentry-dart/flutter/example/android/app/src/main/java/io/flutter/plugins/GeneratedPluginRegistrant.java:16: error: package io.flutter.plugins.packageinfo does not exist
>     flutterEngine.getPlugins().add(new io.flutter.plugins.packageinfo.PackageInfoPlugin());
> ```
> 
> which is also from `GeneratedPluginRegistrant`.
> 
> But I have no idea why :(

if you create a sample project from scratch and add an Android/iOS plugin (eg packageinfo), does it compile for Android and iOS? using the beta channel obviously.
[...]

PR #399, opened 2021-04-05
Fix: Re-enable Flutter Beta

ueman:
## :scroll: Description
Re-enable Flutter Beta CI 

_#skip-changelog_

https://github.com/getsentry/sentry-dart/issues/377

We needed to remove the plugin part of the example because it is not a plugin but an application.

CI builds again.

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes


flutter/example/lib/main.dart
```diff
@@ -231,26 +230,6 @@ class CocoaExample extends StatelessWidget {
...
-  final channel = const MethodChannel('example.flutter.sentry.io');
-
-  @override
-  Widget build(BuildContext context) {
-    return Column(
-      children: [
-        RaisedButton(
-          child: const Text('Web: console.log'),
-          onPressed: () async {
-            await channel.invokeMethod<void>('console.log', 'log me');
```
ueman, line 246: I think we don't need this. It just called `Sentry.captureMessage`.

ueman: For Web we also don't need method channels, just platform dependent imports.

.github/workflows/flutter.yml
```diff
@@ -18,7 +18,7 @@ jobs:
         os: [ubuntu-latest, windows-latest, macos-latest]
         target: ['ios', 'android', 'web']
         # 'beta' channel is not working https://github.com/flutter/flutter/issues/78785
-        channel: ['stable']
+        channel: ['stable', 'beta']
```
marandaneto, line 20: remove the comment above

flutter/example/lib/main.dart
marandaneto: mm makes sense, but what's about if I have a JS file that does something for me, not written in Dart? would that be done thru method channels, right?

.github/workflows/flutter.yml
```diff
@@ -18,7 +18,7 @@ jobs:
         os: [ubuntu-latest, windows-latest, macos-latest]
         target: ['ios', 'android', 'web']
         # 'beta' channel is not working https://github.com/flutter/flutter/issues/78785
```
ueman:
```suggestion
```

flutter/example/lib/main.dart
ueman: Nope, as far as I know you can use https://api.dart.dev/stable/2.12.2/dart-js/dart-js-library.html for that. On web there is no method channel because Dart gets transpiled to JS.

ueman: I guess we could offer support for that in Sentry even without Flutter.

marandaneto: yep, thanks

flutter/example/pubspec.yaml
```diff
@@ -21,9 +21,4 @@ dependency_overrides:
     path: ../
 
 flutter:
-  uses-material-design: true
-  plugin:
```
marandaneto, line 25: I'm afraid that this would remove the `web` badge compatibility from pub.dev
how can we check that?

see an example:
https://github.com/flutter/plugins/blob/master/packages/url_launcher/url_launcher_web/pubspec.yaml

do we need to add at least an empty plugin or not?

ueman: This is the pubspec file of the example not of Sentry itself. To support web, we don't need to do something special. Also  pana in CI would complain if it's not compatible with web anymore. We also have different imports for web and io where needed.

ueman: url_launcher has that because it is a federated plugin which Sentry is not.

ueman: The web build in CI is also green :D

marandaneto: yep, you're totally right, I need a coffee :)

marandaneto (approved):
LGTM

PR #389, opened 2021-03-29
Feature: Initial support for Flutter on Desktop

ueman:
## :scroll: Description

Initial desktop support.
Currently only MacOS. 

See https://sentry.io/organizations/sentry-sdks/issues/2149575550/?query=is%3Aunresolved for an example event.

What currently works:

- Dart exceptions
- Breadcrumbs for navigation
- Release information
- Native Integration just like iOS

What does not work:

- Linux and Windows
- `SentryWidgetsBindingObserver.didChangeMetrics` get called way too much. It could replace all breadcrumbs with just screen size changes.

## :bulb: Motivation and Context
- https://github.com/getsentry/sentry-dart/issues/362
- https://github.com/getsentry/sentry-dart/projects/1#card-50187067

## :green_heart: How did you test it?
New tests and manual tests to make sure MacOS works.

## :pencil: Checklist

- [x] I reviewed submitted code
- [x] I added tests to verify changes
- [x] I updated the docs if needed
- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps
- Linux, see https://github.com/getsentry/sentry-dart/pull/402
- Windows (I don't own a windows devices)
- Debouncing of `SentryWidgetsBindingObserver.didChangeMetrics`. As seen in this [crash report](https://sentry.io/organizations/sentry-sdks/issues/2149575550/?query=is%3Aunresolved) it gets called way too much. Issue created at https://github.com/getsentry/sentry-dart/issues/400
- Read OS and device information (another PR) when no native integration is available, see https://github.com/getsentry/sentry-dart/issues/401

PR #402, opened 2021-04-06
Feat: Support Linux

ueman:
## :scroll: Description
This adds support for Linux (tested on Ubuntu) and the corresponding example.
This branch currently targets https://github.com/getsentry/sentry-dart/pull/389 because it is a follow up.

The current tests make already sure it works. Linux has no native integration yet, so there are no native integrations to test.

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps
- On Linux we don't have any kind of OS and device information


flutter/example/linux/flutter/CMakeLists.txt
```diff
@@ -0,0 +1,91 @@
+cmake_minimum_required(VERSION 3.10)
+
+set(EPHEMERAL_DIR "${CMAKE_CURRENT_SOURCE_DIR}/ephemeral")
+
+# Configuration provided via flutter tool.
+include(${EPHEMERAL_DIR}/generated_config.cmake)
+
+# TODO: Move the rest of this into files in ephemeral. See
+# https://github.com/flutter/flutter/issues/57146.
```
ueman, line 9: This was autogenerated with this todo

flutter/example/linux/flutter/generated_plugin_registrant.cc
```diff
@@ -0,0 +1,13 @@
+//
```
ueman: There are a couple generated files. They aren't gitignored according to the template. I'm not sure yet wether we can ignore them.

flutter/linux/sentry_flutter_plugin.cc
```diff
@@ -0,0 +1,62 @@
...
+
+// Called when a method call is received from Flutter.
+static void sentry_flutter_plugin_handle_method_call(
+    SentryFlutterPlugin* self,
+    FlMethodCall* method_call) {
+  g_autoptr(FlMethodResponse) response = nullptr;
+
+  const gchar* method = fl_method_call_get_name(method_call);
+
+  response = FL_METHOD_RESPONSE(fl_method_not_implemented_response_new());
```
ueman, line 25: I think we can add support for https://github.com/getsentry/sentry-native here.

flutter/example/linux/flutter/generated_plugin_registrant.cc
ueman: They can't be ignored

.github/workflows/flutter.yml
```diff
@@ -82,6 +94,9 @@ jobs:
             web)
             flutter build web
             ;;
+            linux)
+            flutter build linux
```
marandaneto, line 102: the other builds have a `--release`, do we need this here?

flutter/linux/sentry_flutter_plugin.cc
marandaneto: yep, let's not do it, for now, let's just be sure that this plugin does nothing instead of throwing unless it never gets called.

flutter/pubspec.yaml
```diff
@@ -41,3 +41,5 @@ flutter:
       web:
         pluginClass: SentryFlutterWeb
         fileName: sentry_flutter_web.dart
+      linux:
+        pluginClass: SentryFlutterPlugin
```
marandaneto, line 45: should we maybe name all the plugins a bit more specific to the platform? like SentryFlutterLinux or, to rename the others to be always generic? eg SentryFlutterPlugin

flutter/linux/sentry_flutter_plugin.cc
ueman: it never get's called at all.

bruno-garcia: It's great this uses CMake so it should be easy to add sentry-native later.

flutter/pubspec.yaml
bruno-garcia: If these names are just for internal use (I believe that's the case) it would be beneficial to align on the side of appending the platform so make it easier to navigate through the project

bruno-garcia (commented):
Exciting stuff!

flutter/pubspec.yaml
ueman: Yep, these are internal names. They however don't necessarily map as is to a class or file name. On linux there is for example a struct named after this, but the file is named `sentry_flutter_plugin.cc` and `sentry_flutter_plugin.h`.
Also MacOS and iOS use the same file.

We can align these name but I don't really think it matters that much, because the files are also in their corresponding platform folders. Your call :D

.github/workflows/flutter.yml
ueman: Yep good catch

flutter/pubspec.yaml
marandaneto: its just a matter of consistency, not really "necessary", I agree with @bruno-garcia that it helps to navigate thru the project, even more, when you take care of more repos, which does the same thing, with similar names

.github/workflows/flutter.yml
```diff
@@ -83,6 +95,9 @@ jobs:
...
```
ueman, line 102:
```suggestion
            flutter build linux --release
```

marandaneto (approved):
nice, thanks for doing this, LGTM

marandaneto:
@ueman see https://github.com/getsentry/sentry-dart/pull/407#issuecomment-815997737

.github/workflows/flutter.yml
```diff
@@ -42,6 +46,14 @@ jobs:
             distribution: 'adopt'
             java-version: '8'
 
+        # Install required dependencies for Flutter on Linux on Ubuntu
+        - name: 'Setup Linux'
+          run: |
+            sudo apt update
+            sudo apt install -y cmake dbus libblkid-dev libgtk-3-dev liblzma-dev ninja-build pkg-config xvfb
+            sudo apt install -y network-manager upower
```
marandaneto, line 57: is this really necessary? I don't see as a requirement here https://flutter.dev/docs/get-started/install/linux#install-flutter-manually

ueman: The first few builds failed because at least ninja was not available on the ubuntu image. I then looked up what the Flutter Plus Plugin maintainers did. I haven't yet tried to reduce the number of installed packages.

I think we can drop at least `sudo apt install -y network-manager upower` but I have yet to try it.

marandaneto: it was just curiosity tbh, if the flutter plus plugin does the same, let's keep like this :)
I've not installed any of those on my ubuntu machine and it worked out ootb, but I might have them installed already

ueman: I think there's also a difference between Ubuntu Desktop and Ubuntu Server

bruno-garcia (approved):
Great stuff. I'll continue the Windows PR based on this

ueman:
I'll leave this open until https://github.com/getsentry/sentry-dart/pull/389 is approved and merged because this depends on the other branch.

marandaneto (approved):
looks realllyy good, thanks.
LGTM

marandaneto:
> I'll leave this open until #389 is approved and merged because this depends on the other branch.

we can also merge everything to feature/support-desktop as the branch's name is called: support desktop

PR #407, opened 2021-04-08
feat: Windows Support

bruno-garcia:
Good news: 

image

Bad news:

image
[Link](https://sentry.io/organizations/sentry-sdks/issues/2137135923/?query=is%3Aunresolved)


ueman:
https://github.com/fluttercommunity/plus_plugins/issues/180 maybe it is because of this issue.

There's also this issue https://github.com/fluttercommunity/plus_plugins/issues/182 but I don't think it's related.

marandaneto:
apparently we need to optimize this:

https://github.com/getsentry/sentry-dart/blob/main/flutter/lib/src/default_integrations.dart#L370-L375

`packageInfo.packageName`
`packageInfo.version`
and `packageInfo.buildNumber`

should be checked if it's not null, should trim it, should only append and assign if it's nonempty.
so we assign the proper value for `release` and also only assign `dist` if there's actually a valid value.

right now `dist=""` and `release=sentry_flutter_example\u0000@1.0.0\u0000+` so we also need to strip this null char `\u0000`, could be a separate PR but so far its only noticed by the Windows impl. so @bruno-garcia you gotta do it :D

flutter/windows/sentry_flutter_plugin.cpp
```diff
@@ -0,0 +1,82 @@
...
+    std::ostringstream version_stream;
+    version_stream << "Windows ";
+    if (IsWindows10OrGreater()) {
+      version_stream << "10+";
+    } else if (IsWindows8OrGreater()) {
+      version_stream << "8";
+    } else if (IsWindows7OrGreater()) {
+      version_stream << "7";
+    }
+    result->Success(flutter::EncodableValue(version_stream.str()));
```
marandaneto: could be removed so the plugin does nothing

marandaneto:
other than that the PR looks good, thanks @bruno-garcia

ueman:
> `packageInfo.packageName`
> `packageInfo.version`
> and `packageInfo.buildNumber`
> 
> should be checked if it's not null, should trim it, should only append and assign if it's nonempty.
> so we assign the proper value for `release` and also only assign `dist` if there's actually a valid value.

Those fields are empty and not null though: https://github.com/fluttercommunity/plus_plugins/blob/main/packages/package_info_plus_platform_interface/lib/package_info_data.dart

marandaneto:
@ueman and @bruno-garcia for Linux, I just ran on my persona Ubuntu and the release is actually `@0.1.2+3`, so missing the `packageName` as it does not exist, maybe a separated PR fixing this for Linux and Windows would be ideal.
I'm not entirely sure if only `0.1.2+3` would make sense though, we could start with this but I'd check it out with the release health people, unless there's another way to get the App's name and use it instead of `packageName` when its empty

ueman:
> @ueman and @bruno-garcia for Linux, I just ran on my persona Ubuntu and the release is actually `@0.1.2+3`, so missing the `packageName` as it does not exist, maybe a separated PR fixing this for Linux and Windows would be ideal.
> I'm not entirely sure if only `0.1.2+3` would make sense though, we could start with this but I'd check it out with the release health people, unless there's another way to get the App's name and use it instead of `packageName` when its empty

flutter/example/windows/runner/Runner.rc
```diff
@@ -0,0 +1,121 @@
...
+#ifdef FLUTTER_BUILD_NUMBER
+#define VERSION_AS_NUMBER FLUTTER_BUILD_NUMBER
+#else
+#define VERSION_AS_NUMBER 1,0,0
+#endif
+
+#ifdef FLUTTER_BUILD_NAME
+#define VERSION_AS_STRING #FLUTTER_BUILD_NAME
+#else
+#define VERSION_AS_STRING "1.0.0"
```
marandaneto, line 72: @bruno-garcia wondering if changing this actually helps with this issue https://github.com/fluttercommunity/plus_plugins/issues/182

if we change it here, does it read the correct version? if so, we could document that.

marandaneto: we could see where it reads from https://github.com/fluttercommunity/plus_plugins/blob/main/packages/package_info_plus_windows/lib/src/package_info_plus_windows_real.dart

bruno-garcia: this will kick in if `FLUTTER_BUILD_NAME` isnt' defined. So Flutter must pipe already the version from the Flutter-land down to the native layer through this `define`

bruno-garcia: Sounds like there's an issue: https://github.com/flutter/flutter/issues/73652

flutter/example/windows/flutter/generated_plugin_registrant.cc
```diff
@@ -0,0 +1,12 @@
+//
```
ueman: I believe this file can be gitignored.

flutter/example/windows/flutter/generated_plugin_registrant.h
```diff
@@ -0,0 +1,13 @@
+//
```
ueman: I believe this file can be gitignored.

flutter/example/windows/flutter/generated_plugins.cmake
```diff
@@ -0,0 +1,16 @@
+#
```
ueman: I believe this file can be gitignored.

ueman (changes requested):

marandaneto:
PR looks good, a few small things to work on, but almost there, thanks @bruno-garcia

.github/workflows/flutter.yml
```diff
@@ -83,6 +86,9 @@ jobs:
             web)
             flutter build web
             ;;
+            windows)
+            flutter build windows --release
```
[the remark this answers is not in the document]
bruno-garcia: They got it right. On `dotnet` it's `--debug` by default, even on a `publish` mode 🤯

flutter/example/windows/runner/Runner.rc
bruno-garcia: Is this something we can live with for now? This version is just for the example though

bruno-garcia:
not sure why macOS is failing there, missing a file?

flutter/example/windows/runner/Runner.rc
marandaneto: yeah, lets add to the docs that Windows requires setting the release thru options or SENTRY_RELEASE

marandaneto:
> not sure why macOS is failing there, missing a file?

also fails on feature/support-desktop, @ueman is gonna look into it, IMO we could merge this into feature/support-desktop

marandaneto (approved):
LGTM

flutter/example/windows/runner/Runner.rc
bruno-garcia: Thanks!

Issue #410, opened 2021-04-08
Work around PackageInfo Limitations


ueman:
Linux und Windows have a few limitations. Notably:
- Linux has no package name
- Windows has an empty distribution

We need to figure out what a sensible default value is, if we can not work around those limitations.

marandaneto:
Current linux release: `@0.1.2+3`
Current windows release: `sentry_flutter_example\u0000@1.0.0\u0000+` -> https://www.fileformat.info/info/unicode/char/0000/index.htm

marandaneto:
macOS does it nicely `io.sentry.flutter.sentryFlutterExample@0.1.2+3`

marandaneto:
windows issue https://github.com/fluttercommunity/plus_plugins/issues/182 and https://github.com/flutter/flutter/issues/73652

marandaneto:
linux issue https://github.com/fluttercommunity/plus_plugins/issues/203

ueman:
https://www.github.com/fluttercommunity/plus_plugins/tree/main/packages%2Fpackage_info_plus_linux%2Flib%2Fsrc%2Fpackage_info.dart

https://www.github.com/fluttercommunity/plus_plugins/tree/main/packages%2Fpackage_info_plus_web%2Flib%2Fpackage_info_plus_web.dart

Web is also missing the package identifier like linux. Maybe we could use the app name if no package identifier is available.

marandaneto:
@ueman sounds a good idea
there are a few chars not allowed chars though, https://docs.sentry.io/platforms/flutter/configuration/releases/#bind-the-version
`packageName` does not allow such chars, so we have not used a regex or so, maybe with App's name this might be necessary.

ueman:
On web and linux and windows the name given in the pubspec.yaml gets used as appName: https://dart.dev/tools/pub/pubspec#name

I'm not sure though about windows and what UWP uses though.

marandaneto:
to close this issue I suggest the following:

take the Windows example:

`sentry_flutter_example\u0000@1.0.0\u0000+`

trim null chars `\u0000`;
only concat `+` if the `buildNumber` is non-empty;

result should be: `sentry_flutter_example@1.0.0`

other than that, on Windows support, we should document that the `release` should be set by either `options.release` or `SENTRY_RELEASE` env. var.

ueman:
Do we want to do 

> trim null chars \u0000;

ourself or do we fix it in package_info_plus?

[...]

PR #411, opened 2021-04-12
Fix: Use App Name if Package Name is not available

ueman:
## :scroll: Description
Not all platforms have a concept of an package id like on Android or iOS. Thus we need some other identifier. This identifier is the app name.

New test

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/flutter/lib/src/default_integrations.dart
+++ b/flutter/lib/src/default_integrations.dart
@@ -367,8 +367,13 @@ class LoadReleaseIntegration extends Integration<SentryFlutterOptions> {
       if (!kIsWeb) {
         if (options.release == null || options.dist == null) {
           final packageInfo = await _packageLoader();
+          var name = packageInfo.packageName;
+          if (name.isEmpty) {
+            name = packageInfo.appName;
+          }
+
           final release =
-              '${packageInfo.packageName}@${packageInfo.version}+${packageInfo.buildNumber}';
+              '${name}@${packageInfo.version}+${packageInfo.buildNumber}';
           options.logger(SentryLevel.debug, 'release: $release');
 
           options.release = options.release ?? release;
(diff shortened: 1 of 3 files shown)


flutter/lib/src/default_integrations.dart
```diff
@@ -367,8 +367,15 @@ class LoadReleaseIntegration extends Integration<SentryFlutterOptions> {
       if (!kIsWeb) {
         if (options.release == null || options.dist == null) {
           final packageInfo = await _packageLoader();
+          var name = packageInfo.packageName;
+          if (name.isEmpty) {
+            // Not all platforms have a packageName
+            // https://github.com/getsentry/sentry-dart/issues/410
```
marandaneto: instead of linking a GH issue, I'd rather describe the problem/solution as a comment, eg:
// not all platforms have a `packageName`, if not, take `appName` instead

flutter/test/default_integrations_test.dart
```diff
@@ -257,6 +257,40 @@ void main() {
       expect(options.release, 'foo.bar@1.2.3+789');
       expect(options.dist, '789');
     });
+
+    test('sets app name as in release if packagename is empty', () async {
+      final options = SentryFlutterOptions(dsn: fakeDsn);
```
marandaneto: we can improve the testing of this class by using the `Fixture` pattern, see file_system_transport_test.dart or load_contexts_integrations_test.dart

a `sut` that returns a new instance of `LoadReleaseIntegration` and takes a `PackageInfo` as param, so we reuse the method creation of `LoadReleaseIntegration` and `SentryFlutterOptions`

flutter/test/default_integrations_test.dart
```diff
@@ -264,3 +290,21 @@ class Fixture {
   final hub = MockHub();
   final options = SentryFlutterOptions();
 }
+
+class LoadReleaseIntegrationFixture {
```
marandaneto: this test class already has a `Fixture` class, you just need to adapt that it'd be suitable for every test, not creating a Fixture class per "test", a better example would be java and ios repos
eg https://github.com/getsentry/sentry-java/blob/main/sentry/src/test/java/io/sentry/TracesSamplerTest.kt
or just search across this repo per `Fixture`

marandaneto (approved):
This PR alone does not close https://github.com/getsentry/sentry-dart/issues/410 entirely, but it's good to be merged, LGTM.

Issue #384, opened 2021-03-25
Set debug true as first thing in the SDKs


ueman:
We need to set `SentryOptions.debug = true;` as first thing in the SDKs (Dart and Flutter) in order to not drop log messages.

See https://github.com/getsentry/sentry-dart/pull/375#discussion_r599359309 for more information.

PR #413, opened 2021-04-13
Fix: Use dartLogger as default in debug mode

ueman:
## :scroll: Description
In debug mode Sentry now uses the dartLogger (which writes to console) as a logger.
The first thing the SDK does is to create an instance if SentryOptions, so the logical place to put this logic is SentryOptions.
This change also works for Sentry and Sentry Flutter.

https://github.com/getsentry/sentry-dart/issues/384

I've written new tests.

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes


dart/test/hub_test.dart
```diff
@@ -122,12 +122,12 @@ void main() {
 
     test('should add breadcrumb to current Scope', () {
       hub.configureScope((Scope scope) {
-        expect(0, scope..breadcrumbs.length);
```
ueman, line 125: This has nothing to do with the other changes. It's just a litte code clean up.

marandaneto:
PR looks good but it's failing due to pod install, probably not related at all

dart/lib/src/sentry_options.dart
```diff
@@ -160,8 +160,18 @@ class SentryOptions {
...
+    if (checker != null) {
+      platformChecker = checker;
+    }
+
+    // In debug mode we want to log everything by default to the console.
+    // In order to do that, this must be the first thing the SDK does
+    // and the first thing the SDK does, is to instantiate SentryOptions
+    if (platformChecker.isDebugMode() && logger == noOpLogger) {
+      _logger = dartLogger;
+    }
```
marandaneto: so I think this works automatically, but what's about if the user changed the `debug` flag manually in the options callback, we still need to take that into consideration.

eg: user mutate the debug flag to `true` or `false`, and it does not matter the Flutter `debugMode`

dart/lib/src/sentry_options.dart
```diff
@@ -160,8 +160,18 @@ class SentryOptions {
...
     sdk.addPackage('pub:sentry', sdkVersion);
+    if (checker != null) {
+      platformChecker = checker;
+    }
+
+    // In debug mode we want to log everything by default to the console.
+    // In order to do that, this must be the first thing the SDK does
+    // and the first thing the SDK does, is to instantiate SentryOptions
+    if (platformChecker.isDebugMode() && logger == noOpLogger) {
+      _logger = dartLogger;
```
marandaneto: you probably want to set `debug=true` here too

dart/lib/src/sentry_options.dart
ueman: That does nothing to the logger, as it happens after the logger was set.
What do you expect to happen if the user sets debug to true/false?

marandaneto: the final decision of the user always has precedence over automatic configurations.
right now if App is in debug mode, it logs everything to the console.
if user init. the Sdk and set `debug` to false, it still logs everything (I guess)? if it works as intended, then please ignore my comment, I might have been mistaken too.

ueman: Yep, currently if the user sets debug to false it still logs everything.

Sooo we should then set the logger again after the sentry init callback. That way no logs get dropped between Sentry initialization and setting the logger based on the users settings. Does that make sense?

marandaneto: yeah that would be the behaviour I'd expect

dart/lib/src/sentry.dart
```diff
@@ -208,10 +208,15 @@ class Sentry {
...
-    // if logger os NoOp, let's set a logger that prints on the console
-    if (options.debug && options.logger == noOpLogger) {
-      options.logger = dartLogger;
+    // If the user set debug to false and the default logger is still
+    // `dartLogger` we set the logger to `noOpLogger`.
+    // `dartLogger` is the default because otherwise the logs won't be to the
+    // user visible between Sentry's initialization and the invocation of the
+    // options configuration callback.
+    if (options.debug == false && options.logger == dartLogger) {
+      options.logger = noOpLogger;
```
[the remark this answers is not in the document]
marandaneto: Dart already has a `DiagnosticLogger`, when you set a `logger` it already wraps up to a `DiagnosticLogger` that takes into consideration the `diagnosticLevel` field.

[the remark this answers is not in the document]
ueman: I think what Bruno meant is to rename the property to `diagnosticLogger` in order to make its purpose more clear.

The logger is also just a function in the Dart/Flutter SDK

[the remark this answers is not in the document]
marandaneto: we cant rename now, this would require a major bump, every call to `options.logger(params)` would change.
what @bruno-garcia meant is, instead of doing such `if` condition to set a `dartLogger` or `noOpLogger ` could be encapsulated within a setter.

so options.debug = true (means setting `dartLogger`)
options.debug = false (means setting `noOpLogger`).

I like that.

the same could happen if one sets options.logger = function(), this would automatically set `options.debug = true` and options.logger = null would set `options.debug = false`, so such conditions are not across the SDK but only setting it enabled or not.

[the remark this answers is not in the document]
bruno-garcia: > what @bruno-garcia meant is, instead of doing such if condition to set a dartLogger or noOpLogger could be encapsulated within a setter.

Yes! ;)

> the same could happen if one sets options.logger = function(), this would automatically set options.debug = true and options.logger = null would set options.debug = false

I wouldn't go this far though. One might change the logger type in an integration but not really be turning debug mode on or off.

> I think what Bruno meant is to rename the property to diagnosticLogger in order to make its purpose more clear.

Also this. What about we do this on a next major release? If so, is there a Discussion that keeps tracks of stuff we want to add to the next major?

> so options.debug = true (means setting dartLogger)
options.debug = false (means setting noOpLogger).

I'm not sure I would override the logger instance when someone sets `false` since next time you set it to `true` you no longer know what instance was set by the user before.

In .NET it's nullable, so that the call to the logger doesn't happen at all and hence no string concat or anything happens in order to invoke the logger. With NoOp you're calling things so any overhead to prepare the log entry is paid even though the logger doesn't do anything. That's just an optimization though, not saying it's something that we need to align. For really expensive operations we can always check: `logger.isEnabled(level)` before actually calling the logger.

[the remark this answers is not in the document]
ueman: > If so, is there a Discussion that keeps tracks of stuff we want to add to the next major?

Now there is: https://github.com/getsentry/sentry-dart/discussions/419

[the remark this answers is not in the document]
ueman: > With NoOp you're calling things so any overhead to prepare the log entry is paid even though the logger doesn't do anything. That's just an optimization though, not saying it's something that we need to align. For really expensive operations we can always check: logger.isEnabled(level) before actually calling the logger.

I feel like a callback for the logger would be more ergonomic than a null check.
For example:
```dart
logger.log(level, () => 'this callback gets invoked from the logger')
```
Than we would not need to have the actual code riddled with ifs and what not and also the logging is only computed when needen.

ueman:
So it's not clear to me how I should proceed here. @marandaneto @bruno-garcia

dart/lib/src/sentry.dart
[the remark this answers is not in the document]
bruno-garcia: There's an allocation for the callback though. So between the callback and NoOp I'd surely go with the latter.

bruno-garcia:
> > So it's not clear to me how I should proceed here. @marandaneto @bruno-garcia
> 
> let's discuss today in our call

Sorry I missed the call. Let me know if my input is needed, otherwise :shipit:

ueman:
Should I set `debug = true` if the user sets a custom logger?

dart/lib/src/sentry_options.dart
```diff
@@ -76,7 +76,19 @@ class SentryOptions {
...
   /// information if something goes wrong. Default is disabled.
-  bool debug = false;
+  bool get debug => _debug;
+
+  set debug(bool newValue) {
+    _debug = newValue;
+    if (_debug == true && logger == noOpLogger) {
+      _logger = dartLogger;
+    }
+    if (_debug == false && logger == dartLogger) {
```
marandaneto, line 86:
```suggestion
    if (_debug == false && logger != noOpLogger) {
```

maybe its better?

dart/lib/src/sentry_options.dart
```diff
@@ -160,9 +172,20 @@ class SentryOptions {
...
+
     sdk = SdkVersion(name: sdkName(platformChecker.isWeb), version: sdkVersion);
     sdk.addPackage('pub:sentry', sdkVersion);
+
+    // In debug mode we want to log everything by default to the console.
+    // In order to do that, this must be the first thing the SDK does
+    // and the first thing the SDK does, is to instantiate SentryOptions
+    if (platformChecker.isDebugMode()) {
+      debug = true;
+    }
```
marandaneto, line 185: maybe we move this line just after setting the `platformChecker`

dart/lib/src/sentry.dart
```diff
@@ -207,10 +205,15 @@ class Sentry {
...
```
marandaneto: this should be removed now, I guess?

ueman: Yep

dart/lib/src/sentry_options.dart
ueman: That would override the logger which the user has set. We only want to override the logger in case it wasn't altered by the user.

Or am I missing something?

marandaneto: yeah makes sense, it does not make sense to reset to default value

marandaneto (approved):
LGTM

Issue #412, opened 2021-04-12
Add request duration to HTTP breadcrumbs


bruno-garcia:
[badges.bar](https://github.com/bruno-garcia/badges.bar) calls an API to retrieve scores from pub.dev.

Would be nice to see how long these requests took. Perhaps as `data` to show under the status code:

![image](https://user-images.githubusercontent.com/1633368/114430658-ab80be00-9b8c-11eb-8de9-1543529635c9.png)

ueman:
I also left this as a todo in the code when I added this feature:
https://github.com/getsentry/sentry-dart/blob/4d42b29de33d7003a55b652230bb929f7eda5715/dart/lib/src/http_client/sentry_http_client.dart#L44-L47

What should happen if an exception gets thrown? 
How would one test that the measured duration is correct?

bruno-garcia:
> I also left this as a todo in the code when I added this feature:
> https://github.com/getsentry/sentry-dart/blob/4d42b29de33d7003a55b652230bb929f7eda5715/dart/lib/src/http_client/sentry_http_client.dart#L44-L47
> 
> What should happen if an exception gets thrown?
> How would one test that the measured duration is correct?

Would it work to just stop the stopwatch on a `finally` block? if the method returned or throw shouldn't affect the ability to track its duration, right?

ueman:
I think it should work. I wondered if it the duration shouldn't be tracked if the request fails, but after thinking twice that doesn't make much sense.

PR #414, opened 2021-04-13
Feat: SentryHttpClient tracks request durations

ueman:
## :scroll: Description
`SentryHttpClient` now tracks the duration of a request. It also logs failed requests and their durations.

https://github.com/getsentry/sentry-dart/issues/412

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/dart/lib/src/http_client/sentry_http_client.dart
+++ b/dart/lib/src/http_client/sentry_http_client.dart
@@ -53,14 +53,10 @@ class SentryHttpClient extends BaseClient {
   @override
   Future<StreamedResponse> send(BaseRequest request) async {
     // See https://develop.sentry.dev/sdk/event-payloads/breadcrumbs/
-    var breadcrumb = Breadcrumb(
-      type: 'http',
-      category: 'http',
-      data: {
-        'url': request.url.toString(),
-        'method': request.method,
-      },
-    );
+
+    var requestHadException = false;
+    int? statusCode;
+    String? reason;
 
     final stopwatch = Stopwatch();
     stopwatch.start();
@@ -68,21 +64,28 @@ class SentryHttpClient extends BaseClient {
     try {
       final response = await _client.send(request);
 
-      breadcrumb.data?.addAll({
-        'status_code': response.statusCode,
-        if (response.reasonPhrase != null) 'reason': response.reasonPhrase,
-      });
+      statusCode = response.statusCode;
+      reason = response.reasonPhrase;
 
       return response;
+    } catch (_) {
+      requestHadException = true;
+      rethrow;
     } finally {
       stopwatch.stop();
-      breadcrumb.data?['duration'] = stopwatch.elapsed.toString();
 
-      // If breadcrum.data does not contain a response status code
-      // it was an erroneous request. Set breadcrumb level to error
-      if (!(breadcrumb.data?.containsKey('status_code') ?? false)) {
-        breadcrumb = breadcrumb.copyWith(level: SentryLevel.error);
-      }
+      var breadcrumb = Breadcrumb(
+        level: requestHadException ? SentryLevel.error : SentryLevel.info,
+        type: 'http',
+        category: 'http',
+        data: {
+          'url': request.url.toString(),
+          'method': request.method,
+          if (statusCode != null) 'status_code': statusCode,
+          if (reason != null) 'reason': reason,
+          'duration': stopwatch.elapsed.toString(),
+        },
+      );
 
       _hub.addBreadcrumb(breadcrumb);
     }
--- a/dart/lib/src/http_client/sentry_http_client.dart
+++ b/dart/lib/src/http_client/sentry_http_client.dart
@@ -52,18 +52,39 @@ class SentryHttpClient extends BaseClient {
 
   @override
   Future<StreamedResponse> send(BaseRequest request) async {
+    // See https://develop.sentry.dev/sdk/event-payloads/breadcrumbs/
+    var breadcrumb = Breadcrumb(
+      type: 'http',
+      category: 'http',
+      data: {
+        'url': request.url.toString(),
+        'method': request.method,
+      },
+    );
+
     final stopwatch = Stopwatch();
     stopwatch.start();
+
     try {
       final response = await _client.send(request);
-      stopwatch.stop();
-      _logSuccessBreadcrumb(request, response, stopwatch);
+
+      breadcrumb.data?.addAll({
+        'status_code': response.statusCode,
+        if (response.reasonPhrase != null) 'reason': response.reasonPhrase,
+      });
 
       return response;
-    } catch (e) {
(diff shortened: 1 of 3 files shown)


dart/lib/src/http_client/sentry_http_client.dart
```diff
@@ -55,7 +52,32 @@ class SentryHttpClient extends BaseClient {
...
+    stopwatch.start();
+    try {
+      final response = await _client.send(request);
+      stopwatch.stop();
+      _logSuccessBreadcrumb(request, response, stopwatch);
+
+      return response;
+    } catch (e) {
+      stopwatch.stop();
+      _logFailedBreadcrumb(request, stopwatch);
```
marandaneto: `stop` and `_logFailedBreadcrumb` could be under a finally block

dart/test/http_client/sentry_http_client_test.dart
```diff
@@ -183,10 +176,37 @@ void main() {
...
+        expect(request.url, url);
+        throw Exception('foo bar');
+      });
+
+      final client = SentryHttpClient(client: mockClient, hub: mockHub);
+
+      try {
+        await client.get(Uri.parse('https://example.com'));
+        fail('Method did not throw');
+      } on Exception catch (_) {}
```
marandaneto: not sure if all but most of this common code across tests could be improved with the fixture pattern, lets always try to keep it clean, so when touching test classes that don't have it, we try to migrate to not duplicate even more things.

ueman: Yup, already on it. I knew you would say that 👍

dart/lib/src/http_client/sentry_http_client.dart
```diff
@@ -55,22 +52,40 @@ class SentryHttpClient extends BaseClient {
...
+      return response;
+    } finally {
+      stopwatch.stop();
+      breadcrumb.data?['duration'] = stopwatch.elapsed.toString();
+
+      // If breadcrum.data does not contain a response status code
+      // it was an erroneous request. Set breadcrumb level to error
+      if (!(breadcrumb.data?.containsKey('status_code') ?? false)) {
+        breadcrumb = breadcrumb.copyWith(level: SentryLevel.error);
+      }
```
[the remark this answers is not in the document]
marandaneto: also `copyWith` would allocate everything again, maybe it makes sense to allocate the mutable vars to local vars and only create the breadcrumb in the finally block.

eg https://github.com/getsentry/sentry-java/blob/main/sentry-android-okhttp/src/main/java/io/sentry/android/okhttp/SentryOkHttpInterceptor.kt#L40-L47

dart/test/http_client/sentry_http_client_test.dart
```diff
@@ -199,7 +180,48 @@ void main() {
...
+class Fixture {
+  Fixture() {
+    hub = MockHub();
+
+    final mockClient = MockClient((request) async {
+      expect(request.url, requestUri);
+      return Response('', response, reasonPhrase: reasonPhrase);
+    });
+
+    client = SentryHttpClient(client: mockClient, hub: hub);
```
marandaneto: the tested class is always created under a `getSut()` method and not a global instance of `Fixture`, this would cause problems if tests are running in parallel.
you only need to create a fresh instance of `Fixture` in `setUp` if you are testing static variables, otherwise, only one instance of `Fixture` is needed, and a new `sut` is created for each test.

dart/test/http_client/sentry_http_client_test.dart
```diff
@@ -199,7 +180,48 @@ void main() {
       expect(mockHub.captureExceptionCalls.length, 0);
       verify(mockClient.close());
     });
+
+    test('Breadcrumb has correct duration', () async {
+      final mockHub = MockHub();
```
marandaneto: this should have come from the fixture

dart/test/http_client/sentry_http_client_test.dart
```diff
@@ -199,7 +180,48 @@ void main() {
...
+
+      final mockClient = MockClient((request) async {
+        expect(request.url, requestUri);
+        await Future.delayed(Duration(seconds: 1));
+        return Response('', 200, reasonPhrase: 'OK');
+      });
+
+      final client = SentryHttpClient(client: mockClient, hub: mockHub);
```
marandaneto: this should come form a getSut() method

dart/test/http_client/sentry_http_client_test.dart
```diff
@@ -199,7 +168,49 @@ void main() {
...
+
+    client = SentryHttpClient(client: mc, hub: hub);
+  }
+
+  late MockHub hub;
+  late SentryHttpClient client;
+
+  int response = 200;
+  String? reasonPhrase;
+}
```
marandaneto: this is still not following the fixture pattern, it should not have a static method, let me refactor for you

marandaneto:
```
// only a single instance of Fixture
// void main() {
//  final fixture = Fixture();
//}
class Fixture {
  
  // every test does final sut = fixture.getSut() either passing MockClient or not
  SentryHttpClient getSut({MockClient? mockClient) {
        final mc = mockClient ?? getClient();
        return SentryHttpClient(client: mc, hub: hub);
  }

  // tests could assert or expect fixture.hub
  late MockHub hub = MockHub();

  // could optionally have reasonPhrase, status etc with defaults values
  // so tests could also do sut = fixture.getSut(fixture.getClient(500, 'my reason'))
  MockClient getClient() {
    MockClient((request) async {
          expect(request.url, requestUri);
          return Response('', 200, reasonPhrase: null);
        });
  }
}
```

marandaneto: guess you can have a better idea now, let me know if its not clear

ueman: Thanks 😅

dart/test/http_client/sentry_http_client_test.dart
```diff
@@ -199,7 +176,45 @@ void main() {
...
+  SentryHttpClient getSut([MockClient? client]) {
+    final mc = client ?? getClient();
+    return SentryHttpClient(client: mc, hub: hub);
+  }
+
+  // tests could assert or expect fixture.hub
+  late MockHub hub = MockHub();
+
+  // could optionally have reasonPhrase, status etc with defaults values
+  // so tests could also do sut = fixture.getSut(fixture.getClient(500, 'my reason'))
```
marandaneto: the goal of the comments were just to explain you, we can remove from source code

marandaneto (approved):
nice, thanks

ueman:
Issue with duration for reference: https://sentry.io/share/issue/2a6d6dd55b5e47c087c576bd2c538413/

PR #420, opened 2021-04-19
Fix: Trim Unicode NULL character for Windows support

ueman:
## :scroll: Description
I replace every \u0000 with `` (empty string).

I've written new tests

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps
@bruno-garcia Could you please test this?

### DIFF
--- a/flutter/lib/src/default_integrations.dart
+++ b/flutter/lib/src/default_integrations.dart
@@ -373,12 +373,24 @@ class LoadReleaseIntegration extends Integration<SentryFlutterOptions> {
           name = _cleanAppName(packageInfo.appName);
         }
 
-        final release =
-            '$name@${packageInfo.version}+${packageInfo.buildNumber}';
+        final version = _cleanAppName(packageInfo.version);
+        final buildNumber = _cleanAppName(packageInfo.buildNumber);
+
+        var release = name;
+        if (version.isNotEmpty) {
+          release = '$release@$version';
+        }
+        // At least windows sometimes does not have a buildNumber
+        if (buildNumber.isNotEmpty) {
+          release = '$release+$buildNumber';
+        }
+
         options.logger(SentryLevel.debug, 'release: $release');
 
         options.release = options.release ?? release;
-        options.dist = options.dist ?? packageInfo.buildNumber;
+        if (buildNumber.isNotEmpty) {
+          options.dist = options.dist ?? packageInfo.buildNumber;
+        }
       }
     } catch (error) {
       options.logger(
@@ -397,6 +409,7 @@ class LoadReleaseIntegration extends Integration<SentryFlutterOptions> {
         .replaceAll('\t', '_')
         .replaceAll('\r\n', '_')
         .replaceAll('\r', '_')
-        .replaceAll('\n', '_');
+        .replaceAll('\n', '_')
+        .replaceAll('\u{0000}', '');
   }
 }
--- a/flutter/lib/src/default_integrations.dart
+++ b/flutter/lib/src/default_integrations.dart
@@ -370,11 +370,11 @@ class LoadReleaseIntegration extends Integration<SentryFlutterOptions> {
         if (name.isEmpty) {
           // Not all platforms have a packageName.
           // If no packageName is available, use the appName instead.
-          name = _cleanAppName(packageInfo.appName);
+          name = _cleanString(packageInfo.appName);
         }
 
-        final version = _cleanAppName(packageInfo.version);
-        final buildNumber = _cleanAppName(packageInfo.buildNumber);
+        final version = _cleanString(packageInfo.version);
+        final buildNumber = _cleanString(packageInfo.buildNumber);
 
         var release = name;
         if (version.isNotEmpty) {
@@ -400,9 +400,13 @@ class LoadReleaseIntegration extends Integration<SentryFlutterOptions> {
     options.sdk.addIntegration('loadReleaseIntegration');
   }
 
-  String _cleanAppName(String appName) {
+  /// This method cleans the given string from characters which should not be
+  /// used.
+  /// For example https://docs.sentry.io/platforms/flutter/configuration/releases/#bind-the-version
+  /// imposes some requirements. Also Windows uses some characters which
+  /// should not be used.
+  String _cleanString(String appName) {
     // Replace disallowed chars with an underscore '_'
-    // https://docs.sentry.io/platforms/flutter/configuration/releases/#bind-the-version
     return appName
         .replaceAll('/', '_')
         .replaceAll('\\', '_')
@@ -410,6 +414,7 @@ class LoadReleaseIntegration extends Integration<SentryFlutterOptions> {
         .replaceAll('\r\n', '_')
         .replaceAll('\r', '_')
         .replaceAll('\n', '_')
+        // replace Unicode NULL character with an empty string
         .replaceAll('\u{0000}', '');
   }
 }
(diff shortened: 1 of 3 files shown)


flutter/lib/src/default_integrations.dart
```diff
@@ -373,12 +373,24 @@ class LoadReleaseIntegration extends Integration<SentryFlutterOptions> {
           name = _cleanAppName(packageInfo.appName);
         }
 
-        final release =
-            '$name@${packageInfo.version}+${packageInfo.buildNumber}';
+        final version = _cleanAppName(packageInfo.version);
+        final buildNumber = _cleanAppName(packageInfo.buildNumber);
```
marandaneto: maybe `_cleanAppName` should be renamed now, its not only about app name

flutter/lib/src/default_integrations.dart
```diff
@@ -397,6 +409,7 @@ class LoadReleaseIntegration extends Integration<SentryFlutterOptions> {
         .replaceAll('\t', '_')
         .replaceAll('\r\n', '_')
         .replaceAll('\r', '_')
-        .replaceAll('\n', '_');
+        .replaceAll('\n', '_')
+        .replaceAll('\u{0000}', '');
```
marandaneto, line 418: we need to be sure that this works on Windows, maybe @bruno-garcia can test it?

flutter/lib/src/default_integrations.dart
```diff
@@ -370,15 +370,27 @@ class LoadReleaseIntegration extends Integration<SentryFlutterOptions> {
         if (name.isEmpty) {
           // Not all platforms have a packageName.
           // If no packageName is available, use the appName instead.
-          name = _cleanAppName(packageInfo.appName);
+          name = _cleanString(packageInfo.appName);
+        }
+
+        final version = _cleanString(packageInfo.version);
```
bruno-garcia, line 376: This block is guarded by _either_ `release` or `dist` being empty on options. But below we only take the `release` built in here, if `options.release` is not empty, we just use that, and same for `dist`. 

So if only `options.release` is set, we'll build a new release (as per this block) but won't use it at all.
Would it make sense to check if `+` exists and append the `buildNumber` we detected in that case?

flutter/lib/src/default_integrations.dart
```diff
@@ -370,15 +370,27 @@ class LoadReleaseIntegration extends Integration<SentryFlutterOptions> {
...
         }
 
-        final release =
-            '$name@${packageInfo.version}+${packageInfo.buildNumber}';
         options.logger(SentryLevel.debug, 'release: $release');
 
         options.release = options.release ?? release;
-        options.dist = options.dist ?? packageInfo.buildNumber;
+        if (buildNumber.isNotEmpty) {
+          options.dist = options.dist ?? packageInfo.buildNumber;
```
bruno-garcia: We might need check the format of build number. Will this always be in a valid format as per Sentry's spec? 
https://docs.sentry.io/platforms/dart/configuration/releases/#bind-the-version

bruno-garcia (commented):
The fix works! :)

image

https://sentry.io/organizations/sentry-sdks/issues/2137135923/?query=is%3Aunresolved

flutter/lib/src/default_integrations.dart
```suggestion
          options.dist = options.dist ?? buildNumber;
```

good catch, maybe use `buildNumber` which is "cleaned" already

marandaneto:
> The fix works! :)
> 
> image
> 
> https://sentry.io/organizations/sentry-sdks/issues/2137135923/?query=is%3Aunresolved

Screenshot 2021-04-19 at 22 11 22

If I opened the right JSON, `sentry_flutter_example\u0000@1.0.0` its still wrong, could you double check @bruno-garcia ?

bruno-garcia:
oof, you're right @marandaneto 

```json
  "release": "sentry_flutter_example\u0000@1.0.0",
```

flutter/lib/src/default_integrations.dart
marandaneto: I think the `??` does the job if I got what you meant eg

`options.release = options.release ?? release;`

but what we could do is to actually check if we need to build the `release` or `dist` only and not execute code that is not necessary, eg, building a new `release` if we're going to use the `options.release` anyway as its not empty

marandaneto:
> There's no general 'error processing your event' on the event though. Wasn't that the case before this change?

before it was `sentry_flutter_example\u0000@1.0.0\u0000+`

ueman:
> There's no general 'error processing your event' on the event though. Wasn't that the case before this change?

I think that's because `options.dist` does not get set to an empty string anymore.

flutter/lib/src/default_integrations.dart
```diff
@@ -370,15 +370,27 @@ class LoadReleaseIntegration extends Integration<SentryFlutterOptions> {
         if (name.isEmpty) {
```
marandaneto, line 370: @bruno-garcia I think the bug is here, should be isNotEmpty

marandaneto, line 370:
```suggestion
        if (name.isNotEmpty) {
```

flutter/lib/src/default_integrations.dart
bruno-garcia: yeah it's already using options.release if it's already set, it's just odd that we build a new release value even though it's not used.

And I was trying to think of ways to combine what we built in the block with what was defined but best not mess around with a value that was explicitly set by the user on options.

So the only improvement here is as you said just to avoid doing the work if not needed though it'll make the logic more complicated

flutter/lib/src/default_integrations.dart
bruno-garcia: I dream of integration test on the target platform that will validate these for us. Mainly because of regressions introduced later.

Thoughts? Q2 is the time to plan and book these tasks

flutter/lib/src/default_integrations.dart
ueman: I don't think so. It would mean that we always use the app name if the package name is not empty.

ueman: `var name = packageInfo.packageName;` This was the culprit.

flutter/lib/src/default_integrations.dart
ueman: See https://github.com/getsentry/sentry-dart/issues/422

flutter/test/default_integrations_test.dart
```diff
@@ -283,6 +284,69 @@ void main() {
...
+          packageName: 'sentry_flutter_example\u{0000}',
+          version: '',
+          buildNumber: '123\u{0000}',
+        ));
+      };
+      await fixture
+          .getIntegration(loader: loader)
+          .call(MockHub(), fixture.options);
+
+      expect(fixture.options.release, 'sentry_flutter_example+123');
```
marandaneto, line 332: I think this is wrong, in this case, if there's no `version`, it should be `sentry_flutter_example@123`

marandaneto: I'll confirm that, let's keep as it is for now

marandaneto (approved):
LGTM

flutter/test/default_integrations_test.dart
marandaneto: this is actually not going to happen because pubspec must have a `version` or it defaults to 0.0.0 https://dart.dev/tools/pub/pubspec#version
its only possible because we've mocked the values, so no special anything case here.

Issue #390, opened 2021-03-29
Improve test which check that integrations are added correctly


ueman:
We do have a lot of tests, which check that `Integration`s are added based on the platform. They work but they should be improved so that they don't rely on magic numbers.

See for some more context: https://github.com/getsentry/sentry-dart/pull/389#discussion_r603084182_

ueman:
In addition to this these options https://github.com/getsentry/sentry-dart/blob/de0a2ce7229bbed9e071a9beddb4153ba65fe46e/flutter/lib/src/sentry_flutter_options.dart#L173 should probably migrated to the `Platform` class introduced by https://github.com/getsentry/sentry-dart/pull/389

PR #421, opened 2021-04-19
Fix: Improve Sentry Integration tests

ueman:
## :scroll: Description
This PR improves the tests which make sure that every platform uses the correct integrations.

__#skip-changelog__

https://github.com/getsentry/sentry-dart/issues/390

I've written new tests.

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes


flutter/test/sentry_flutter_util.dart
```diff
@@ -9,60 +9,53 @@ import 'package:sentry_flutter/src/version.dart';
...
+  required List<Type> shouldHaveIntegrations,
+  required List<Type> shouldNotHaveIntegrations,
+  required bool hasFileSystemTransport,
 }) =>
     (options) async {
-      assert(options is SentryFlutterOptions);
-
       options.dsn = fakeDsn;
 
       expect(kDebugMode, options.debug);
```
ueman: I think this should now probably go into a dedicated test.

flutter/test/sentry_flutter_test.dart
```diff
@@ -1,128 +1,143 @@
...
-
 import 'mocks.dart';
-import 'mocks.mocks.dart';
 import 'sentry_flutter_util.dart';
 
-void main() {
-  TestWidgetsFlutterBinding.ensureInitialized();
+/// These are the integrations which should be added on every platform.
+/// They don't depend on the underlying platform.
+final platformAgnosticIntegrations = [
```
ueman, line 13: I added a lists for every platform and type of integration. Based on these lists we know which platform should have which integration. I think it quite easy to change if add support for more integrations (or add integrations on existing platforms).

flutter/test/sentry_flutter_test.dart
```diff
@@ -1,128 +1,143 @@
...
+        platformChecker: getPlatformChecker(platform: MockPlatform.windows()),
       );
-
-      await Sentry.captureMessage('a message');
-
-      final event =
-          verify(transport.send(captureAny)).captured.first as SentryEvent;
-
-      expect(event.sdk!.integrations.length, 7);
-      expect(event.sdk!.integrations.contains('loadContextsIntegration'), true);
```
marandaneto, line 81: this should be tested on load_contexts_integrations_test if not yet

flutter/test/sentry_flutter_test.dart
```diff
@@ -1,128 +1,143 @@
...
-
-      await Sentry.captureMessage('a message');
-
-      final event =
-          verify(transport.send(captureAny)).captured.first as SentryEvent;
-
-      expect(event.sdk!.integrations.length, 7);
-      expect(
-        event.sdk!.integrations.contains('loadContextsIntegration'),
-        false,
```
marandaneto, line 102: this should be tested on load_contexts_integrations_test if not yet

flutter/test/sentry_flutter_test.dart
```diff
@@ -1,128 +1,143 @@
...
-
-      await Sentry.captureMessage('a message');
-
-      final event =
-          verify(transport.send(captureAny)).captured.first as SentryEvent;
-
-      expect(event.sdk!.integrations.length, 7);
-      expect(
-        event.sdk!.integrations.contains('loadAndroidImageListIntegration'),
-        false,
```
marandaneto, line 125: should be tested on load_android_image_list_test

flutter/test/sentry_flutter_util.dart
marandaneto: makes sense

flutter/test/sentry_flutter_test.dart
ueman: This is covered by the rewritten tests.

flutter/test/load_android_image_list_test.dart
```diff
@@ -37,6 +37,19 @@ void main() {
     called = false;
   });
 
+  test('$LoadAndroidImageListIntegration adds itself to sdk.integrations',
+      () async {
+    final options = SentryFlutterOptions()..dsn = fakeDsn;
+    final hub = Hub(options);
+
+    LoadAndroidImageListIntegration(_channel)(hub, options);
```
marandaneto: let's make a fixture

ueman: Sure

flutter/test/load_android_image_list_test.dart
```diff
@@ -125,3 +132,21 @@ SentryEvent getEvent({bool symbolicated = false}) {
...
+    channel = MethodChannel('sentry_flutter');
+    options = SentryFlutterOptions()..dsn = fakeDsn;
+    hub = Hub(options);
+  }
+
+  late final MethodChannel channel;
+  late final SentryFlutterOptions options;
+
+  late final Hub hub;
+  var called = false;
```
marandaneto: I believe this should not be part of the Fixture, this is per test I'd say, fixture should not have state

flutter/test/load_android_image_list_test.dart
```diff
@@ -125,3 +132,21 @@ SentryEvent getEvent({bool symbolicated = false}) {
   );
   return SentryEvent(exception: ex);
 }
+
+class Fixture {
+  Fixture() {
+    channel = MethodChannel('sentry_flutter');
+    options = SentryFlutterOptions()..dsn = fakeDsn;
+    hub = Hub(options);
```
marandaneto: every other fixture initializes the fields directly, why is it within a ctor and making them `late`, is there any benefit in Dart?

marandaneto: could `Hub` be a mock instead or the real impl?

ueman: I can't initialize `Hub` in a field because it needs `SentryOptions`. I think all other properties could be initialized in a field.

ueman: Nope, i tried that and the tests failed. Some tests need the event processor to get executed which the mock does not have.

ueman:
I would like to land #413 before this, so I can implement https://github.com/getsentry/sentry-dart/issues/390#issuecomment-810164155 in this PR.

flutter/test/load_android_image_list_test.dart
```diff
@@ -125,3 +148,18 @@ SentryEvent getEvent({bool symbolicated = false}) {
...
+
+class Fixture {
+  Fixture() {
+    hub = Hub(options);
+  }
+
+  late final channel = MethodChannel('sentry_flutter');
+  late final options = SentryFlutterOptions(dsn: fakeDsn);
+
+  late final Hub hub;
```
marandaneto: in theory, `hub` is created right away, `hub` requires `options`, so using `late` for both of them is just semantically wrong

marandaneto (approved):
left a suggestion, other than that, LGTM

flutter/test/load_android_image_list_test.dart
```diff
@@ -125,3 +148,18 @@ SentryEvent getEvent({bool symbolicated = false}) {
...
+
+class Fixture {
+  Fixture() {
+    hub = Hub(options);
+  }
+
+  final channel = MethodChannel('sentry_flutter');
+  final options = SentryFlutterOptions(dsn: fakeDsn);
+
+  late final Hub hub;
```
marandaneto, line 160:
```suggestion
  late final channel = MethodChannel('sentry_flutter');
  final options = SentryFlutterOptions(dsn: fakeDsn);

  final Hub hub;
```

marandaneto: that would be semantically correct

ueman: I can't make the Hub final though and the MethodChannel doesn't need to be late.

marandaneto: `hub` is already final, I just removed the `late`.
`channel` does not have to be `late`, but it can, its used in `setUp` so indeed, making it `late` isn't any helpful

ueman: The `Hub` is inited in the constructor so it has to be `late final` otherwise it would cause a compiler error.
The `Hub` can't be inited outside the constructor because it needs the SentryOptions. 
So it is correct as is :D

marandaneto: all good then, my bad, thanks :D

PR #439, opened 2021-04-27
Feature: Log `print()` calls as Breadcrumbs

### LINKED ISSUE #423: Log calls to `print()` as Breadcrumbs
One can add a [`PrintHandler`](https://api.dart.dev/stable/2.12.4/dart-async/PrintHandler.html) to a [`Zone`](https://api.dart.dev/stable/2.12.4/dart-async/Zone/print.html). This allows us to intercept all(?) calls to `print()` and log them as Breadcrumbs. 

Also see https://api.dart.dev/stable/2.12.4/dart-async/runZonedGuarded.html

Also see https://github.com/getsentry/sentry-dart/projects/1#card-50166788

ueman:
## :scroll: Description
We can now record calls to `print()` as Breadcrumbs.

- One can easily run into recursion. I added a tests so that it should not happen. See comments.
- Should this be an option in `SentryOptions`? -> Yes

Example event: https://sentry.io/share/issue/9aba382a0ae649baafaea54522729709/

Closes https://github.com/getsentry/sentry-dart/issues/423

I've written new tests.

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/dart/lib/src/default_integrations.dart
+++ b/dart/lib/src/default_integrations.dart
@@ -36,9 +36,9 @@ class RunZonedGuardedIntegration extends Integration {
       },
       zoneSpecification: ZoneSpecification(
         print: (self, parent, zone, line) {
-          if (!hub.isEnabled) {
+          if (!hub.isEnabled || !options.enablePrintBreadcrumbs) {
             // early bail out, in order to better guard against the recursion
-            // as decribed below.
+            // as described below.
             parent.print(zone, line);
             return;
           }
--- a/dart/lib/src/default_integrations.dart
+++ b/dart/lib/src/default_integrations.dart
@@ -13,22 +13,59 @@ class RunZonedGuardedIntegration extends Integration {
 
   final Future<void> Function() _runner;
 
+  /// Needed to check if we somehow caused a `print()` recursion
+  bool _isPrinting = false;
+
   @override
   FutureOr<void> call(Hub hub, SentryOptions options) {
-    runZonedGuarded(() async {
-      await _runner();
-    }, (exception, stackTrace) async {
-      // runZonedGuarded doesn't crash the App.
-      final mechanism = Mechanism(type: 'runZonedGuarded', handled: true);
-      final throwableMechanism = ThrowableMechanism(mechanism, exception);
+    runZonedGuarded(
+      () async {
+        await _runner();
+      },
+      (exception, stackTrace) async {
+        // runZonedGuarded doesn't crash the App.
+        final mechanism = Mechanism(type: 'runZonedGuarded', handled: true);
+        final throwableMechanism = ThrowableMechanism(mechanism, exception);
 
-      final event = SentryEvent(
-        throwable: throwableMechanism,
-        level: SentryLevel.fatal,
-      );
+        final event = SentryEvent(
+          throwable: throwableMechanism,
+          level: SentryLevel.fatal,
+        );
 
-      await hub.captureEvent(event, stackTrace: stackTrace);
-    });
+        await hub.captureEvent(event, stackTrace: stackTrace);
+      },
+      zoneSpecification: ZoneSpecification(
+        print: (self, parent, zone, line) {
+          if (!hub.isEnabled) {
+            // early bail out, in order to better guard against the recursion
+            // as decribed below.
+            parent.print(zone, line);
+            return;
+          }
+
+          if (_isPrinting) {
+            // We somehow landed in a recursion.
+            // This happens for example if:
+            // hub.addBreadcrumb() called print() itself.
+            // This happens for example if hub.isEnabled == false and
+            // options.logger == dartLogger
+            //
+            // Anyway, in order to not cause a stack overflow due to recursion
+            // we drop any print() call while adding a breadcrumb.
+            return;
+          }
+          _isPrinting = true;
+          try {
+            hub.addBreadcrumb(Breadcrumb(
+              message: line,
+            ));
+            parent.print(zone, line);
+          } finally {
+            _isPrinting = false;
+          }
+        },
+      ),
+    );
 
     options.sdk.addIntegration('runZonedGuardedIntegration');
   }
--- a/dart/lib/src/default_integrations.dart
+++ b/dart/lib/src/default_integrations.dart
@@ -6,8 +6,12 @@ import 'protocol.dart';
 import 'sentry_options.dart';
 import 'throwable_mechanism.dart';
 
-/// integration that runs runner function within runZonedGuarded and capture
-/// errors on the runZonedGuarded error handler
+/// Integration that runs runner function within `runZonedGuarded` and capture
+/// errors on the `runZonedGuarded` error handler.
+/// See https://api.dart.dev/stable/dart-async/runZonedGuarded.html
+///
+/// This integration also records calls to `print()` as Breadcrumbs.
+/// This can be configured with [SentryOptions.enablePrintBreadcrumbs]
 class RunZonedGuardedIntegration extends Integration {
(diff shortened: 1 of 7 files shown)


dart/test/default_integrations_test.dart
```diff
@@ -108,9 +108,65 @@ void main() {
     expect(true, throwableMechanism.mechanism.handled);
     expect(throwable, throwableMechanism.throwable);
   });
+
+  test('Run zoned guarded logs calls to print as breadcrumb', () async {
+    Future<void> callback() async {
+      print('foo bar');
+    }
+
+    final integration = RunZonedGuardedIntegration(callback);
```
marandaneto: move this to a getSut in the fixture class

dart/test/default_integrations_test.dart
```diff
@@ -108,9 +108,65 @@ void main() {
...
+
+    await integration(fixture.hub, fixture.options);
+
+    expect(fixture.hub.addBreadcrumbCalls.length, 0);
+  });
+
+  test('Run zoned guarded: No recursion for print() calls', () async {
+    final options = SentryOptions(dsn: fakeDsn);
+    final hub = PrintRecursionMockHub();
```
marandaneto, line 148: should be read from fixture

dart/lib/src/default_integrations.dart
```diff
@@ -13,22 +13,59 @@ class RunZonedGuardedIntegration extends Integration {
...
+          }
+          _isPrinting = true;
+          try {
+            hub.addBreadcrumb(Breadcrumb(
+              message: line,
+            ));
+            parent.print(zone, line);
+          } finally {
+            _isPrinting = false;
+          }
```
marandaneto: `hub.addBreadcrumb` never throws, does `parent.print` may throw? otherwise we'd not need a try/catch

dart/lib/src/default_integrations.dart
```diff
@@ -13,22 +13,59 @@ class RunZonedGuardedIntegration extends Integration {
...
+
+        final event = SentryEvent(
+          throwable: throwableMechanism,
+          level: SentryLevel.fatal,
+        );
+
+        await hub.captureEvent(event, stackTrace: stackTrace);
+      },
+      zoneSpecification: ZoneSpecification(
+        print: (self, parent, zone, line) {
```
marandaneto, line 42: lets also add a option check -> https://github.com/getsentry/sentry-dart/issues/423#issuecomment-827410560

dart/lib/src/default_integrations.dart
```diff
@@ -13,22 +13,59 @@ class RunZonedGuardedIntegration extends Integration {
 
   final Future<void> Function() _runner;
 
+  /// Needed to check if we somehow caused a `print()` recursion
+  bool _isPrinting = false;
```
marandaneto, line 21: this might need to be an atomic var if print could be called concurrently across different threads (isolates)?

dart/lib/src/default_integrations.dart
ueman: I think it could throw if we cause a recursion.

dart/lib/src/default_integrations.dart
marandaneto: not even sure if it'd be called again from a different isolate anyway, nor I know if Dart has Atomic vars, food for thought though

dart/lib/src/default_integrations.dart
```diff
@@ -13,22 +13,59 @@ class RunZonedGuardedIntegration extends Integration {
...
+        final event = SentryEvent(
+          throwable: throwableMechanism,
+          level: SentryLevel.fatal,
+        );
+
+        await hub.captureEvent(event, stackTrace: stackTrace);
+      },
+      zoneSpecification: ZoneSpecification(
+        print: (self, parent, zone, line) {
+          if (!hub.isEnabled || !options.enablePrintBreadcrumbs) {
```
```suggestion
          if (!options.enablePrintBreadcrumbs || !hub.isEnabled) {
```
its cheaper to check the boolean first, I guess

dart/lib/src/default_integrations.dart
```diff
@@ -13,22 +13,59 @@ class RunZonedGuardedIntegration extends Integration {
...
+
+          if (_isPrinting) {
+            // We somehow landed in a recursion.
+            // This happens for example if:
+            // hub.addBreadcrumb() called print() itself.
+            // This happens for example if hub.isEnabled == false and
+            // options.logger == dartLogger
+            //
+            // Anyway, in order to not cause a stack overflow due to recursion
+            // we drop any print() call while adding a breadcrumb.
```
marandaneto: maybe lets drop a message here:

`parent.print(zone,'recursion during print call');` or something like that

dart/lib/src/default_integrations.dart
```diff
@@ -13,22 +13,59 @@ class RunZonedGuardedIntegration extends Integration {
...
+            return;
+          }
+          _isPrinting = true;
+          try {
+            hub.addBreadcrumb(Breadcrumb(
+              message: line,
+            ));
```
marandaneto: maybe we set level as `debug`? unfortunately `ZoneSpecification` does not expose a `level`, so lets assume the lowest instead of `info` which is our default.
for type, probably `debug` or keep without it which defaults to `default`, see https://develop.sentry.dev/sdk/event-payloads/breadcrumbs/#breadcrumb-types

for category, either without it or `console`?

@bruno-garcia ideas?

dart/lib/src/default_integrations.dart
ueman: Yeah, that's probably a good idea

dart/lib/src/default_integrations.dart
ueman: I also asked on discord.
Someone should finally do https://github.com/getsentry/develop/issues/213 :D

bruno-garcia: I have no better ideas here. Makes sense what Manoel said.
`info` sounds fair as a default. One idea: `category` could be `print`?

Imagine that you want to implement `beforeBreadcrumb` and want to filter out some of these `print` statements. You could filter first by category=print then regex on the content for example

dart/lib/src/default_integrations.dart
bruno-garcia: the `finally` here also communicates intent really well: No matter way we need to unflip this flag not to be in a broken state

dart/lib/src/default_integrations.dart
marandaneto: I'd still make it with existent values which are:

level = debug
type = debug
category = console

those values are already pre-defined here https://develop.sentry.dev/sdk/event-payloads/breadcrumbs/#breadcrumb-types

`category=print` would make sense but isn't documented so UI defaults to `default` mode

dart/lib/src/default_integrations.dart
ueman: Because this is private you can't set this in any other Isolate. Different Isolates have different Zones, so this won't be called.

flutter/example/lib/main.dart
```diff
@@ -113,6 +113,12 @@ class MainScaffold extends StatelessWidget {
               child: const Text('Dart: Web request'),
               onPressed: () => makeWebRequest(context),
             ),
+            RaisedButton(
+              child: const Text('Record print() as breadcrumb'),
+              onPressed: () {
+                print('Foo bar');
```
marandaneto: maybe we also Sentry.captureMessage('a message with a foo ar breadcrumb') so we can check on Sentry directly

```suggestion
                print('A print breadcrumb');
```

marandaneto (approved):
LGTM

PR #445, opened 2021-05-02
Bump: Sentry Cocoa to 7.0.3

ueman:
## :scroll: Description
Bump Sentry Cocoa to 7.0.3

New Cocoa Release

With the example app

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 * Feat: Support for Out-of-Memory-Tracking on macOS/iOS (#424)
 * Fix: Trim `\u0000` from Windows package info (#420)
 * Fix: `dist` was read from `SENTRY_DSN`, now it's read from `SENTRY_DIST` (#442)
+* Bump: sentry-cocoa to v7.0.3 (#443)
 
 # 5.0.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@
 * Feat: Support for Out-of-Memory-Tracking on macOS/iOS (#424)
 * Fix: Trim `\u0000` from Windows package info (#420)
 * Fix: `dist` was read from `SENTRY_DSN`, now it's read from `SENTRY_DIST` (#442)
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -12,6 +12,7 @@
 * Feat: Support for Out-of-Memory-Tracking on macOS/iOS (#424)
 * Fix: Trim `\u0000` from Windows package info (#420)
 * Fix: `dist` was read from `SENTRY_DSN`, now it's read from `SENTRY_DIST` (#442)
+* Bump: sentry-cocoa to v7.0.3 (#443)
```
```suggestion
* Bump: sentry-cocoa to v7.0.3 (#445)
```

flutter/macos/sentry_flutter.podspec
```diff
@@ -12,7 +12,7 @@ Sentry SDK for Flutter with support to native through sentry-cocoa.
                          :tag => s.version.to_s }
   s.source_files     = 'Classes/**/*'
   s.public_header_files = 'Classes/**/*.h'
-  s.dependency 'Sentry', '~> 7.0.0'
+  s.dependency 'Sentry', '~> 7.0.3'
```
marandaneto, line 15: are these 2 files totally equal? ios and macos, can we make a symlink too? or this is actually required because of publishing

marandaneto (approved):

flutter/macos/sentry_flutter.podspec
ueman: Hm, that's a good question. I think technically we can make the content work for iOS and macOS with just one file.
Two possible issues I see are
- CocoaPods has some issues with symlinks, that's the reason why I didn't symlink the source folder but all files individually.
- The files are copied into the app project either via Cocoapods or via Flutter itself. I don't know if that works with symlinks.

I guess we'll have to try and see if it works. 
I created a ticket here: https://github.com/getsentry/sentry-dart/issues/447

marandaneto: makes sense, thanks

PR #446, opened 2021-05-02
Fix: Enable tests on browser

### LINKED ISSUE #434: Make the tests from sentry_flutter also to be run on the "browser"
https://github.com/getsentry/sentry-dart/blob/a9979b0adc18b2545ad907ebae8b1f4b558844d6/flutter/test/sentry_flutter_web_test.dart

get the inspiration from:
https://github.com/getsentry/sentry-dart/blob/d24cb5039f38845d73f613454704c2904cf0a94c/dart/test/sentry_browser_test.dart

ueman:
## :scroll: Description

This change adds a job which runs the tests on chrome (browser).

_#skip-changelog_

Closes https://github.com/getsentry/sentry-dart/issues/434

This just adds more tests runs.

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes


marandaneto:
34 successful checks
nice :D

flutter/test/sentry_flutter_test.dart
```diff
@@ -53,7 +53,7 @@ void main() {
         packageLoader: loadTestPackage,
         platformChecker: getPlatformChecker(platform: MockPlatform.android()),
       );
-    });
+    }, testOn: 'vm');
```
ueman, line 56: All these tests with `, testOn: 'vm'` doesn't need to be run on browser because they're for not browser targets.
Also they fail when run on browser.
I think it's okay to not run them on browser.

.github/workflows/flutter.yml
```diff
@@ -105,6 +105,16 @@ jobs:
...
+  
+  web-test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: subosito/flutter-action@v1
+      - run: |
+          cd flutter
+          flutter pub get
+          flutter test --platform chrome
```
marandaneto: do we need a new Job? what;s about just reusing the `Test` job?

flutter/test/sentry_flutter_test.dart
marandaneto: maybe we delete `sentry_flutter_web_test.dart` if we dont need it?

.github/workflows/flutter.yml
```diff
@@ -67,6 +67,7 @@ jobs:
           run: |
             cd flutter
             flutter pub get
+            flutter test --platform chrome
```
marandaneto, line 70: lets also add the browser test to https://github.com/getsentry/sentry-dart/blob/main/flutter/tool/presubmit.sh#L12

marandaneto (approved):

PR #450, opened 2021-05-04
Fix integration adding in browser

ueman:
## :scroll: Description
Fixes a bug where iOS/macOS integrations were added on an iOS/macOS browser. The same happened also on Android.

I've added regression tests

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/flutter/lib/src/sentry_flutter.dart
+++ b/flutter/lib/src/sentry_flutter.dart
@@ -89,12 +89,14 @@ mixin SentryFlutter {
     }
 
     // Will enrich events with device context, native packages and integrations
-    if (options.platformChecker.platform.isIOS ||
-        options.platformChecker.platform.isMacOS) {
+    if (!options.platformChecker.isWeb &&
+        (options.platformChecker.platform.isIOS ||
+            options.platformChecker.platform.isMacOS)) {
       integrations.add(LoadContextsIntegration(channel));
     }
 
-    if (options.platformChecker.platform.isAndroid) {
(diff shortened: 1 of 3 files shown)


flutter/lib/src/sentry_flutter.dart
```diff
@@ -89,12 +89,14 @@ mixin SentryFlutter {
...
-        options.platformChecker.platform.isMacOS) {
+    if (!options.platformChecker.isWeb &&
+        (options.platformChecker.platform.isIOS ||
+            options.platformChecker.platform.isMacOS)) {
       integrations.add(LoadContextsIntegration(channel));
     }
 
-    if (options.platformChecker.platform.isAndroid) {
+    if (!options.platformChecker.isWeb &&
+        options.platformChecker.platform.isAndroid) {
```
marandaneto, line 99: can we double-check the usage of `.isIOS`, `isMacOS`, and `isAndroid` and see if we've not missed anywhere else the `!isWeb` check? if you've not done it yet :)

ueman: Yep, I'll double check it.

ueman: Okay, we're good. The only usages are for adding integrations

marandaneto: thanks

marandaneto (approved):
LGTM

PR #452, opened 2021-05-04
Feature: Enrich events with information

### LINKED ISSUE #262: Flutter Web/Desktop missing OS/Device/Browser info
_Platform:_
- [ ]  Dart
- [ ]  Flutter Android or iOS
- [x]  Flutter Web

_IDE:_
- [x]  VSCode
- [ ]  IntelliJ/AS
- [ ]  XCode
- [ ]  Other, which one?

- [ ]  Enabled
- [x]  Disabled (CanvasKit)

_Platform installed with:_
- [x] pub.dev
- [ ] GitHub

❯ flutter doctor -v
[✓] Flutter (Channel unknown, 1.24.0-10.2.pre, on macOS 11.1 20C69 darwin-x64, locale en-US)
    • Flutter version 1.24.0-10.2.pre at /usr/local/flutter
    • Framework revision 022b333a08 (7 weeks ago), 2020-11-18 11:35:09 -0800
    • Engine revision 07c1eed46b
    • Dart version 2.12.0 (build 2.12.0-29.10.beta)

[✓] Android toolchain - develop for Android devices (Android SDK version 29.0.3)
    • Android SDK at /Users/adam/Library/Android/sdk
    • Platform android-29, build-tools 29.0.3
    • Java binary at: /Applications/Android Studio.app/Contents/jre/jdk/Contents/Home/bin/java
    • Java version OpenJDK Runtime Environment (build 1.8.0_212-release-1586-b4-5784211)

[✓] Xcode - develop for iOS and macOS (Xcode 12.3)
    • Xcode 12.3, Build version 12C33
    • CocoaPods version 1.10.0

[✓] Android Studio (version 3.6)
    • Flutter plugin version 45.1.1
    • Dart plugin version 192.7761
    • Java version OpenJDK Runtime Environment (build 1.8.0_212-release-1586-b4-5784211)

[✓] VS Code (version 1.52.1)
    • Flutter extension version 3.18.0

[✓] Connected device (2 available)
    • Web Server (web)       • web-server                               • web-javascript • Flutter Tools
    • Chrome (web)           • chrome                                   • web-javascript • Google Chrome 87.0.4280.88

• No issues found!

**4.0.1**

---
I have the following issue:

On previous versions of the Sentry integration with Flutter Web we were able to see OS and Browser info but after the upgrade, this info is no longer captured or available in error reports.

**Steps to reproduce:**
- Upgrade package to `sentry_flutter: ^4.0.1`

**Actual result:**
- No mobile OS info is captured with errors
- No mobile Browser info captured with errors

**Expected result:**
- Result

### LINKED ISSUE #315: Web missing OS/Device/Browser info
I believe the Sentry javascript client automatically collects and adds os/device/browser info. The Dart web client should do the same.

ueman:
## :scroll: Description
Enrich events with information

Closes #315
Closes #262

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps
- [x] Add `Compiler` or `runtime type` field for Runtimes (AOT, VM, dart2js, dartdevc)
- [x] Write tests
- [x] https://github.com/getsentry/sentry-dart/issues/464


dart/lib/src/enricher/_io_enricher.dart
```diff
@@ -0,0 +1,84 @@
...
+    final args = Platform.executableArguments;
+    final packageConfig = Platform.packageConfig;
+    final cpuCount = Platform.numberOfProcessors;
+    final environment = Platform.environment;
+    final locale = Platform.localeName;
+    final localHostname = Platform.localHostname;
+    // TODO What about:
+    // Platform.executeable
+    // Platform.resolvedExecuteable
+    // Platform.pathSeperator
```
bruno-garcia: Not sure it's relevant:
```suggestion
```

dart/lib/src/enricher/_io_enricher.dart
```diff
@@ -0,0 +1,84 @@
...
+  Map<String, dynamic> _getExtras(Map<String, dynamic>? extras) {
+    final args = Platform.executableArguments;
+    final packageConfig = Platform.packageConfig;
+    final cpuCount = Platform.numberOfProcessors;
+    final environment = Platform.environment;
+    final locale = Platform.localeName;
+    final localHostname = Platform.localHostname;
+    // TODO What about:
+    // Platform.executeable
+    // Platform.resolvedExecuteable
```
bruno-garcia: What are these about? The FQPN of the executable?
Is that something we want to add to the protocol?

dart/lib/src/enricher/_web_enricher.dart
```diff
@@ -0,0 +1,149 @@
...
+    } else {
+      return os.copyWith(
+        name: _platform.operatingSystem,
+      );
+    }
+  }
+
+  List<SentryRuntime> _getRuntimes(List<SentryRuntime>? runtimes) {
+    final dartRuntime = SentryRuntime(name: 'Dart');
+    final flutterRuntime = SentryRuntime(name: 'Flutter');
```
ueman: This is not Flutter

dart/lib/src/enricher/_io_enricher.dart
```diff
@@ -0,0 +1,84 @@
...
+
+    return event.copyWith(
+      contexts: contexts,
+      extra: _getExtras(event.extra),
+    );
+  }
+
+  List<SentryRuntime> _getRuntimes(List<SentryRuntime>? runtimes) {
+    final dartRuntime = SentryRuntime(name: 'Dart', version: Platform.version);
+    final flutterRuntime = SentryRuntime(name: 'Flutter');
```
ueman: Not Flutter

flutter/lib/src/flutter_enricher.dart
```diff
@@ -0,0 +1,123 @@
...
+  final SingletonFlutterWindow _window;
+  final PlatformChecker _checker;
+  final DeviceInfoPlugin _deviceInfoPlugin;
+  final Enricher _dartEnricher;
+  WindowsDeviceInfo? _windowsDeviceInfo;
+  LinuxDeviceInfo? _linuxDeviceInfo;
+
+  @override
+  FutureOr<SentryEvent> apply(SentryEvent event) async {
+    if (_checker.hasNativeIntegration) {
```
ueman: Even for platforms with native integrations this should be executed  for those Flutter & Dart specific informations

dart/lib/src/enricher/_io_enricher.dart
ueman: Platform.executable:
> The path returned is the literal path used to run the script. This path might be relative or just be a name from which the executable was found by searching the PATH.

Platform.resolvedExecutable:
> This is the absolute path, with all symlinks resolved, to the executable used to run the script.

ueman: So i guess this should just be included if PII is enabled.

flutter/pubspec.yaml
```diff
@@ -15,6 +15,7 @@ dependencies:
     sdk: flutter
   sentry: ^5.0.0
   package_info_plus: ^1.0.0
+  device_info_plus: ^1.0.0
```
ueman: If we don't want to depend on this we could consider making this available as a seperate package.

flutter/lib/src/flutter_enricher.dart
```diff
@@ -0,0 +1,123 @@
...
+    // adding information which is already avaiable in Dart.
+    return _dartEnricher.apply(event);
+  }
+
+  Map<String, dynamic> _getExtras(Map<String, dynamic>? extras) {
+    if (extras == null) {
+      return {
+        'window_is_visible': _window.viewConfiguration.visible,
+        // dark mode or light mode
+        'brightness': describeEnum(_window.platformBrightness),
```
ueman: This would probably a good candidate for an addition to the protocol.

dart/lib/src/enricher/_io_enricher.dart
```diff
@@ -0,0 +1,84 @@
...
+    // Platform.executeable
+    // Platform.resolvedExecuteable
+    // Platform.pathSeperator
+
+    final moreExtras = <String, dynamic>{
+      if (args.isNotEmpty) 'executableArguments': Platform.executableArguments,
+      if (packageConfig != null) 'packageConfig': packageConfig,
+      'numberOfProcessors': cpuCount,
+      if (environment.isNotEmpty) 'environment': environment,
+      'localName': locale,
```
bruno-garcia:
```suggestion
      'locale': locale,
```

Maybe something we can align here with .NET.
There we have an "unofficial context interface" [called "Current Culture"](https://docs.microsoft.com/en-us/dotnet/api/system.globalization.cultureinfo.currentculture?view=net-5.0)

image

Maybe we can get together to spec it out (aka: someone starts with a PR to https://github.com/getsentry/develop with a proposal) and we consider extensions to the protocol to all this data you were able to grab here.

dart/lib/src/enricher/_web_enricher.dart
bruno-garcia: Here we're on dart2js or the other one for dev, right?
Maybe worth adding a key under the Runtime interface like: `Compiler: dart2js` or something?

flutter/example/lib/main.dart
```diff
@@ -120,6 +120,17 @@ class MainScaffold extends StatelessWidget {
                 Sentry.captureMessage('A message with a print() Breadcrumb');
               },
             ),
+            RaisedButton(
```
bruno-garcia: Will this be deleted before finalizing it? I assume the idea is to run this on all outgoing events, right?

flutter/lib/src/flutter_enricher.dart
```diff
@@ -0,0 +1,123 @@
...
+        'locale': _window.locale.toString(),
+      };
+    }
+    extras.putIfAbsent(
+        'window_is_visible', () => _window.viewConfiguration.visible);
+
+    extras.putIfAbsent(
+        'brightness', () => describeEnum(_window.platformBrightness));
+
+    extras.putIfAbsent('locale', () => _window.locale.toString());
```
bruno-garcia: Sorry if here I'm being repetitive but just so we can discuss: Lets avoid using Extra ourselves, and `locale` is a good chance for us to chat, align across SDKs and document on `develop`. 

Maybe `App` or `OS` (depending where the locale is define, I believe what makes sense the most in this case?

flutter/pubspec.yaml
bruno-garcia: We try to keep SDKs dependency-free as much as possible but here I don't see a way around it. Since usually this is data we get from the platform itself, and in Flutter (correct me if I'm wrong please) `device_info_plus` became the facto platform information library. Folks from [Flutter Community maintaining this library](https://github.com/fluttercommunity/plus_plugins) as very responsive and approachable so landing a patch there it's surely easier than trying to land a patch some library that's unmaintained or not so mainstream.

@marandaneto what do you think?
Also, for visibility: @mitsuhiko

bruno-garcia (commented):
This has so much great information. I'd love to see what this looks like in Sentry. Care to attach a screenshot?

dart/lib/src/enricher/_io_enricher.dart
```diff
@@ -0,0 +1,84 @@
...
+    final environment = Platform.environment;
+    final locale = Platform.localeName;
+    final localHostname = Platform.localHostname;
+    // TODO What about:
+    // Platform.executeable
+    // Platform.resolvedExecuteable
+    // Platform.pathSeperator
+
+    final moreExtras = <String, dynamic>{
+      if (args.isNotEmpty) 'executableArguments': Platform.executableArguments,
```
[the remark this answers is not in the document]
rhcarvalho: Including env vars is dangerous, it could even unintentionally expose secrets and credentials.
I don't think we do it elsewhere (typing from my phone now)

dart/lib/src/enricher/_io_enricher.dart
ueman: This is the Dart SDK, not the Flutter SDK. So including it here would be wrong. It should go into the FlutterEnricher. If at all, because, as you said, there's no Flutter version info.

dart/lib/src/enricher/_web_enricher.dart
ueman: Good point. I believe on web there is only dart2js. On the other platforms there is the VM for debug and AOT for release mode. I'm not sure what's used in profile mode. I'll have to look it up.

flutter/example/lib/main.dart
ueman: Yes, it's currently more in a proof of concept stage. I'm also still searching for available information. So not plugging this into a proper place in the SDK makes things a little easier.

There are also this Accessibility information available:
https://api.flutter.dev/flutter/dart-ui/AccessibilityFeatures-class.html 

Yeah, I'll attach a screenshot 👍

flutter/pubspec.yaml
marandaneto: there are 2 ways to go and both have their trade-offs.
we're not depending only on `device_info_plus` if we want to go fully "Flutter native" with Web/Desktop support, we'll depend on a bunch of them that might conflict with App's dependencies, that's one of them.

Indeed, the `plus` plugins are well maintained and welcoming new contributors, and accepting/merging/releasing PRs quite fast.

Going this way, we'd be missing features that the Native SDKs offer ootb, eg, Offline caching, Release health, etc, also maintaining another code base.

Going with the Native SDKs, like adding sentry-native and sentry-javascript for JS and Desktop support seems less code and easier to do, but obviously, it has its trade-offs as well.

Let's discuss this before going forward.

ueman:
> This has so much great information. I'd love to see what this looks like in Sentry. Care to attach a screenshot?

https://sentry.io/share/issue/8215a456950349b48031230203c4b9de/ 
Well it's not a screenshot but I guess it works too.
This is what it currently looks like on Flutter for Web

dart/lib/src/enricher/_io_enricher.dart
ueman: Done: https://github.com/getsentry/develop/pull/329

flutter/lib/src/flutter_enricher.dart
ueman: Also done: https://github.com/getsentry/develop/pull/330

ueman:
This is what it looks like on Ubuntu (without device info):
https://sentry.io/share/issue/025c1729f3194e0e8ec3b4ef72c84b03/

dart/lib/src/enricher/_io_enricher.dart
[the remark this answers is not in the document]
ueman: yep, you are right

flutter/pubspec.yaml
ueman: Until we decided wether we want to use `device_info_plus` or write the native code ourself, we could also just include information given by Dart and Flutter.

So the question is, is this `Enricher` approach something which we want to move on with?

kuhnroyal: I think it would be great to have something like this available. Maybe it should be in a separate extension package so developers are not forced to use it (in regard to conflicting packages).
This could still serve as template for developers who want to use different or their own dependencies.

ueman: @kuhnroyal You can already use event processors to add information to your crash reports and events. See https://pub.dev/documentation/sentry/latest/sentry/EventProcessor.html

bruno-garcia:
For reference: (you can paste an image from the clip board into comments):

image

ueman:
# Browser
sentry io_organizations_sentry-sdks_issues__project=5428562

# Linux (Ubuntu)
sentry io_organizations_sentry-sdks_issues_2384348813__project=5428562 query=is%3Aunresolved

dart/lib/src/enricher/_web_enricher.dart
```diff
@@ -0,0 +1,161 @@
...
+      }
+      if (screenOrientation.type?.startsWith('landscape') ?? false) {
+        orientation = SentryOrientation.landscape;
+      }
+    }
+
+    String? screenResolution;
+    var screen = _screen;
+    if (screen != null) {
+      screenResolution = '${screen.width}x${screen.height}';
```
ueman: This should be `screen_height_pixels` and `screen_width_pixels`:
https://github.com/getsentry/develop/pull/335

flutter/lib/src/flutter_enricher.dart
```diff
@@ -0,0 +1,173 @@
...
+        'brightness', () => describeEnum(_window.platformBrightness));
+    return extras;
+  }
+
+  SentryDevice _applyDevice(SentryDevice? device) {
+    final orientation = _window.physicalSize.width > _window.physicalSize.height
+        ? SentryOrientation.landscape
+        : SentryOrientation.portrait;
+
+    final screenResolution =
```
ueman: This should be `screen_height_pixels` and `screen_width_pixels`:
https://github.com/getsentry/develop/pull/335

ueman:
# Android

On Android  the native SDK doesn't seem to fill in the missing bits. Is that intended?

sentry io_organizations_sentry-sdks_issues_2391189370__project=5428562 query=is%3Aunresolved

dart/lib/src/enricher/_io_enricher.dart
```diff
@@ -0,0 +1,72 @@
...
+    }
+    extras.addAll(moreExtras);
+    return extras;
+  }
+
+  SentryDevice _getDevice(SentryDevice? device) {
+    return (device ?? SentryDevice()).copyWith(
+      language: Platform.localeName,
+      // this is the Name given to the device by the user of the device
+      name: Platform.localHostname,
```
ueman: This is potential PII

flutter/lib/src/flutter_enricher.dart
```diff
@@ -0,0 +1,168 @@
...
+
+  List<SentryRuntime> _getRuntimes(List<SentryRuntime>? runtimes) {
+    var flutterRuntimeDescription = '';
+
+    // See
+    // - https://flutter.dev/docs/testing/build-modes
+    // - https://github.com/flutter/flutter/wiki/Flutter%27s-modes
+    if (_checker.isWeb) {
+      if (_checker.isDebugMode()) {
+        flutterRuntimeDescription = 'Flutter with dartdevc';
```
ueman: Do we want to add a `compiler` (or something along those lines) field to `Runtime`?  It could then hold the information which compiler (AOT, VM, dart2js, dartdevc) was used.

dart/lib/src/enricher/_web_enricher.dart
```diff
@@ -0,0 +1,96 @@
...
+
+  List<SentryRuntime> _getRuntimes(List<SentryRuntime>? runtimes) {
+    // Pure Dart doesn't have specific runtimes per build mode
+    // like Flutter: https://flutter.dev/docs/testing/build-modes
+    final dartRuntime = SentryRuntime(
+      name: 'Dart',
+      rawDescription: 'Dart on browser',
+    );
+
+    final browserRuntime = SentryRuntime(
```
ueman: I'm pretty sure this is not the best way to report it but I also don't know how the JavaScript SDK does it

dart/lib/src/enricher/_io_enricher.dart
```diff
@@ -0,0 +1,82 @@
...
+      contexts: contexts,
+    );
+  }
+
+  List<SentryRuntime> _getRuntimes(List<SentryRuntime>? runtimes) {
+    // Pure Dart doesn't have specific runtimes per build mode
+    // like Flutter: https://flutter.dev/docs/testing/build-modes
+    final dartRuntime = SentryRuntime(
+      name: 'Dart',
+      rawDescription: Platform.version,
```
ueman: > An unprocessed description string obtained by the runtime. For some well-known runtimes, Sentry will attempt to parse name and version from this string, if they are not explicitly given.

From https://develop.sentry.dev/sdk/event-payloads/contexts/#runtime-context

What do I have to do for that?

CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 # Unreleased
 
+* Feat: Enrich events with more context
```
```suggestion
* Feat: Enrich events with more context (#452)
```

ueman:
Example event: https://sentry.io/organizations/sentry-sdks/issues/2149575550/events/190d80a9e2414d1cb5ea25f56892d36b/?project=5428562

flutter/lib/src/flutter_enricher.dart
```diff
@@ -0,0 +1,185 @@
...
+    contexts['culture'] = _getCulture();
+
+    contexts['packages'] = await _getPackages();
+
+    return event.copyWith(
+      contexts: contexts,
+    );
+  }
+
+  Future<Map<String, dynamic>> _getPackages() async {
```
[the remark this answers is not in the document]
ueman: And because we're getting these from the LicenseRegistry, these are only the packages which have a license.

dart/lib/src/enricher/_io_enricher.dart
marandaneto: yes, should only be attached if `sendDefaultPii` is enabled

flutter/lib/src/flutter_enricher.dart
marandaneto: sounds useful

dart/lib/src/enricher/_web_enricher.dart
marandaneto: Screenshot 2021-05-18 at 09 21 04

just an example, pretty simple

marandaneto: thats the browser though, for JS maybe it does not make sense to report a Runtime

dart/lib/src/enricher/_io_enricher.dart
marandaneto: Screenshot 2021-05-18 at 09 23 04

I guess giving name and version is enough if the Platform does not offer a description, so server infers it automatically

dart/lib/src/enricher/_web_enricher.dart
ueman: I guess we would then need to parse the User-Agent. Do we want to do that?

flutter/lib/src/flutter_enricher.dart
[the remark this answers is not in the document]
marandaneto: I think version is mandatory, https://develop.sentry.dev/sdk/event-payloads/sdk/
so if it does not work, set the version as "unknown"
worth adding a comment to the code about packages not being available because there's no license

dart/lib/src/enricher/_io_enricher.dart
```diff
@@ -0,0 +1,82 @@
+import 'dart:async';
+import 'dart:io';
+
+import '../protocol.dart';
+
+import 'enricher.dart';
+
+final Enricher instance = IoEnricher();
```
marandaneto: would be possible to do this with event processors instead? that's the way to mutate/enrich events

dart/lib/src/enricher/_io_enricher.dart
ueman: `Platfom.version` returns a string in the form of `2.12.3 (stable) (Wed Apr 14 11:02:39 2021 +0200) on "android_arm64"`.
 So ideally I would like that the `Version` field just contains `2.12.3 (stable)`.

Or at least that what I thought how it works after reading the documentation.

dart/lib/src/enricher/_web_enricher.dart
marandaneto: I'd not parse but rather set it as it is, there's no standard format for user agent string

flutter/lib/src/flutter_enricher.dart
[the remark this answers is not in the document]
ueman: Isn't the SDK just for the version information of the Sentry SDK?

flutter/lib/src/flutter_enricher.dart
```diff
@@ -0,0 +1,185 @@
...
+    this._checker,
+    this._dartEnricher,
+    this._getWidgetsBinding,
+  );
+
+  factory FlutterEnricher.simple(PlatformChecker checker) {
+    return FlutterEnricher(
+      checker,
+      Enricher(),
+      () => WidgetsFlutterBinding.ensureInitialized(),
```
marandaneto: I prefer to not call `WidgetsFlutterBinding.ensureInitialized()` here, this could lead to bugs like https://github.com/getsentry/sentry-dart/issues/263
I'd rather prefer that our event processor is executed once the event is captured and then we're pretty sure that the bindings are already inited.

ueman: That's exactly the reason it's a callback. It just gets called when an event is captured and not at the time of initializing the `FlutterEnricher`.

dart/lib/src/enricher/_io_enricher.dart
ueman: I guess. Do we need to make sure that the information is added before or after the user's event processors?

marandaneto: event processors are executed in order, our event processors should be always before the user's event processors, because when it comes to the user's event processors, every field should be already set so the user can erase/mutate the values.

dart/lib/src/enricher/_io_enricher.dart
marandaneto: yes, that's ideal when you can grab the version separately, but if in Flutter you just get this huge String, and you can't guarantee that the format is always the same, its just hard to parse it.
eg this String might have a different format for different platforms, that's why setting such plain String isn't that bad.

flutter/lib/src/flutter_enricher.dart
marandaneto: the idea is that this is not even called, the SDK init. already takes care of `WidgetsFlutterBinding.ensureInitialized`, so at this point, you don't need to take care of it.

flutter/lib/src/flutter_enricher.dart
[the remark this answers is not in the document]
marandaneto: yes, indeed, for the App's dependencies, it'd be `event.modules`

flutter/lib/src/flutter_enricher.dart
ueman: Yeah, but I need the `WidgetsBinding` which is returned from that call. If `WidgetsFlutterBinding` is already initialized it just returns the same instance.

See https://api.flutter.dev/flutter/widgets/WidgetsFlutterBinding/ensureInitialized.html

flutter/lib/src/flutter_enricher.dart
```diff
@@ -0,0 +1,199 @@
...
+      var packages = <String>{};
+      await LicenseRegistry.licenses.forEach(
+        (entry) => packages.addAll(
+          entry.packages.toList(),
+        ),
+      );
+
+      _packages = Map.fromEntries(
+        packages.map(
+          (e) => MapEntry(e, 'unknown'),
```
ueman: Is `unknown` or `n/a` (for not available) better?
I kinda like `n/a` better because that's what it is. 
(It's also shorter and saves some bytes though it probably doesn't matter much)

flutter/lib/src/flutter_enricher.dart
```diff
@@ -0,0 +1,199 @@
...
+  /// - No version information is available
+  /// - Flutter's native dependencies are also included.
+  FutureOr<Map<String, String>> _getPackages() async {
+    if (_packages.isEmpty) {
+      // This can take some time.
+      // Therefore we cache this after running
+      var packages = <String>{};
+      await LicenseRegistry.licenses.forEach(
+        (entry) => packages.addAll(
+          entry.packages.toList(),
```
ueman: I probably should add a comment what's happening here:

The license registry has a list of licenses entries (MIT, Apache...).
Each license entry has a list of packages which licensed under this particular license.
Libraries can be dual licensed.

We don't care about those license issues, we just want each package name once.
Therefore I add each name to a set to make sure I only have it once.

marandaneto: sounds good, is that the only way to get the packages list? because as. you said, if a package does not have a package, it will not appear there

flutter/lib/src/flutter_enricher.dart
ueman: Then we keep `unknown`.

dart/lib/src/enricher/io_enricher_event_processor.dart
```diff
@@ -0,0 +1,106 @@
...
+    options.platformChecker.hasNativeIntegration,
+    options.sendDefaultPii,
+  );
+  return processor.apply;
+}
+
+/// Enriches [SentryEvents] with various kinds of information.
+/// Uses Darts [Platform](https://api.dart.dev/stable/dart-io/Platform-class.html)
+/// class to read information.
+class IoEnricherEventProcessor {
```
ueman: All of these enrichers are kinda `EventProcessor` classes. I created https://github.com/getsentry/sentry-dart/issues/479 . So this should be pretty straight forward to migrate in the future.

ueman:
@marandaneto Any idea how to test if an event processor was added without checking the count of them?

flutter/lib/src/flutter_enricher_event_processor.dart
```diff
@@ -0,0 +1,200 @@
...
+  /// This however has some drawbacks:
+  /// - Only packages with licenses are known
+  /// - No version information is available
+  /// - Flutter's native dependencies are also included.
+  FutureOr<Map<String, String>> _getPackages() async {
+    if (_packages.isEmpty) {
+      // This can take some time.
+      // Therefore we cache this after running
+      var packages = <String>{};
+      await LicenseRegistry.licenses.forEach(
```
ueman, line 87:
```suggestion
      // The license registry has a list of licenses entries (MIT, Apache...).
      // Each license entry has a list of packages which licensed under this particular license.
      // Libraries can be dual licensed.
      // 
      // We don't care about those license issues, we just want each package name once.
      // Therefore we add each name to a set to make sure we only add it once.
      await LicenseRegistry.licenses.forEach(
```

marandaneto:
> @marandaneto Any idea how to test if an event processor was added without checking the count of them?

checking the options.eventProcessors and see if the desired type is within the list? we might do it already iirc

dart/lib/src/enricher/io_enricher_event_processor.dart
marandaneto: sounds good, so we can get unblocked with this, I will do a fine review tomorrow

dart/lib/src/enricher/io_enricher_event_processor.dart
```diff
@@ -0,0 +1,106 @@
+import 'dart:async';
+import 'dart:io';
+
+import '../protocol.dart';
+import '../sentry_options.dart';
+
+EventProcessor enricherEventProcessor(SentryOptions options) {
+  final processor = IoEnricherEventProcessor(
+    options.platformChecker.hasNativeIntegration,
+    options.sendDefaultPii,
```
marandaneto: this will probably cause a bug I guess, if `sendDefaultPii` is false by default and you change it in the configuration callback, at this point you passed the boolean as false to the processor as a value copy.
maybe just pass the raw options to IoEnricherEventProcessor

dart/lib/src/enricher/io_enricher_event_processor.dart
```diff
@@ -0,0 +1,106 @@
...
+  Map<String, dynamic> _getDartContext(bool includePii) {
+    final args = Platform.executableArguments;
+    final packageConfig = Platform.packageConfig;
+
+    String? executable;
+    if (includePii) {
+      try {
+        // This throws sometimes for some reason
+        executable = Platform.executable;
+      } catch (_) {}
```
marandaneto: maybe log it?

dart/lib/src/enricher/web_enricher_event_processor.dart
```diff
@@ -0,0 +1,102 @@
...
+    );
+  }
+
+  List<SentryRuntime> _getRuntimes(List<SentryRuntime>? runtimes) {
+    // Pure Dart doesn't have specific runtimes per build mode
+    // like Flutter: https://flutter.dev/docs/testing/build-modes
+    final dartRuntime = SentryRuntime(
+      key: 'sentry_dart_runtime',
+      name: 'Dart',
+    );
```
marandaneto: wondering if its useful if `Platform.version` does not exist here

dart/test/enricher/io_enricher_test.dart
```diff
@@ -0,0 +1,157 @@
...
+
+    test('adds Dart context with PII', () async {
+      final enricher = fixture.getSut(includePii: true);
+      final event = await enricher.apply(fixture.event);
+
+      final dartContext = event.contexts['dart_context'];
+      expect(dartContext, isNotNull);
+      expect(dartContext['number_of_processors'], isNotNull);
+      // Getting the executable sometimes throws
+      //expect(dartContext['executable'], isNotNull);
```
marandaneto, line 83: should we do smt about it?

dart/test/enricher/io_enricher_test.dart
```diff
@@ -0,0 +1,157 @@
...
+      expect(
+        event.contexts.operatingSystem?.version,
+        fakeEvent.contexts.operatingSystem?.version,
+      );
+    });
+  });
+}
+
+class Fixture {
+  SentryEvent event = SentryEvent();
```
marandaneto: this should not be part of the fixture

dart/test/enricher/web_enricher_test.dart
```diff
@@ -0,0 +1,138 @@
...
+      expect(
+        event.contexts.operatingSystem?.name,
+        fakeEvent.contexts.operatingSystem?.name,
+      );
+    });
+  });
+}
+
+class Fixture {
+  SentryEvent event = SentryEvent();
```
marandaneto: also not part of the fixture

flutter/lib/src/flutter_enricher_event_processor.dart
```diff
@@ -0,0 +1,206 @@
...
+    this.hasNativeIntegration,
+  );
+
+  factory FlutterEnricherEventProcessor.simple({
+    required PlatformChecker checker,
+    required bool hasNativeIntegration,
+  }) {
+    return FlutterEnricherEventProcessor(
+      checker,
+      () => WidgetsBinding.instance!,
```
marandaneto: is it safe to do `!`? wouldnt cause an issue if its null?

flutter/lib/src/flutter_enricher_event_processor.dart
```diff
@@ -0,0 +1,206 @@
...
+            debugDefaultTargetPlatformOverride,
+      if (_window.initialLifecycleState.isNotEmpty)
+        'initial_lifecycle_state': _window.initialLifecycleState,
+      if (_window.defaultRouteName.isNotEmpty)
+        'default_route_name': _window.defaultRouteName,
+      if (currentLifecycle != null)
+        'current_lifecycle_state': describeEnum(currentLifecycle),
+      // Seems to always return false.
+      // Also always fails in tests.
+      // 'window_is_visible': _window.viewConfiguration.visible,
```
marandaneto, line 136: should we do smt here?

flutter/test/flutter_enricher_test.dart
```diff
@@ -0,0 +1,262 @@
...
+      expect(
+        event.contexts.device?.theme,
+        fakeEvent.contexts.device?.theme,
+      );
+    });
+  });
+}
+
+class Fixture {
+  final event = SentryEvent();
```
marandaneto: also not part of the fixture

flutter/test/flutter_enricher_test.dart
```diff
@@ -0,0 +1,262 @@
...
+  }) {
+    return FlutterEnricherEventProcessor(
+      checker ?? PlatformChecker(),
+      binding,
+      hasNativeIntegration,
+    );
+  }
+}
+
+class MockPlatformChecker implements PlatformChecker {
```
marandaneto: should we move this to the mock class? we may reuse somewhere else

dart/test/enricher/io_enricher_test.dart
ueman: I want to create a bug report with a reproduceable example but haven't finished it yet.

flutter/lib/src/flutter_enricher_event_processor.dart
ueman: At this point it's safe because `WidgetsFlutterBinding.ensureInitialized()` was already called in order to pass Sentry config to native.

I can change it to `?`, it's also probably better.

marandaneto: yeah better safe than sorry

ueman: Just for reference: https://github.com/flutter/flutter/issues/64830

dart/lib/src/enricher/web_enricher_event_processor.dart
ueman: Your call, I just included it because all other have it too.

flutter/lib/src/flutter_enricher_event_processor.dart
ueman: https://github.com/flutter/flutter/issues/83919

dart/test/enricher/io_enricher_test.dart
ueman: https://github.com/flutter/flutter/issues/83921

ueman:
I would like to build on https://github.com/getsentry/sentry-dart/pull/489 and https://github.com/getsentry/sentry-dart/pull/491

dart/lib/src/enricher/web_enricher_event_processor.dart
marandaneto: mm, yeah I just don't see the benefit for hardcoded values, I'd leave the comment for future reference but the `dartRuntime` would be gone as it does not say anything

marandaneto (approved):
LGTM

venkatd:
@ueman does this code run for you on debug mode with Flutter web? For me the whole browser crashes so I've had to downgrade to `5.1.0`. If it runs for you without problems, I can try to share a reproduction.

ueman:
> @ueman does this code run for you on debug mode with Flutter web? For me the whole browser crashes so I've had to downgrade to `5.1.0`. If it runs for you without problems, I can try to share a reproduction.

Are you talking about https://github.com/getsentry/sentry-dart/issues/521 or something else? If it's not https://github.com/getsentry/sentry-dart/issues/521 I appreciate a reproduction. Thanks!

PR #458, opened 2021-05-10
DO NOT MERGE: Snippet to add ID to User

### LINKED ISSUE #449: Add option for Android device ID generation
See https://github.com/getsentry/sentry-java/pull/1455#issuecomment-831257971

ueman:
## :scroll: Description
See https://github.com/getsentry/sentry-dart/issues/449

https://sentry.io/share/issue/9aba382a0ae649baafaea54522729709/

Fixes https://github.com/getsentry/sentry-dart/issues/449

## :pencil: Checklist

- [ ] All tests passing
- [ ] No breaking changes

### DIFF
--- a/flutter/example/lib/main.dart
+++ b/flutter/example/lib/main.dart
@@ -5,6 +5,7 @@ import 'package:flutter/material.dart';
 import 'package:flutter/services.dart';
 import 'package:sentry_flutter/sentry_flutter.dart';
 import 'package:universal_platform/universal_platform.dart';
+import 'package:device_info_plus/device_info_plus.dart';
 
 // ATTENTION: Change the DSN below with your own to see the events in Sentry. Get one at sentry.io
 const String _exampleDsn =
@@ -14,13 +15,27 @@ Future<void> main() async {
   await SentryFlutter.init(
     (options) {
       options.dsn = _exampleDsn;
-      // use breadcrumb tracking of WidgetsBindingObserver
-      // options.useFlutterBreadcrumbTracking();
-      // use breadcrumb tracking of platform Sentry SDKs
-      // options.useNativeBreadcrumbTracking();
+      if (UniversalPlatform.isAndroid) {
+        options.addEventProcessor(addAndroidDeviceId);
+      }
     },
     // Init your App.
     appRunner: () => runApp(MyApp()),
+  );
+}
+
+FutureOr<SentryEvent> addAndroidDeviceId(
+  SentryEvent event, {
+  dynamic hint,
+}) async {
+  var info = await DeviceInfoPlugin().androidInfo;
+  var user = (event.user ?? SentryUser()).copyWith(id: info.fingerprint);
+
+  // todo there's no ID field on SentryDevice
+  // var device = (event.contexts.device ?? SentryDevice()).copyWith();
+
+  return event.copyWith(
+    user: user,
   );
 }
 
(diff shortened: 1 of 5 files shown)


flutter/example/lib/main.dart
```diff
@@ -14,16 +15,30 @@ Future<void> main() async {
...
     appRunner: () => runApp(MyApp()),
   );
 }
 
+FutureOr<SentryEvent> addAndroidDeviceId(
+  SentryEvent event, {
+  dynamic hint,
+}) async {
+  var info = await DeviceInfoPlugin().androidInfo;
+  var user = (event.user ?? SentryUser()).copyWith(id: info.fingerprint);
```
ueman: There's a couple of candidates for Ids and I'm not quite sure which is should be user for which field.
I guess User.Id should be [fingerprint](https://pub.dev/documentation/device_info_plus_platform_interface/latest/model_android_device_info/AndroidDeviceInfo/fingerprint.html)
and Device.id should be [androidId](https://pub.dev/documentation/device_info_plus_platform_interface/latest/model_android_device_info/AndroidDeviceInfo/androidId.html).

Is [id](https://pub.dev/documentation/device_info_plus_platform_interface/latest/model_android_device_info/AndroidDeviceInfo/id.html) relevant? I guess not.

flutter/example/lib/main.dart
```diff
@@ -14,16 +15,30 @@ Future<void> main() async {
   await SentryFlutter.init(
     (options) {
       options.dsn = _exampleDsn;
-      // use breadcrumb tracking of WidgetsBindingObserver
```
ueman, line 17: This is not needed anymore.

flutter/example/lib/main.dart
```diff
@@ -14,16 +15,30 @@ Future<void> main() async {
...
 
+FutureOr<SentryEvent> addAndroidDeviceId(
+  SentryEvent event, {
+  dynamic hint,
+}) async {
+  var info = await DeviceInfoPlugin().androidInfo;
+  var user = (event.user ?? SentryUser()).copyWith(id: info.fingerprint);
+
+  // todo there's no ID field on SentryDevice
+  // var device = (event.contexts.device ?? SentryDevice()).copyWith();
```
marandaneto: that's ok, it's not gonna affect the number of unique users or anything, so its not a breaking change as it does not even exist in the Dart/Flutter SDK

ueman: Do we want to add the ID field for Device?

ueman: No, see https://github.com/getsentry/sentry-dart/pull/458#discussion_r629100195

flutter/example/lib/main.dart
marandaneto: we only care for `user.id` which is using the value of the SDK 4.x
we must set only the value read from https://developer.android.com/reference/android/provider/Settings.Secure#ANDROID_ID

ignore `device.id`

flutter/example/lib/main.dart
marandaneto: could be removed in a different PR as this is not going to be merged anyway

marandaneto: could be removed in a different PR, this PR is not going to be merged anyway

marandaneto:
the other option instead of using event processor would be something like:

```
SentryUser user = new SentryUser();
user.setId(Settings.Secure.ANDROID_ID); // read using device_info

Sentry.setUser(user);
```

but its not a good idea because it'd affect iOS users, so Event processors would make more sense

PR #463, opened 2021-05-10
Feature: `withScope` callback

### LINKED ISSUE #427: Add scope callback to capture methods
Currently there's no way to modify a scope for a single event without global side effect.

The latest approach on allowing a scope to be mutated is by taking a scope callback as a second argument to the capture methods. Lets add this to the SDK spec and implement in Dart.

~The Unified API includes a couple of ways to do that: `pushScope/popPop` and `withScope`.~

~Works like~
```
Sentry.withScope(scope => {
    scope.setTag(..);
    Sentry.captureException(..)
```

~Turns out this isn't on develop.sentry.dev so we need to add there, or figure out why it's not there anymore~

ueman:
## :scroll: Description
This adds a callback to the `capture*` methods to allow the user to add information for a specific event.

Fixes https://github.com/getsentry/sentry-dart/issues/427

I've written new tests

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes


dart/lib/src/hub.dart
```diff
@@ -65,12 +66,18 @@ class Hub {
       );
     } else {
       final item = _peek();
+      var scope = item.scope;
+
+      if (withScope != null) {
+        scope = item.scope.clone();
```
```suggestion
        scope = scope.clone();
```

could be `scope` directly as you already assigned up there, right? if this is possible, fix the other cases too

ueman: Good catch

dart/lib/src/hub.dart
```diff
@@ -105,12 +113,18 @@ class Hub {
...
+        withScope(scope);
+      }
```
marandaneto: this probably can be a private method that returns the mutated `scope` if `withScope` is given instead of dup. the same code block across methods

dart/test/hub_test.dart
```diff
@@ -165,4 +165,69 @@ void main() {
...
+    late Hub hub;
+    late MockSentryClient client;
+    SentryOptions options;
+
+    setUp(() {
+      options = SentryOptions(dsn: fakeDsn);
+      hub = Hub(options);
+      client = MockSentryClient();
+      hub.bindClient(client);
+    });
```
marandaneto: f i x t u r e :P

flutter/example/lib/main.dart
```diff
@@ -116,6 +116,18 @@ class MainScaffold extends StatelessWidget {
                 Sentry.captureMessage('A message with a print() Breadcrumb');
               },
             ),
+            RaisedButton(
+              child:
+                  const Text('Capture message with scope without breadcrumbs'),
```
```suggestion
                  const Text('Capture message with scope'),
```

flutter/example/lib/main.dart
```diff
@@ -116,6 +116,18 @@ class MainScaffold extends StatelessWidget {
...
+              onPressed: () {
+                Sentry.captureMessage(
+                  'This event has no breadcrumbs',
+                  withScope: (scope) {
+                    scope.clearBreadcrumbs();
```
marandaneto: instead of only removing data which is always nice to see, lets do something else, like, adding a tag, an extra, or something else

bruno-garcia (approved):
This looks great!
Something to add to Java and .NET next :)

PR #467, opened 2021-05-13
Fix: Merge user from scope and event

### LINKED ISSUE #466: sendDefaultPii stops sending of scope.user details
Platform: Dart
SDK: **5.0.0**

---

I tried enabling `SentryOptions.sendDefaultPii` due to the recommendation in the docs and also using `Sentry.configureScope` to set the `user` details.
When sendDefaultPii is enabled the scope.user details are never sent in error reports.

```
Sentry.init(
  (SentryOptions options) async {
    options
      ..dsn = dsn
      ..sendDefaultPii = true;
  },
);
Sentry.configureScope(
  (scope) => scope.user = SentryUser(username: "Tester")
);
```

The cause is [/dart/lib/src/scope.dart#L159 - applyToEvent](https://github.com/getsentry/sentry-dart/blob/main/dart/lib/src/scope.dart#L159) as the user with Pii details have already been created in SentryClient._prepareEvent
```
event = event.copyWith(
...
user: event.user ?? user,
...
);
```

My expectation is that this would have attempted to merge the details, in my case scoped user details would have more precedence.

If this is expected behaviour a note in the docs would have saved a lot of time.

Many thanks!

ueman:
## :scroll: Description
This PR introduces the merge of the user from event and scope.
The user from the event takes precedence over the scope.

Fixes https://github.com/getsentry/sentry-dart/issues/466

Added a regression test

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes


dart/lib/src/scope.dart
```diff
@@ -213,6 +213,26 @@ class Scope {
...
+    if (eventUser == null && scopeUser != null) {
+      return scopeUser;
+    }
+    // otherwise the user of scope takes precedence over the event user
+    return scopeUser?.copyWith(
+      id: eventUser?.id,
+      email: eventUser?.email,
+      ipAddress: eventUser?.ipAddress,
+      username: eventUser?.username,
+      // TODO should extras be merged?
```
ueman: I'm not quite sure if the extras of the user should be merged

CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 # Unreleased
 
+* Fix: Merge user from event and scope
```
```suggestion
* Fix: Merge user from event and scope (#467)
```

dart/lib/src/scope.dart
marandaneto: I guess we can do, yes, but event extras keys have precedence

dart/lib/src/scope.dart
```diff
@@ -213,6 +213,26 @@ class Scope {
...
+      extras: eventUser?.extras,
```
marandaneto: I don't think this would solve the problem, if I understand the code.
if `eventUser?.id` is null and `scopeUser.id` isn't null, the copied user.id will contain a null value.
merging should respect precedence, but if its null, we should use the one that has a value at least.

so I believe we could do something like:
`final id = eventUser?.id ?? scopeUser?.id`
for all properties, and then, create a new user, wdyt?

dart/test/scope_test.dart
```diff
@@ -308,7 +308,10 @@ void main() {
 
       final updatedEvent = await scope.applyToEvent(event, null);
 
-      expect(updatedEvent?.user, eventUser);
+      // We can't check for equality of the user because it
+      // gets merged from event and scope.
+      expect(updatedEvent?.user, isNotNull);
+      expect(updatedEvent?.user?.id, eventUser.id);
```
marandaneto: instead of adding such comment, we could assert field per field, as we know `eventUser` and `scopeUser`, we know what should be the output of each field, or maybe even removing this assert from this test because we have a specific one for the user merging, your call

dart/test/sentry_client_test.dart
```diff
@@ -394,6 +394,25 @@ void main() {
       expect(capturedEvent.fingerprint, eventFingerprint);
       expect(capturedEvent.breadcrumbs, eventCrumbs);
     });
+
+    test('should apply the scope user to null event user fields ', () async {
+      final client = SentryClient(options);
```
marandaneto: great opportunity to create a fixture pattern, lets always do it when we're touching test code and it's not using such a pattern that helps us writing more concise tests, also, reusing code

dart/lib/src/scope.dart
ueman: Using `null` with a `copyWith` keeps the original value because it already does `other ?? this` itself.

marandaneto: oh yeah, makes sense, my bad

dart/test/sentry_client_test.dart
```diff
@@ -373,6 +374,7 @@ void main() {
     );
 
     setUp(() {
+      fixture = Fixture();
       options = SentryOptions(dsn: fakeDsn);
       options.transport = MockTransport();
       scope = Scope(options)
```
marandaneto: if those things are used across all the tests, they should be part of the fixture too or to be declared within the test if its not the case, the idea of having a fixture is also to encapsulate every property

marandaneto (approved):
left a suggestion for improving the fixture, but other than that LGTM

dart/test/sentry_client_test.dart
```diff
@@ -352,47 +352,75 @@ void main() {
...
-        ..fingerprint = fingerprint
-        ..addBreadcrumb(crumb);
+      fixture = Fixture();
     });
 
     test('should not apply the scope to non null event fields ', () async {
-      final client = SentryClient(options);
-      await client.captureEvent(event, scope: scope);
+      final transport = MockTransport();
+      final client = fixture.getSut(true, transport);
```
marandaneto:
```suggestion
      final client = fixture.getSut(true, fixture.transport);
```

should not this be part of the fixture?

dart/test/sentry_client_test.dart
```diff
@@ -613,6 +641,38 @@ SentryEvent? eventProcessorDropEvent(SentryEvent event, {dynamic hint}) {
...
+  late SentryEvent event;
+
+  final transaction = '/test/scope';
+  final eventTransaction = '/event/transaction';
+  final fingerprint = ['foo', 'bar', 'baz'];
+  final eventFingerprint = ['123', '456', '798'];
+  final user = SentryUser(id: '123');
+  final eventUser = SentryUser(id: '987');
+  final crumb = Breadcrumb(message: 'bread');
+  final eventCrumbs = [Breadcrumb(message: 'bread')];
```
marandaneto: I dont think this should be part of the fixture declaration, this could be specific per test and values can change, I guess?

dart/test/sentry_client_test.dart
```diff
@@ -613,6 +641,38 @@ SentryEvent? eventProcessorDropEvent(SentryEvent event, {dynamic hint}) {
...
+      level: SentryLevel.warning,
+      transaction: eventTransaction,
+      user: eventUser,
+      fingerprint: eventFingerprint,
+      breadcrumbs: eventCrumbs,
+    );
+  }
+
+  late SentryOptions options;
+  late Scope scope;
```
marandaneto: this sounds more like a createScope(): Scope method than a global field in the fixture

dart/test/sentry_client_test.dart
ueman: true

dart/test/sentry_client_test.dart
ueman: Those won't change though.

dart/test/sentry_client_test.dart
```diff
@@ -613,11 +638,39 @@ SentryEvent? eventProcessorDropEvent(SentryEvent event, {dynamic hint}) {
...
     return SentryClient(options);
   }
+
+  Scope createScope(SentryOptions options) {
+    return Scope(options)
+      ..user = user
+      ..transaction = transaction
+      ..fingerprint = fingerprint
+      ..addBreadcrumb(crumb);
+  }
```
marandaneto: this should be a method within the `sentry_client_test.dart ` file but not really within the Fixture class

dart/test/sentry_client_test.dart
```diff
@@ -613,11 +638,39 @@ SentryEvent? eventProcessorDropEvent(SentryEvent event, {dynamic hint}) {
...
+  final crumb = Breadcrumb(message: 'bread');
+  final eventCrumbs = [Breadcrumb(message: 'bread')];
+
+  late SentryEvent event = SentryEvent(
+    level: SentryLevel.warning,
+    transaction: eventTransaction,
+    user: eventUser,
+    fingerprint: eventFingerprint,
+    breadcrumbs: eventCrumbs,
+  );
```
marandaneto: similar to https://github.com/getsentry/sentry-dart/pull/467/files#r638548589

dart/test/sentry_client_test.dart
marandaneto: still, its not part of the fixture pattern, explained a bit more here https://github.com/getsentry/sentry-dart/pull/467/files#r638550478

dart/test/sentry_client_test.dart
marandaneto: https://github.com/getsentry/sentry-dart/pull/467/files#r638548589

PR #473, opened 2021-05-25
Feat: Failed requests HTTP client

### LINKED ISSUE #292: SentryHttpClient captures errored requests and set the 'request' field accordingly
The SentryHttpClient should add the last request to crash reports. This should work for all supported platform.

I feel like this should be an easy change because requests are already recorded as breadcrumbs.

ueman:
## :scroll: Description
This is a HTTP client wich automatically add failed requests to an event.

Closes https://github.com/getsentry/sentry-dart/issues/292

It still needs tests.

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes


dart/lib/src/http_client/failed_request_client.dart
```diff
@@ -0,0 +1,174 @@
...
+/// which records events for failed requests.
+///
+/// Configured with default values, this captures requests which throw an
+/// exception.
+/// This can be for example for the following reasons:
+/// - In an browser environment this can be requests which fail because of CORS.
+/// - In an mobile or desktop application this can be requests which failed
+///   because the connection was interrupted.
+///
+/// Additionally you can configre specific HTTP response codes to be consideres
```
ueman:
```suggestion
/// Additionally you can configure specific HTTP response codes to be considered
```

dart/lib/src/http_client/failed_request_client.dart
```diff
@@ -0,0 +1,174 @@
...
+
+      return response;
+    } catch (exception, stackTrace) {
+      await _captureException(
+        exception: exception,
+        stackTrace: stackTrace,
+        request: request,
+      );
+
+      rethrow;
```
ueman, line 114: There needs to be a good hint about this. If the user captures this himself he gets two crash reports.

dart/lib/src/http_client/failed_request_client.dart
```diff
@@ -0,0 +1,174 @@
...
+    );
+    return _hub.captureEvent(event, stackTrace: stackTrace);
+  }
+
+  // Types of Request can be found here:
+  // https://pub.dev/documentation/http/latest/http/http-library.html
+  Object? _getDataFromRequest(BaseRequest request) {
+    if (request is MultipartRequest) {
+      final data = <String, String>{...request.fields};
+      // TODO request.files
```
ueman: Do we want to include files?

dart/lib/src/http_client/failed_request_client.dart
```diff
@@ -0,0 +1,174 @@
...
+      return request.body;
+    }
+
+    // There's nothing we can do for a StreamedRequest
+    return null;
+  }
+}
+
+/// Describes the size of http request bodies which should be added to an event
+enum MaxRequestBodySize {
```
ueman: Link https://docs.sentry.io/platforms/dotnet/guides/aspnetcore/configuration/options/#max-request-body-size in the comment.

dart/lib/src/http_client/sentry_http_client.dart
```diff
@@ -43,57 +44,22 @@ import '../hub_adapter.dart';
 /// }
 /// ```
 class SentryHttpClient extends BaseClient {
```
ueman, line 73: I've splitted `SentryHttpClient` into `FailedRequestClient` and `BreadcrumbClient`. This makes it easier to test and also easier to add more functionality or to remove functionality. Also the user can choose to only use the ones he needs.

I still need to make `SentryHttpClient` configurable so that `FailedRequestClient` is opt-in by default.

dart/lib/src/http_client/failed_request_client.dart
marandaneto: a way to solve this is via deduplication, see Java
https://github.com/getsentry/sentry-java/blob/b45944c93d53f68bc1e64f5ac4c9a5400b252d44/sentry/src/main/java/io/sentry/DuplicateEventDetectionEventProcessor.java

we can also make this in another PR btw if we see its necessary

dart/lib/src/http_client/failed_request_client.dart
marandaneto: don't think so, could be a future improvement, but definitely not worth it now

dart/lib/src/http_client/failed_request_client.dart
marandaneto: this goes under the protocol folder

dart/lib/src/http_client/failed_request_client.dart
ueman: Ah nice

dart/lib/src/http_client/sentry_http_client.dart
marandaneto: mmm, I'm wondering how this would work, today people add `SentryHttpClient`, does this change?
how I opt in only for failed requests or breadcrumbs? how do I use both?
I imagined this with a sort of optional ctor params for `SentryHttpClient`, I need to check how friendly this would be for the end user now

dart/lib/src/http_client/failed_request_client.dart
ueman: Our event processor is just a type alias for a method though. So that would be another breaking change for v6

dart/lib/src/http_client/breadcrumb_client.dart
```diff
@@ -0,0 +1,98 @@
...
+///
+/// The `BreadcrumbClient` can also be used as a wrapper for your own
+/// HTTP [Client](https://pub.dev/documentation/http/latest/http/Client-class.html):
+/// ```dart
+/// import 'package:sentry/sentry.dart';
+/// import 'package:http/http.dart' as http;
+///
+/// final myClient = http.Client();
+///
+/// var client = BreadcrumbClient(client: myClient);
```
marandaneto, line 35: this would be a breaking change, not sure if its worth it

dart/lib/src/http_client/failed_request_client.dart
```diff
@@ -0,0 +1,174 @@
...
+///
+/// Additionally you can configure specific HTTP response codes to be considered
+/// as a failed request. In the following example, the status codes 404 and 500
+/// are considered a failed request.
+///
+/// ```dart
+/// import 'package:sentry/sentry.dart';
+///
+/// var client = FailedRequestClient(
+///   failedRequestStatusCodes: [404, 500]
```
marandaneto: wondering if it makes sense to be a range, instead of a List of codes, like 400..599, one could -> 
400..403, 405..599

dart/lib/src/http_client/breadcrumb_client.dart
ueman: No not really, because the new SentryHttpClient combines both clients.

dart/lib/src/http_client/failed_request_client.dart
ueman: Yeah, that's a good idea

dart/lib/src/http_client/failed_request_client.dart
```diff
@@ -0,0 +1,174 @@
...
+      data: _getDataFromRequest(request),
+      other: {
+        'Content-Length': request.contentLength.toString(),
+      },
+    );
+    final event = SentryEvent(
+      throwable: exception,
+      request: sentryRequest,
+      level: SentryLevel.error,
+    );
```
marandaneto: let's add a mechanism to it, like we do with other type of events

https://github.com/getsentry/sentry-dart/blob/main/dart/lib/src/default_integrations.dart#L31-L37

so people could filter by the mechanism

dart/lib/src/http_client/failed_request_client.dart
```diff
@@ -0,0 +1,174 @@
...
+  final Hub _hub;
+
+  /// Configures up to which size request bodies should be included in events
+  final MaxRequestBodySize maxRequestBodySize;
+
+  /// Describes which HTTP status codes should be considered as a failed
+  /// requests.
+  ///
+  /// Per default no status code is considered a failed request.
+  final List<int> failedRequestStatusCodes;
```
marandaneto: they could be internal

dart/lib/src/http_client/failed_request_client.dart
```diff
@@ -0,0 +1,174 @@
...
+    try {
+      final response = await _client.send(request);
+
+      if (failedRequestStatusCodes.contains(response.statusCode)) {
+        await _captureException(request: request);
+      }
+
+      return response;
+    } catch (exception, stackTrace) {
+      await _captureException(
```
marandaneto: you call `_captureException` twice, as you can use optional params, maybe we could do it in the finally block only once

dart/lib/src/http_client/failed_request_client.dart
```diff
@@ -0,0 +1,174 @@
...
+    Client? client,
+    Hub? hub,
+  })  : _hub = hub ?? HubAdapter(),
+        _client = client ?? Client();
+
+  final Client _client;
+  final Hub _hub;
+
+  /// Configures up to which size request bodies should be included in events
+  final MaxRequestBodySize maxRequestBodySize;
```
marandaneto, line 88: currently not being checked, right?

dart/lib/src/http_client/sentry_http_client.dart
marandaneto: one thing to consider is, both clients call `send`, does it mean it executes the URL request twice? another thing to consider is, if the 1st client rethrows, does the other client still executes correctly? I guess spitting up may create a bit more complexity than making it easier to test, we can discuss this offline

marandaneto: also, this is a breaking change, setting `SentryHttpClient` which is the latest stable version would not capture errored HTTP calls, now it will, and its a breaking change, ideally those features are opt-in and not opt-out, till we get it stable and do a major bump

dart/lib/src/http_client/failed_request_client.dart
ueman: No not really. The first call is for non-throwing "bad" events like 500 status codes. The second call is for failing throwing requests. 

Also I don't have access to the exception and it's stacktrace in the finally block so I would need to save them to variables. For bad requests I would also need to save it to a variable. How would that be better?

dart/lib/src/http_client/failed_request_client.dart
ueman: Correct, it's still WIP :D

dart/lib/src/http_client/sentry_http_client.dart
ueman: Side note: What I've done here is to apply the decorator pattern. Each client (decorator) adds functionality. Each decorator just calls its inner object and the innermost object is a Client (either supplied by the user or created by us).

dart/lib/src/http_client/failed_request_client.dart
marandaneto: lets check also if WeakRef exists and works for Web too

ueman: If there's no WeakRef, I think we could just use a Queue which only keeps the last x events. Would that work?

ueman:
@bruno-garcia 
What do you think of splitting each HTTP client related functionality in its own client?
So recording breadcrumbs goes into the `BreadcrumbClient`, capturing "bad" requests (requests which throw or have a 400/500 status code for example) goes into the `FailedRequestClient`. 
Additionally there's the `SentryHttpClient` which then uses both internally and it's the user facing class.

In the future there would then be a third client for the distributed tracing thingy.

dart/lib/src/http_client/failed_request_client.dart
ueman: I created an issue for this: https://github.com/getsentry/sentry-dart/issues/480

dart/lib/src/http_client/failed_request_client.dart
marandaneto: I'd not say its better, its just a conventional if you want to do something within the try and catch block, see https://github.com/getsentry/sentry-java/blob/main/sentry-android-okhttp/src/main/java/io/sentry/android/okhttp/SentryOkHttpInterceptor.kt

also, let's say you have a bug in the `_captureException` method that throws if the request is broken, it throws and go to `catch` block, it calls `_captureException` again, its a side effect.

marandaneto:
> @bruno-garcia
> What do you think of splitting each HTTP client related functionality in its own client?
> So recording breadcrumbs goes into the `BreadcrumbClient`, capturing "bad" requests (requests which throw or have a 400/500 status code for example) goes into the `FailedRequestClient`.
> Additionally there's the `SentryHttpClient` which then uses both internally and it's the user facing class.
> 
> In the future there would then be a third client for the distributed tracing thingy.

@bruno-garcia opinions?

dart/lib/src/http_client/failed_request_client.dart
ueman: Ah that's a good point. I haven't considered that.

bruno-garcia:
Sorry I'll take a look in a bit

bruno-garcia:
The decorator approach LGTM as its indeed easier to test and compose features as needed.

A couple of things to keep in mind is the usability though, the user should have a simple way into the API without having to worry about the decoration composition. This will also to allow us to add new decorators later without affecting the user.

So if we can keep: `SentryHttpClient()` that bundles all decorators by default, it's a great approach.
The order also is important, for example add the breadcrumb after capturing the event (so the event doesn't include double crumbs, the artificial one created by Sentry plus the one from the integration).

ueman:
> So if we can keep: SentryHttpClient() that bundles all decorators by default, it's a great approach.

That's the plan.

> The order also is important, for example add the breadcrumb after capturing the event (so the event doesn't include double crumbs, the artificial one created by Sentry plus the one from the integration).

That's a good point. Just changed it accordingly.

bruno-garcia:
The integration creating events directly will open a new precedence for other SDKs. We don't do this yet from anywhere AFAIK. At least not in the Mobile team's SDKs. Including .NET and Java bases.

Would love to hear your learnings once this is live and some customer feedback came in

bruno-garcia (approved):

marandaneto (approved):
LGTM

marandaneto:
@bruno-garcia would you like to do a final review before we merge and release it as beta? as this PR drives the PoC

ueman:
> There's the risk of creating events too large for ingestion.

What is the largest allowed size? I'll add another test for that.

> Finally, PII is another concern so docs should clearly state that there's this risk and point to the scrubbing data parts of the docs. I'm not too familiar with the automatic scrubbing we already have, worth pointing to that too since there's some level of default protection against PII getting in already.

Gonna do that, too

dart/lib/src/http_client/failed_request_client.dart
```diff
@@ -0,0 +1,239 @@
...
+///      body: {'name': 'doodle', 'color': 'blue'});
+///  print(await client.get(uriResponse.bodyFields['uri']));
+/// } finally {
+///  client.close();
+/// }
+/// ```
+class FailedRequestClient extends BaseClient {
+  FailedRequestClient({
+    this.maxRequestBodySize = MaxRequestBodySize.small,
+    this.failedRequestStatusCodes = const [],
```
marandaneto, line 68: maybe we have a sort of default value here? otherwise, it won't capture anything ootb.
eg 400...599

dart/lib/src/http_client/failed_request_client.dart
```diff
@@ -0,0 +1,239 @@
...
+      }
+
+      if (failedRequestStatusCodes.containsStatusCode(statusCode)) {
+        // Capture an exception if the status code is considered bad
+        await _captureEvent(
+          request: request,
+          reason: failedRequestStatusCodes.toDescription(),
+          requestDuration: stopwatch.elapsed,
+        );
+      }
```
marandaneto: does it mean its gonna capture 2 events? I see 2 calls to `_captureEvent`

marandaneto: should it be either if there was an exception or if it was within the given range?

dart/lib/src/http_client/failed_request_client.dart
ueman: Hm, I'm not sure.

The [OpenApiGenerator](https://openapi-generator.tech/docs/generators/dart/) throws exceptions if the sever responds with 4xx to 5xx status codes. I'm not quite sure if we could deduplicate those errors.
Then there's also [chopper](https://pub.dev/packages/chopper) and [retrofit](https://pub.dev/packages/retrofit). I would have to look up, how they handle those status code.

My concern is that we would generate way to much noise by including them by default.

We could provide however a couple named constructors for common scenarios:
- server failed (5xx)
- client sent wrong request (4xx)
- all (4xx to 5xx)

marandaneto (commented):
> > There's the risk of creating events too large for ingestion.
> 
> What is the largest allowed size? I'll add another test for that.

as we are using `MaxRequestBodySize`, I guess we don't need to do anything more, if the user is using `always` its their own risk to get events dropped, see a similar issue https://github.com/getsentry/sentry-java/issues/910
checking the size of the event before sending has runtime cost so that's why we don't do it.
just in case of know-how, https://develop.sentry.dev/sdk/envelopes/#size-limits
worth noting that this is for envelopes, which is gonna be into the next major as well

dart/lib/src/http_client/failed_request_client.dart
```diff
@@ -0,0 +1,239 @@
...
+    StackTrace? stackTrace,
+    String? reason,
+    required Duration requestDuration,
+    required BaseRequest request,
+  }) {
+    // As far as I can tell there's no way to get the uri without the query part
+    // so we replace it with an empty string.
+    final urlWithoutQuery = request.url.replace(query: '').toString();
+
+    final query = request.url.query.isEmpty ? null : request.url.query;
```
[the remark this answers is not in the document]
ueman: This is just for having it in the correct fields of the `Request` class.

dart/lib/src/http_client/failed_request_client.dart
marandaneto: I see, I think the named ctors would be nice, but not adding anything by default sounds suboptimal too, you're setting up the client already, which means you already want to capture events, the granularity could be configured but should offer something by default IMO.

iirc retrofit only throws if there was a network error, parsing error, or something like that, but if an API returns 400 to 599, you can still read the response, so it does not throw, I believe the `http` Dart is the same, if the Library throws when it returns 4xx or 5xx, this would be almost useless if you send a body response, you'd not be able to read it.

the way how it is, I believe it'd never capture anything when returning 4xx or 5xx by default

dart/lib/src/http_client/failed_request_client.dart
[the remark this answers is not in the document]
marandaneto: ah ok

dart/lib/src/http_client/failed_request_client.dart
```diff
@@ -0,0 +1,239 @@
...
+
+    final sentryRequest = SentryRequest(
+      method: request.method,
+      headers: request.headers,
+      url: urlWithoutQuery,
+      queryString: query,
+      cookies: request.headers['Cookie'],
```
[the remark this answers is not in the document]
ueman: What would be the best way to get the SentryOptions in order to check for sendDefaultPii? There's no way to get it currently.

dart/lib/src/http_client/failed_request_client.dart
```diff
@@ -0,0 +1,245 @@
...
+  }
+
+  // See https://develop.sentry.dev/sdk/event-payloads/request/
+  Future<void> _captureEvent({
+    Object? exception,
+    StackTrace? stackTrace,
+    String? reason,
+    required Duration requestDuration,
+    required BaseRequest request,
+    required bool withPii,
```
marandaneto: this should be read from `options.sendDefaultPii`, the thing is, there's no way to access it, in this case, you might need to expose the `Hub#getOptions` like Java does https://github.com/getsentry/sentry-java/blob/main/sentry/src/main/java/io/sentry/Hub.java#L438-L441

so instead of taking  `withPii` as param, you actually call `hub.getOptions.sendDefaultPii`

dart/lib/src/http_client/failed_request_client.dart
```diff
@@ -0,0 +1,245 @@
...
+        await _captureEvent(
+          exception: exception,
+          stackTrace: stackTrace,
+          request: request,
+          requestDuration: stopwatch.elapsed,
+          withPii: true,
+        );
+      }
+
+      if (failedRequestStatusCodes.containsStatusCode(statusCode)) {
```
marandaneto: maybe this becomes an `else if` branch from the if above? so we avoid 2 checks when only 1 can be true

dart/lib/src/http_client/failed_request_client.dart
ueman: Yeah, I thought so too.
How would that work for the HubAdapter?
The HubAdapter only passes through to the static methods.

marandaneto: damn, Java does `Sentry.getCurrentHub().getOptions()` and we just removed `getCurrentHub` cus its not part of the unified API but it'd be fine as a private method, wondering if it makes sense to rollback https://github.com/getsentry/sentry-dart/pull/490/files and mark it as internal (meta annotation), and use it to access the options.

I don't see another way of sharing the `defaultDefaultPii` flag, maybe @bruno-garcia has ideas?

ueman: We could also just have a `includePii` for the constructor. I mean the user has to use the client manually anyway.

Though I'm not saying it's necessarily better 🤷‍♂️

dart/lib/src/http_client/breadcrumb_client.dart
```diff
@@ -0,0 +1,98 @@
...
+        type: 'http',
+        category: 'http',
+        data: {
+          'url': request.url.toString(),
+          'method': request.method,
+          if (statusCode != null) 'status_code': statusCode,
+          if (reason != null) 'reason': reason,
+          'duration': stopwatch.elapsed.toString(),
+        },
+      );
```
marandaneto: this could be a factory ctor called `http` that takes the optional properties, eg https://github.com/getsentry/sentry-java/blob/main/sentry/src/main/java/io/sentry/Breadcrumb.java#L59-L91

dart/lib/src/http_client/failed_request_client.dart
```diff
@@ -0,0 +1,243 @@
...
+      data: _getDataFromRequest(request),
+      other: {
+        'content_length': request.contentLength.toString(),
+        'duration': requestDuration.toString(),
+      },
+    );
+
+    final mechanism = Mechanism(
+      type: 'SentryHttpClient',
+      handled: true,
```
marandaneto: setting to `true` or not setting at all is the same, we could remove and save a few chars

dart/lib/src/http_client/failed_request_client.dart
```diff
@@ -0,0 +1,243 @@
...
+      description: reason,
+    );
+    final throwableMechanism = ThrowableMechanism(mechanism, exception);
+
+    final event = SentryEvent(
+      throwable: throwableMechanism,
+      request: sentryRequest,
+      level: SentryLevel.error,
```
marandaneto: backend infers `error` automatically, we could remove and save a few chars

dart/test/http_client/failed_request_client_test.dart
```diff
@@ -0,0 +1,277 @@
...
+
+      await sut.get(requestUri, headers: {'Cookie': 'foo=bar'});
+
+      expect(fixture.hub.captureEventCalls.length, 1);
+    });
+
+    test('close does get called for user defined client', () async {
+      final mockHub = MockHub();
+
+      final mockClient = CloseableMockClient();
```
marandaneto, line 129: why not using the fixture for this test?

ueman: It's a different one.

dart/lib/src/http_client/failed_request_client.dart
marandaneto: works for me

marandaneto (approved):
LGTM

Issue #476, opened 2021-05-28
Better grouping / in-app detection


HazAT:
Looking through a few errors of our own mobile app:

https://sentry.io/organizations/sentry/issues/2422252071/events/4beecc04944246ddaf284cb4119c5b1a/?project=5645511
https://sentry.io/organizations/sentry/issues/2339061202/events/41cea4bcd0384197b045f8a6db840bf7/?project=5645511
https://sentry.io/organizations/sentry/issues/2422252066/events/9db886252e5d418d846abfe372b60082/?project=5645511

The only difference is always the 3rd frame:
`sentry_api.dart in SentryApi.organizations at line 31 within sentry_mobile`
`sentry_api.dart in SentryApi.sessions at line 190 within sentry_mobile`
`sentry_api.dart in SentryApi.projectIdsWithSessions at line 70 within sentry_mobile`

![CleanShot 2021-05-28 at 10 46 40](https://user-images.githubusercontent.com/363802/119957032-0128f280-bfa2-11eb-854c-d24d665b62ae.png)
![CleanShot 2021-05-28 at 10 46 53](https://user-images.githubusercontent.com/363802/119957055-07b76a00-bfa2-11eb-8bf9-e84a019eec1f.png)
![CleanShot 2021-05-28 at 10 47 02](https://user-images.githubusercontent.com/363802/119957086-0e45e180-bfa2-11eb-8403-f1b9cff63d02.png)

I understand that those are different error and therefore not grouped together but since we use the first frame to describe the issue in the list it looks like this:
![CleanShot 2021-05-28 at 10 50 10](https://user-images.githubusercontent.com/363802/119957637-88766600-bfa2-11eb-98bb-d345f657afe4.png)

which is not helpful. If we would have better `in-app` frame detection this should be way more helpful.

ueman:
The last frame is also different.
`io_client.dart` and `base_client.dart` are definitely not in app because they're coming from [http](https://pub.dev/packages/http).

marandaneto:
indeed, these 2 frames should not be considered inApp, something to be improved on https://github.com/getsentry/sentry-dart/blob/main/dart/lib/src/sentry_stack_trace_factory.dart#L145-L171

ueman:
[...]

PR #482, opened 2021-05-31
Feat: Allow setting of default values for in-app-frames via options

ueman:
## :scroll: Description

This adds the option for setting the default value for in app frames.

Relates to https://github.com/getsentry/sentry-dart/issues/476

Added a test

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes


CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # Unreleased
 
 * Fix: Merge user from event and scope (#467)
+* Feature: Allow setting of default values for in-app-frames via `SentryOptions.isStackFrameInAppDefault`
```
```suggestion
* Feature: Allow setting of default values for in-app-frames via `SentryOptions.isStackFrameInAppDefault` (#482)
```

dart/lib/src/sentry_stack_trace_factory.dart
```diff
@@ -17,7 +17,6 @@ class SentryStackTraceFactory {
 
   /// returns the [SentryStackFrame] list from a stackTrace ([StackTrace] or [String])
   List<SentryStackFrame> getStackFrames(dynamic stackTrace) {
-    // TODO : fix : in release mode on Safari passing a stacktrace object fails, but works if it's passed as String
```
ueman, line 20: This isn't needed anymore.

dart/lib/src/sentry_stack_trace_factory.dart
```diff
@@ -147,26 +148,32 @@ class SentryStackTraceFactory {
...
+      // Early bail out.
+      return _options.isStackFrameInAppDefault;
     }
+    // The following code depends on the scheme being set.
 
-    for (final include in _options.inAppIncludes) {
-      if (frame.package != null && frame.package == include) {
+    final package = frame.package;
+    if (package != null) {
+      if (_options.inAppIncludes.contains(package)) {
```
bruno-garcia, line 158: Should this be `startsWith` instead?

dart/lib/src/sentry_options.dart
```diff
@@ -136,6 +136,15 @@ class SentryOptions {
   /// example : ['sentry'] will include exception from 'package:sentry/sentry.dart'
   List<String> get inAppIncludes => List.unmodifiable(_inAppIncludes);
 
+  /// Configures wether stack trace frames are considered in app frames as default.
+  /// You can use this to essentially make [inAppIncludes] or [inAppExcludes]
+  /// a white or black list.
```
marandaneto: allow/deny list instead https://develop.sentry.dev/inclusion/

dart/test/stack_trace_test.dart
```diff
@@ -88,6 +88,31 @@ void main() {
...
+
+      expect(serializedFrame.inApp, true);
+
+      // make sure the inverse also applies
+      options.isStackFrameInAppDefault = false;
+
+      serializedFrame =
+          SentryStackTraceFactory(options).encodeStackTraceFrame(frame)!;
+
+      expect(serializedFrame.inApp, false);
```
marandaneto: lets make a new test for this one

dart/test/stack_trace_test.dart
```diff
@@ -88,6 +88,31 @@ void main() {
...
+      // - frame.package is null
+      // These conditions triggers the default value being used
+      final frame = Frame.parseVM('#0 Foo (async/future.dart:0:0)');
+
+      // default is true
+      final options = SentryOptions(dsn: fakeDsn)
+        ..isStackFrameInAppDefault = true;
+
+      var serializedFrame =
+          SentryStackTraceFactory(options).encodeStackTraceFrame(frame)!;
```
marandaneto: this could be under a fixture and getSut

dart/test/stack_trace_test.dart
```diff
@@ -31,7 +31,7 @@ void main() {
       );
     });
 
-    test('cleanses absolute paths', () {
+    test('cleanes absolute paths', () {
```
marandaneto:
```suggestion
    test('cleans absolute paths', () {
```

dart/lib/src/sentry_options.dart
```diff
@@ -136,6 +136,15 @@ class SentryOptions {
...
+  /// This value is only used if Sentry can not find the frames origin.
```
```suggestion
  /// This value is only used if Sentry can not find the origin frame.
```

dart/lib/src/sentry_stack_trace_factory.dart
ueman: Could you elaborate on that? I think I don't get what you'r hinting at.

Though I'm pretty sure it shouldn't be a `startsWith`.
`inAppIncludes` is a list of Strings (package names) which should be considered as in app frames.
So I think the name should match exactly. If we're using `startsWith` and use `flutter` as an include
the `flutter_svg` package would also be considered in app.

dart/lib/src/sentry_options.dart
ueman: Dang, I forgot about that :(

dart/lib/src/sentry_stack_trace_factory.dart
```diff
@@ -147,26 +148,32 @@ class SentryStackTraceFactory {
...
     }
 
-    if (frame.isCore ||
-        (frame.uri.scheme == 'package' && frame.package == 'flutter')) {
+    if (frame.isCore) {
+      // This is a Dart frame
+      return false;
+    }
+
+    if (frame.package == 'flutter') {
```
marandaneto, line 172: could still be a `||` operation, right separating it?

dart/lib/src/sentry_stack_trace_factory.dart
```diff
@@ -147,26 +148,32 @@ class SentryStackTraceFactory {
...
-    for (final exclude in _options.inAppExcludes) {
-      if (frame.package != null && frame.package == exclude) {
+      if (_options.inAppExcludes.contains(package)) {
         return false;
       }
     }
 
-    if (frame.isCore ||
-        (frame.uri.scheme == 'package' && frame.package == 'flutter')) {
+    if (frame.isCore) {
```
marandaneto, line 167: why don't we check `frame.uri.scheme == 'package'` anymore`? isn't needed anymore?

ueman: The `frame.uri.scheme == 'package'` check is redundant. If the scheme isn't package, `frame.package` would be null.

dart/lib/src/sentry_stack_trace_factory.dart
ueman: Yep, but I think it's worth being more explicit here because it's such an important part.

dart/lib/src/sentry_options.dart
```diff
@@ -136,6 +136,15 @@ class SentryOptions {
...
   List<String> get inAppIncludes => List.unmodifiable(_inAppIncludes);
 
+  /// Configures wether stack trace frames are considered in app frames as default.
+  /// You can use this to essentially make [inAppIncludes] or [inAppExcludes]
+  /// an allow or deny list.
+  /// This value is only used if Sentry can not find the origin of the frame.
+  ///
+  /// E.g. if [isStackFrameInAppDefault] is true you only need to maintain
+  /// [inAppExcludes].
+  bool isStackFrameInAppDefault = true;
```
ueman: I feel like there could be a better name for this, though I can't think of any

kuhnroyal:
Looking forward to this! Thanks!

dart/lib/src/sentry_options.dart
marandaneto: we don't add `is` to the other bools in this class, whats about:
`inAppFrames`only? or `inAppFramesInclusive` or `inAppInclusiveFrames`.
@bruno-garcia ?

dart/lib/src/sentry_stack_trace_factory.dart
marandaneto: on Flutter and Dart it makes sense to match exactly because we match only the package which isn't hierarquical.
on Java/Android is the package, which is like `io.sentry` so  `startsWith` makes sense because I may have `io.sentry.a` and `io.sentry.b` but I don't want to maintain a list of my endless packages, so I just add the root of my package.

dart/lib/src/sentry_stack_trace_factory.dart
marandaneto: I think a `||` operation is just as clean as 2 ifs, the downside is that this has a cost, well, its probably not even noticeable, computers are designed for this, I don't have strong feelings here so feel free to keep it as it is

dart/lib/src/sentry_options.dart
bruno-garcia: I'll throw more ideas but ultimately it's your call folks:
* `defaultInAppInclude` - As in, by default it's in app included.
* `preferInApp`

sorry, I don't feel like my suggestions are necessarily less crappy :)

ueman: `considerStackFramesInApp` 😁

Note to self: Edit changelog accordingly

dart/lib/src/sentry_stack_trace_factory.dart
bruno-garcia: RIght, I didn't think this through. Java and .NET is `startsWith`. But here as @marandaneto indicated, I believe it should be an exact match as opposed to `contains`

marandaneto: the `contains` here does an exact match, the `contains` is in the list instance and not in the string, so it's correct already

dart/lib/src/sentry_options.dart
```diff
@@ -136,6 +136,15 @@ class SentryOptions {
   /// example : ['sentry'] will include exception from 'package:sentry/sentry.dart'
   List<String> get inAppIncludes => List.unmodifiable(_inAppIncludes);
 
+  /// Configures wether stack trace frames are considered in app frames as default.
```
```suggestion
  /// Configures whether stack trace frames are considered in app frames by default.
```

dart/lib/src/sentry_options.dart
```diff
@@ -136,6 +136,15 @@ class SentryOptions {
...
+  /// You can use this to essentially make [inAppIncludes] or [inAppExcludes]
+  /// an allow or deny list.
+  /// This value is only used if Sentry can not find the origin of the frame.
+  ///
+  /// E.g. if [isStackFrameInAppDefault] is true you only need to maintain
+  /// [inAppExcludes].
```
marandaneto: maybe add the example for `false` too

dart/lib/src/sentry_options.dart
```suggestion
  bool inAppFramesByDefault = true;
```

this sounds better to me tbh

marandaneto:
let's consider adding the new options in https://docs.sentry.io/platforms/dart/configuration/options/ near https://docs.sentry.io/platforms/dart/configuration/options/#in-app-include
same for flutter docs

ueman:
Docs PR is here https://github.com/getsentry/sentry-docs/pull/3684

dart/lib/src/sentry_options.dart
```diff
@@ -136,6 +136,17 @@ class SentryOptions {
...
+  /// Configures whether stack trace frames are considered in app frames by default.
+  /// You can use this to essentially make [inAppIncludes] or [inAppExcludes]
+  /// an allow or deny list.
+  /// This value is only used if Sentry can not find the origin of the frame.
+  ///
+  /// - If [considerFramesInAppByDefault] is true you only need to maintain
+  /// [inAppExcludes].
+  /// - If [considerFramesInAppByDefault] is false you only need to maintain
+  /// [inAppIncludes].
+  bool considerFramesInAppByDefault = true;
```
```suggestion
  bool considerInAppFramesByDefault = true;
```

looks like this reads better, wdyt?

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # Unreleased
 
 * Fix: Merge user from event and scope (#467)
+* Feature: Allow setting of default values for in-app-frames via `SentryOptions.isStackFrameInAppDefault` (#482)
```
```suggestion
* Feature: Allow setting of default values for in-app-frames via `SentryOptions.considerInAppFramesByDefault` (#482)
```

if we agree on the naming

dart/lib/src/sentry_options.dart
```diff
@@ -136,6 +136,17 @@ class SentryOptions {
...
 
+  /// Configures whether stack trace frames are considered in app frames by default.
+  /// You can use this to essentially make [inAppIncludes] or [inAppExcludes]
+  /// an allow or deny list.
+  /// This value is only used if Sentry can not find the origin of the frame.
+  ///
+  /// - If [considerFramesInAppByDefault] is true you only need to maintain
+  /// [inAppExcludes].
+  /// - If [considerFramesInAppByDefault] is false you only need to maintain
+  /// [inAppIncludes].
```
```suggestion
  /// - If [considerInAppFramesByDefault] is true you only need to maintain
  /// [inAppExcludes].
  /// - If [considerInAppFramesByDefault] is false you only need to maintain
  /// [inAppIncludes].
```

if we agree on the naming

dart/lib/src/sentry_options.dart
ueman: I don't think so but the docs are good so I'm fine with both names.

dart/test/stack_trace_test.dart
```diff
@@ -181,17 +202,33 @@ void main() {
...
+      ]);
     });
   });
 }
+
+class Fixture {
+  SentryStackTraceFactory getSut({
+    List<String> inAppIncludes = const [],
+    List<String> inAppExcludes = const [],
+    bool considerFramesInAppByDefault = true,
```
marandaneto:
```suggestion
    bool considerInAppFramesByDefault = true,
```

marandaneto: also its usages

marandaneto (approved):
left a last comment but rather than that LGTM

dart/test/stack_trace_test.dart
ueman: fixed 💯

Issue #480, opened 2021-05-31
DeduplicationEventProcessor


ueman:
In order to remove duplicated events, for example caused by [this](https://github.com/getsentry/sentry-dart/pull/473#discussion_r638492342) we need an event processor which removes duplicated events.

See https://github.com/getsentry/sentry-java/blob/b45944c93d53f68bc1e64f5ac4c9a5400b252d44/sentry/src/main/java/io/sentry/DuplicateEventDetectionEventProcessor.java for reference

Depends on https://github.com/getsentry/sentry-dart/issues/479

bruno-garcia:
The Java one was based on this but for reference: https://github.com/getsentry/sentry-dotnet/blob/626d9966d600615a125da5d90cf8d25931aefb94/src/Sentry/Internal/DuplicateEventDetectionEventProcessor.cs#L15

marandaneto:
Dart does not have a WeakRef type so we'd need to figure this out

PR #489, opened 2021-06-07
Feat: Change EventProcessor to an interface

### LINKED ISSUE #479: Make `EventProcessor` an interface
Currently `EventProcessor` is just a typealias. However we need it to be an interface.

This is needed for the added processors in https://github.com/getsentry/sentry-dart/pull/452 
and also for a future `DeduplicationEventProcessor` for [this](https://github.com/getsentry/sentry-dart/pull/473#discussion_r638559139) from https://github.com/getsentry/sentry-dart/pull/473 .

ueman:
## :scroll: Description
EventProcessor is now an interface instead of a pure function

Closes https://github.com/getsentry/sentry-dart/issues/479

Unblocks
- https://github.com/getsentry/sentry-dart/issues/480
- https://github.com/getsentry/sentry-dart/pull/473
- https://github.com/getsentry/sentry-dart/pull/452

## :pencil: Checklist

- [x] All tests passing
- [ ] No breaking changes


CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 # Unreleased
 
+* Feat: EventProcessor changed to an interface
```
```suggestion
* Feat: EventProcessor changed to an interface (#489)
```

CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 # Unreleased
 
+* Feat: EventProcessor changed to an interface (#489)
```
```suggestion
* Ref: EventProcessor changed to an interface (#489)
```

technically a refactoring I guess

flutter/test/load_contexts_integrations_test.dart
```diff
@@ -74,23 +72,22 @@ void main() {
...
-        as FutureOr<SentryEvent>);
+    final event = await fixture.options.eventProcessors.first.apply(e);
 
     expect(called, true);
-    expect(event.contexts.device!.name, 'Device1');
-    expect(event.contexts.app!.name, 'test-app');
-    expect(event.contexts.operatingSystem!.name, 'os1');
-    expect(event.contexts.gpu!.name, 'gpu1');
-    expect(event.contexts.browser!.name, 'browser1');
+    expect(event?.contexts.device!.name, 'Device1');
```
marandaneto: would it make sense to be `!` instead of `?` here and all the following tests?

dart/lib/src/protocol/event_processor.dart
```diff
@@ -0,0 +1,13 @@
+import 'dart:async';
+
+import '../protocol.dart';
+
+/// [EventProcessor]s are callbacks that run for every event. They can either
+/// return a new event which in most cases means just adding data *or* return
+/// null in case the event will be dropped and not sent.
+abstract class EventProcessor {
```
marandaneto: event processor is part of the unified API but not the protocol, protocol is the data classes, so I'd move it

flutter/test/load_contexts_integrations_test.dart
ueman: I'm wondering the same but I think having `?` is slightly better because you'll get the error or failure message from the `excpect` instead of Dart runtime error message.

marandaneto: makes sense, but then it should be `?` everywhere, right, if `device` is null and using `!` wont work either

marandaneto:
do we need to add `export 'event_processor.dart';` to the sentry.dart file?

marandaneto (approved):
LGTM

PR #494, opened 2021-06-13
Feat: Make `beforeSend` async

### LINKED ISSUE #459: Consider making `beforeSend` and `beforeBreadcrumb` async
Currently `beforeSend` and `beforeBreadcrumb` are not async which makes it hard to use any libraries like device_info_plus hard to use in those callbacks.

`beforeSend` is already only used in an async context. `beforeBreadcrumb` is used in a non-async context.

Event processors are already async.

ueman:
Closes https://github.com/getsentry/sentry-dart/issues/459

## :pencil: Checklist

- [x] All tests passing
- [ ] No breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # Unreleased
 
 * Feat: Add Culture Context (#491)
-* Feat: `beforeSend` callback accepts async code
+* Feat: `beforeSend` callback accepts async code (#494)
 
 ## Breaking Changes:
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Unreleased
 
 * Feat: Add Culture Context (#491)
+* Feat: `beforeSend` callback accepts async code
 
 ## Breaking Changes:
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # Unreleased
 
 * Feat: Add Culture Context (#491)
+* Feat: `beforeSend` callback accepts async code
```
```suggestion
* Feat: `beforeSend` callback accepts async code (#494)
```

dart/lib/src/sentry_options.dart
```diff
@@ -242,8 +243,10 @@ class SentryOptions {
 
 /// This function is called with an SDK specific event object and can return a modified event
 /// object or nothing to skip reporting the event
-typedef BeforeSendCallback = SentryEvent? Function(SentryEvent event,
-    {dynamic hint});
+typedef BeforeSendCallback = FutureOr<SentryEvent?> Function(
```
marandaneto, line 246: should we update the tests in sentry_client_test?

ueman: They're still valid and green, so I'm not quite sure what you're trying to say.

marandaneto: while they don't break because of `FutureOr` which is a good thing, we could either add a new test with the new signature (having an async callback), or make the current ones async., I'm fine either way.

marandaneto (approved):
LGTM

PR #495, opened 2021-06-13
Ref: Improve SentryFlutterOptions and add tests for NativeSdkIntegration

### LINKED ISSUE #440: Write tests for passing SentryOptions via method channel to native
There are currently no tests which check that options are passed correctly or at all to the native side.
We should write some for the `NativeSdkIntegration`.

### LINKED ISSUE #485: Rename `cacheDirSize` to `maxCacheItems`
Similar to https://github.com/getsentry/sentry-java/issues/1162

The key for communicating the changes to the native side needs to be renamed too.

ueman:
Closes https://github.com/getsentry/sentry-dart/issues/485
Closes https://github.com/getsentry/sentry-dart/issues/440

Added new tests

## :pencil: Checklist

- [x] All tests passing
- [ ] No breaking changes


CHANGELOG.md
```diff
@@ -8,6 +8,9 @@
 * Feat: Support envelope based transport for events (#391)
   * The method signature of `Transport` changed from `Future<SentryId> send(SentryEvent event)` to `Future<SentryId> send(SentryEnvelope envelope)`
 * Remove `Sentry.currentHub` (#490)
+* Ref: Rename `cacheDirSize` to `maxCacheItems` and add `maxCacheItems` for iOS (#495)
+* Ref: `autoSessionTrackingIntervalMillis` renamed to `autoSessionTrackingInterval` (#495)
+* Ref: `anrTimeoutIntervalMillis` renamed to `anrTimeoutInterval` (#495)
```
marandaneto: all the other SDKs have a simple Long as an interval in Millis.
not all the languages have a Duration builtin, not on Java 6/7 which is available on Android, for example, keeping it as a Long makes it easier to unify the options.
By comparing on Android, just noticed this issue https://github.com/getsentry/sentry-java/issues/1530

I'd not do this on this PR as it's not described in the 2 related issues.
I'll bring this up though in the unified API meeting, maybe more people would be interested in changing this as well.

flutter/test/native_sdk_integration_test.dart
```diff
@@ -0,0 +1,160 @@
...
+
+    test('test default values', () async {
+      final channel = const MethodChannel('initNativeSdk');
+
+      String? methodName;
+      dynamic arguments;
+      channel.setMockMethodCallHandler((call) async {
+        methodName = call.method;
+        arguments = call.arguments;
+      });
```
marandaneto: this part looks duplicated across tests, probably better if extracted to a method

flutter/test/native_sdk_integration_test.dart
```diff
@@ -0,0 +1,160 @@
...
+        'release': null,
+        'enableAutoSessionTracking': true,
+        'enableNativeCrashHandling': true,
+        'attachStacktrace': true,
+        'attachThreads': false,
+        'autoSessionTrackingIntervalMillis': 30000,
+        'dist': null,
+        'integrations': [],
+        'packages': [
+          {'name': 'pub:sentry', 'version': '5.1.1'}
```
marandaneto: this will fail when we bump the version every time I believe, better reading the version from `sdkVersion` property

flutter/test/native_sdk_integration_test.dart
```diff
@@ -0,0 +1,160 @@
...
+
+      String? methodName;
+      dynamic arguments;
+      channel.setMockMethodCallHandler((call) async {
+        methodName = call.method;
+        arguments = call.arguments;
+      });
+      var sut = Fixture().getSut(channel);
+
+      await sut.call(HubAdapter(), createOptions());
```
marandaneto, line 25: should `HubAdapter()` be a simple mock or fake? otherwise calls will be forwarded to the sentry static class that has a state

CHANGELOG.md
ueman: At least C# uses `TimeSpan`s for time units too: https://github.com/getsentry/sentry-dotnet/blob/main/src/Sentry/SentryOptions.cs#L276

ueman: Yep, I'll do another PR/issue for that though.

flutter/test/native_sdk_integration_test.dart
ueman: Good catch!

flutter/test/native_sdk_integration_test.dart
ueman: NativeSDKIntegration doesn't do anything with the Hub so it doesn't matter what I pass. But yeah, it should probably be Hub, just so that it reads better.

CHANGELOG.md
marandaneto: mmm, interesting, TIL, I will bring this to the meeting too then, thanks.

marandaneto: the downside of using such builtin durations as well is when reading configuration from external resources eg file system (spring boot etc), we'd need 2 fields to set such property, one with the duration in time and another one with the duration type (millis, seconds, or whatever the user wanted to set).
we can discuss this in the other issue though.

flutter/test/native_sdk_integration_test.dart
marandaneto: good point, lets just either make a dummy fake or mock, so when we read it, we know it does nothing, I thought it was bounded with the HubAdapter behavior somehow.

flutter/test/mocks.dart
```diff
@@ -63,3 +63,54 @@ class MockPlatform implements Platform {
...
+  @override
+  void addBreadcrumb(Breadcrumb crumb, {hint}) {}
+
+  @override
+  void bindClient(SentryClient client) {}
+
+  @override
+  Future<SentryId> captureEvent(SentryEvent event,
+      {stackTrace, hint, ScopeCallback? withScope}) {
+    throw UnimplementedError();
```
marandaneto: to make this a bit more useful then I'd return `SentryId.empty()` instead of `UnimplementedError` for all methods

flutter/test/native_sdk_integration_test.dart
```diff
@@ -0,0 +1,163 @@
...
+    });
+
+    test('test default values', () async {
+      String? methodName;
+      dynamic arguments;
+      final channel = createChannelWithCallback((call) async {
+        methodName = call.method;
+        arguments = call.arguments;
+      });
+      var sut = Fixture().getSut(channel);
```
marandaneto: maybe create a new instance of the `Fixture` in the `setUp` as the other tests, so you don't have to repeat

CHANGELOG.md
marandaneto: this should be gone I believe as you removed the interval refactoring

marandaneto (approved):
LGTM

PR #498, opened 2021-06-15
Feat: DeduplicationEventProcessor

### LINKED ISSUE #480: DeduplicationEventProcessor

ueman:
Closes https://github.com/getsentry/sentry-dart/issues/480

new tests

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes


dart/lib/src/event_processor/deduplication_event_processor.dart
```diff
@@ -0,0 +1,56 @@
...
+      _options.logger(
+        SentryLevel.debug,
+        'Duplicate Exception detected. Event ${event.eventId} will be discarded.',
+      );
+      return null;
+    }
+    // No duplication detected
+    exceptionMap[exception] = null;
+    if (exceptionMap.length > eventsToKeepForDeduplication) {
+      exceptionMap.remove(exceptionMap.entries.last.key);
```
ueman: As we don't have a weak map or similar I think having a max length is an acceptable solution.
I haven't yet really thought about what a good length might be and just choose 20 for now.

dart/lib/src/sentry_options.dart
```diff
@@ -188,6 +189,10 @@ class SentryOptions {
   /// Whether to send personal identifiable information along with events
   bool sendDefaultPii = false;
 
+  /// Wether deduplication is enabled. Shoud be set to true if
+  /// [SentryHttpClient] is used with
```
ueman: This doc comment needs to be finished :D

dart/test/event_processor/deduplication_event_processor_test.dart
```diff
@@ -0,0 +1,46 @@
+import 'package:sentry/sentry.dart';
+import 'package:sentry/src/event_processor/deduplication_event_processor.dart';
+import 'package:test/test.dart';
+
+import '../mocks.dart';
+
+void main() {
+  group('$DeduplicationEventProcessor', () {
+    test('deduplicates', () {
```
ueman: I'm pretty sure I'm missing some test cases.
Which events need deduplication and which don't?
As far as I can tell, Events without an exception aren't deduplicated and only events with the same exception are deduplicated.

ueman: Also maybe a test case for the max length of the map.

dart/lib/src/event_processor/deduplication_event_processor.dart
```diff
@@ -0,0 +1,56 @@
...
+/// ```dart
+/// var fooOne = Exception('foo');
+/// var fooTwo = Exception('foo');
+/// ```
+/// because (fooOne == fooTwo) equals false
+class DeduplicationEventProcessor extends EventProcessor {
+  DeduplicationEventProcessor(this._options);
+
+  // Map from exception to object
+  final Map<Object, Object?> exceptionMap = {};
```
marandaneto: should we be specific about key and type?

marandaneto: can probably be internal too

dart/lib/src/event_processor/deduplication_event_processor.dart
marandaneto: not sure if I understand this part, I don't see when the exception gets added to the map

dart/lib/src/event_processor/deduplication_event_processor.dart
ueman: The basetype of Exceptions in Dart is Object (the key) because you can throw anything. The value is never used. This is still just a naive conversion of the Java and C# implementations.

Maybe using a [LinkedHashSet](https://api.dart.dev/stable/2.13.3/dart-collection/LinkedHashSet-class.html) is a better option.

dart/lib/src/event_processor/deduplication_event_processor.dart
ueman: It's line 50

dart/lib/src/event_processor/deduplication_event_processor.dart
marandaneto: looks like we don't use the value at all, wondering if `Queue` or `Dequeue` would make more sense

marandaneto:
@bruno-garcia please review this one, which is a different approach due to the lack of WeakRef in Dart

dart/test/event_processor/deduplication_event_processor_test.dart
marandaneto: the main use case that should be deduplicated is like:

```
  private void test() {
    try {
      final int test = 1/0;
    } catch (Exception e) {
      // `e` is captured
      Sentry.captureException(e);
      throw e;
    }
  }
  
  private void myTest() throws Exception {
    // `test` throws and UncaughtExceptionHandlerIntegration gets called capturing `e` again
    // as `e` was already captured in the method `test`, the 2nd capture should be deduplicated
    test();
  }
```

dart/lib/src/event_processor/deduplication_event_processor.dart
ueman: Yep, I wondered about that in the Java and C# implementation, too.

marandaneto: do we have a thread-safe type? Dart is sort of single-threaded but when using multiple isolates, this could be racey

dart/lib/src/event_processor/deduplication_event_processor.dart
marandaneto: oh damn, need coffee, indeed.
If you see the `Scope#_breadcrumbs`, I've used a `Queue` which is a circular buffer, looks like this approach could be similar to that, we could even offer the fixed size `20` configurable too.

dart/lib/src/event_processor/deduplication_event_processor.dart
marandaneto: https://github.com/getsentry/sentry-java/issues/1533 lets see if there was a reason

ueman: I think it's just to make look ups fast. For maps and sets you can just do a 'contains' lookup. While you can do that for lists too, it's not O(1) but O(n) in the worst case. At least if I remember the big O notation correctly 😅

ueman: Nope, there's no such thread safe class. Errors should be propagated via `Ports` (see https://api.dart.dev/stable/2.13.3/dart-isolate/dart-isolate-library.html) back to the main isolate. See for example Flutters [`compute` methods source code](https://github.com/flutter/flutter/blob/d79295af24c3ed621c33713ecda14ad196fd9c31/packages/flutter/lib/src/foundation/_isolates_io.dart).

marandaneto: nice, so we just need to choose the right data structure for such a case.

marandaneto: yep, that's a good point, let's see if it's the case :D

dart/lib/src/event_processor/deduplication_event_processor.dart
marandaneto: lets make 10, deduplicating events its a matter of micro/milli seconds, so 20 is quite high already, lets do 10 for now

dart/lib/src/event_processor/deduplication_event_processor.dart
marandaneto: as we don't want to hold references, nor to leak objects, after discussing, what would make more sense is having a data structure that stores the `hashCode` as an Int, it's way cheaper than the reference of the object in memory.
let's make a circular buffer similar to Scope#breadcrumbs

dart/lib/src/event_processor/deduplication_event_processor.dart
marandaneto: see https://github.com/getsentry/sentry-dart/pull/498#discussion_r653314510

dart/lib/src/event_processor/deduplication_event_processor.dart
marandaneto: the reason was that Java and .NET don't have a List interface of weak references, but only a `WeakHashMap`, so when the GC is doing its job, it'd clean up the data automatically from the HashMap, also, because of lookup time.
to improve lookup time, we can use a data structure that is sorted or O(1), but instead of the object itself, we insert and check the hashCode of the object.
I believe Java could be also improved, see https://github.com/getsentry/sentry-java/issues/1533#issuecomment-863016739

dart/lib/src/event_processor/deduplication_event_processor.dart
```diff
@@ -0,0 +1,60 @@
...
+  FutureOr<SentryEvent?> _deduplicate(SentryEvent event) {
+    // cast to Object? in order to enable better type checking
+    final exception = event.throwable as Object?;
+    if (exception == null) {
+      // If no exception is given, just return the event
+      return event;
+    }
+    if (_exceptionToDeduplicate.contains(exception)) {
+      _options.logger(
+        SentryLevel.debug,
```
bruno-garcia: Might be worthy of `info` level (warnings is a stretch, right?)

dart/lib/src/event_processor/deduplication_event_processor.dart
```diff
@@ -0,0 +1,60 @@
+import 'dart:async';
+import 'dart:collection';
+import '../event_processor.dart';
+import '../protocol.dart';
+import '../sentry_options.dart';
+
+const eventsToKeepForDeduplication = 20;
```
bruno-garcia: I'd suggest something like 5

The use case is when there's a:

```
Sentry.captureException(e);
throw e;
```
And the caller also does `captureException` so these usually happen one after the other.

bruno-garcia (commented):
Differently then Java and .NET this approach doesn't rely on weakref (I understand it isn't an option here) but it means we'll be holding the reference of the N exceptions and hence 'leaking' all the object tree they hold for the duration of the process. Makes me wonder if this should be an opt-in.

ueman:
> Differently then Java and .NET this approach doesn't rely on weakref (I understand it isn't an option here) but it means we'll be holding the reference of the N exceptions and hence 'leaking' all the object tree they hold for the duration of the process. Makes me wonder if this should be an opt-in.

@marandaneto already suggested to store the hashes of the exception instead of the exception itself. 

This is only needed for the Http Client, which is also opt in. So there's nothing which speaks against making this opt in.

dart/lib/src/event_processor/deduplication_event_processor.dart
ueman: I'll make it configureable and set the default value to 5.

dart/lib/src/sentry_options.dart
```diff
@@ -189,6 +190,25 @@ class SentryOptions {
   /// Whether to send personal identifiable information along with events
   bool sendDefaultPii = false;
 
+  /// Wether [SentryEvent] deduplication is enabled.
```
```suggestion
  /// Whether [SentryEvent] deduplication is enabled.
```

dart/lib/src/sentry_options.dart
```diff
@@ -189,6 +190,25 @@ class SentryOptions {
...
+  int _exceptionsToKeepForDeduplication = 5;
+
+  /// Describes how many exceptions are kept to be checked for deduplication.
+  /// This should be a small positiv integer in order to keep deduplication
+  /// performant.
+  /// Is only in effect if [enableDeduplication] is set to true.
+  int get exceptionsToKeepForDeduplication => _exceptionsToKeepForDeduplication;
+
+  set exceptionsToKeepForDeduplication(int count) {
+    assert(count >= 0);
```
marandaneto: if 0 should be `enableDeduplication=false`, so I'd remove `=`?

dart/test/event_processor/deduplication_event_processor_test.dart
```diff
@@ -0,0 +1,114 @@
...
+  });
+}
+
+SentryEvent createEvent(var message) {
+  return SentryEvent(throwable: Exception(message));
+}
+
+class Fixture {
+  DeduplicationEventProcessor getSut(bool enabled,
+      [int? exceptionsToKeepForDeduplication]) {
```
marandaneto: I'd not make `exceptionsToKeepForDeduplication` nullable, but its a test, your call :)

dart/test/event_processor/deduplication_event_processor_test.dart
```diff
@@ -0,0 +1,114 @@
...
+import 'package:sentry/src/event_processor/deduplication_event_processor.dart';
+import 'package:test/test.dart';
+
+import '../mocks.dart';
+import '../mocks/mock_transport.dart';
+
+void main() {
+  group('$DeduplicationEventProcessor', () {
+    test('deduplicates if enabled', () {
+      final sut = Fixture().getSut(true);
```
marandaneto: if you need a new Fixture every time which is not the case, use a lateinit var and setup, otherwise a global Fixture that creates new Sut only

dart/test/event_processor/deduplication_event_processor_test.dart
```diff
@@ -0,0 +1,114 @@
...
+      var barEvent = createEvent('bar');
+      var fooBarEvent = createEvent('foo bar');
+
+      expect(sut.apply(fooEvent), isNotNull);
+      expect(sut.apply(barEvent), isNotNull);
+      expect(sut.apply(fooBarEvent), isNotNull);
+      expect(sut.apply(fooEvent), isNotNull);
+    });
+
+    test('$DeduplicationEventProcessor is added on init', () async {
```
marandaneto: goes to sentry_test.dart probably

dart/test/event_processor/deduplication_event_processor_test.dart
```diff
@@ -0,0 +1,114 @@
...
+        (options) {
+          options.dsn = fakeDsn;
+          final count = options.eventProcessors
+              .whereType<DeduplicationEventProcessor>()
+              .length;
+          expect(count, 1);
+        },
+      );
+
+      await Sentry.close();
```
marandaneto: use a `setUp` or `tearDown`

dart/test/event_processor/deduplication_event_processor_test.dart
```diff
@@ -0,0 +1,114 @@
...
+
+      await Sentry.init(
+        (options) {
+          options.dsn = fakeDsn;
+          options.transport = transport;
+          options.enableDeduplication = true;
+        },
+      );
+
+      // doesn't work with outerTestMethod as appRunner Callback
```
marandaneto: I dont get this comment

dart/lib/src/sentry_options.dart
ueman: Good catch

dart/test/event_processor/deduplication_event_processor_test.dart
ueman: Why though? This is the only test in this file which needs it.

dart/test/event_processor/deduplication_event_processor_test.dart
ueman: The test doesn't work if `outerTestMethod` is passed as `appRunner` callback

dart/test/event_processor/deduplication_event_processor_test.dart
marandaneto: not really, I see multiple tests using `Sentry.close()`
but I'd say this is less important when doing https://github.com/getsentry/sentry-dart/pull/498/files#r656023034

dart/lib/src/sentry_options.dart
```diff
@@ -189,6 +190,25 @@ class SentryOptions {
...
+  /// [SentryHttpClient] is used to capture failed requests.
+  bool enableDeduplication = true;
+
+  int _exceptionsToKeepForDeduplication = 5;
+
+  /// Describes how many exceptions are kept to be checked for deduplication.
+  /// This should be a small positiv integer in order to keep deduplication
+  /// performant.
+  /// Is only in effect if [enableDeduplication] is set to true.
+  int get exceptionsToKeepForDeduplication => _exceptionsToKeepForDeduplication;
```
marandaneto: tbh the name is a bit off for me, but I am not sure what to suggest either.
maybe @bruno-garcia or @philipphofmann ?

`maxDeduplicationQueueItems` looks better IMO

marandaneto (approved):
pre-approving, lets just figure out this before merging https://github.com/getsentry/sentry-dart/pull/498/files#r659610845

dart/lib/src/sentry_options.dart
ueman: Maybe `maxDeduplicationItems`. It's not really relevant for the user how the exceptions are stored, isn't it?

philipphofmann: `maxDeduplicationItems` also makes sense to me. Do we even need to make this option public?

ueman: No, not necessarily. The value of it is just a guess for what should work well. So that's why I made it configureable.

marandaneto:
sounds good, merge it :)

dart/lib/src/sentry_options.dart
marandaneto: other languages have a WeakRef list so it's not necessary having such an option.
here the approach is different, and depending on how you use the SDK, 5 items may be enough, or not.

PR #504, opened 2021-06-20
Ref: Use Duration class for time intervals in SentryOptions

### LINKED ISSUE #497: Use `Duration`s for time spans in SentryOptions
Ideally we would like to use to `Duration`s where we use time spans in SentryOptions.
More context: https://github.com/getsentry/sentry-dart/pull/495#discussion_r650677884

ueman:
## :scroll: Description
Use Duration instead of ints for time spans in SentryOptions.
Old options are still existing but marked as deprecated, so not breaking.

Closes #497

Existing tests

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 * Feat: Add Culture Context (#491)
 * Feat: Capture failed requests as event (#473)
 * Feat: `beforeSend` callback accepts async code (#494)
+* Feat: Change timespans to Durations in SentryOptions (#504)
 
 ## Breaking Changes:
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,6 @@
 * Feat: Add Culture Context (#491)
 * Feat: Capture failed requests as event (#473)
 * Feat: `beforeSend` callback accepts async code (#494)
-* Feat: Change timespans to Durations in SentryOptions (#504)
 
 ## Breaking Changes:
 
@@ -14,6 +13,7 @@
   * The method signature of `Transport` changed from `Future<SentryId> send(SentryEvent event)` to `Future<SentryId> send(SentryEnvelope envelope)`
 * Remove `Sentry.currentHub` (#490)
 * Ref: Rename `cacheDirSize` to `maxCacheItems` and add `maxCacheItems` for iOS (#495)
+* Feat: Change timespans to Durations in SentryOptions (#504)
 
 # 5.1.0
 
--- a/flutter/lib/src/default_integrations.dart
+++ b/flutter/lib/src/default_integrations.dart
@@ -209,7 +209,7 @@ class NativeSdkIntegration extends Integration<SentryFlutterOptions> {
         'attachStacktrace': options.attachStacktrace,
         'attachThreads': options.attachThreads,
         'autoSessionTrackingIntervalMillis':
-            options.autoSessionTrackingIntervalMillis,
+            options.autoSessionTrackingInterval.abs().inMilliseconds,
         'dist': options.dist,
         'integrations': options.sdk.integrations,
         'packages':
@@ -217,7 +217,8 @@ class NativeSdkIntegration extends Integration<SentryFlutterOptions> {
         'diagnosticLevel': options.diagnosticLevel.name,
         'maxBreadcrumbs': options.maxBreadcrumbs,
         'anrEnabled': options.anrEnabled,
-        'anrTimeoutIntervalMillis': options.anrTimeoutIntervalMillis,
+        'anrTimeoutIntervalMillis':
+            options.anrTimeoutInterval.abs().inMilliseconds,
         'enableAutoNativeBreadcrumbs': options.enableAutoNativeBreadcrumbs,
         'maxCacheItems': options.maxCacheItems,
         'sendDefaultPii': options.sendDefaultPii,
--- a/flutter/lib/src/default_integrations.dart
+++ b/flutter/lib/src/default_integrations.dart
@@ -209,7 +209,7 @@ class NativeSdkIntegration extends Integration<SentryFlutterOptions> {
         'attachStacktrace': options.attachStacktrace,
         'attachThreads': options.attachThreads,
         'autoSessionTrackingIntervalMillis':
-            options.autoSessionTrackingInterval.abs().inMilliseconds,
+            options.autoSessionTrackingInterval.inMilliseconds,
         'dist': options.dist,
         'integrations': options.sdk.integrations,
         'packages':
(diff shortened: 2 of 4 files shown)


flutter/lib/src/sentry_flutter_options.dart
```diff
@@ -25,37 +25,52 @@ class SentryFlutterOptions extends SentryOptions {
...
+        ? Duration(milliseconds: value)
+        : autoSessionTrackingInterval;
   }
 
+  /// The session tracking interval. This is the interval to end a
+  /// session if the App goes to the background.
+  /// Default is 30 seconds/30000 milliseconds.
+  /// See: [enableAutoSessionTracking]
+  /// Always uses the given duration as a positiv timespan.
+  Duration autoSessionTrackingInterval = Duration(milliseconds: 30000);
```
bruno-garcia: The default in all mobile SDKs is 5 seconds. Was it agreed to changed to 30?

ueman: It was like this before, I only changed the data type. If it should be 5 seconds, we should change it now because we're doing a major update anyway

ueman: Actually it's 30 seconds.
At least on iOS and Android (Java), see 
https://github.com/getsentry/sentry-cocoa/blob/2bfbbcdc11f84efd54498120e882d42e8bedd30e/Sources/Sentry/SentryOptions.m#L42

and 

https://github.com/getsentry/sentry-java/blob/a5f30b43b1dad2634ce020809f3b52e0d564a22a/sentry/src/main/java/io/sentry/SentryOptions.java#L205

marandaneto: its been always 30 seconds, never agreed to be 5

flutter/lib/src/sentry_flutter_options.dart
```diff
@@ -25,37 +25,52 @@ class SentryFlutterOptions extends SentryOptions {
...
+  @Deprecated('Use autoSessionTrackingInterval instead')
   int get autoSessionTrackingIntervalMillis =>
-      _autoSessionTrackingIntervalMillis;
+      autoSessionTrackingInterval.inMilliseconds;
 
+  @Deprecated('Use autoSessionTrackingInterval instead')
   set autoSessionTrackingIntervalMillis(int value) {
-    _autoSessionTrackingIntervalMillis =
-        value >= 0 ? value : _autoSessionTrackingIntervalMillis;
+    autoSessionTrackingInterval = value >= 0
```
marandaneto, line 37: 0 never makes sense, add an `assert`

flutter/lib/src/sentry_flutter_options.dart
```diff
@@ -25,37 +25,52 @@ class SentryFlutterOptions extends SentryOptions {
...
   /// Available only for Android.
   /// See: [anrEnabled]
-  int get anrTimeoutIntervalMillis => _anrTimeoutIntervalMillis;
+  @Deprecated('Use anrTimeoutInterval instead')
+  int get anrTimeoutIntervalMillis => anrTimeoutInterval.inMilliseconds;
 
+  @Deprecated('Use anrTimeoutInterval instead')
   set anrTimeoutIntervalMillis(int value) {
-    _anrTimeoutIntervalMillis = value >= 0 ? value : _anrTimeoutIntervalMillis;
+    anrTimeoutInterval =
```
marandaneto, line 72: similar to https://github.com/getsentry/sentry-dart/pull/504/files#r656028709

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 * Feat: Add Culture Context (#491)
 * Feat: Capture failed requests as event (#473)
 * Feat: `beforeSend` callback accepts async code (#494)
+* Feat: Change timespans to Durations in SentryOptions (#504)
```
marandaneto: lets add a `Breaking Changes` section as before

flutter/lib/src/sentry_flutter_options.dart
bruno-garcia: I'm getting old, yes it was always 30. I used to set to 5 to test while developing.

flutter/lib/src/default_integrations.dart
```diff
@@ -209,15 +209,16 @@ class NativeSdkIntegration extends Integration<SentryFlutterOptions> {
         'attachStacktrace': options.attachStacktrace,
         'attachThreads': options.attachThreads,
         'autoSessionTrackingIntervalMillis':
-            options.autoSessionTrackingIntervalMillis,
+            options.autoSessionTrackingInterval.abs().inMilliseconds,
```
marandaneto: why do we need to `abs`?

flutter/lib/src/sentry_flutter_options.dart
```diff
@@ -25,17 +25,33 @@ class SentryFlutterOptions extends SentryOptions {
...
+  /// The session tracking interval. This is the interval to end a
+  /// session if the App goes to the background.
+  /// Default is 30 seconds/30000 milliseconds.
+  /// See: [enableAutoSessionTracking]
+  /// Always uses the given duration as a positiv timespan.
+  Duration get autoSessionTrackingInterval => _autoSessionTrackingInterval;
+
+  set autoSessionTrackingInterval(Duration value) {
+    assert(!value.isNegative);
+    assert(value > Duration.zero);
```
marandaneto:
```suggestion
    assert(value > Duration.zero);
```

both asserts do the same thing right

flutter/lib/src/sentry_flutter_options.dart
```diff
@@ -45,15 +61,31 @@ class SentryFlutterOptions extends SentryOptions {
...
+  Duration _anrTimeoutInterval = Duration(milliseconds: 5000);
+
+  /// ANR Timeout internal. Default is 5000 milliseconds or 5 seconds.
+  /// Used by Androids AnrIntegration. Available only on Android.
+  /// See: [anrEnabled]
+  /// Always uses the given duration as a positiv timespan.
+  Duration get anrTimeoutInterval => _anrTimeoutInterval;
+
+  set anrTimeoutInterval(Duration value) {
+    assert(!value.isNegative);
```
marandaneto: same as https://github.com/getsentry/sentry-dart/pull/504/files#r659552218

flutter/lib/src/default_integrations.dart
ueman: Since there are asserts now, this is not needed anymore

flutter/lib/src/sentry_flutter_options.dart
ueman: yep, you are right

marandaneto (approved):
LGTM

Issue #194, opened 2020-11-27
Sentry.configureScope not synchronize in native


nEdAy:
sentry_flutter: ^4.0.0-alpha.2

```
Sentry.configureScope((Scope scope) async {
    scope.user = User(id: xxx);
});
```

The user information was set in the flutter and did not synchronize in native(Android).

Hope to add this feature, thanks

marandaneto:
@nEdAy thanks for this.

True, we still don't synchronize the Scope to the native layer, but we'll definitely do it.

bruno-garcia:
This is needed also in order to get the scope IP address to the session in native.

rayliverified:
Is this related to a race condition that happens between the native and Flutter SDKs? 
If await is used on SentryFlutter.init(), only the Flutter scope details are sent. 
If await is not called on SentryFlutter.init(), only the native details are sent.
I'm currently getting both the native and Flutter details to send by calling Sentry.configureScope before every error report call.

marandaneto:
@searchy2 not sure if I follow, this is a feature request, not a bug, at least related to this issue.
could you open an issue describing the problem? using the issue template would be very helpful too, thanks,

rayliverified:
Will do, thanks!

mafreud:
any update?

marandaneto:
no, you could always call the Native SDKs yourself via Method channels, we'll eventually do it, just not a high priority right now, PRs are also welcome, happy to guide & review.

ueman:
Do we want to implement it with ScopeObserver (like the [Java implementation](https://github.com/getsentry/sentry-java/blob/e1a7226bead691b2a824332f3d3a0930e68e33cc/sentry/src/main/java/io/sentry/IScopeObserver.java))?

marandaneto:
@ueman it's not a unified behavior across SDK but usually helpful when you have multiple SDKs packaged together.
it sounds like a good idea for me, you can experiment and see how it goes, obviously, we can adapt if necessary.
when Dart is used with Flutter, it should propagate the changes (possible to turn on/off via options).
when Dart is used alone, it should not, so that's why the Interface exists.

ueman:
A few thoughts on this:

### Option 1a:
[...]

PR #505, opened 2021-06-20
Feat: Attachment Support

### LINKED ISSUE #289: Add Attachment support
The native SDKs (sentry-android, sentry-cocoa) added support to Attachments.

The [spec is in the develop docs](https://develop.sentry.dev/sdk/features/#attachments).

Thumbs up if this is the feature you've been waiting for.

ueman:
Closes #289

## :pencil: Checklist

- [x] All tests passing
- [ ] No breaking changes


dart/lib/src/scope_extensions/scope_extensions.dart
```diff
@@ -0,0 +1,51 @@
+export '_io_scope_extensions.dart'
+    if (dart.library.html) '_web_scope_extensions.dart';
+
+import 'dart:typed_data';
+
+import '../../attachment.dart';
+import '../scope.dart';
+
+extension ScopeExtensions on Scope {
+  void addAttachmentBytes(
```
ueman: There are a couple data types in Dart which can be used as `bytes`:
- `Uint8List`
- `List<int>`
- `ByteData`

I think the names of the methods should reflect which data type they take.

For Flutter we could also add a `addAttachmentFromAsset` method, which reads attachments from an [`AssetBundle`](https://api.flutter.dev/flutter/services/AssetBundle-class.html)

dart/lib/src/scope_extensions/_web_scope_extensions.dart
```diff
@@ -0,0 +1,7 @@
+import '../scope.dart';
+
+extension IoScopeExtensions on Scope {
+  Future<void> addAttachementPath(String path) async {
```
ueman: I think we don't need this method.
Or it could add an attachment from an URL, I'm not sure which makes more sense.

dart/lib/src/scope_extensions/_io_scope_extensions.dart
```diff
@@ -0,0 +1,38 @@
+import 'dart:io';
+
+import '../../attachment.dart';
+import '../scope.dart';
+
+extension IoScopeExtensions on Scope {
+  Future<void> addAttachementPath(
```
ueman: Actually I think `Attachment` should have a couple of factory constructors.
- `Attachment.fromPath()`
- `Attachment.fromFile()`

and so on

ueman:
This does not yet include offline caching of attachments.

ueman:
I think there's currently no way to add attachments which need to be awaited. E.g. files and paths. The others are theoretically good, though I expect in practice it's not because they need to be awaited previously.

This is because the scope callback does not accept futures.

dart/lib/src/attachment.dart
```diff
@@ -0,0 +1,112 @@
...
+// https://develop.sentry.dev/sdk/envelopes/#attachment
+
+/// Arbitrary content which gets attached to an event.
+class Attachment {
+  Attachment({
+    required this.content,
+    required this.fileName,
+    AttachmentType? type,
+    String? mimeType,
+  })  : mimeType = mimeType ?? 'application/octet-stream',
```
ueman: This probably shouldn't default to `application/octet-stream`, but to `null` in order to be inferred in the backend.

marandaneto:
> This does not yet include offline caching of attachments.

what do you mean by that?
once you add an attachment to the scope and call eg captureException(x), the generated envelope will contain the attachment in it so it'd be cached in the disk and sent on restart. (Android/iOS/macOS at least)

marandaneto:
> I think there's currently no way to add attachments which need to be awaited. E.g. files and paths. The others are theoretically good, though I expect in practice it's not because they need to be awaited previously.
> 
> This is because the scope callback does not accept futures.

not sure if I follow, what do you mean? I didn't review the PR yet so I miss context

dart/lib/src/attachment.dart
```diff
@@ -0,0 +1,112 @@
+import 'dart:typed_data';
+export 'attachment_extensions/attachment_extensions.dart';
+
+// https://develop.sentry.dev/sdk/envelopes/#attachment
+
+/// Arbitrary content which gets attached to an event.
+class Attachment {
```
marandaneto: probably SentryAttachment to avoid import coalision

dart/lib/src/attachment.dart
```diff
@@ -0,0 +1,112 @@
...
+
+  /// Attachment file name.
+  final String fileName;
+
+  /// Attachment content type.
+  final String mimeType;
+}
+
+/// Attachment type.
+enum AttachmentType {
```
marandaneto: on Java is a String due to being easier to expand types without releasing a new version
maybe @bruno-garcia has opinions, I'm fine as it is as it covers all the cases right now

dart/lib/src/attachment.dart
```diff
@@ -0,0 +1,112 @@
...
+  /// During event ingestion, the last logs are extracted into event
+  /// breadcrumbs.
+  unrealLogs
+}
+
+extension AttachmentTypeX on AttachmentType {
+  String toSentryIdentifier() {
+    switch (this) {
+      case AttachmentType.attachment:
+        return 'event.attachment';
```
marandaneto: maybe we do this one as a `default` branch of the `switch`? because its the default type

dart/lib/src/attachment_extensions/_io_attachment_extensions.dart
```diff
@@ -0,0 +1,40 @@
+import 'dart:io';
+
+import '../attachment.dart';
+import '../scope.dart';
+
+extension IoScopeExtensions on Scope {
```
marandaneto: `Scope` only offers `addAttachment(attachment)`

the `Attachment` has enough factories for it, right?

dart/lib/src/attachment_extensions/attachment_extensions.dart
```diff
@@ -0,0 +1,2 @@
+export '_io_attachment_extensions.dart'
+    if (dart.library.html) '_web_scope_extensions.dart';
```
marandaneto: sometimes we check for `html` and sometimes for `if (dart.library.io)`, lets standardized and use the same check aways, just a nit though

dart/lib/src/scope.dart
```diff
@@ -81,6 +81,10 @@ class Scope {
 
   final SentryOptions _options;
 
+  final List<Attachment> _attachements = [];
+
+  List<Attachment> get attachements => UnmodifiableListView(_attachements);
```
marandaneto: What's the benefit over `List.unmodifiable`? I'd stick to `List.unmodifiable` for now and raise an issue to use `UnmodifiableListView`, also for Map instead of `Map.unmodifiable` if they bring benefits

dart/lib/src/sentry_client.dart
```diff
@@ -106,8 +106,13 @@ class SentryClient {
         return _sentryId;
       }
     }
-    final envelope = SentryEnvelope.fromEvent(preparedEvent, _options.sdk);
-    return await _options.transport.send(envelope);
+    final envelope = SentryEnvelope.fromEvent(
+      preparedEvent,
+      _options.sdk,
+      attachments: scope?.attachements,
```
marandaneto, line 114: @philipphofmann does attachment require to be in the same envelope as the event?

otherwise https://github.com/getsentry/sentry-java/issues/1508 will be a blocker

dart/lib/src/sentry_client.dart
```diff
@@ -199,7 +204,7 @@ class SentryClient {
   }
 
   /// Reports the [envelope] to Sentry.io.
-  Future<SentryId?> captureEnvelope(SentryEnvelope envelope) {
+  Future<SentryId> captureEnvelope(SentryEnvelope envelope) {
```
marandaneto: not every envelope has an id, so it should be nullable

dart/lib/src/sentry_envelope_item_header.dart
```diff
@@ -13,17 +18,18 @@ class SentryEnvelopeItemHeader {
...
-    }
-    if (fileName != null) {
-      json['filename'] = fileName;
-    }
-    json['type'] = type;
-    json['length'] = await length();
+    final json = <String, dynamic>{
+      if (contentType != null) 'content_type': contentType,
+      if (fileName != null) 'filename': fileName,
+      if (attachmentType != null) 'attachment_type': type,
```
marandaneto:
```suggestion
      if (attachmentType != null) 'attachment_type': attachmentType,
```

flutter/example/pubspec.yaml
```diff
@@ -13,6 +13,7 @@ dependencies:
     sdk: flutter
   sentry_flutter: any
   universal_platform: ^1.0.0-nullsafety
+  feedback: 2.0.0
```
marandaneto, line 16: I saw it coming :D

marandaneto:
nice PR @ueman its a Draft but I already have a great feeling about that, thanks.

marandaneto:
we will talk about https://github.com/getsentry/sentry-java/issues/1517 this week, depending of the output of the meeting, we can address this issue in this PR too

dart/lib/src/attachment.dart
ueman: A string probably makes more sense. We already use strings for envelope types, so it would be consistent.

dart/lib/src/attachment_extensions/_io_attachment_extensions.dart
ueman: This is tricky. On web it doesn't make sense to have the `fromPath` and `fromFile` methods (and it's actually not available). So I put them in extension methods to make them available depending on if its web or not. Though it doesn't work quite right yet.

flutter/example/pubspec.yaml
ueman: I'll remove it later. It's nice for testing attachments though.

ueman:
> > I think there's currently no way to add attachments which need to be awaited. E.g. files and paths. The others are theoretically good, though I expect in practice it's not because they need to be awaited previously.
> > This is because the scope callback does not accept futures.
> 
> not sure if I follow, what do you mean? I didn't review the PR yet so I miss context

You'll have to use async/await 
```dart
Sentry.configureScope((scope) async {
    scope.addAttachment(await Attachment.fromFile(file));
}):
```
in order to add an attachment from a file. Though the callback doesn't allow it IIRC.

You have to do this instead:
```dart
final attachment = await Attachment.fromFile(file);
Sentry.configureScope((scope) {
    scope.addAttachment(attachment);
}):
```

Same thing for `Sentry.captureEvent(event, withScope: (scope){...})`

I mean you can use it without, so it's not a blocker.

flutter/example/pubspec.yaml
marandaneto: we dont have to, its an example :)

dart/lib/src/sentry_client.dart
marandaneto: yep, ideally, it'd be within the same envelope, but Android opens up and break it down because it needs to enrich the event before sending.

dart/lib/src/attachment_extensions/_io_attachment_extensions.dart
marandaneto: the behaviour is correct, what I mean is, the extension should be in the Attachment class, not in the Scope

marandaneto:
spec https://develop.sentry.dev/sdk/features/#attachments
we dont have to consider `addToTransactions` right now

marandaneto:
> > > I think there's currently no way to add attachments which need to be awaited. E.g. files and paths. The others are theoretically good, though I expect in practice it's not because they need to be awaited previously.
> > > This is because the scope callback does not accept futures.
> > 
> > 
> > not sure if I follow, what do you mean? I didn't review the PR yet so I miss context
> 
> You'll have to use async/await
> 
> ```dart
> Sentry.configureScope((scope) async {
>     scope.addAttachment(await Attachment.fromFile(file));
> }):
> ```
> 
> in order to add an attachment from a file. Though the callback doesn't allow it IIRC.
> 
> You have to do this instead:
> 
> ```dart
> final attachment = await Attachment.fromFile(file);
> Sentry.configureScope((scope) {
>     scope.addAttachment(attachment);
> }):
> ```
> 
> Same thing for `Sentry.captureEvent(event, withScope: (scope){...})`
> 
> I mean you can use it without, so it's not a blocker.

See https://develop.sentry.dev/sdk/features/#attachments

> The SDK should read the file when an event gets captured and not when the user adds an attachment to the scope.

bruno-garcia:
Disclaimer: I didn't read/review the PR. Just wanted to throw in some challenges we learned since releasing the attachment API in some SDKs so far: 

* How can we attach a file when an event is captured through the uncaught exception handler?
* Is it possible to remove an attachment in a callback of sorts?
* How easy will it be once we add a feature such as `attachScreenshot` (could be an extension method) that will result in a screenshot taken at the time the event is being created (and not at the time the method is called). [See this workaround](https://github.com/getsentry/symbol-collector/blob/ead54e797709ea2bc4f96297aa59ac35ae5c10b5/src/SymbolCollector.Android.Library/Host.cs#L157-L158). 
* Does the API expose a way to add an attachment 'only once'?

Ideas that could solve these cases ^ would be welcomed

ueman:
> * How can we attach a file when an event is captured through the uncaught exception handler?

Flutter apps don't quit on uncaught exceptions in Dart, so this is a non issue. If an exception happens in the native part the Dart uncaught exception handler is never called at all I guess. We would need to sync the scope across SDKs for that.

> * Is it possible to remove an attachment in a callback of sorts?

No, though @marandaneto already raised this issue, too.

> * How easy will it be once we add a feature such as `attachScreenshot` (could be an extension method) that will result in a screenshot taken at the time the event is being created (and not at the time the method is called). [See this workaround](https://github.com/getsentry/symbol-collector/blob/ead54e797709ea2bc4f96297aa59ac35ae5c10b5/src/SymbolCollector.Android.Library/Host.cs#L157-L158). 

This is quite an interesting problem.
Taking a screenshot is easy (given we hooked into the UI), because Flutter does not have app quitting exceptions, so it should always be meaningful.
However there is no automatic way to hook into the Flutter UI to take the screenshot at all. Also screenshots don't work on Flutter Web with the HTML renderer 
and Platform Views are invisible on all platforms.

It probably makes sense to use the native part, if the time it takes to call the native platform has a low enough latency.

> * Does the API expose a way to add an attachment 'only once'?

I think you can use `Sentry.captureEvent(event, withScope: (scope){...})` for that if I understand the question correctly.

> Ideas that could solve these cases ^ would be welcomed

marandaneto:
@ueman 

> If an exception happens in the native part the Dart uncaught exception handler is never called at all I guess. We would need to sync the scope across SDKs for that.

that's right, could be worked on #194

> This is quite an interesting problem.
> Taking a screenshot is easy (given we hooked into the UI), because Flutter does not have app quitting exceptions, so it should always be meaningful.
> However there is no automatic way to hook into the Flutter UI to take the screenshot at all. Also screenshots don't work on Flutter Web with the HTML renderer
> and Platform Views are invisible on all platforms.

indeed, the workaround in the symbol-collector right now would not be possible as `configureScope` isn't `async` so you cant read the stream and await for it.
about Flutter Web, we can make this available only for Flutter io, its fine.
the idea is that if the options is enabled and they hooked up our integration manually when initing the SDK, we take a screenshot when `runZonedGuarded` is called, `runZonedGuarded` allows async code so I think its fine to take the screenshot and read the stream, add an attachment and capture it.
We just need to be sure that the zone in `runZonedGuarded` allows taking the screenshot or not (because of main/UI thread)

> I think you can use Sentry.captureEvent(event, withScope: (scope){...}) for that if I understand the question correctly.

it'd be a workaround but that's not quite the same thing.
the idea is that `addAttachment` gets an optional param to be sent only once and the event can be captured later at some point and the attachment delete itself from the attachment list.
with the `withScope` you pretty much need to capture the event right away which is not the case most of the times.

dart/lib/src/attachment.dart
```diff
@@ -0,0 +1,112 @@
...
+      fileName: fileName,
+      mimeType: mimeType,
+    );
+  }
+
+  /// Attachment type.
+  final AttachmentType type;
+
+  /// Attachment content.
+  final Uint8List content;
```
philipphofmann: `m`: In Java, we call this `bytes`, in Objective-C `data` and in [Python](https://github.com/getsentry/sentry-python/blob/master/sentry_sdk/attachments.py#L14) `bytes`. In C# though content. I'm not familiar on how to name such things in Dart, but maybe `content` is the way to go.

dart/lib/src/attachment.dart
philipphofmann: On Cocoa and Java, we don't have the `attachmentType` in the API. We just have the content type.

dart/lib/src/attachment.dart
```diff
@@ -0,0 +1,112 @@
...
+
+  /// Attachment file name.
+  final String fileName;
+
+  /// Attachment content type.
+  final String mimeType;
```
philipphofmann: `h`: We call this `contentType` in all the other APIs.

dart/lib/src/attachment.dart
```diff
@@ -0,0 +1,112 @@
...
+  Attachment({
+    required this.content,
+    required this.fileName,
+    AttachmentType? type,
+    String? mimeType,
+  })  : mimeType = mimeType ?? 'application/octet-stream',
+        type = type ?? AttachmentType.attachment;
+
+  /// Creates an [Attachment] from a [Uint8List]
+  factory Attachment.fromUint8List(
```
philipphofmann: `m`: Can we rename `fromUint8List`, `fromIntList` and `fromByteData` to `fromBytes` and use method overloading? Maybe it's also fine to have only one method with `Uint8List`.

philipphofmann (commented):
Didn't finish yet, will do another pass tomorrow.

dart/lib/src/attachment.dart
ueman: Dart has no method overloading, so that's not possible :(

There is no single data type which is used to represent bytes, so that's why I added multiple different constructors.
The Dart HTTP library we use, for example uses `List<int>`and `UInt8List`.

dart/lib/src/attachment.dart
ueman: As long as you're not dealing with emails :D

dart/lib/src/scope_extensions/_web_scope_extensions.dart
marandaneto: no URL afaik, only a path of the local storage, otherwise we'd have to download a file from a URL, we prefer that they do it themselves

dart/lib/src/attachment.dart
marandaneto: yep, thats why `SentryAttachment`

dart/lib/src/attachment.dart
marandaneto: we do, it was a bug and I fixed it a while ago
https://github.com/getsentry/sentry-java/blob/47b88ecea6dbd34af44ba3a442cfd03115758b79/sentry/src/main/java/io/sentry/Attachment.java#L16-L17

dart/lib/src/attachment.dart
marandaneto: yeah I'd go with factories for the different types too

marandaneto:
consider adding a method called `clearAttachments ` to the Scope, similar to https://github.com/getsentry/sentry-dotnet/issues/1088

philipphofmann:
> consider adding a method called `clearAttachments ` to the Scope, similar to [getsentry/sentry-dotnet#1088](https://github.com/getsentry/sentry-dotnet/issues/1088)

Just wanted to add a comment for this. Thanks @marandaneto 🚀

dart/lib/src/sentry_client.dart
philipphofmann: > @philipphofmann does attachment require to be in the same envelope as the event?

No, they don't, but should, see https://develop.sentry.dev/sdk/envelopes/#attachment

> For minidump and apple crash report attachments, the corresponding "event" Item must be sent within the same Envelope.
> Generic attachments can be ingested separately from their events. We recommend sending them in the same Envelope, which allows for more efficient rate limiting and filtering.

marandaneto: yep, I know about minidumps, so at least https://github.com/getsentry/sentry-java/issues/1508 isn't a blocker for now

dart/lib/src/attachment.dart
philipphofmann: `h`: Can we guess the MIME content type in dart based on the filename, as described here: https://develop.sentry.dev/sdk/features/#attachments

> and guess the content type of the attachment via the filename.

philipphofmann (commented):
Just found one more API issue.

dart/lib/src/attachment.dart
philipphofmann: We set the default in Java and Objective-C, but you are right; we could remove them and let the backend decide. I created an issue for Cocoa https://github.com/getsentry/sentry-cocoa/issues/1196

ueman: For reference we had a conversation [here](https://discord.com/channels/621778831602221064/654883298736340994/856597058093187152) on Discord about that.

dart/lib/src/attachment.dart
philipphofmann: As already pointed out, different in the other SDKs and also [Python](https://github.com/getsentry/sentry-python/blob/master/sentry_sdk/attachments.py). I disagree with naming it `SentryAttachment`, but it's up to you to decide. @bruno-garcia what's your opinion ⬆️?

ueman:
@philipphofmann Just to be sure:
h is for high, m for medium and l stands for low, right?

marandaneto:
> @philipphofmann Just to be sure:
> h is for high, m for medium and l stands for low, right?

yes, the LOGAF scale I've told you about a while ago.

dart/lib/src/scope.dart
ueman: I think `UnmodifiableListX` is probably the way to go: https://stackoverflow.com/a/61404767/12002462

dart/lib/src/sentry_client.dart
ueman: `Transport.send` can't return a `SentryId?`, so I assume that's wrong too?

dart/lib/src/scope.dart
marandaneto: nice, open an issue to discuss and replace with `UnmodifiableListX` (milestone 6.0.0), but let's use the `List.unmodifiable` for consistency here and not get blocked by it, the reason is while the `view` list makes sense, it will be different from the other SDKs that just give a snapshot of the list at the given moment.

dart/lib/src/sentry_client.dart
marandaneto: probably, `captureEnvelope` should always return `SentryId?`, but eg `captureMessage`, `captureEvent`, etcc should return `SentryId` because it defaults to `SentryId.empty()` if not set.
but `captureEnvelope` can capture eg an attachment that does not require a `sentryId` that's why its nullable and don't default to `SentryId.empty()`

marandaneto:
We should also expose the `maxAttachmentSize` options as Java/Apple

dart/lib/src/sentry_attachment_extensions/io_attachment_extensions.dart
```diff
@@ -0,0 +1,42 @@
+import 'dart:io';
+
+import '../sentry_attachment.dart';
+import '../scope.dart';
+
+extension IoAttachmentExtensions on Scope {
+  /// Creates an attachment from a given path.
+  /// Only available on `dart:io` platforms.
+  /// Not available on web.
+  static SentryAttachment fromPath(
```
ueman: This needs test for dropping an attachment in case it's not readable.

dart/lib/src/sentry_envelope.dart
```diff
@@ -27,9 +38,7 @@ class SentryEnvelope {
     final newLineData = utf8.encode('\n');
     for (final item in items) {
       yield newLineData;
```
ueman: Should this newline be added if `yield await item.envelopeItemStream();` is empty?

dart/lib/src/sentry_attachment/io_sentry_attachment.dart
```diff
@@ -0,0 +1,35 @@
+import 'dart:io';
+
+import 'sentry_attachment.dart';
+
+class IoSentryAttachment extends SentryAttachment {
```
ueman, line 5: Dart doesn't allow static methods nor factories in extensions. So we need to subclass here :(

ueman:
> We should also expose the `maxAttachmentSize` options as Java/Apple

What should happen if it's bigger?

For reference, the option is here:
https://github.com/getsentry/sentry-java/blob/79fab9fe98de0673176abdda9d09327a336c9150/sentry/src/main/java/io/sentry/SentryOptions.java#L261

marandaneto:
> > We should also expose the `maxAttachmentSize` options as Java/Apple
> 
> What should happen if it's bigger?
> 
> For reference, the option is here:
> https://github.com/getsentry/sentry-java/blob/79fab9fe98de0673176abdda9d09327a336c9150/sentry/src/main/java/io/sentry/SentryOptions.java#L261

its spec'ed here https://develop.sentry.dev/sdk/features/#max-attachment-size
you can also check the Java/iOS code

dart/lib/src/scope.dart
```diff
@@ -285,6 +298,10 @@ class Scope {
       }
     });
 
+    for (final attachment in _attachements) {
+      clone._attachements.add(attachment);
```
```suggestion
      clone.addAttachment.add(attachment);
```

dart/lib/src/sentry_attachment/sentry_attachment.dart
```diff
@@ -0,0 +1,97 @@
...
+  /// Attachment file name.
+  final String filename;
+
+  /// Attachment content type.
+  /// Inferred by Sentry if it's not given.
+  final String? contentType;
+}
+
+/// Attachment type.
+class AttachmentType {
```
marandaneto: should it be a different file?

dart/lib/src/sentry_attachment/sentry_attachment.dart
```diff
@@ -0,0 +1,97 @@
+import 'dart:async';
+import 'dart:typed_data';
+
+// https://develop.sentry.dev/sdk/features/#attachments
+// https://develop.sentry.dev/sdk/envelopes/#attachment
+
+typedef ContentLoader = FutureOr<Uint8List> Function();
```
marandaneto, line 7: why do we need this? we dont have on Java, and we still read it async.

dart/lib/src/sentry_envelope.dart
marandaneto: not sure if I follow, you mean if the envelope has nothing?

dart/test/sentry_attachment_test.dart
```diff
@@ -0,0 +1,85 @@
...
+      await expectLater(await attachment.bytes, [0, 0, 0, 0]);
+    });
+  });
+
+  group('$Scope $SentryAttachment tests', () {
+    test('Sending with attachments', () async {
+      final options = SentryOptions(dsn: fakeDsn);
+      final transport = MockTransport();
+      options.transport = transport;
+      final hub = Hub(options);
```
marandaneto: I see fixtures

dart/test/transport/http_transport_test.dart
```diff
@@ -92,7 +92,7 @@ void main() {
           SentryEnvelope.fromEvent(sentryEvent, fixture.options.sdk);
       final eventId = await sut.send(envelope);
 
-      expect(eventId, SentryId.empty());
```
marandaneto: SentryId.empty() is a `static final `, why is it failing if not `toString`?

dart/lib/src/sentry_attachment/sentry_attachment.dart
ueman: I'm not sure either. Maybe they should be fields of the attachment itself?

dart/lib/src/sentry_envelope.dart
ueman: An envelop has multiple items. If one (attachment) item is skipped because it's file no longer exists. Should the newline before it be added? I would guess no, though I'm not sure.

dart/lib/src/sentry_attachment/sentry_attachment.dart
marandaneto: on Java we have as a property in the Attachment class called `DEFAULT_ATTACHMENT_TYPE`, which defaults to `event.attachment`, we don't have the concept of `AttachmentType` unless someone else passes a custom string.
so maybe those constants should be part of the SentryAttachment itself

dart/lib/src/sentry_envelope.dart
marandaneto: if the attachment no longer exists (content), even its item header should not be part of the envelope, the envelope should still be formed correctly otherwise it will be malformed and rejected

dart/lib/sentry_io.dart
```diff
@@ -0,0 +1,2 @@
+export 'sentry.dart';
+export 'src/sentry_attachment_extensions/io_attachment_extensions.dart';
```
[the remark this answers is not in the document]
marandaneto: can we make export conditional as well? or only import?

dart/lib/src/sentry_attachment/sentry_attachment.dart
marandaneto: mm so if Web does not have this concept, and they need to pass eg an array of byte, there's nothing to be loaded async., its everything in memory already, so I'd say the concept of `ContentLoader` isn't needed at all? or maybe I am missing something still

ueman: Yeah, that's right. But then you can't add file loading on demand for io platforms because attachment doesn't expose a method to do so.

dart/lib/sentry_io.dart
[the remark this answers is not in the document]
ueman: Conditional exports/imports must have the same api, which in this case is not possible.

See https://dart.dev/guides/libraries/create-library-packages#conditionally-importing-and-exporting-library-files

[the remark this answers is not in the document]
ueman: So exposing a new import is the only way :(

Alternatively we can remove file support for attachments, but I don't think that's the way to go.

dart/test/transport/http_transport_test.dart
ueman: This got me wondering and actually `SentryId` is buggy.
The following tests do not pass:
```dart
test('id roundtrip', () {
    final id = SentryId.newId();
    expect(id, SentryId.fromId(id.toString()));
});

test('empty id equals from empty id', () {
    expect(
      SentryId.empty(),
      SentryId.fromId('00000000000000000000000000000000'),
    );
});
```
This is because it doesn't override the equality operator/hashCode.

The line you commented on failed because the second case in my example above was hit.
By calling `toString()` it passed because the string equality check then returns that they're equal.

marandaneto: mm, so it's fine that it's not the same object but with the same id, but the `SentryId.empty()` is a static final var so it should be always the same, if compared with another `SentryId.empty()`

marandaneto: I don't think we need to override the equality operators either, it'd be useful for tests, but for the SDK itself, it won't make any difference.

ueman: yep, but it was converted to JSON and back to a SentryId, so that is why.

marandaneto:
@ueman is it ready for final code review? @bruno-garcia or @philipphofmann will do it, thanks

ueman:
> @ueman is it ready for final code review? @bruno-garcia or @philipphofmann will do it, thanks

Yep, it's ready now.

dart/lib/src/attachment.dart
bruno-garcia: Dart is a bit sensitive to clashing types (though latest version had some improvements to deal with these cases).
No strong opinions here. Given the history of name clashing issues in this repo I suppose SentryAttachment is probably safer

dart/lib/src/sentry_attachment.dart
```diff
@@ -0,0 +1,102 @@
...
+
+  /// Attachment type.
+  /// Should be one of types given in [AttachmentType].
+  final String attachmentType;
+
+  /// Attachment content.
+  /// Is loaded while sending this attachment.
+  FutureOr<Uint8List> get bytes => _loader();
+
+  final ContentLoader _loader;
```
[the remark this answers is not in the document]
marandaneto: lets keep as it is for now and work on it later

marandaneto (approved):
LGTM

Issue #441, opened 2021-04-28
Support onCrashedLastRun


jguz88:
I'm using `sentry_flutter 5.0.0` and the new functionality is proposed for Flutter. I would like to enrich crash report by adding custom data. Unfortunately, existing `beforeSend` callback is never called if a Flutter application launched again after a crash. 

```
SentryEvent beforeSend(SentryEvent event, {dynamic hint}) {
  if (event.level == SentryLevel.fatal) {
    event = event.copyWith(environment: '', tags: ..., extra: ...);
  }
  return event;
}

await SentryFlutter.init((options) {
    options.dsn = ...;
    options.beforeSend = beforeSend;
});
```

### Proposal 1

Invoke the `beforeSend` callback inside the SDK in case of opening a Flutter application just after a crash.

### Proposal 2

Create separate callback which is called if the Flutter application crashed last time.
For instance, in a similar way as it's now for native iOS SDK

```
SentryEvent onCrashedLastRun(SentryEvent event) {
   return event.copyWith(environment: '', tags: ..., extra: ...);
}

await SentryFlutter.init((options) {
    options.dsn = ...;
    options.onCrashedLastRun = onCrashedLastRun;
});
```

Thank you for your brilliant work so far!

marandaneto:
@jguz88 indeed, thanks for raising this.

we'd go probably with `onCrashedLastRun` as its the way how we do it on other SDKs, but Flutter rarely hard crashes and when it does, it does via signal handlers which are not handled by the Flutter SDK but the underlying SDKs, either sentry-cocoa or sentry-native

bruno-garcia:
As a work around, you can write a `beforeSend` in Java. There's an issue tracking a better way to allow you initialize the Android SDK programatically: #265 
Until this one is solved it's a bit complicated to "go manual".

jguz88:
Yes, I did this in that way and provide a ```beforeSend``` in Java as soon as an application started. But this configuration will be overwritten in the ```initNativeSdk``` method  which also invokes ```SentryAndroid.init```. I would like to preserve your configuration and add my changes upon that. I also realize it's not so simple.

larssn:
Would this allow us to enrich native crashes?

It's a challenge currently to figure out, say, who was logged into Firebase if there's a hard crash.

bruno-garcia:
[...]

PR #506, opened 2021-06-20
Feat: Support for UserFeedback

### LINKED ISSUE #215: Support for user feedback
I would like to request support for [Sentry's User Feedback](https://docs.sentry.io/platforms/javascript/enriching-events/user-feedback/).

I see at least two possibilities on how to do this:

## Proposal 1:
By providing a more elaborate `SentryFlutter.captureException` like for example:
```dart
SentryFlutter.captureException(
  exception: error,
  stackTrace: stackTrace,
  context: context, // this is a BuildContext
);
```
With a `BuildContext` one can show a dialog (see [showDialog](https://api.flutter.dev/flutter/material/showDialog.html)) which can be used to gather the user's feedback.

## Proposal 2:
Another option would be by overriding [ErrorWidget.builder](https://api.flutter.dev/flutter/widgets/ErrorWidget/builder.html) (aka Flutter's infamous [Red Screen of Death](https://stackoverflow.com/questions/53903928/disable-flutters-red-screen-of-death)).
This way one could could also capture and report the errors which get passed into `ErrorWidget.builder`. These are errors
which are happening during the creation of the UI.

An example (more or less pseudo code):
```dart
// We assume SentryOptions has a ErrorWidgetBuilder named errorWidgetBuilder which can be defined by the user.
// By providing the same callback it is easy to migrate to this.

class UserFeedback {
  // ...
}

class UserFeedbackErrorWidget extends StatefulWidget {
  const YellowBird({ Key key, this.details, this.options }) : super(key: key);

  final FlutterErrorDetails details;
  final SentryOptions options;

  @override
  _UserFeedbackErrorWidgetState createState() => _UserFeedbackErrorWidgetState();
}

class _UserFeedbackErrorWidgetState extends State<UserFeedbackErrorWidget> {

  @override
  void initState(){
    super.initState();
    // one can't open a dialog in initState
    // see https://stackoverflow.com/questions/51766977/flutter-showdialogcontext-in-initstate-method
   Future.delayed(Duration.zero, askForUserFeedback)
  }

  @override
  Widget build(BuildContext context) {
    // This way a user defined widget is shown inline in the ui.
    // It's nice for the user, because it is not very intrusive.
    return options.errorWidgetBuilder(widget.details);
  }

  Future<void> askForUserFeedback() {
    final error = widget.details;
    final sentryId = await Sentry.captureException(error);

    // showDialog returns what is given to Navigator.pop(...)
    final feedback = await showDialog<UserFeedback>(context, (innerContext){ 
      // show dialog, where the user can input additional information and return it
     return UserFeedbackDialog(
       onSubmit: (feedback) => Navigator.pop(innerContext, feedback)
     );
    });
    await Sentry.sendFeedback(sentryId, feedback);
  }
}
```
Now we can set `ErrorWidget.builder` inside an `Integration` with 
```dart
ErrorWidget.builder = (FlutterErrorDetails details) => UserFeedbackErrorWidget(details: details, options: sentryOptions);
```

I guess providing both options would be optimal.

If there are any questions, I'm happy to help.

ueman:
## :scroll: Description
Adds support for user feedback.
Builds on https://github.com/getsentry/sentry-dart/pull/505 and thus should be merged after it.

Closes #215

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes


ueman:
Also includes a UI.
It works for light and dark mode. The Sentry logo switches colors too. The highlight and button colors are taken from the users app theme. So if the primary color is orange, the button would be orange.

Bildschirmfoto 2021-06-20 um 20 10 56
Bildschirmfoto 2021-06-20 um 20 10 31

Bildschirmfoto 2021-06-21 um 11 39 43

Bildschirmfoto 2021-06-21 um 11 39 32

flutter/lib/src/ui/user_feedback_dialog.dart
```diff
@@ -0,0 +1,458 @@
...
+      child: CustomPaint(
+        size: Size(222, 66),
+        painter: _SentryLogoCustomPainter(color),
+      ),
+    );
+  }
+}
+
+// Created with https://fluttershapemaker.com/
+class _SentryLogoCustomPainter extends CustomPainter {
```
ueman: This way we don't have to include it as an bitmap. Flutter doesn't has support for vector images by default.

ueman: The image was taken from https://sentry.io/branding/

dart/lib/src/hub.dart
```diff
@@ -176,6 +178,13 @@ class Hub {
     return sentryId;
   }
 
+  Future<void> captureUserFeedback(UserFeedback userFeedback) async {
+    final item = _peek();
```
marandaneto: I know its a draft but just to not forget, requires checks like `captureEvent` does, if SDK is enabled, etc

dart/lib/src/hub.dart
```diff
@@ -176,6 +178,13 @@ class Hub {
...
+    await item.client.captureEnvelope(
```
marandaneto: `client` could expose `captureUserFeedback` that does the `SentryEnvelope.fromUserFeedback` internally

dart/lib/src/sentry_envelope_item.dart
```diff
@@ -20,6 +21,24 @@ class SentryEnvelopeItem {
...
 
+  factory SentryEnvelopeItem.fromUserFeedback(UserFeedback feedback) {
+    final cachedItem = _CachedItem(() async {
+      final jsonEncoded = jsonEncode(feedback.toJson());
+      return utf8.encode(jsonEncoded);
+    });
+
+    final getLength = () async {
+      return (await cachedItem.getData()).length;
+    };
```
marandaneto, line 39: part of this block is duplicated across methods, maybe we extract to a private method

dart/lib/src/user_feedback.dart
```diff
@@ -0,0 +1,54 @@
+import 'protocol.dart';
+
+class UserFeedback {
```
marandaneto: SentryUserFeedback so we avoid import coalision

flutter/example/lib/main.dart
```diff
@@ -338,3 +397,22 @@ Future<void> makeWebRequest(BuildContext context) async {
     },
   );
 }
+
+class ThemeProvider extends ChangeNotifier {
```
marandaneto, line 439: we probably dont need this as its just part of the example for user feedback, keeping it simple make it easier for people to understand the important bits

marandaneto:
> What tests do we want to have for the UI, if the UI is wanted at all? Do we want to have widget and golden image tests?

we actually don't need a UI at all, but its nice that you have added to the sample, so we can keep it.

spec is here https://develop.sentry.dev/sdk/features/#user-feedback
we can tackle https://github.com/getsentry/sentry-dart/issues/441 later as its not a blocker and it'd require changes on native SDKs too probably.

marandaneto:
this PR branches from `feat/attachments`, I'd not get blocked by this though because user feedback and attachments are different features that could be used together, but don't depend on each other.

overall nice PR, unit tests and so on are missing but I like where this is going :) thanks @ueman

dart/lib/src/hub.dart
ueman: And additionally it probably needs a check that the `userFeedback.sentryId` is not equal to `SentryId.empty()`.

dart/lib/src/user_feedback.dart
philipphofmann: We also called it `UserFeedback` on [.NET](https://github.com/getsentry/sentry-dotnet/blob/main/src/Sentry/UserFeedback.cs) and [Java](https://github.com/getsentry/sentry-java/blob/main/sentry/src/main/java/io/sentry/UserFeedback.java). On Cocoa, it's `SentryUserFeedback` for Objective-C and `UserFeedback` for Swift. We did this\ because `UserFeedback` is a not so common class. So I would say `UserFeedback` is fine here.

philipphofmann (approved):
API looks good. Didn't check the dart details, though.

dart/lib/src/user_feedback.dart
marandaneto: unless you use the library that @ueman developed himself, https://github.com/ueman/feedback/blob/master/lib/src/user_feedback.dart hehe
so `SentryUserFeedback` is the way to go, Dart is like objC for imports, so better safer than sorry

ueman:
The UI is in the example app for now. I do intend to clean it up and write tests for it in a follow-up PR.

marandaneto:
@ueman is it ready for final code review? @bruno-garcia or @philipphofmann will do it, thanks

ueman:
> @ueman is it ready for final code review? @bruno-garcia or @philipphofmann will do it, thanks

Yep, it's ready now.

dart/lib/src/hub.dart
marandaneto: still pending

ueman: There's a check for a disabled SDK and the UserFeedback ctor disallows the use of an empty ID. Are there more conditions which I need to consider?

dart/lib/src/sentry_envelope_item.dart
ueman: I think we should do some refactoring in a different PR.
Currently, we try to stream the envelope's payload, but that's not really possible because we need to know the size of the payload for the envelope header. We should instead stream the whole envelope. That way we could get rid of this kinda weird code alltogether.

bruno-garcia (approved):
LGTM

dart/lib/src/hub.dart
marandaneto: lets add a check anyway because at runtime we don't know if its empty or not, so we don't even call the API if its an empty eventId

marandaneto (approved):
LGTM

Issue #492, opened 2021-06-10
Don't enable `print()` breadcrumbs by default on Flutter


ueman:
_Platform:_
- [ ]  Dart
- [x]  Flutter Android or iOS
- [x]  Flutter Web

`Sentry Flutter Version 5.1.0`

By default Flutter calls [`dumpErrorToConsole`](https://api.flutter.dev/flutter/foundation/FlutterError/dumpErrorToConsole.html) for every exceptions which it catches. If you go down the rabbit hole, you'll end up [here](https://github.com/flutter/flutter/blob/02c026b03c/packages/flutter/lib/src/foundation/print.dart#L29). As you can see it calls `print()`. 

This results in pretty bad breadcrumbs because each stacktrace frame is written as a single breadcrumb.
![Bildschirmfoto 2021-06-10 um 09 18 01](https://user-images.githubusercontent.com/1270149/121481766-c3c65b00-c9cc-11eb-8601-72c161bfde85.png)

This seems to be an issue on debug and release builds.

ueman:
It should be disabled by default (or just on Flutter).
I haven't come around to do this yet, though. The documentation should probably mention this behavior, too, in case a user manually enables this.

I don't think we need further investigation as the reason for this behavior is already explained in the issue. Or do you mean something else?

marandaneto:
@ueman I meant, we should fix this, sorry :D

PR #515, opened 2021-07-02
Ref: Disable print breadcrumbs on Flutter

ueman:
## :scroll: Description
This disables print() breadcrumbs on Flutter

#492

New test 

## :pencil: Checklist

- [x] All tests passing
- [ ] No breaking changes

### DIFF
--- a/flutter/lib/src/sentry_flutter_options.dart
+++ b/flutter/lib/src/sentry_flutter_options.dart
@@ -9,6 +9,7 @@ class SentryFlutterOptions extends SentryOptions {
   SentryFlutterOptions({String? dsn, PlatformChecker? checker})
       : super(dsn: dsn, checker: checker) {
     enableBreadcrumbTrackingForCurrentPlatform();
+    enablePrintBreadcrumbs = false;
   }
 
   /// Enable or disable the Auto session tracking on the Native SDKs (Android/iOS)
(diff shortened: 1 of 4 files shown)


flutter/lib/src/sentry_flutter_options.dart
```diff
@@ -9,6 +9,7 @@ class SentryFlutterOptions extends SentryOptions {
   SentryFlutterOptions({String? dsn, PlatformChecker? checker})
       : super(dsn: dsn, checker: checker) {
     enableBreadcrumbTrackingForCurrentPlatform();
+    enablePrintBreadcrumbs = false;
```
marandaneto, line 12: is that the only option? I'd prefer rather a workaround if possible :D

ueman: We could intercept https://github.com/flutter/flutter/blob/f4abaa0735eba4dfd8f33f73363911d63931fe03/packages/flutter/lib/src/foundation/assertions.dart#L919

and disable recording when it's called.
Kinda like what we do it for FlutterError.onError.
Does that sound better?

marandaneto: sounds better to me, rather than disabling the feature by default.
we can check if its the `FlutterExceptionHandler` type and do not log it if its a release build, because, on debug level, printing to console is rather useful

ueman:
Closed in favor of #618

Issue #511, opened 2021-06-25
replace `dynamic` to `Object?` whenever possible


marandaneto:
we never know if a dynamic is a nullable field, also dynamic disables type checking.
should we replace it for better type checking?

@ueman @bruno-garcia

ueman:
I wonder how that works with JSON (de) serialization, since that historically always has been Map<String,dynamic>. Though I suspect that it works. Other than that I see nothing which should stop us from doing that.

marandaneto:
yeah I think at runtime would not change anything, and I also think that we don't call any method directly on `dynamic` by skipping type check, we always `cast` before using the dynamic objects anyway.

marandaneto:
this would be ideally done with https://github.com/getsentry/sentry-dart/issues/510 so we do a single breaking change in the hints signature, as they are all `dynamic`

ueman:
Json decoding returns a `Map<String, dynamic>` so `x.fromJson(Map<String, dynamic>)` has to stay that way. A couple of things like extras will have to stay because of that, too. However, there's still a lot of code to improve left.

marandaneto:
`dynamic decode` returns `dynamic` but can be always cast right, why do we need to keep `fromJson` as `dynamic`? I don't see why it'd be a problem.

marandaneto:
In favor of https://github.com/getsentry/sentry-dart/issues/510

PR #516, opened 2021-07-05
Ref: Make every use of dynamic explicit

ueman:
## :scroll: Description
This enables Dart's strong mode and fixes the resulting errors/warnings.
Every use of `dynamic` must be explicit now.
Cast must be explicit, too.

In a follow-up PR, we can now search and replace for every use of `dynamic` and replace it.

Part of #511

The current test suite covers these changes.

## :pencil: Checklist

- [x] All tests passing
- [ ] No breaking changes

## :crystal_ball: Next steps
Replace every `dynamic` with `Object` or `Object?`


dart/lib/src/protocol/mechanism.dart
```diff
@@ -52,8 +54,8 @@ class Mechanism {
     this.synthetic,
     Map<String, dynamic>? meta,
     Map<String, dynamic>? data,
-  })  : _meta = meta != null ? Map.from(meta) : null,
-        _data = data != null ? Map.from(data) : null;
+  })  : _meta = meta != null ? Map<String, dynamic>.from(meta) : null,
+        _data = data != null ? Map<String, dynamic>.from(data) : null;
```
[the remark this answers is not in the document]
marandaneto: `_meta` needs to be mutable, but the given param may be immutable, so we have to create a new one as we don't know if its mutable or not

dart/lib/src/protocol/contexts.dart
```diff
@@ -28,41 +28,49 @@ class Contexts extends MapView<String, dynamic> {
...
 
   /// Deserializes [Contexts] from JSON [Map].
-  factory Contexts.fromJson(Map<String, dynamic> data) {
+  // ignore: strict_raw_type
+  factory Contexts.fromJson(Map data) {
+    // This class should be deserializable from Map<String, dynamic> and Map<Object?, Object?>,
+    // because it comes from json.decode which is a Map<String, dynamic> and from
+    // methodchannels which is a Map<Object?, Object?>.
+    // Map<String, dynamic> and Map<Object?, Object?> only have
+    // Map<dynamic, dynamic> as common type constraint
```
marandaneto, line 37: the leys are always a String though, can we force the type here somehow?

ueman: I've tried various things, but the only way which worked was iterating through each map and casting each individual entry which seems like overkill.

Issue #513, opened 2021-06-30
Flutter Frame Tracking Integration


ueman:
Flutter allows the developer to read how long it took for a frame to draw. See [FrameTiming](https://api.flutter.dev/flutter/dart-ui/FrameTiming-class.html).

It would be nice if Sentry attaches a listener and sends events or add breadcrumbs if a frame took to long to draw.

Since there's no performance tracing yet, I can imagine that it's not that useful since you don't really know what caused dropped frames. Though the NavigatorObserver should help with that.

marandaneto:
I don't think this as events or breadcrumbs are useful as they won't know where/how to fix it, this would fit either our measurements or metrics product which is being worked on as we speak.
See eg https://github.com/getsentry/sentry-cocoa/pull/1123
it requires the performance API so its blocked by it

ueman:
In one of the recent Flutter versions (either 2.5 or 2.8) your now able to read the current [frame number](https://api.flutter.dev/flutter/dart-ui/FrameData/frameNumber.html) and `FrameTiming` also got a [frame number](https://api.flutter.dev/flutter/dart-ui/FrameTiming/frameNumber.html) property. 

This should make it possible to get at least somewhat useful metrics.

ueman:
There's also https://api.flutter.dev/flutter/scheduler/SchedulerBinding/addPersistentFrameCallback.html which gets called every frame.

ueman:
After raising the min versions of Dart and Flutter, the detection of slow and frozen frames is possible in pure Dart code without having to rely on native code. Thus that feature could be implemented for Linux and Windows. (Maybe on the canvaskit renderer on web too?) I believe the current implementation needs some changes to enable that though.

marandaneto:
> After raising the min versions of Dart and Flutter, the detection of slow and frozen frames is possible in pure Dart code without having to rely on native code. Thus that feature could be implemented for Linux and Windows. (Maybe on the canvaskit renderer on web too?) I believe the current implementation needs some changes to enable that though.

Yes, I believe that would be the goal, getting the values directly from `Skia` instead of the Native SDKs.

marandaneto:
PR is closed due to https://github.com/getsentry/sentry-dart/pull/1271#issuecomment-1562970140

marandaneto:
Closed because of https://github.com/getsentry/sentry-dart/pull/1271#issuecomment-1562970140
We can reevaluate this in case we learn more or find out that this is actually possible within our current data model.

PR #519, opened 2021-07-05
Proof of Concept: Frame Timing Integration

ueman:
## :scroll: Description
Proof of concept for tracking frame durations.

It's not really useful without being able to better determine what's causing it.

#513

Not yet

## :pencil: Checklist

- [ ] All tests passing
- [ ] No breaking changes

### DIFF
--- a/flutter/example/lib/main.dart
+++ b/flutter/example/lib/main.dart
@@ -1,4 +1,5 @@
 import 'dart:async';
+import 'dart:io';
 
 import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
@@ -58,6 +59,10 @@ class MainScaffold extends StatelessWidget {
             RaisedButton(
               child: const Text('Open another Scaffold'),
               onPressed: () => SecondaryScaffold.openSecondaryScaffold(context),
+            ),
+            RaisedButton(
+              child: const Text('Open Scaffold with slow drawing'),
+              onPressed: () => SlowScaffold.open(context),
             ),
             RaisedButton(
               child: const Text('Dart: try catch'),
@@ -326,4 +331,45 @@ Future<void> makeWebRequest(BuildContext context) async {
       );
     },
   );
+}
+
+class SlowScaffold extends StatelessWidget {
+  static Future<void> open(BuildContext context) {
+    return Navigator.push(
+      context,
+      MaterialPageRoute<void>(
+        settings:
+            const RouteSettings(name: 'SlowScaffold', arguments: 'sloooowwwww'),
+        builder: (context) {
+          return SlowScaffold();
+        },
+      ),
+    );
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    sleep(Duration(milliseconds: 20));
+    return Scaffold(
+      appBar: AppBar(
+        title: const Text('SecondaryScaffold'),
+      ),
+      body: Center(
+        child: Column(
+          children: [
+            const Text(
+              'You have added a navigation event '
+              'to the crash reports breadcrumbs.',
+            ),
+            MaterialButton(
+              child: const Text('Go back'),
+              onPressed: () {
+                Navigator.pop(context);
+              },
+            ),
+          ],
+        ),
+      ),
+    );
+  }
 }
--- a/flutter/lib/src/integrations/frame_timing_integration.dart
+++ b/flutter/lib/src/integrations/frame_timing_integration.dart
@@ -6,10 +6,27 @@ import 'package:sentry/sentry.dart';
 
 import '../sentry_flutter_options.dart';
 
+/// Records the time which a frame takes to draw, if it's above
+/// a certain threshold, i.e. [FrameTimingIntegration.badFrameThreshold].
+///
+/// Should not be added in debug mode because the performance of the debug mode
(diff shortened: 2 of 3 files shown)


flutter/lib/src/integrations/frame_timing_integration.dart
```diff
@@ -0,0 +1,117 @@
...
+/// See [SchedulerBinding.addTimingsCallback](https://api.flutter.dev/flutter/scheduler/SchedulerBinding/addTimingsCallback.html)
+/// to learn more about the performance impact of using this.
+///
+/// Adding a `timingsCallback` has a real significant performance impact as
+/// noted above. Thus this integration should only be added if it's enabled.
+/// The enabled check should not happen inside the `timingsCallback`.
+class FrameTimingIntegration extends Integration<SentryFlutterOptions> {
+  FrameTimingIntegration({
+    required this.reporter,
+    this.badFrameThreshold = const Duration(milliseconds: 16),
```
ueman, line 25: I haven't yet found a way to determine the device current refresh rate.

flutter/lib/src/integrations/frame_timing_integration.dart
```diff
@@ -0,0 +1,117 @@
...
+      level: SentryLevel.warning,
+    ));
+  }
+
+  String _message(
+    int badFrameCount,
+    Duration worstFrameDuration,
+    Duration totalDuration,
+  ) {
+    return '$badFrameCount frames exceeded ${_formatMS(badFrameThreshold)} '
```
ueman, line 99: Average frame time would be nice to know, too.

ueman:
Now that we have performance tracing, we can add frame metrics just like [Android](https://github.com/getsentry/sentry-java/blob/1a409209b77d5e28915c2f3df49e6941bed3d79c/sentry-android-core/src/main/java/io/sentry/android/core/ActivityFramesTracker.java#L96-L98) or [iOS](https://github.com/getsentry/sentry-cocoa/blob/94ec24f58b12c5c7dfae7d9f6430c06655026201/Sources/Sentry/SentryFramesTracker.m)

flutter/example/lib/main.dart
```diff
@@ -1,4 +1,5 @@
 import 'dart:async';
+import 'dart:io';
```
marandaneto, line 2: that requires a custom import since it should run on Web too

ueman:
@marandaneto Is there some documentation on how these values should be added to transactions? Or do I have to go by the Android/iOS source code?

marandaneto:
> @marandaneto Is there some documentation on how these values should be added to transactions? Or do I have to go by the Android/iOS source code?

there are no official docs yet since mobile vitals is a preview, but if you're willing to work on that, I can guide you.

ueman:
@marandaneto that would be nice

ueman:
If I understand the source code of the Android SDK correctly, I need to add the frame metrics in the following format to a transaction?
```json
{
...
"measurements":{
      "frames_frozen":{
         "value":0.0
      },
      "frames_slow":{
         "value":2.0
      },
      "frames_total":{
         "value":2.0
      }
   }
},
...
```
I'm assuming the data collection should start (and end) with the start (and end) of a transaction. Is that correct?

marandaneto:
> If I understand the source code of the Android SDK correctly, I need to add the frame metrics in the following format to a transaction?
> 
> ```json
> {
> ...
> "measurements":{
>       "frames_frozen":{
>          "value":0.0
>       },
>       "frames_slow":{
>          "value":2.0
>       },
>       "frames_total":{
>          "value":2.0
>       }
>    }
> },
> ...
> ```
> 
> I'm assuming the data collection should start (and end) with the start (and end) of a transaction. Is that correct?

that's correct, good finding, sorry, we didn't have the time to document it yet.

ueman:
Are there any hooks for a transaction start and end which I can use to turn on and off frame time recording? The docs state that it has a small performance impact, so I really don't want to have it enabled the whole time.

If there's no hooks, I can think of three ways to enable us to go further with this:
- Add some kind of hooks for transactions. Maybe something like integrations/event processors but just internal?
- A transaction factory which an SDK can register. The Flutter SDK could then subclass the transaction to add frame time recordings
- A method on `SentryFlutter` to start a subclassed transaction which enables frame time recordings.

marandaneto:
> Are there any hooks for a transaction start and end which I can use to turn on and off frame time recording? The docs state that it has a small performance impact, so I really don't want to have it enabled the whole time.
> 
> If there's no hooks, I can think of three ways to enable us to go further with this:
> 
> * Add some kind of hooks for transactions. Maybe something like integrations/event processors but just internal?
> * A transaction factory which an SDK can register. The Flutter SDK could then subclass the transaction to add frame time recordings
> * A method on `SentryFlutter` to start a subclassed transaction which enables frame time recordings.

ideally, like Android, the measurements are dependent on this issue https://github.com/getsentry/sentry-dart/pull/611
so we start checking slow/frozen frames when a transaction is created by the navigation transactions.
and then thru the navigation transaction, we can start and stop the measurements monitoring, so no need for hooks or anything, just an integration that handles everything, well, we do need hooks to start and finish a transaction during the screen loading (or routes).

PR #520, opened 2021-07-06
Fix: Incorrect serialization caused by Flutter Enricher

ueman:
## :scroll: Description
An enum was assigned. Dart's JSON serializer can't handle enums, so it broke.

https://github.com/getsentry/sentry-dart/issues/521

Improved test

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes


flutter/lib/src/flutter_enricher_event_processor.dart
```diff
@@ -115,19 +115,23 @@ class FlutterEnricherEventProcessor extends EventProcessor {
...
         'debug_default_target_platform_override':
-            debugDefaultTargetPlatformOverride,
-      if (_window?.initialLifecycleState.isNotEmpty ?? false)
-        'initial_lifecycle_state': _window?.initialLifecycleState,
-      if (_window?.defaultRouteName.isNotEmpty ?? false)
-        'default_route_name': _window?.defaultRouteName,
+            describeEnum(debugPlatformOverride),
+      if (initialLifecycleState != null && initialLifecycleState.isNotEmpty)
+        'initial_lifecycle_state': initialLifecycleState,
+      if (defaultRouteName != null && defaultRouteName.isNotEmpty)
```
marandaneto, line 133: What's about using the pattern `var?.isNotEmpty ?? false` as other places in the code, I mean, it was like this before, why changing it?

marandaneto (approved):
@ueman left a suggestion, but approving it

flutter/lib/src/flutter_enricher_event_processor.dart
ueman: This way there's Dart type inference and I don't need to add a `!`.

Issue #518, opened 2021-07-05
Report `PlatformExceptions` with stacktraces from Dart and the native platform as 'chained exception'


ueman:
[`PlatformException`](https://api.flutter.dev/flutter/services/PlatformException-class.html)s can have a native stacktrace.

As per [doc](https://develop.sentry.dev/sdk/event-payloads/exception/) an event should be able to report a chained exception, see the [examples](https://develop.sentry.dev/sdk/event-payloads/exception/#examples).

Ideally this is done with https://github.com/getsentry/sentry-dart/issues/321

Java implementation for reference: https://github.com/getsentry/sentry-java/blob/a33afef96fc5d051b27ffc81f2bb429585b3488a/sentry/src/main/java/io/sentry/SentryEvent.java#L39

marandaneto:
@ueman this sounds like a dup of https://github.com/getsentry/sentry-dart/issues/321 to me, or what would be the difference?

ueman:
I understood #321 like the parsing step of it, appending the stacktrace and sending it as one stacktrace.
Also we need to add the possibility of having multiple exceptions/stacktraces at all.

Though I'm fine with closing this as a dupe.

marandaneto:
@ueman ok got it, its a more specialized issue then, on what to do with the `PlatformExceptions`, that makes sense if a PlatformExceptions can hold another native stack trace, I'm fine, and they are linked already.

marandaneto:
Closing this one in favor of https://github.com/getsentry/sentry-dart/issues/321

PR #524, opened 2021-07-12
Feat: Events with multiple exceptions and threads

ueman:
## :scroll: Description
Changed SentryEvent in accordance to 
https://develop.sentry.dev/sdk/event-payloads/exception/ and
https://develop.sentry.dev/sdk/event-payloads/threads/

This is a breaking change.

Part of #518

Not yet because I have questions

## :pencil: Checklist

- [x] All tests passing
- [ ] No breaking changes

### DIFF
--- a/dart/lib/src/protocol/sentry_event.dart
+++ b/dart/lib/src/protocol/sentry_event.dart
@@ -3,6 +3,8 @@ import 'package:meta/meta.dart';
 import '../protocol.dart';
 import '../throwable_mechanism.dart';
 import '../utils.dart';
+import '../scope.dart';
+import 'sentry_thread.dart';
 
 /// An event to be reported to Sentry.io.
 @immutable
@@ -16,6 +18,8 @@ class SentryEvent {
     Map<String, dynamic>? extra,
     List<String>? fingerprint,
     List<Breadcrumb>? breadcrumbs,
+    List<SentryException>? exceptions,
+    List<SentryThread>? threads,
     this.sdk,
     this.platform,
     this.logger,
@@ -26,8 +30,6 @@ class SentryEvent {
     this.message,
     this.transaction,
     dynamic throwable,
-    this.stackTrace,
-    this.exception,
     this.level,
     this.culprit,
     this.user,
@@ -42,6 +44,8 @@ class SentryEvent {
         extra = extra != null ? Map.from(extra) : null,
         fingerprint = fingerprint != null ? List.from(fingerprint) : null,
         breadcrumbs = breadcrumbs != null ? List.from(breadcrumbs) : null,
+        exceptions = exceptions != null ? List.from(exceptions) : null,
+        threads = threads != null ? List.from(threads) : null,
         _throwable = throwable;
 
   /// Refers to the default fingerprinting algorithm.
@@ -88,23 +92,25 @@ class SentryEvent {
   ///
   /// It's `runtimeType` and `toString()` are logged.
   /// If it's an Error, with a stackTrace, the stackTrace is logged.
-  /// If this behavior is undesirable, consider using a custom formatted [message] instead.
+  /// If this behavior is undesirable, consider using a custom formatted
+  /// [message] instead.
   dynamic get throwable => (_throwable is ThrowableMechanism)
       ? (_throwable as ThrowableMechanism).throwable
       : _throwable;
 
-  /// A Throwable decorator that holds a Mechanism related to the decorated Throwable
+  /// A throwable decorator that holds a [Mechanism] related to the decorated
+  /// [throwable]
   ///
-  /// Use the 'throwable' directly if you don't want the decorated Throwable
+  /// Use the [throwable] directly if you don't want the decorated throwable
   dynamic get throwableMechanism => _throwable;
 
-  /// an optional attached StackTrace
-  /// used when event has no throwable or exception, see [SentryOptions.attachStacktrace]
-  final SentryStackTrace? stackTrace;
+  /// One or multiple chained (nested) exceptions that occurred in a program.
+  final List<SentryException>? exceptions;
 
-  /// an exception or error that occurred in a program
-  /// TODO more doc
-  final SentryException? exception;
+  /// The Threads Interface specifies threads that were running at the time an
+  /// event happened. These threads can also contain stack traces.
+  /// Typically not needed in Dart applications.
+  final List<SentryThread>? threads;
 
   /// The name of the transaction which generated this event,
   /// for example, the route name: `"/users/<username>/"`.
@@ -128,7 +134,8 @@ class SentryEvent {
   /// List of breadcrumbs for this event.
   ///
   /// See also:
-  /// * https://docs.sentry.io/enriching-error-data/breadcrumbs/?platform=javascript
+  /// * https://docs.sentry.io/platforms/dart/enriching-events/breadcrumbs/
+  /// * https://docs.sentry.io/platforms/flutter/enriching-events/breadcrumbs/
   final List<Breadcrumb>? breadcrumbs;
 
   /// Information about the current user.
@@ -151,22 +158,26 @@ class SentryEvent {
   /// [defaultFingerprint] to the list in addition to your custom values.
   ///
   /// Examples:
-  ///
-  ///     // A completely custom fingerprint:
-  ///     var custom = ['foo', 'bar', 'baz'];
-  ///     // A fingerprint that supplements the default one with value 'foo':
-  ///     var supplemented = [SentryEvent.defaultFingerprint, 'foo'];
+  /// ```dart
+  /// // A completely custom fingerprint:
+  /// var custom = ['foo', 'bar', 'baz'];
+  /// // A fingerprint that supplements the default one with value 'foo':
+  /// var supplemented = [SentryEvent.defaultFingerprint, 'foo'];
+  /// ```
   final List<String>? fingerprint;
 
-  /// The SDK Interface describes the Sentry SDK and its configuration used to capture and transmit an event.
+  /// The SDK Interface describes the Sentry SDK and its configuration used
+  /// to capture and transmit an event.
   final SdkVersion? sdk;
 
-  ///  contains information on a HTTP request related to the event.
-  ///  In client, this can be an outgoing request, or the request that rendered the current web page.
-  ///  On server, this could be the incoming web request that is being handled
+  /// Contains information on a HTTP request related to the event.
+  /// In client, this can be an outgoing request, or the request that rendered
+  /// the current web page.
+  /// On server, this could be the incoming web request that is being handled
   final SentryRequest? request;
 
-  /// The debug meta interface carries debug information for processing errors and crash reports.
+  /// The debug meta interface carries debug information for processing errors
+  /// and crash reports.
   final DebugMeta? debugMeta;
 
   SentryEvent copyWith({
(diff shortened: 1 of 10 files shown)


dart/lib/src/sentry_client.dart
```diff
@@ -124,13 +124,18 @@ class SentryClient {
 
     event = _applyDefaultPii(event);
 
-    if (event.exception != null) return event;
+    if (event.exceptions != null) return event;
```
ueman: Could need an isEmpty check too

dart/test/sentry_event_test.dart
```diff
@@ -292,18 +298,26 @@ void main() {
...
+          name: 'Isolate',
+        )
+      ]).toJson();
+      expect(serialized['threads']['values'], isNotNull);
     });
 
-    test('should not serialize event.stacktrace if event.exception is set', () {
+    /*
+    test(
+        'should not serialize event.threads.stacktrace if event.exception is set',
```
ueman, line 311: I'm not quite sure what the desired behavior of this test (`should not serialize event.threads.stacktrace if event.exception is set`) and the one below it (`should not serialize stacktrace if not SentryStacktrace`) should be.

dart/lib/src/protocol/sentry_event.dart
```diff
@@ -3,6 +3,8 @@ import 'package:meta/meta.dart';
 import '../protocol.dart';
 import '../throwable_mechanism.dart';
 import '../utils.dart';
+import '../scope.dart';
+import 'sentry_thread.dart';
```
marandaneto: can be part of protocol.dart

dart/lib/src/protocol/sentry_event.dart
```diff
@@ -88,23 +92,25 @@ class SentryEvent {
...
+  /// One or multiple chained (nested) exceptions that occurred in a program.
+  final List<SentryException>? exceptions;
 
-  /// an exception or error that occurred in a program
-  /// TODO more doc
-  final SentryException? exception;
+  /// The Threads Interface specifies threads that were running at the time an
+  /// event happened. These threads can also contain stack traces.
+  /// Typically not needed in Dart applications.
+  final List<SentryThread>? threads;
```
marandaneto, line 111: as Dart is single-threaded or errors are not shared across multiple Isolates, maybe we don't add it? that was the reason in the 1st place

dart/lib/src/protocol/sentry_event.dart
```diff
@@ -353,23 +354,20 @@ class SentryEvent {
...
-            'crashed': true,
-            'name': 'Current Isolate',
-          }
-        ]
-      };
+    final exceptionsJson = exceptions
+        ?.map((e) => e.toJson())
+        .where((e) => e.isNotEmpty)
+        .toList(growable: false);
+    if (exceptionsJson != null && exceptionsJson.isNotEmpty) {
```
marandaneto: why not the pattern `?.isNotEmpty ?? false`?

dart/lib/src/protocol/sentry_event.dart
ueman: I thought that we could add the Java thread (or whatever the platforms language is) when adding mixed stacktraces. Though I don't know if that would work.

marandaneto: nvm, its useful because sometimes we add the threads interface because of `Current Isolate`

dart/lib/src/protocol/sentry_event.dart
```diff
@@ -88,23 +92,25 @@ class SentryEvent {
...
   ///
-  /// Use the 'throwable' directly if you don't want the decorated Throwable
+  /// Use the [throwable] directly if you don't want the decorated throwable
   dynamic get throwableMechanism => _throwable;
 
-  /// an optional attached StackTrace
-  /// used when event has no throwable or exception, see [SentryOptions.attachStacktrace]
-  final SentryStackTrace? stackTrace;
+  /// One or multiple chained (nested) exceptions that occurred in a program.
+  final List<SentryException>? exceptions;
```
marandaneto, line 106: on java we have a class called SentryValues that encapsulates the list, https://github.com/getsentry/sentry-java/blob/93f3c221c8923ee6894d7f8edcaffb0a16cb6f14/sentry/src/main/java/io/sentry/SentryValues.java
not sure if its necessary as we don't use reflection for serialization, but worth saying it.

ueman: Yeah, I also saw that. .NET has it, too, but with a comment saying it could be removed because they do serialization themselves, too. Sooo I didn't bother to implement it as we do serialization, too.

dart/lib/src/protocol/sentry_event.dart
ueman: We could replace `Current Isolate` with [Isolate.debugName](https://api.flutter.dev/flutter/dart-isolate/Isolate/debugName.html) if available.

> This name does not uniquely identify an isolate. Multiple isolates in the same process may have the same debugName.
> For a given isolate, this value will be the same as the values returned by Dart_DebugName in the C embedding API and the debugName property in IsolateMirror.

That could be a problem, though.

dart/lib/src/protocol/sentry_thread.dart
```diff
@@ -0,0 +1,72 @@
...
+  /// See https://develop.sentry.dev/sdk/event-payloads/stacktrace/
+  final SentryStackTrace? stacktrace;
+
+  Map<String, dynamic> toJson() {
+    return {
+      if (id != null) 'id': id,
+      if (name != null) 'name': name,
+      if (crashed != null) 'crashed': crashed,
+      if (current != null) 'current': current,
+      if (stacktrace != null) 'stacktrace': stacktrace!.toJson(),
```
marandaneto: should we cache to an extra var instead of `!`?

dart/test/sentry_event_test.dart
marandaneto: if `attachStacktrace` is enabled, the event has a `stackTrace` set, but also a `exception`, and the `exception` also has a `stackTrace`, they are the same since they read the current stack trace from the Isolate or the stackTrace from the error, so we only set the threads.stacktrace if there's not a exception (useful for captureMessage(x) since there's no exception), reducing the amount of sent data

dart/lib/src/protocol/sentry_event.dart
```diff
@@ -26,8 +30,6 @@ class SentryEvent {
     this.message,
     this.transaction,
     dynamic throwable,
-    this.stackTrace,
```
marandaneto, line 29: why is stack trace removed?

marandaneto: oh maybe because when building the event and `attachStackTrace` is enabled, we can add it directly to the thread list instead of during serialization

ueman: Exactly

dart/test/sentry_event_test.dart
ueman: Thanks, that clears it up

ueman: Both test never did anything. Both checked for a `stacktrace` key in the json, but that doesn't exist. Well, that's gonna change now :D

dart/lib/src/protocol/sentry_event.dart
marandaneto: yeah I guess I checked that before, iirc

dart/test/sentry_event_test.dart
marandaneto: damn, luckily we have you :D

dart/lib/src/protocol/sentry_exception.dart
```diff
@@ -29,7 +29,7 @@ class SentryException {
     this.module,
     this.stackTrace,
     this.mechanism,
-    this.threadId,
+    this.threadId = 0,
```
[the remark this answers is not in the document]
marandaneto: only if its a mandatory field, then 0, because isolates don't have an id right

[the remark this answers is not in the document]
ueman: But then how do we link threads to exceptions?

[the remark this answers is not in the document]
marandaneto: there's always only 1 thread, right? so we can default to 0?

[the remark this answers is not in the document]
ueman: Well, you can have different isolates, but you can't identify them with an id of some sorts. Also, because the user is responsible for collecting exceptions in another isolate, I would assume that it's safe to default to 0.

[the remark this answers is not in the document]
bruno-garcia: @ueman we'd only have a single isolate in the `threads` list or many? Could we differentiate between the currently running one where the exception was created? If so we should use their `id` and if no `id` exist we can do like Manoel suggested and add an artificial `id` to the exception and the the 1 isolate that did have an issue.

We just need to keep in mind if we're merging data (adding Java threads for example) things can get messy but AFAIK we don't merge that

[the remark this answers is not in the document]
bruno-garcia: Shouldn't this default to `null`?

Are we going to always send a threads list by default so the concern is that we'll add the thread's stack-trace instead of the exception? If that's the case we need to match the id so Sentry's UI prioritizes the exceptions'.

But if we're adding threads but not sending it (then are we adding this for some cross-SDK support feature?) it would be be better to ensure this is `null`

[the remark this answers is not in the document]
ueman: > we'd only have a single isolate in the threads list or many?

We can't automatically populate `threads` with other isolates except for the current isolate. The user could do that though.

> Could we differentiate between the currently running one where the exception was created?

No, isolates don't have ids and exceptions don't propagate from one isolate to another.

> Are we going to always send a threads list by default so the concern is that we'll add the thread's stack-trace instead of the exception?

Kinda, what I want to do is to only send a stacktrace with the exception if `options.attachStacktrace` is true, so that we don't add the same stacktrace two times.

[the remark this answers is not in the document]
bruno-garcia: > Kinda, what I want to do is to only send a stacktrace with the exception if options.attachStacktrace is true, so that we don't add the same stacktrace two times.

If `attachStacktrace` is true, we create a thread with stacktrace, not an exception though.
If we're capturing an exception (with a stacktrace) then the `attachStacktrace` option doesn't play a role. It's useful for a `captureMessage` for example.


> No, isolates don't have ids and exceptions don't propagate from one isolate to another.

I'm just trying to push for not defauling to `0`. It's best not to send anything then sending `0` on each event. If we don't know the `id` lets not send any?

dart/lib/src/protocol/sentry_event.dart
```diff
@@ -224,29 +231,26 @@ class SentryEvent {
...
-
-    final stackTraceValuesJson = json['threads']?['values'];
-    Map<String, dynamic>? stackTraceValuesStacktraceJson;
-    if (stackTraceValuesJson?.isNotEmpty == true) {
-      stackTraceValuesStacktraceJson = {};
-      stackTraceValuesStacktraceJson =
-          stackTraceValuesJson?.first['stacktrace'];
-    }
+    final breadcrumbs =
+        breadcrumbsJson?.map((e) => Breadcrumb.fromJson(e)).toList();
```
marandaneto:
```suggestion
        breadcrumbsJson?.map((e) => Breadcrumb.fromJson(e)).toList(growable: false);
```

marandaneto: probably missing in some other fields too?

marandaneto (approved):
LGTM

PR #533, opened 2021-07-19
Fix: Always call Flutter.onError

### LINKED ISSUE #522: sentry_flutter: Invoke default FlutterError.onError even if reportSilentFlutterErrors is false
Right now the original FlutterError.onError is only called for **silent** errors if `reportSilentFlutterErrors` is true. This is a bit weird, I would expect that the error messages are simply passed through, always.

https://github.com/getsentry/sentry-dart/blob/679dce91d23ebcf2c0ee405add942685c32d3a20/flutter/lib/src/default_integrations.dart#L54-L68

I suggest to move `_defaultOnError!(errorDetails);` basically to the start of the method.. maybe surrounded in a try{} block?

ueman:
## :scroll: Description
See motivation

Closes #522

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # Unreleased
 
 * Enhancement: Call `toString()` on all non-serializable fields (#528)
-* Fix: Always call `Flutter.onError` in order to not swallow messages
+* Fix: Always call `Flutter.onError` in order to not swallow messages (#533)
 
 # 6.0.0-beta.2
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Unreleased
 
 * Enhancement: Call `toString()` on all non-serializable fields (#528)
+* Fix: Always call `Flutter.onError` in order to not swallow messages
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # Unreleased
 
 * Enhancement: Call `toString()` on all non-serializable fields (#528)
+* Fix: Always call `Flutter.onError` in order to not swallow messages
```
```suggestion
* Fix: Always call `Flutter.onError` in order to not swallow messages (#533)
```

flutter/lib/src/default_integrations.dart
```diff
@@ -78,6 +72,12 @@ class FlutterErrorIntegration extends Integration<SentryFlutterOptions> {
           'if you wish to capture silent errors',
         );
       }
+      // Call original handler, regardless of `errorDetails.silent` or
+      // `reportSilentFlutterErrors`. This ensures, that we doen't swallow
```
```suggestion
      // `reportSilentFlutterErrors`. This ensures, that we don't swallow
```

marandaneto (approved):

PR #537, opened 2021-07-24
Replace package_info_plus instead by package_info

tuanitpro:
## :scroll: Description
package_info_plus-1.0.4/android/src/main/java/dev/fluttercommunity/plus/packageinfo/PackageInfoPlugin.java uses or overrides a deprecated API.

## :pencil: Checklist

- [ ] All tests passing
- [ ] No breaking changes


ueman:
Thank you for this PR.
However, package_info is deprecated and package_info_plus is it's successor. So we will not accept this PR.

But I'm curious, why do you want to use package_info?

PR #538, opened 2021-07-24
Feat: Add some more information for exceptions collected via `FlutterError.onError`

ueman:
## :scroll: Description
Bildschirmfoto 2021-07-24 um 17 26 41

The information in the red square is added by this PR.

Changed test accordingly

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/flutter/example/lib/main.dart
+++ b/flutter/example/lib/main.dart
@@ -147,6 +147,26 @@ class MainScaffold extends StatelessWidget {
                   () => throw Exception('Throws in Future.delayed')),
             ),
             RaisedButton(
+              child: const Text('Capture from FlutterError.onError'),
+              onPressed: () {
+                // modeled after a real exception
+                FlutterError.onError?.call(FlutterErrorDetails(
+                  exception: Exception('A really bad exception'),
+                  silent: false,
+                  context: DiagnosticsNode.message('while handling a gesture'),
+                  library: 'gesture',
+                  informationCollector: () => [
+                    DiagnosticsNode.message(
+                        'Handler: "onTap" Recognizer: TapGestureRecognizer'),
+                    DiagnosticsNode.message(
+                        'Handler: "onTap" Recognizer: TapGestureRecognizer'),
+                    DiagnosticsNode.message(
+                        'Handler: "onTap" Recognizer: TapGestureRecognizer'),
+                  ],
+                ));
+              },
+            ),
+            RaisedButton(
               child: const Text('Dart: Web request'),
               onPressed: () => makeWebRequest(context),
             ),
--- a/flutter/lib/src/default_integrations.dart
+++ b/flutter/lib/src/default_integrations.dart
@@ -57,19 +57,27 @@ class FlutterErrorIntegration extends Integration<SentryFlutterOptions> {
         final collector = errorDetails.informationCollector?.call() ?? [];
         final information =
             (StringBuffer()..writeAll(collector, '\n')).toString();
+        // errorDetails.library defaults to 'Flutter framework' even though it
+        // is nullable. We do null checks anyway, just to be sure.
         final library = errorDetails.library;
+
+        final flutterErrorDetails = <String, String>{
+          // This is a message which should make sense if written after the
+          // word `thrown`:
+          // https://api.flutter.dev/flutter/foundation/FlutterErrorDetails/context.html
+          if (context != null) 'context': 'thrown $context',
+          if (collector.isNotEmpty) 'information': information,
+          if (library != null) 'library': library,
+        };
 
         // FlutterError doesn't crash the App.
         final mechanism = Mechanism(
           type: 'FlutterError',
           handled: true,
           data: {
-            // This is a message which should make sense if written after the
-            // word `thrown`:
-            // https://api.flutter.dev/flutter/foundation/FlutterErrorDetails/context.html
-            if (context != null) 'context': 'thrown $context',
-            if (collector.isNotEmpty) 'information': information,
-            if (library != null) 'library': library,
+            if (flutterErrorDetails.isNotEmpty)
+              'hint':
+                  'See "flutter_error_details" down below for more information'
           },
         );
         final throwableMechanism = ThrowableMechanism(mechanism, exception);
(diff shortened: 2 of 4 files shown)


flutter/lib/src/default_integrations.dart
```diff
@@ -44,16 +44,31 @@ class FlutterErrorIntegration extends Integration<SentryFlutterOptions> {
...
+            (StringBuffer()..writeAll(collector, '\n')).toString();
+        final library = errorDetails.library;
+
         // FlutterError doesn't crash the App.
-        final mechanism = Mechanism(type: 'FlutterError', handled: true);
+        final mechanism = Mechanism(
+          type: 'FlutterError',
+          handled: true,
+          data: {
+            if (context != null) 'context': context,
```
ueman: As per [docs](https://api.flutter.dev/flutter/foundation/FlutterErrorDetails/context.html) this is a message which should make sense if written after the word `thrown`.
```suggestion
            if (context != null) 'context': 'thrown $context',
```
or maybe 
```suggestion
            if (context != null) 'thrown': context,
```

bruno-garcia: Then what about adding ur comment to it too:

```suggestion
            // This is a message which should make sense if written after the word `thrown`: https://api.flutter.dev/flutter/foundation/FlutterErrorDetails/context.html
            if (context != null) 'context': 'thrown $context',
```

flutter/test/default_integrations_test.dart
```diff
@@ -46,6 +46,9 @@ void main() {
     final details = FlutterErrorDetails(
       exception: throwable,
       silent: silent,
+      context: DiagnosticsNode.message('while handling a gesture'),
+      library: 'sentry',
+      informationCollector: () => [DiagnosticsNode.message('foo bar')],
```
bruno-garcia, line 52: Worth adding a test to validate the behavior with multiple lines (since this is an array).
Also good to check what it looks like in Sentry.

bruno-garcia (commented):
How often can the `information` field be larger than what Sentry can display properly? It transforms an array into a single line, so it could end up truncated or not fitting the UI very well. Perhaps we need to find another way to present it?

ueman:
> How often can the `information` field be larger than what Sentry can display properly? It transforms an array into a single line, so it could end up truncated or not fitting the UI very well. 

Hard to say as its origin can be the Flutter framework, libraries and developers themselves.
So it's probably better to assume that it can be very long.

> Perhaps we need to find another way to present it?

What would be a better way to display this?
As far as I can tell, the only data which is displayed as a text block is the exception description.

The events data could be another place for this, but then the information would be visually disconnected in the Sentry UI :(

ueman:
It doesn't play nicely if it's really long :( I raised an issue for this little UI bug here: https://github.com/getsentry/sentry/issues/27737
Bildschirmfoto 2021-07-26 um 08 47 01

flutter/example/lib/main.dart
```diff
@@ -146,6 +146,26 @@ class MainScaffold extends StatelessWidget {
...
+                    DiagnosticsNode.message(
+                        'Handler: "onTap" Recognizer: TapGestureRecognizer'),
+                    DiagnosticsNode.message(
+                        'Handler: "onTap" Recognizer: TapGestureRecognizer'),
+                    DiagnosticsNode.message(
+                        'Handler: "onTap" Recognizer: TapGestureRecognizer'),
+                  ],
+                ));
+              },
+            ),
```
marandaneto, line 168: the example `Flutter error : Scaffold.of()` already triggers a `FlutterError.onError` error, isnt that enough? is it common that people trigger `FlutterError.onError` manually too?

ueman: > the example Flutter error : Scaffold.of() already triggers a FlutterError.onError error, isnt that enough?

This is just to test how it looks in the Sentry UI when it has a long trail of information, which `Flutter error : Scaffold.of()` doesn't have (as seen in the screenshot of the PR description). We can remove it before merging.

> is it common that people trigger FlutterError.onError manually too?

No, but it's not uncommon in libraries.

marandaneto:
@ueman is this ready to be merged or the multiple line thing should be tested?

ueman:
> @ueman is this ready to be merged or the multiple line thing should be tested?

It works and is tested but it could be that it doesn't look that good in the Sentry UI for long information. So unless you or @bruno-garcia has a better idea where to put the information this is good to go.

I tried to add it to the events extra data, but then this information kinda gets lost.

marandaneto:
> I tried to add it to the events extra data, but then this information kinda gets lost.

what do you mean by kinda get lost? does it make to Sentry or not easy to find since the information is down there in the additional data?
just trying to understand if we have a bug or not.

ueman:
> > I tried to add it to the events extra data, but then this information kinda gets lost.
> 
> what do you mean by kinda get lost? does it make to Sentry or not easy to find since the information is down there in the additional data?
> just trying to understand if we have a bug or not.

It gets visually lost when it at the same place as all the device, os and app information. It does make to Sentry so it's not a bug.

I'll add some pictures later to clarify what I mean :D

ueman:
This event has the data added to the mechanism and as the `crash_context`. At the top I think it's really useful, though it might not look good for long text, just like @bruno-garcia said. At the bottom, the information gets kinda visually lost next to all the meta data. 
sentry io_organizations_sentry-sdks_issues_2535239328__project=5428562 query=is%3Aunresolved

bruno-garcia:
I'm a bit wary of adding data that'll be potentially large in that field in this form. Perhaps best we find a way to break in up in more 'key/value' pairs

bagedevimo:
This is looks amazing for us, as we have 3 different flutter apps all using this context data and we've only just realized it's missing. Very keen to help out in any way that I can.

ueman:
> This is looks amazing for us, as we have 3 different flutter apps all using this context data and we've only just realized it's missing. Very keen to help out in any way that I can.

Thanks! That's really appreciated. 
Do you mind answering the following questions?

1. How long are your `context` and `information` values?
1. What length of these fields have you seen in the wild?
1. How would you like to see these fields in the Sentry UI?

bagedevimo:
Our context fields are typically 3-4 words, similar length to your example. We have none longer than 36 characters. I'm afraid we've never seen information values in practice, as Sentry isn't capturing them :P. Looking at the definition of FlutterErrorDetails, it seems reasonable to suggest these should be key/value pairs, but as far as I can tell you could only split on the newlines that may or may not exist. Seems fine for most use-cases? I don't know a lot about the Sentry UI though sorry. A neat UX might be that if a tag is too long up the top, it is cut off and can be clicked to scroll down to the full value.

ueman:
Okay, I've changed it so that the mechanism has a tag which tells the user to look for a `flutter_error_details` context down below. Down there, everything displays well. I hope that is satisfying for all :D

marandaneto (approved):

PR #553, opened 2021-08-17
Feat: Add max attachment size option

### LINKED ISSUE #531: Maximum attachment size option
We need to filter out attachments which are of a certain size.
Like https://docs.sentry.io/platforms/dotnet/enriching-events/attachments/#maximum-attachment-size for example.

ueman:
## :scroll: Description
Add max attachment size option according to https://develop.sentry.dev/sdk/features/#max-attachment-size

https://develop.sentry.dev/sdk/features/#max-attachment-size

Closes https://github.com/getsentry/sentry-dart/issues/531

New tests

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/dart/lib/src/sentry_envelope.dart
+++ b/dart/lib/src/sentry_envelope.dart
@@ -1,4 +1,7 @@
 import 'dart:convert';
+
+import 'sentry_item_type.dart';
+import 'sentry_options.dart';
 import 'utils.dart';
 import 'sentry_attachment/sentry_attachment.dart';
 import 'sentry_envelope_header.dart';
@@ -17,7 +20,7 @@ class SentryEnvelope {
   /// All items contained in the envelope.
   final List<SentryEnvelopeItem> items;
 
-  /// Create an `SentryEnvelope` with containing one `SentryEnvelopeItem` which holds the `SentyEvent` data.
+  /// Create an [SentryEnvelope] with containing one [SentryEnvelopeItem] which holds the [SentryEvent] data.
   factory SentryEnvelope.fromEvent(
     SentryEvent event,
     SdkVersion sdkVersion, {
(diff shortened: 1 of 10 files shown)


dart/lib/src/sentry_options.dart
```diff
@@ -53,6 +53,22 @@ class SentryOptions {
...
+
+  /// Maximum allowed file size of attachments, in bytes.
+  /// Attachments above this size will be discarded
+  ///
+  /// Remarks: Regardless of this setting, attachments are also limited to 20mb
+  /// (compressed) on Relay.
+  int get maxAttachmentSize => _maxAttachmentSize;
+
+  set maxAttachmentSize(int maxAttachmentSize) {
+    assert(maxAttachmentSize >= 0);
```
```suggestion
    assert(maxAttachmentSize > 0);
```

since 0 does not make sense?

dart/lib/src/sentry_envelope_item.dart
```diff
@@ -11,11 +11,15 @@ class SentryEnvelopeItem {
...
-      return await attachment.bytes;
-    });
+    final cachedItem = _CachedItem(() async => await attachment.bytes);
 
-    final getLength = () async => (await cachedItem.getData()).length;
+    final getLength = () async {
+      try {
+        return (await cachedItem.getData()).length;
+      } catch (_) {
+        return 0;
```
marandaneto: an attachment that has problems getting its data should not be added to the envelope, so returning 0 looks like a valid attachment, and its wrong.

dart/lib/src/sentry_envelope.dart
```diff
@@ -44,13 +47,18 @@ class SentryEnvelope {
...
       header.toJson(),
       toEncodable: jsonSerializationFallback,
     ));
     final newLineData = utf8.encode('\n');
     for (final item in items) {
+      if (item.header.type == SentryItemType.attachment) {
+        if (await item.header.length() > options.maxAttachmentSize) {
+          continue;
+        }
+      }
```
marandaneto, line 69: on Java, the validation happens on `SentryEnvelopeItem.fromAttachment` since the `envelopeStream` is too late, we know in advance if the envelope is bigger or not

ueman: In Dart we do not know the size in advance :(

dart/lib/src/sentry_envelope_item.dart
marandaneto: maybe we return -1 here and who calls this function test if its -1 and drop it? we don't want to add an empty line or break the envelope

marandaneto (approved):

PR #560, opened 2021-08-20
Set name of current route as transaction

ueman:
## :scroll: Description
This change adds the current route as `SentryEvent.transaction`.

https://develop.sentry.dev/sdk/event-payloads/ suggests to use the current route as transaction name.

Unit tests

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/flutter/lib/src/navigation/sentry_navigator_observer.dart
+++ b/flutter/lib/src/navigation/sentry_navigator_observer.dart
@@ -35,13 +35,16 @@ const _navigationKey = 'navigation';
 ///   - [RouteObserver](https://api.flutter.dev/flutter/widgets/RouteObserver-class.html)
 ///   - [Navigating with arguments](https://flutter.dev/docs/cookbook/navigation/navigate-with-arguments)
 class SentryNavigatorObserver extends RouteObserver<PageRoute<dynamic>> {
-  SentryNavigatorObserver({Hub? hub}) : hub = hub ?? HubAdapter();
+  SentryNavigatorObserver({Hub? hub, this.setTransaction = true})
+      : _hub = hub ?? HubAdapter();
 
-  final Hub hub;
+  final Hub _hub;
+  final bool setTransaction;
 
   @override
   void didPush(Route<dynamic> route, Route<dynamic>? previousRoute) {
     super.didPush(route, previousRoute);
+    setCurrentRoute(route.settings.name);
     _addBreadcrumb(
       type: 'didPush',
       from: previousRoute?.settings,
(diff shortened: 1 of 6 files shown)


marandaneto:
although this makes sense, it'd conflict with the performance API that comes next, since this field kinda got changed, let's not merge for now

ueman:
@marandaneto Should this still be considered or can this be closed?

I mean it kinda makes sense if performance tracing is disabled.

flutter/test/sentry_navigator_observer_test.dart
```diff
@@ -226,5 +227,47 @@ void main() {
         ).data,
       );
     });
+
+    test('route name as transaction', () {
+      final hub = _MockHub();
+      final observer = SentryNavigatorObserver(hub: hub, setTransaction: true);
```
marandaneto, line 233: we can make a fixture

flutter/lib/src/navigation/sentry_navigator_observer.dart
```diff
@@ -35,13 +35,16 @@ const _navigationKey = 'navigation';
 ///   - [RouteObserver](https://api.flutter.dev/flutter/widgets/RouteObserver-class.html)
 ///   - [Navigating with arguments](https://flutter.dev/docs/cookbook/navigation/navigate-with-arguments)
 class SentryNavigatorObserver extends RouteObserver<PageRoute<dynamic>> {
-  SentryNavigatorObserver({Hub? hub}) : hub = hub ?? HubAdapter();
+  SentryNavigatorObserver({Hub? hub, this.setTransaction = true})
+      : _hub = hub ?? HubAdapter();
 
-  final Hub hub;
+  final Hub _hub;
+  final bool setTransaction;
```
marandaneto, line 42: we can make it private and probably give a better naming, also writing in the ctor docs that enabling this overwrites the current transaction name in the scope.

about the name, maybe `routeNameAsTransactionName` not sure, but `setTransaction` only is too subjective IMO

marandaneto:
> @marandaneto Should this still be considered or can this be closed?
> 
> I mean it kinda makes sense if performance tracing is disabled.

still makes sense, I did a review, would u like to work on it? so we can merge, thanks for doing this btw.

flutter/lib/src/navigation/sentry_navigator_observer.dart
```diff
@@ -76,12 +79,20 @@ class SentryNavigatorObserver extends RouteObserver<PageRoute<dynamic>> {
...
   }) {
-    hub.addBreadcrumb(RouteObserverBreadcrumb(
+    _hub.addBreadcrumb(RouteObserverBreadcrumb(
       navigationType: type,
       from: from,
       to: to,
     ));
   }
+
+  void setCurrentRoute(String? name) {
```
ueman, line 89: What should happen if the route name is `null`?
Resetting the transaction name to null or just nothing?

flutter/lib/src/navigation/sentry_navigator_observer.dart
ueman: This can't be made private because it's a named constructor argument. Though because it's final it's basically readonly.

ueman:
Closed this in favor of #615 because I don't have the rights anymore.

flutter/lib/src/navigation/sentry_navigator_observer.dart
marandaneto: nothing I'd say, but I guess the route is never null anyway

PR #567, opened 2021-08-24
Feat: Read environment variables on IO platforms

### LINKED ISSUE #557: Check if String.fromEnvironment works
as Ive used `Platform.environment` to make it work on the Sentry Dart Plugin, maybe `String.fromEnvironment` only works for dart-defines and not machine env. var., we have to check it.

ueman:
## :scroll: Description
The user can now configure the Sentry SDK via environment variables on Dart IO platforms.

Closes https://github.com/getsentry/sentry-dart/issues/557

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 * Feat: Enable or disable reporting of packages (#563)
 * Bump: Sentry-Cocoa to 7.2.2 (#566)
 * Bump: Sentry-Android to 5.1.1 (#566)
+* Feat: Read Sentry config from environment variables as fallback
 
 # 6.0.0-beta.4
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 * Feat: Enable or disable reporting of packages (#563)
 * Bump: Sentry-Cocoa to 7.2.2 (#566)
 * Bump: Sentry-Android to 5.1.1 (#566)
-* Feat: Read Sentry config from environment variables as fallback
+* Feat: Read Sentry config from environment variables as fallback (#567)
 
 # 6.0.0-beta.4
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 * Feat: Enable or disable reporting of packages (#563)
 * Bump: Sentry-Cocoa to 7.2.2 (#566)
 * Bump: Sentry-Android to 5.1.1 (#566)
-* Feat: Read Sentry config from environment variables as fallback (#567)
+* Fix: Read Sentry config from environment variables as fallback (#567)
 
 # 6.0.0-beta.4
 
--- a/dart/lib/src/environment/_io_environment_variables.dart
+++ b/dart/lib/src/environment/_io_environment_variables.dart
@@ -1,0 +1,42 @@
+import 'dart:io';
+
+import 'environment_variables.dart';
+import 'keys.dart';
+
+final EnvironmentVariables envs = IoEnvironmentVariables();
+
+/// In addition to dart defines this io implementation can read from the
+/// environment variables.
+class IoEnvironmentVariables extends EnvironmentVariables {
+  @override
+  String? get environment {
+    if (const bool.hasEnvironment(sentryEnvironment)) {
+      return const String.fromEnvironment(sentryEnvironment);
+    }
+    return Platform.environment[sentryEnvironment];
+  }
+
+  @override
+  String? get dsn {
+    if (const bool.hasEnvironment(sentryDsn)) {
+      return const String.fromEnvironment(sentryDsn);
+    }
+    return Platform.environment[sentryDsn];
+  }
+
+  @override
+  String? get release {
+    if (const bool.hasEnvironment(sentryRelease)) {
+      return const String.fromEnvironment(sentryRelease);
+    }
+    return Platform.environment[sentryRelease];
+  }
+
+  @override
+  String? get dist {
+    if (const bool.hasEnvironment(sentryDist)) {
+      return const String.fromEnvironment(sentryDist);
+    }
+    return Platform.environment[sentryDist];
+  }
+}
--- a/dart/lib/src/environment/_io_environment_variables.dart
+++ b/dart/lib/src/environment/_io_environment_variables.dart
@@ -1,5 +1,6 @@
 import 'dart:io';
 
+import '_web_environment_variables.dart';
 import 'environment_variables.dart';
 import 'keys.dart';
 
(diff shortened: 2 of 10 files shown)


dart/lib/src/environment/_io_environment_variables.dart
```diff
@@ -0,0 +1,42 @@
+import 'dart:io';
+
+import 'environment_variables.dart';
+import 'keys.dart';
+
+final EnvironmentVariables envs = IoEnvironmentVariables();
```
marandaneto, line 7:
```suggestion
final EnvironmentVariables _envs = IoEnvironmentVariables();
```

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 * Feat: Enable or disable reporting of packages (#563)
 * Bump: Sentry-Cocoa to 7.2.2 (#566)
 * Bump: Sentry-Android to 5.1.1 (#566)
+* Feat: Read Sentry config from environment variables as fallback (#567)
```
```suggestion
* Fix: Read Sentry config from environment variables as fallback (#567)
```

dart/lib/src/environment/_web_environment_variables.dart
```diff
@@ -0,0 +1,26 @@
+import 'environment_variables.dart';
+import 'keys.dart';
+
+final EnvironmentVariables envs = WebEnvironmentVariables();
```
marandaneto, line 4:
```suggestion
final EnvironmentVariables _envs = WebEnvironmentVariables();
```

dart/lib/src/environment/_io_environment_variables.dart
ueman: I need this to be available in `environment_variables.dart` so it can not be private.

dart/lib/src/environment/_web_environment_variables.dart
ueman: Same as above

dart/lib/src/environment/_io_environment_variables.dart
marandaneto: ah ok, makes sense

dart/lib/src/environment/_io_environment_variables.dart
```diff
@@ -0,0 +1,42 @@
...
+
+/// In addition to dart defines this io implementation can read from the
+/// environment variables.
+class IoEnvironmentVariables extends EnvironmentVariables {
+  @override
+  String? get environment {
+    if (const bool.hasEnvironment(sentryEnvironment)) {
+      return const String.fromEnvironment(sentryEnvironment);
+    }
```
marandaneto: not sure if it makes sense or if it makes even more confusing, but we could simply do like:

```
final web = WebEnvironmentVariables()

String? get environment {
  return web.environment ?? Platform.environment[sentryEnvironment];
}
```

ueman: Good point

marandaneto:
looks good, some testing and thats it :)

ueman:
> looks good, some testing and thats it :)

How do you suggest to test it? There's no way to mock `Platform` or the dart defines.

marandaneto:
> > looks good, some testing and thats it :)
> 
> How do you suggest to test it? There's no way to mock `Platform` or the dart defines.

it'd require a level of abstraction on top of `hasEnvironment`, `fromEnvironment` and `Platform.environment`, not sure if its worth though, your call to add the level of abstraction and unit tests or to test it manually that it picks up either the dart defines or the env var

ueman:
@marandaneto  What do you think about using the e2e test to verify that behavior? We could set the values and verify that they match the expected values.

marandaneto:
> @marandaneto What do you think about using the e2e test to verify that behavior? We could set the values and verify that they match the expected values.

sounds also good, set eg the DSN via dart-defines and environment via env. var. or the other way around

e2e_test/bin/e2e_test.dart
```diff
@@ -17,10 +17,26 @@ void main(List<String> arguments) async {
...
+  // check those after init.
+  if (_options?.environment != 'e2e') {
+    print('Environment is not e2e');
+    exit(1);
+  }
+
+  if (_options?.dist != '1') {
+    print('Dist is not 1');
+    exit(1);
+  }
```
marandaneto: the other option would be to `SentryEvent.fromJson` with the API response and assert the fields of the event, eg `event.dist`, `event.environment`, etc..., what do you think?

ueman: That are two tests for the price of one :D

marandaneto: indeed, but the idea is that we'd be expanding the assertion as discussed here https://github.com/getsentry/sentry-dart/issues/514#issuecomment-905830314
so asserting `environment` and `dist` would not be a problem, your call, the current solution also works, but holding the reference of `_options` is also very hacky :)

marandaneto (approved):

e2e_test/bin/e2e_test.dart
```diff
@@ -10,8 +10,7 @@ const _exampleDsn =
 const _org = 'sentry-sdks';
 const _projectSlug = 'sentry-flutter';
 
-final _token =
```
marandaneto: if this is the token from `SENTRY_AUTH_TOKEN`, lets rotate it, if its a personal one, rotate as well please

e2e_test/bin/e2e_test.dart
```diff
@@ -51,11 +78,12 @@ Future<bool> _waitForEventToShowUp(Uri url) async {
     );
     print('${response.statusCode}: ${response.body}');
     if (response.statusCode == 200) {
-      return true;
+      // The json does not match what `SentryEvent.fromJson` expects
+      return jsonDecode(utf8.decode(response.bodyBytes));
```
marandaneto, line 82: how so? it should, can you send me the payload?

e2e_test/bin/e2e_test.dart
ueman: Yeah, that was a personal

e2e_test/bin/e2e_test.dart
ueman:
```json
{"event_id":"1a83e0da67914827909cc8dee3ca28d9","project":5428562,"release":null,"dist":null,"platform":"other","message":"","datetime":"2021-08-31T09:57:24.247000Z","tags":[["environment","e2e"],["level","error"],["os","linux Linux 5.8.0-1039-azure #42~20.04.1-Ubuntu SMP Thu Jul 15 14:11:07 UTC 2021"],["os.name","linux"]],"_metrics":{"bytes.ingested.event":1150,"bytes.stored.event":2226},"contexts":{"dart_context":{"number_of_processors":2,"package_config":"file:///home/runner/work/sentry-dart/sentry-dart/e2e_test/.dart_tool/package_config.json","type":"dart_context"},"device":{"name":"fv-az399-758","timezone":"UTC","language":"C.UTF-8","type":"device"},"os":{"name":"linux","version":"Linux 5.8.0-1039-azure #42~20.04.1-Ubuntu SMP Thu Jul 15 14:11:07 UTC 2021","type":"os"},"sentry_dart_runtime":{"name":"Dart","raw_description":"2.15.0-65.0.dev (dev) (Fri Aug 27 11:53:11 2021 -0700) on \"linux_x64\"","type":"sentry_dart_runtime"}},"culprit":"e2e_test.dart in main","environment":"e2e","exception":{"values":[{"type":"_Exception","value":"Exception: E2E Test Message","stacktrace":{"frames":[{"function":"main","filename":"e2e_test.dart","abs_path":"e2e_test.dart","lineno":27,"colno":5,"in_app":true}]}}]},"fingerprint":["{{ default }}"],"grouping_config":{"enhancements":"eJybzDRxY3J-bm5-npWRgaGlroGxrpHxpEkTNxXkF6XrpiQWleQk5qXrFqdka008y8yIS4IBANd-Gx4","id":"newstyle:2019-10-29"},"hashes":["f470a3c7079f68d620bd4204ac51ff1a"],"ingest_path":[{"public_key":"XE7QiyuNlja9PZ7I9qJlwQotzecWrUIN91BAO7Q5R38","version":"21.8.0"}],"key_id":"1764334","level":"error","location":"e2e_test.dart","logger":"","metadata":{"filename":"e2e_test.dart","function":"main","type":"_Exception","value":"Exception: E2E Test Message"},"received":1630403844.498328,"sdk":{"name":"sentry.dart","version":"6.0.0-beta.5","integrations":["isolateErrorIntegration"],"packages":[{"name":"pub:sentry","version":"6.0.0-beta.5"}]},"timestamp":1630403844.247,"title":"_Exception: Exception: E2E Test Message","type":"error","version":"7"}
```

from https://sentry.io/share/issue/b378131b171643b28ebd70c8f0c430a6/ 
aka https://sentry.io/organizations/sentry-sdks/issues/2590383852/events/1a83e0da67914827909cc8dee3ca28d9/json/

ueman: So for example tags in the event (our event and docs) look like
```json
{
  "tags": {
    "ios_version": "4.0",
    "context": "production"
  }
}
```
but actually look like
```json
{
"tags":[
      [
         "environment",
         "e2e"
      ],
      [
         "level",
         "error"
      ],
      [
         "os",
         "linux Linux 5.8.0-1039-azure #42~20.04.1-Ubuntu SMP Thu Jul 15 14:11:07 UTC 2021"
      ],
      [
         "os.name",
         "linux"
      ]
   ],
}
```

ueman: It is also kinda documented that they look kinda different: https://docs.sentry.io/api/events/retrieve-an-event-for-a-project/

marandaneto: oh got it, indeed, as it returns a Map, we could assert by key name then following the contract described above, TIL

PR #569, opened 2021-08-25
Upload source maps for web example

marandaneto:
## :scroll: Description
_#skip-changelog_

## :bulb: Motivation and Context
Closes https://github.com/getsentry/sentry-dart/issues/430

## :pencil: Checklist

- [ ] I reviewed submitted code
- [ ] I added tests to verify changes
- [ ] I updated the docs if needed
- [ ] All tests passing
- [ ] No breaking changes

PR #571, opened 2021-08-25
Update run script

marandaneto:
## :scroll: Description
_#skip-changelog_

## :pencil: Checklist

- [ ] All tests passing
- [ ] No breaking changes

### DIFF
--- a/flutter/example/run.sh
+++ b/flutter/example/run.sh
@@ -1,7 +1,10 @@
 #!/bin/bash
 set -e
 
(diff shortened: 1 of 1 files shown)


bruno-garcia (approved):
LGTM!

Not going to use the plugin instead?

marandaneto:
> LGTM!
> 
> Not going to use the plugin instead?

we use the plugin #569, but the script is convenience and its linked across a few SS posts as well, so keeping it up to date at least for now, when the plugin goes GA, we can drop it

PR #573, opened 2021-08-30
Enha: Refactoring and improvements for the performance API on top of the performance PoC

marandaneto:
- Performance
- wip
- Feat perf new arch classes

## :scroll: Description
_#skip-changelog_

## :pencil: Checklist

- [ ] All tests passing
- [ ] No breaking changes


dart/lib/src/sentry_tracer.dart
```diff
@@ -0,0 +1,20 @@
...
+  String? name;
+  @override
+  late SentrySpanContext context;
+  DateTime? timestamp;
+  DateTime? startTimestamp;
+  late bool isFinished;
+
+  // find out how to pass its own instance to super
+  SentryTracer(this._hub, this._transactionContext)
+      : super(_transactionContext);
```
marandaneto: @ueman do you know if this is even possible?

`SentrySpan` having an instance of `SentryTracer` via `super` call

maybe its not even necessary, but objC does it

ueman: I don't see why it shouldn't be possible.

Though maybe it has to look like this:
```dart
SentryTracer(this._hub, SentryTransactionContext transactionContext)
      : _transactionContext = transactionContext, super(transactionContext);
```

dart/lib/src/sentry_transaction.dart
```diff
@@ -0,0 +1,24 @@
+import '../sentry.dart';
+
+class SentryTransaction extends SentryEvent {
```
ueman, line 4: What if we make a `EventLike` mixin:

```dart
mixin EventLike<T> {
// basically all fields of an event

  // abstract
  T copyWith();
}
```

Then we could do:
```dart 
class SentryEvent with EventLike<SentryEvent> {
  @override
  SentryEvent copyWith() { ... }
}
```

```dart 
class SentryTransaction with EventLike<SentryTransaction> {
  @override
  SentryTransaction copyWith() { ... }
}
```

That way we could have immutable SentryEvent and SentryTransaction.
The same could also work for ISentrySpan.

ueman: Well if we want to have immutable transaction and span classes

marandaneto: yep, Sentry Java has SentryBaseEvent so all common properties go there since many properties from SentryEvent does not make much sense for SentryTransaction, we'd end up at some point with either abstract class or `mixin`, I'm just trying to make it work the whole setup for now but not changing much, I find it easier to refactor when its already working, but its a good idea

dart/lib/src/hub.dart
```diff
@@ -309,43 +314,95 @@ class Hub {
...
+        SentryTransactionContext(
+          name,
+          operation,
+          description: description,
+        ),
+        bindToScope: bindToScope,
+      );
 
-  void startTransaction() {}
+  ISentrySpan startTransactionWithContext(
```
marandaneto, line 332: to discuss: since Dart does not have method overloadings, either we use different method names or fit every possible parameter as optional, but this could cause confusion since you may pass `transactionContext` that takes a `name` and `name` but with different values

dart/lib/src/hub.dart
```diff
@@ -309,43 +314,95 @@ class Hub {
...
-      final item = _peek();
-      final event = await item.scope.applyToEvent(transaction.data, null);
-      if (event == null) {
-        return SentryId.empty();
+      if (!transaction.finished) {
+        _options.logger(
+          SentryLevel.warning,
+          'Capturing unfinished transaction: ${transaction.eventId}',
+        );
       }
```
marandaneto, line 392: similar to https://github.com/getsentry/sentry-java/issues/1690 maybe, does it make sense to send a transaction if unfinished?

dart/lib/src/protocol/contexts.dart
```diff
@@ -123,7 +122,8 @@ class Contexts extends MapView<String, dynamic> {
 
   set gpu(SentryGpu? gpu) => this[SentryGpu.type] = gpu;
 
-  /// Trace context describes the GPU of the device.
+  /// looks like trace should not be SentrySpanContext but instead something simpler
+  /// which is not with all the fields from Span
   SentrySpanContext? get trace => this[SentrySpanContext.type];
```
marandaneto: maybe we need a SentryTraceContext which only has the trace fields instead of the whole span context?

dart/lib/src/scope.dart
```diff
@@ -14,7 +15,8 @@ class Scope {
   String? transactionName;
 
   /// Returns active transaction or null if there is no active transaction.
-  SentryTransaction? transaction;
+  // SentryTransaction? transaction;
+  ISentrySpan? span;
```
marandaneto, line 18: doing it like iOS, theres no scope.setTransaction(x) but rather scope.setSpan(x) which is the transaction

dart/lib/src/sentry_span_interface.dart
```diff
@@ -0,0 +1,33 @@
...
+    String? description,
+  });
+
+  void setTag(String key, String value);
+
+  void removeTag(String key);
+  // exposed to Transaction too but it does not get sent to the backend since transactions dont have the data bag
+  void setData(String key, dynamic value);
+// exposed to Transaction too but it does not get sent to the backend since transactions dont have the data bag
+  void removeData(String key);
```
marandaneto: `setData` is available under the transaction instance but there's no such `data` field, do we ignore it or set it in the `extra` field? on Android, we don't have the data field at all because of that

dart/lib/src/sentry_tracer.dart
```diff
@@ -0,0 +1,104 @@
+import '../sentry.dart';
+import 'tracing.dart';
+
+class SentryTracer extends ISentrySpan {
+  final Hub _hub;
+  late String _name;
+
+  // missing waitForChildren
```
marandaneto, line 8: do we need this feature for the manual API?

dart/lib/src/sentry_tracer.dart
```diff
@@ -0,0 +1,104 @@
...
+        scope.span = null;
+      }
+    });
+
+    final transaction = _toTransaction();
+    await _hub.captureTransaction(transaction);
+  }
+
+  SentryTransaction _toTransaction() {
+    // filter unfinished spans _children
```
marandaneto: add filtering of non finbished spans

dart/lib/src/sentry_transaction.dart
```diff
@@ -0,0 +1,47 @@
...
+      : super(
+          timestamp: trace.timestamp,
+          transaction: name,
+        ) {
+    _startTimestamp = trace.startTimestamp;
+
+    // tags belong to the event and not trace for transactions
+    final context = trace.context;
+    _tags.addAll(context.tags);
+    context.tags.clear();
```
marandaneto: thats just a workaround for now, Java also sets the trace.tags but its wrong

dart/lib/src/hub.dart
marandaneto: let's keep like this, similar to iOS

dart/lib/src/hub.dart
marandaneto: thats the right behaviour, drop it

dart/lib/src/protocol/contexts.dart
marandaneto: makes sense to have its own SentryTraceContext since SentrySpanContext is polluted with fields that don't belong to the `trace` object

dart/lib/src/scope.dart
marandaneto: let's not add the `getLatestActiveSpan` and always return the span bound to the scope, since the UI adjusts the chart timeline by the timestamps, all spans will be children from the root span (bound to the scope)

dart/lib/src/sentry_span_context.dart
```diff
@@ -0,0 +1,72 @@
...
+  SpanId? parentId;
+  bool? sampled;
+  late String operation;
+  String? description;
+  SpanStatus? status;
+  late Map<String, String> tags;
+
+  factory SentrySpanContext.fromJson(Map<String, dynamic> json) {
+    // assign empty id if non existent instead of null and new id
+    // missing sampled
```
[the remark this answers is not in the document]
marandaneto: we should drop and log it since its probably a bug in the serialization, such mandatory fields should not be missing

dart/lib/src/sentry_span_interface.dart
marandaneto: on iOS:
transaction.setData(x, y)
transaction.contexts.trace.data[x] = y

on .NET
transaction.setData(x, y)
transaction.extras[x] = y

Java does not have the data field at all

Dart?

dart/lib/src/sentry_span_interface.dart
```diff
@@ -0,0 +1,33 @@
...
+
+  // Future because finish calls hub.captureTransaction, should we do a Future or not?
+  Future<void> finish({
+    SpanStatus? status,
+  });
```
marandaneto: makes sense to keep it as a Future since the user can decide whether to wait or not otherwise it could lead to back pressure problems

dart/lib/src/sentry_tracer.dart
marandaneto: let's skip it for now but let's start a DACI about it

dart/lib/src/sentry_tracer.dart
marandaneto: lets finish them with `timeline_exceed` and use the timestamp from the transaction timestamp

dart/lib/src/sentry_transaction.dart
marandaneto: tranasction.contexts.trace should not have the tags property, so this is automatically solved

dart/lib/src/sentry_tracer.dart
marandaneto: typo: deadline_exceeded

dart/lib/src/sentry_span_interface.dart
marandaneto: it's done as the .NET API, if you set transaction.setData(x, y), it ends up in the extras field

PR #588, opened 2021-09-16
Tests for the Performance API

marandaneto:
## :scroll: Description
_#skip-changelog_

unit tests for the performance API

## :pencil: Checklist

- [X] All tests passing
- [X] No breaking changes

### DIFF
--- a/dart/lib/src/event_processor/deduplication_event_processor.dart
+++ b/dart/lib/src/event_processor/deduplication_event_processor.dart
@@ -63,7 +63,7 @@ class DeduplicationEventProcessor extends EventProcessor {
     // No duplication detected
     _exceptionToDeduplicate.add(exceptionHashCode);
     if (_exceptionToDeduplicate.length > _options.maxDeduplicationItems) {
-      _exceptionToDeduplicate.removeLast();
+      _exceptionToDeduplicate.removeFirst();
     }
     return event;
   }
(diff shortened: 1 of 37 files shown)


dart/lib/src/event_processor/deduplication_event_processor.dart
```diff
@@ -63,7 +63,7 @@ class DeduplicationEventProcessor extends EventProcessor {
     // No duplication detected
     _exceptionToDeduplicate.add(exceptionHashCode);
     if (_exceptionToDeduplicate.length > _options.maxDeduplicationItems) {
-      _exceptionToDeduplicate.removeLast();
+      _exceptionToDeduplicate.removeFirst();
```
marandaneto, line 66: @ueman since `add(x)` adds to the tail, we need to remove the `first` that is the oldest one, I guess its a bug, right, test only passed because the test that was doing that, didn't enable the `enableDeduplication` feature

dart/test/event_processor/deduplication_event_processor_test.dart
```diff
@@ -1,45 +1,55 @@
...
+      final sut = fixture.getSut(true);
+      final transaction = _createTransaction(fixture.hub);
+
+      expect(sut.apply(transaction), isNotNull);
+      expect(sut.apply(transaction), isNotNull);
+    });
+
     test('exceptions to keep for deduplication', () {
-      final sut = fixture.getSut(false, 2);
+      final sut = fixture.getSut(true, 2);
```
marandaneto, line 48: @ueman thats why the test passed, `enableDeduplication` was enabled

PR #600, opened 2021-09-23
Enrich Dart context with isolate name

### LINKED ISSUE #597: Enrich events with isolate name
Events should contain the name of the isolate they originate from.
This can be achieved via `Isolate.current.debugName`. The name is not guaranteed to be unique but could be helpful in some scenarios.

This is not Flutter specific and should probably be done by default in the Dart plugin.

kuhnroyal:
## :scroll: Description
Enrich Dart context with isolate name

* fixes #597 

Add a test :)

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps
none

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Unreleased
 
-* enrich Dart context with isolate name
+* enrich Dart context with isolate name (#600)
 
 # 6.1.0-alpha.1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+* enrich Dart context with isolate name
+
 # 6.1.0-alpha.1
 
 * Performance API for Dart/Flutter (#530)
--- a/dart/lib/src/enricher/io_enricher_event_processor.dart
+++ b/dart/lib/src/enricher/io_enricher_event_processor.dart
@@ -64,6 +64,7 @@ class IoEnricherEventProcessor extends EventProcessor {
   Map<String, dynamic> _getDartContext(bool includePii) {
     final args = Platform.executableArguments;
     final packageConfig = Platform.packageConfig;
+    final isolate = Isolate.current.debugName;
 
     String? executable;
     if (includePii) {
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 # Unreleased
 
+* enrich Dart context with isolate name
```
```suggestion
* enrich Dart context with isolate name (#600)
```

dart/lib/src/enricher/io_enricher_event_processor.dart
```diff
@@ -83,6 +84,7 @@ class IoEnricherEventProcessor extends EventProcessor {
     return <String, dynamic>{
       if (packageConfig != null) 'package_config': packageConfig,
       'number_of_processors': Platform.numberOfProcessors,
+      'isolate': Isolate.current.debugName,
```
marandaneto: worth testing on Flutter/Dart Web, since Isolates have a few limitations on the web

kuhnroyal: Well this is the IO processor. It is not used on the web?
But could check if we can test the same on Flutter.

kuhnroyal: Don't see how to add this to any of the flutter tests.
But `Isolate.current.debugName` works in Flutter, I use it heavily and currently have a custom processor which adds this as well.

Isolates are not supported on Web target.

marandaneto: Yep, this is IO only just wanted to check, all good then

ueman: It's nullable, so we should just add it, in case it's not null
```suggestion
      if(Isolate.current.debugName != null)
      'isolate': Isolate.current.debugName,
```

kuhnroyal: From what I have seen, it is never `null`, as it will default to the entry point name.

```dart
void foo(dynamic message) {
  print(Isolate.current.debugName);
  print(message);
}

void main() {
  Isolate.spawn(
    foo,
    'yay',
  );
}
```
```
[+1752 ms] flutter: foo
[        ] flutter: yay
```

That's why the main Flutter isolate is called `main`.

I would actually default it to something in case it actually is `null`.
Something like `Isolate.current.debugName ?? 'unknown'`?

ueman: Huh, interesting TIL. I would say, not adding it is better than setting it to `unknown` because you could actually set it to `unknown`. I don't believe anybody would actually do that, but you'll never know.

```dart
void foo(dynamic message) {
  print(Isolate.current.debugName);
  print(message);
}

void main() {
  Isolate.spawn(
    foo,
    'unknown',
  );
}
```

marandaneto (approved):
LGTM
thanks @kuhnroyal

PR #603, opened 2021-09-29
Feat: Sentry Performance for HTTP client

### LINKED ISSUE #595: Add traceHeaders support for performance API
https://develop.sentry.dev/sdk/performance/#header-sentry-trace

marandaneto:
## :scroll: Description
Feat: Sentry Performance for HTTP client

Closes https://github.com/getsentry/sentry-dart/issues/595

running and unit tests

## :pencil: Checklist

- [X] All tests passing
- [X] No breaking changes

### DIFF
--- a/dart/lib/src/http_client/breadcrumb_client.dart
+++ b/dart/lib/src/http_client/breadcrumb_client.dart
@@ -90,7 +90,6 @@ class BreadcrumbClient extends BaseClient {
     }
   }
 
-  /// See https://github.com/getsentry/sentry-dart/pull/226#discussion_r536984785
   @override
   void close() => _client.close();
 }
--- a/dart/lib/src/http_client/breadcrumb_client.dart
+++ b/dart/lib/src/http_client/breadcrumb_client.dart
@@ -90,9 +90,7 @@ class BreadcrumbClient extends BaseClient {
     }
   }
 
+  /// See https://github.com/getsentry/sentry-dart/pull/226#discussion_r536984785
   @override
-  void close() {
-    // See https://github.com/getsentry/sentry-dart/pull/226#discussion_r536984785
-    _client.close();
-  }
+  void close() => _client.close();
 }
--- a/dart/lib/src/http_client/breadcrumb_client.dart
+++ b/dart/lib/src/http_client/breadcrumb_client.dart
@@ -58,8 +58,8 @@ class BreadcrumbClient extends BaseClient {
     String? reason;
     int? responseBodySize;
 
-    final stopwatch = Stopwatch();
-    stopwatch.start();
+    // final stopwatch = Stopwatch();
+    // stopwatch.start();
 
     try {
       final response = await _client.send(request);
@@ -73,7 +73,7 @@ class BreadcrumbClient extends BaseClient {
       requestHadException = true;
       rethrow;
     } finally {
-      stopwatch.stop();
+      // stopwatch.stop();
 
       var breadcrumb = Breadcrumb.http(
         level: requestHadException ? SentryLevel.error : SentryLevel.info,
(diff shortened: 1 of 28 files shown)


dart/lib/src/http_client/sentry_http_client.dart
```diff
@@ -101,6 +103,10 @@ class SentryHttpClient extends BaseClient {
       innerClient = BreadcrumbClient(client: innerClient, hub: _hub);
     }
 
+    if (networkTracing) {
+      innerClient = TracingClient(client: innerClient, hub: _hub);
```
ueman: I believe the TracingClient should be the innermost one. So it should be above the BreadcrumClient in the code. 

Also if the BreadcrumClient is a child of the TracingClient the current request is also in the breadcrumb list of this trace which probably is not needed because the user looks at a trace for it.

dart/lib/src/http_client/tracing_client.dart
```diff
@@ -0,0 +1,46 @@
...
+
+      response = await _client.send(request);
+      span?.status = SpanStatus.fromHttpStatusCode(response.statusCode);
+    } catch (exception) {
+      span?.throwable = exception;
+      span?.status = SpanStatus.internalError();
+
+      rethrow;
+    } finally {
+      await span?.finish();
```
ueman, line 43: I wonder if this should be awaited. 

It would make the users HTTP call potentially take longer just because we're potentially sending a trace to Sentry. 

Also if the BreadcrumClient is around this it would skew the recorded time of it.

What do you think?

dart/lib/src/http_client/breadcrumb_client.dart
```diff
@@ -58,8 +58,8 @@ class BreadcrumbClient extends BaseClient {
     String? reason;
     int? responseBodySize;
 
-    final stopwatch = Stopwatch();
-    stopwatch.start();
+    // final stopwatch = Stopwatch();
```
marandaneto: should we drop such things since we have span support now?

ueman: It's in the docs as expected feature for HTTP integrations.

marandaneto: makes sense

dart/lib/src/http_client/tracing_client.dart
marandaneto: its up to the user to `await` the client get or not, I'd assume, eg

`await client.get(Uri.parse('https://flutter.dev/'));`

or `client.get(Uri.parse('https://flutter.dev/'));`, otherwise, if there's a wrapped transaction, we might finish the transaction before the HTTP call has been finished, not getting the span of it

dart/lib/src/http_client/sentry_http_client.dart
marandaneto: makes sense, just moved it, thanks

dart/lib/src/http_client/tracing_client.dart
```diff
@@ -0,0 +1,50 @@
...
+    );
+
+    StreamedResponse? response;
+    try {
+      if (span != null) {
+        final traceHeader = span.toSentryTrace();
+        request.headers[traceHeader.name] = traceHeader.value;
+      }
+
+      // TODO: tracingOrigins support
```
ueman, line 33: Maybe create an issue for this todo?

ueman (approved):
LGTM
Though there are some failing tests it seems

ueman:
Isn't there a SentryOption which decides wether tracing is enabled or not?

marandaneto:
> Isn't there a SentryOption which decides wether tracing is enabled or not?

yes, but `SentryHttpClient` nor `Hub` have access to the SentryOptions, Java exposed Hub#getOptions as workaround, we can do that maybe later, I'd not block this PR right now tho

dart/lib/src/http_client/tracing_client.dart
marandaneto: https://github.com/getsentry/sentry-dart/issues/604

PR #615, opened 2021-10-07
Set name of current route as transaction

ueman:
## :scroll: Description
This change adds the current route as `SentryEvent.transaction`.

https://develop.sentry.dev/sdk/event-payloads/ suggests to use the current route as transaction name.

Unit tests

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes


flutter/lib/src/navigation/sentry_navigator_observer.dart
```diff
@@ -35,13 +35,19 @@ const _navigationKey = 'navigation';
...
 class SentryNavigatorObserver extends RouteObserver<PageRoute<dynamic>> {
-  SentryNavigatorObserver({Hub? hub}) : hub = hub ?? HubAdapter();
+  SentryNavigatorObserver({Hub? hub, this.setRouteNameAsTransaction = true})
+      : _hub = hub ?? HubAdapter();
 
-  final Hub hub;
+  final Hub _hub;
+
+  /// Enabling this option overrides the current [Scope.transaction].
+  /// So be careful when this is used together with performance tracing.
```
```suggestion
  /// Enabling this option overrides the current [Scope.transaction].
  /// So be careful when this is used together with performance monitoring.
```

flutter/lib/src/navigation/sentry_navigator_observer.dart
```diff
@@ -35,13 +35,19 @@ const _navigationKey = 'navigation';
...
+  final bool setRouteNameAsTransaction;
```
```suggestion
  final bool _setRouteNameAsTransaction;
```

marandaneto: can we make it private?

marandaneto (approved):
left a comment other than that, LGTM

flutter/lib/src/navigation/sentry_navigator_observer.dart
ueman: No, because it's a constructor param. But because it's a final property it's read only.

marandaneto: I mean, we can do the same as `_hub` and `hub`, so we dont expose the `setRouteNameAsTransaction`

PR #618, opened 2021-10-12
Feat: `debugPrint` integration

### LINKED ISSUE #492: Don't enable `print()` breadcrumbs by default on Flutter

ueman:
## :scroll: Description
This is an integration which intercepts [`debugPrint`](https://api.flutter.dev/flutter/foundation/debugPrint.html) and adds them as breadcrumbs.
This fixes #492 by basically removing the call to print done by debugPrint.

Fixes #492

New unit tests

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes


flutter/lib/src/default_integrations.dart
```diff
@@ -484,3 +484,49 @@ class LoadReleaseIntegration extends Integration<SentryFlutterOptions> {
         .replaceAll('\u{0000}', '');
   }
 }
+
+/// Integration which intercepts Flutters [debugPrint] method.
+/// If this integration is added, all calls to [debugPrint] a redirected to
+/// add a [Breadcrumb]. [debugPrint] is not outputting to the console anymore!
+/// This integration fixes the issue described in
+/// [#492](https://github.com/getsentry/sentry-dart/issues/492).
+class DebugPrintIntegration extends Integration<SentryFlutterOptions> {
```
ueman: I wonder if this should be added by default. I don't think so because it's kinda invasive.
Though we could include it by default and just enable it in release mode.

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # Unreleased
 
 * Feat: Add current route as transaction (#615)
+* Feat: Add Breadcrumbs for Flutters `debugPrint`
```
```suggestion
* Feat: Add Breadcrumbs for Flutters `debugPrint` (#618)
```

flutter/lib/src/default_integrations.dart
```diff
@@ -484,3 +484,49 @@ class LoadReleaseIntegration extends Integration<SentryFlutterOptions> {
...
+    // This makes it impossible to just disable adding print() breadcrumbs
+    // before debugPrint is called and re-enable it after debugPrint was called.
+    // See the docs for more information.
+    // https://api.flutter.dev/flutter/foundation/debugPrint.html
+    debugPrint = _debugPrint;
+  }
+
+  @override
+  void close() {
+    debugPrint = debugPrintBackup;
```
ueman: This needs a test.

dart/lib/src/protocol/breadcrumb.dart
```diff
@@ -66,6 +66,22 @@ class Breadcrumb {
     );
   }
 
+  factory Breadcrumb.console({
```
ueman, line 69: I should add a test.

flutter/lib/src/default_integrations.dart
```diff
@@ -484,3 +484,49 @@ class LoadReleaseIntegration extends Integration<SentryFlutterOptions> {
...
   }
 }
+
+/// Integration which intercepts Flutters [debugPrint] method.
+/// If this integration is added, all calls to [debugPrint] a redirected to
+/// add a [Breadcrumb]. [debugPrint] is not outputting to the console anymore!
+/// This integration fixes the issue described in
+/// [#492](https://github.com/getsentry/sentry-dart/issues/492).
+class DebugPrintIntegration extends Integration<SentryFlutterOptions> {
+  late DebugPrintCallback debugPrintBackup;
```
```suggestion
  late DebugPrintCallback _debugPrintBackup;
```

flutter/lib/src/default_integrations.dart
marandaneto: maybe we can extract to its own class, I feel that default_integrations.dart is getting bloated

marandaneto:
@ueman does that mean that if calling `print(message)` it'll forward to our integration?

the other question is, the integration lives in the flutter package, but the `ZoneSpecification` with the `print` call is in the dart package, so it'd not work for the dart package?

should we only add this integration by default if `enablePrintBreadcrumbs` is enabled?

flutter/lib/src/default_integrations.dart
marandaneto: see https://github.com/getsentry/sentry-dart/pull/618#issuecomment-942205322

ueman:
The problem this solves is, that FlutterErrors are collected by the existing FlutterErrorIntegration and then forwarded to the default FlutterError.onError implementation. The default FlutterError.onError implementation calls `debugPrint` and `debugPrint` calls `print()`. And we also intercept calls to `print()`. The default implementation of `debugPrint` does throttling and line splitting so that we end up with https://github.com/getsentry/sentry-dart/issues/492 . Errors are collected twice, once as an event and once as possibly multiple breadcrumbs.

With this we may still end up with duplication, but it's at least well formatted and it may add breadcrumbs for logs which are done directly with `debugPrint`.

By replacing `debugPrint` with our own implementation we avoid that `print()` is called. Thats's why this shouldn't be enabled in debug builds, because that would effectivly disable most of Flutters logging.

--- 

> @ueman does that mean that if calling `print(message)` it'll forward to our integration?

No, but `debugPrint` calls are collected as breadcrumbs.

> the other question is, the integration lives in the flutter package, but the `ZoneSpecification` with the `print` call is in the dart package, so it'd not work for the dart package?

`debugPrint` is a Flutter method and can't live in the dart package.

> should we only add this integration by default if `enablePrintBreadcrumbs` is enabled?

That does make sense. Though as mentioned before I'm not sure we ever want to enable it in debug builds.

marandaneto:
@ueman 

> That does make sense. Though as mentioned before I'm not sure we ever want to enable it in debug builds.

maybe we do both guards then, only add if release build & `enablePrintBreadcrumbs` is enabled

flutter/lib/src/integrations/debug_print_integration.dart
```diff
@@ -0,0 +1,61 @@
...
+    // This makes it impossible to just disable adding print() breadcrumbs
+    // before debugPrint is called and re-enable it after debugPrint was called.
+    // See the docs for more information.
+    // https://api.flutter.dev/flutter/foundation/debugPrint.html
+    debugPrint = _debugPrint;
+  }
+
+  @override
+  void close() {
+    debugPrint = _debugPrintBackup;
```
marandaneto: worth checking if `_debugPrintBackup` isnt null tho, if `isDebug` or `!enablePrintBreadcrumbs`, `_debugPrintBackup` isnt initialized and will throw, or `debugPrint` will be null

flutter/lib/src/integrations/debug_print_integration.dart
```diff
@@ -0,0 +1,61 @@
...
+    // We're simply replacing debugPrint here. The default implementation is a
+    // a throttling system which prints using Darts print method. It's basically
+    // a fire and forget method which completes sometime in the future. We can't
+    // observe when it's done.
+    //
+    // This makes it impossible to just disable adding print() breadcrumbs
+    // before debugPrint is called and re-enable it after debugPrint was called.
+    // See the docs for more information.
+    // https://api.flutter.dev/flutter/foundation/debugPrint.html
+    debugPrint = _debugPrint;
```
marandaneto, line 40: add the integration to the `sdk` list:

`options.sdk.addIntegration('debugPrintIntegration');`

flutter/lib/src/integrations/debug_print_integration.dart
ueman: `_debugPrintBackup` is set before I check for `isDebug` or `!enablePrintBreadcrumbs` so it's not an issue. I can still add a test case for that though.

flutter/lib/src/integrations/debug_print_integration.dart
ueman: oh shoot, you're right.

flutter/lib/src/integrations/debug_print_integration.dart
marandaneto: indeed, you're right, wondering if it makes sense to only assign after checking the vars, otherwise, we are pointing to an object that we won't ever use, your call, if it's not throwing I am happy already

flutter/lib/src/integrations/debug_print_integration.dart
```diff
@@ -0,0 +1,66 @@
...
+
+  @override
+  FutureOr<void> call(Hub hub, SentryFlutterOptions options) {
+    _hub = hub;
+    _options = options;
+
+    final isDebug = options.platformChecker.isDebugMode();
+    final enablePrintBreadcrumbs = options.enablePrintBreadcrumbs;
+    if (isDebug || !enablePrintBreadcrumbs) {
+      return Future.value();
```
marandaneto, line 26: just `return` would work since its not an `async` method or?

marandaneto (approved):
thanks @ueman, looks good, LGTM

flutter/lib/src/integrations/debug_print_integration.dart
ueman: No, because a single `return;` does not return `void` :(

PR #619, opened 2021-10-18
Remove any versioning

marandaneto:
_#skip-changelog_

prefer `path` rather than `dependency_overrides` and `any` versioning


ueman (approved):
That makes sense, I recently did that for my libraries, too. So this LGTM!

Issue #577, opened 2021-09-01
Change Flutter init to provide a way to hook into the widget tree


ueman:
Currently the setup for Flutter works like this:

```dart
import 'package:flutter/widgets.dart';
import 'package:sentry_flutter/sentry_flutter.dart';

typedef AppRunner = FutureOr<void> Function();

Future<void> main() async {
  await SentryFlutter.init(
    (options) {
      options.dsn = 'https://example@sentry.io/add-your-dsn-here';
    },
    // Init your App.
    appRunner: () => runApp(MyApp()),
  );
}
```

If we change it to 

```dart
import 'package:flutter/widgets.dart';
import 'package:sentry_flutter/sentry_flutter.dart';

typedef AppRunner = FutureOr<Widget?> Function();

Future<void> main() async {
  await SentryFlutter.init(
    (options) {
      options.dsn = 'https://example@sentry.io/add-your-dsn-here';
    },
    // Init your App.
    appRunner: () => MyApp(),
  );
}

// inside SentryFlutter.init
final widget = await appRunner();
if(widget != null) {
  runApp(SentryWidget(child: widget));
}
```

we can wrap the users widget and call runApp ourself.
If the user doesn't want to do that, he can still call `runApp` himself and return null.
[`runApp`](https://api.flutter.dev/flutter/widgets/runApp.html) returns just void and is a fire and forget method, so it should work.

This would enable us to hook into the widget tree for the Screenshot feature (https://github.com/getsentry/sentry-dart/issues/556) and also for the dialog for crashes on the last run (https://github.com/getsentry/sentry-dart/issues/441)

RohanSengupta326:
I would like to work on this issue .

marandaneto:
the other option is just to use your own `runZonedGuarded`, but we'd ship in the SDK the `SentryWidget` class so its less boilerplate for people, how does that sound @ueman and @RohanSengupta326 ?

the example using a custom `runZonedGuarded` can be seen in the readme https://github.com/getsentry/sentry-dart/tree/main/flutter

ueman:
I would say that at the current state of the SDK this is not needed.
It's becomes useful as soon as the Sentry SDK should provide #556 and #441 .

Shipping a widget is also a very good way to hook into the widget tree. I think that people would actually prefer that.

@marandaneto  I'm not quite sure what you mean by using your own `runZonedGuarded`?

marandaneto:
[...]

PR #620, opened 2021-10-19
changed flutter init to provide a way to hook into the widget tree. closes issue #577

RohanSengupta326:

Screenshot 2021-10-19 at 11 39 15 AM

we can wrap the users widget and call runApp ourself.
If the user doesn't want to do that, he can still call runApp himself and return null.
runApp returns just void and is a fire and forget method, so it should work.

This would enable us to hook into the widget tree for the Screenshot feature (#556) and also for the dialog for crashes on the last run (#441)

closes issue : #577 

## :pencil: Checklist


marandaneto:
@RohanSengupta326 this is a breaking change and needs to be well thought, since we still don't ship the `SentryWidget` class, I'd rather not merge this PR which is not even compiling/addressing the issues.

PR #635, opened 2021-11-15
Fix: Remove is Enum check to support older Dart versions

### LINKED ISSUE #634: version 6.1.1 Error: 'Enum' isn't a type.
_Platform:_
- [x]  Dart
- [x]  Flutter Android or iOS
- [ ]  Flutter Web

_IDE:_
- [x]  VSCode
- [ ]  IntelliJ/AS
- [ ]  XCode
- [ ]  Other, which one?

- [ ]  Enabled
- [x]  Disabled

_Platform installed with:_
- [x] pub.dev
- [ ] GitHub

```
[√] Flutter (Channel stable, 2.2.3, on Microsoft Windows [versÃ£o 10.0.18363.1916], locale pt-BR)
    • Flutter version 2.2.3 at D:\Dev\flutter\flutter
    • Framework revision f4abaa0735 (4 months ago), 2021-07-01 12:46:11 -0700
    • Engine revision 241c87ad80
    • Dart version 2.13.4

[√] Android toolchain - develop for Android devices (Android SDK version 30.0.3)
    • Android SDK at C:\Users\Igor\AppData\Local\Android\sdk
    • Platform android-30, build-tools 30.0.3
    • Java binary at: C:\Program Files\Android\Android Studio\jre\bin\java
    • Java version OpenJDK Runtime Environment (build 1.8.0_242-release-1644-b01)

[√] Chrome - develop for the web
    • Chrome at C:\Program Files (x86)\Google\Chrome\Application\chrome.exe

[√] Android Studio (version 4.0)
    • Android Studio at C:\Program Files\Android\Android Studio
    • Flutter plugin version 48.1.2
    • Dart plugin version 193.7361
    • Java version OpenJDK Runtime Environment (build 1.8.0_242-release-1644-b01)

[√] VS Code (version 1.62.2)
    • VS Code at C:\Users\Igor\AppData\Local\Programs\Microsoft VS Code
    • Flutter extension version 3.28.0

[√] Connected device (2 available)
    • Chrome (web) • chrome • web-javascript • Google Chrome 95.0.4638.69
    • Edge (web)   • edge   • web-javascript • Microsoft Edge 95.0.1020.44

• No issues found!
```
**>=2.12.0 <3.0.0**

---
I have the following issue:

I have the following issue: yesterday, when i was trying to build my app i was getting this error, since the rest of the week everything was normal, i started to debug and see where the error was coming from, and then i found a new release of sentry, (version 6.1.1 - released 11/11/2021)

the problem is comming from sentry-6.1.1/lib/src/utils.dart:35:23

maybe something broke after the release?
```
$ flutter build apk

 Building with sound null safety 

registerResGeneratingTask is deprecated, use registerGeneratedResFolders(FileCollection)
registerResGeneratingTask is deprecated, use registerGeneratedResFolders(FileCollection)
registerResGeneratingTask is deprecated, use registerGeneratedResFolders(FileCollection)
../flutter/.pub-cache/hosted/pub.dartlang.org/sentry-6.1.1/lib/src/utils.dart:35:23: Error: 'Enum' isn't a type.
  if (nonEncodable is Enum) {
                      ^^^^
                                                                        
                                                                        
FAILURE: Build failed with an exception.

* Where:
Script 'D:\Dev\flutter\flutter\packages\flutter_tools\gradle\flutter.gradle' line: 1035

* What went wrong:
Execution failed for task ':app:compileFlutterBuildRelease'.
> Process 'command 'D:\Dev\flutter\flutter\bin\flutter.bat'' finished with non-zero exit value 1

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.

* Get more help at https://help.gradle.org

BUILD FAILED in 27s
Running Gradle task 'assembleRelease'...
Running Gradle task 'assembleRelease'... Done                      28,4s
Gradle task assembleRelease failed with exit code 1
```
the temporary solution is to downgrade to 5.1.0
**Steps to reproduce:**
run the command: flutter build apk

**Actual result:**
error
**Expected result:**
build should work

marandaneto:
## :scroll: Description
Fix: Remove is Enum check to support older Dart versions

Closes #634

## :pencil: Checklist

- [X] All tests passing
- [X] No breaking changes

## :crystal_ball: Next steps
https://github.com/getsentry/sentry-dart/issues/636

### DIFF
--- a/dart/lib/src/utils.dart
+++ b/dart/lib/src/utils.dart
@@ -18,22 +18,9 @@ String formatDateAsIso8601WithMillisPrecision(DateTime date) {
   return '${iso}Z';
 }
 
-/// borrowed from flutter diagnostics describeEnum since enum.toString() is
-/// too verbose, Dart 2.15 has enum.name
-/// https://github.com/dart-lang/sdk/issues/30021
-String _describeEnum(Object enumEntry) {
-  final description = enumEntry.toString();
-  final indexOfDot = description.indexOf('.');
-
-  return description.substring(indexOfDot + 1);
-}
-
 Object? jsonSerializationFallback(Object? nonEncodable) {
   if (nonEncodable == null) {
     return null;
-  }
-  if (nonEncodable is Enum) {
-    return _describeEnum(nonEncodable);
   }
   return nonEncodable.toString();
 }
--- a/dart/test/utils_test.dart
+++ b/dart/test/utils_test.dart
@@ -23,8 +23,8 @@ void main() {
     });
 
     test('enum returns described enum during serialization', () {
-      expect(
-          jsonSerializationFallback(SentryOrientation.landscape), 'SentryOrientation.landscape');
+      expect(jsonSerializationFallback(SentryOrientation.landscape),
+          'SentryOrientation.landscape');
     });
 
     test('null Object returns null during serialization', () {
--- a/dart/test/utils_test.dart
+++ b/dart/test/utils_test.dart
@@ -24,7 +24,7 @@ void main() {
 
     test('enum returns described enum during serialization', () {
       expect(
-          jsonSerializationFallback(SentryOrientation.landscape), 'landscape');
+          jsonSerializationFallback(SentryOrientation.landscape), 'SentryOrientation.landscape');
     });
 
     test('null Object returns null during serialization', () {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+* Fix: Remove is Enum check to support older Dart versions (#)
+
 # 6.1.1
 
 * Fix: Transaction serialization if not encodable (#633)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Unreleased
 
-* Fix: Remove is Enum check to support older Dart versions (#)
+* Fix: Remove is Enum check to support older Dart versions (#635)
 
 # 6.1.1
 
--- a/dart/pubspec.yaml
+++ b/dart/pubspec.yaml
@@ -6,6 +6,8 @@ description: >
 homepage: https://docs.sentry.io/platforms/dart/
 repository: https://github.com/getsentry/sentry-dart
 
+# when bumping the min. sdk version, upgrade the e2e_test project too, it should compile using
+# the min. supported sdk version to check back compatibility.
 environment:
   sdk: '>=2.12.0 <3.0.0'
 
--- a/dart/pubspec.yaml
+++ b/dart/pubspec.yaml
@@ -6,8 +6,6 @@ description: >
 homepage: https://docs.sentry.io/platforms/dart/
 repository: https://github.com/getsentry/sentry-dart
 
-# when bumping the min. sdk version, upgrade the e2e_test project too, it should compile using
-# the min. supported sdk version to check back compatibility.
 environment:
   sdk: '>=2.12.0 <3.0.0'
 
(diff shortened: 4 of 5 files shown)


dart/lib/src/utils.dart
```diff
@@ -18,22 +18,9 @@ String formatDateAsIso8601WithMillisPrecision(DateTime date) {
...
-  final indexOfDot = description.indexOf('.');
-
-  return description.substring(indexOfDot + 1);
-}
-
 Object? jsonSerializationFallback(Object? nonEncodable) {
   if (nonEncodable == null) {
     return null;
   }
-  if (nonEncodable is Enum) {
```
marandaneto, line 35: didn't find a trustable way to identify if it's an enum using Dart < 2.14

dart/test/utils_test.dart
```diff
@@ -23,8 +23,8 @@ void main() {
     });
 
     test('enum returns described enum during serialization', () {
-      expect(
-          jsonSerializationFallback(SentryOrientation.landscape), 'landscape');
+      expect(jsonSerializationFallback(SentryOrientation.landscape),
+          'SentryOrientation.landscape');
```
marandaneto, line 27: for known SDK types this is not a problem
See https://github.com/getsentry/sentry-dart/blob/main/dart/lib/src/protocol/sentry_device.dart#L178-L194
this is only for unknown types set dynamically by the user.

kuhnroyal:
Can we add some way for the user to specify the fallback themselves?

marandaneto:
> Can we add some way for the user to specify the fallback themselves?

since you are setting types that the SDK don't know during ser/deser, instead of setting the object itself, you could simply set the serialized version already.
eg
`event.contexts['myEnumType'] = describeEnum(myEnum);`
or mutating the event with your own description of the object type during beforeSend https://docs.sentry.io/platforms/dart/configuration/filtering/

dart/lib/src/utils.dart
philipphofmann: Can't we maybe still use toString somehow as a fallback?

bruno-garcia (approved):
No way to ifdef?

philipphofmann (commented):
It's hard for me to evaluate if this change is fine as I have no experience with Dart. The only thing I can point out is that we should add tests for this if possible. If there are tests, why didn't we have to change them?

marandaneto:
> No way to ifdef?

no afaik https://github.com/dart-lang/sdk/issues/7238

dart/lib/src/utils.dart
marandaneto: the line below does it

marandaneto:
> It's hard for me to evaluate if this change is fine as I have no experience with Dart. The only thing I can point out is that we should add tests for this if possible. If there are tests, why didn't we have to change them?

tests were modified, see utils_test.dart

philipphofmann (approved):
LGTM 🚀

ueman:
> Can we add some way for the user to specify the fallback themselves?

That should be pretty straight forward to implement, in the SDK. 
I think `jsonSerializationFallback` is currently a top level function somewhere but it could be changed to a property of `SentryOptions` and be customizable.

marandaneto:
> > Can we add some way for the user to specify the fallback themselves?
> 
> That should be pretty straight forward to implement, in the SDK. I think `jsonSerializationFallback` is currently a top level function somewhere but it could be changed to a property of `SentryOptions` and be customizable.

it'd be easy, but not sure if it's worth it, there are many ways of solving it on client-side only, another one would be overwriting `toString()` plus the ones mentioned above https://github.com/getsentry/sentry-dart/pull/635#issuecomment-968694292

kuhnroyal:
I agree, it's fine to handle manually on my side. Just need to know.

PR #641, opened 2021-11-23
Log errors captured in integrations with SentryLogger

### LINKED ISSUE #640: Consider making Dart stack trace available in events
I want to log some events to console in the `beforeSend` callback (for example from the `runZonedGuarded` integration) but the original `StackTrace` is not available and I can't find a good way to log the `SentryStackTrace`.

Does this make sense?

kuhnroyal:
This is just a very quick PR for discussion.

## :scroll: Description
* add logger name parameter
* log errors from Zone, Isolate and Flutter integrations

Fixes https://github.com/getsentry/sentry-dart/issues/640

## :pencil: Checklist

- [ ] All tests passing
- [ ] No breaking changes

### DIFF
--- a/dart/lib/src/isolate_error_integration.dart
+++ b/dart/lib/src/isolate_error_integration.dart
@@ -31,7 +31,7 @@ class IsolateErrorIntegration extends Integration {
 
 RawReceivePort _createPort(Hub hub, SentryOptions options) {
   return RawReceivePort(
-    (Object? error) async {
+    (dynamic error) async {
       await handleIsolateError(hub, options, error);
     },
   );
@@ -42,14 +42,20 @@ RawReceivePort _createPort(Hub hub, SentryOptions options) {
 Future<void> handleIsolateError(
   Hub hub,
   SentryOptions options,
-  Object? error,
+  dynamic error,
 ) async {
   options.logger(SentryLevel.debug, 'Capture from IsolateError $error');
 
   // https://api.dartlang.org/stable/2.7.0/dart-isolate/Isolate/addErrorListener.html
   // error is a list of 2 elements
   if (error is List && error.length == 2) {
-    final Object? throwable = error.first;
+    /// The errors are sent back as two-element lists.
+    /// The first element is a `String` representation of the error, usually
+    /// created by calling `toString` on the error.
+    /// The second element is a `String` representation of an accompanying
+    /// stack trace, or `null` if no stack trace was provided.
+    /// To convert this back to a [StackTrace] object, use [StackTrace.fromString].
+    final String throwable = error.first;
     final String? stackTrace = error.last;
 
     options.logger(
--- a/dart/lib/src/isolate_error_integration.dart
+++ b/dart/lib/src/isolate_error_integration.dart
@@ -52,6 +52,14 @@ Future<void> handleIsolateError(
     final dynamic throwable = error.first;
     final dynamic stackTrace = error.last;
 
+    options.logger(
+      SentryLevel.error,
+      'Uncaught isolate error',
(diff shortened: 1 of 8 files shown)


dart/lib/src/isolate_error_integration.dart
```diff
@@ -52,6 +52,14 @@ Future<void> handleIsolateError(
     final dynamic throwable = error.first;
     final dynamic stackTrace = error.last;
 
+    options.logger(
+      SentryLevel.error,
+      'Uncaught isolate error',
+      logger: 'sentry.isolateError',
+      exception: throwable,
+      stackTrace: StackTrace.fromString(stackTrace),
```
marandaneto: can this throw? if `stackTrace` is null or it cant be a `StackTrace`, or any other reason? otherwise we need to wrap up in a try catch, maybe this is an instance of `StackTrace` already

kuhnroyal: Yea this needs to be checked for null but IIRC this is a String because it has to be sent across isolates.

marandaneto: yep, I'd then check all the cases, if it's null, if it's a StackTrace already, and do the `StackTrace.fromString` last case, also worth checking if `fromString` does not throw with an invalid string that is not a stack trace.

kuhnroyal: From the isolate error handler doc:
>   /// The second element is a `String` representation of an accompanying
  /// stack trace, or `null` if no stack trace was provided.
  /// To convert this back to a [StackTrace] object, use [StackTrace.fromString].

kuhnroyal:
The sentry_flutter builds seem to be failing because they don't use the correct dart dependency.

dart/lib/src/isolate_error_integration.dart
```diff
@@ -31,7 +31,7 @@ class IsolateErrorIntegration extends Integration {
 
 RawReceivePort _createPort(Hub hub, SentryOptions options) {
   return RawReceivePort(
-    (dynamic error) async {
+    (Object? error) async {
```
marandaneto: I'd rather rollback those changes because of https://github.com/getsentry/sentry-dart/issues/511
so ideally we'd do it all together and go thru pre-releases to avoid errors

kuhnroyal: I reverted all the dynamic changes except for the list types.

marandaneto:
@kuhnroyal CI fails ->

> ERROR: ../lib/src/default_integrations.dart:77:11: Error: No named parameter with the name 'logger'.
> ERROR:           logger: 'sentry.flutterError',

flutter/example/pubspec.yaml
```diff
@@ -11,15 +11,21 @@ environment:
 dependencies:
   flutter:
     sdk: flutter
+  sentry:
   sentry_flutter:
```
marandaneto, line 15: instead of removing the relative path, I'd rather add for `sentry` too, I recall having issues when using `dependency_overrides` and using `pana`, so we keep the same as every other example too.

kuhnroyal: That doesn't work because `sentry_flutter` pulls `sentry: ^x.x.x` - that results in a dependency conflict if it is specified as path dependency. It needs to be an override.

Also pana ist not really relevant for an app since it is not being published to `pub.dev` or am I missing something ?

marandaneto: yeah, I didnt run the dependency conflict before so its fine.

marandaneto:
@kuhnroyal add changelog entry and its good to go, thanks for doing this

PR #664, opened 2021-12-14
Fix: ConcurrentModificationError in when finishing span

### LINKED ISSUE #662: `ConcurrentModificationError` in when finishing span
I have a lot of reports of a `ConcurrentModificationError` coming from this for loop:
https://github.com/getsentry/sentry-dart/blob/58de5ac0943ec9bd5e15a4fdcd7a4b5ae9411585/dart/lib/src/sentry_tracer.dart#L31-L35

I believe this happens when a new span starts before the finish method is done.
It is hard to reproduce manually, but a fix I see is to make a copy of children before starting to iterate over them.

marandaneto:
## :scroll: Description
Fix: ConcurrentModificationError in when finishing span

Closes https://github.com/getsentry/sentry-dart/issues/662

## :pencil: Checklist

- [X] All tests passing
- [X] No breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+* Fix: ConcurrentModificationError in when finishing span #
+
 # 6.2.1
 
 * Fix: `sentry_logging` works now on web (#660)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Unreleased
 
-* Fix: ConcurrentModificationError in when finishing span #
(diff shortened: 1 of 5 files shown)


bruno-garcia (approved):
Lgtm but whst abour adding a diagnostics warn log

ueman:
Also, a test would be good

marandaneto:
> Also, a test would be good

yep, will come.
edit @ueman done

marandaneto:
> Lgtm but whst abour adding a diagnostics warn log

right now we don't have access to the log in this class, will check if its easy, good idea.
edit: will follow up with another PR so we don't block the bugfix release

PR #676, opened 2021-12-22
Allow to set startTimestamp & endTimestamp manually to SentrySpan

fatihergin:
## :scroll: Description
Allows to set `startTimestamp`(while creating) and `endTimestamp`(while finishing) manually to `SentrySpan`

- It will allow to trace events which Sentry does not have control over starting point of that events. 

	e.g. When a 3rd party SDK returns only duration of an internal execution which is wanted to trace.
	
    p.s. It's already implemented (`startTimestamp`) on [sentry-java](https://github.com/getsentry/sentry-java/blob/3ec06985c47758ec9cf8d33592349bd5137d5620/sentry/src/main/java/io/sentry/Sentry.java#L661) with the motivation of tracing app-start time where Sentry not initialized yet

- It will give more control over transactions in terms of creating&finishing them.

	It might be desired to decide whether to start a transaction or not conditionally which will be clear at a later time after starting the operation.

Unit tests 

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+* Feat: Allow to set startTimestamp & endTimestamp manually to SentrySpan (#676)
 * Bump: Sentry-Cocoa to 7.10.0 (#777)
 * Feat: Additional Dart/Flutter context information (#778)
 * Bump: Kotlin plugin to 1.5.31 (#763)
--- a/dart/lib/src/noop_sentry_span.dart
+++ b/dart/lib/src/noop_sentry_span.dart
@@ -41,7 +41,11 @@ class NoOpSentrySpan extends ISentrySpan {
   void setTag(String key, String value) {}
 
   @override
-  ISentrySpan startChild(String operation, {String? description}) =>
+  ISentrySpan startChild(
+    String operation, {
+    String? description,
+    DateTime? startTimestamp,
+  }) =>
       NoOpSentrySpan();
 
   @override
--- a/dart/lib/src/noop_sentry_span.dart
+++ b/dart/lib/src/noop_sentry_span.dart
@@ -26,7 +26,7 @@ class NoOpSentrySpan extends ISentrySpan {
   }
 
   @override
-  Future<void> finish({SpanStatus? status}) async {}
+  Future<void> finish({SpanStatus? status, DateTime? endTimestamp}) async {}
 
   @override
   void removeData(String key) {}
--- a/dart/lib/src/protocol/sentry_span.dart
+++ b/dart/lib/src/protocol/sentry_span.dart
@@ -9,7 +9,7 @@ import '../utils.dart';
 
 class SentrySpan extends ISentrySpan {
   final SentrySpanContext _context;
-  DateTime? _timestamp;
+  DateTime? _endTimestamp;
   late final DateTime _startTimestamp;
   final Hub _hub;
 
@@ -19,7 +19,7 @@ class SentrySpan extends ISentrySpan {
 
   SpanStatus? _status;
   final Map<String, String> _tags = {};
-  void Function()? _finishedCallback;
+  void Function({DateTime? endTimestamp})? _finishedCallback;
 
   @override
   bool? sampled;
@@ -30,7 +30,7 @@ class SentrySpan extends ISentrySpan {
     this._hub, {
     DateTime? startTimestamp,
     bool? sampled,
-    Function()? finishedCallback,
+    Function({DateTime? endTimestamp})? finishedCallback,
   }) {
     _startTimestamp = startTimestamp?.toUtc() ?? getUtcDateTime();
     this.sampled = sampled;
@@ -43,17 +43,28 @@ class SentrySpan extends ISentrySpan {
       return;
     }
 
+    final utcDateTime = getUtcDateTime();
+
     if (status != null) {
       _status = status;
     }
-    _timestamp = endTimestamp ?? getUtcDateTime();
+
+    if (endTimestamp?.isBefore(_startTimestamp) ?? false) {
+      _hub.options.logger(
+        SentryLevel.warning,
+        'End timestamp ($endTimestamp) cannot be before start timestamp ($_startTimestamp)',
+      );
+      _endTimestamp = utcDateTime;
+    } else {
+      _endTimestamp = endTimestamp?.toUtc() ?? utcDateTime;
+    }
 
     // associate error
     if (_throwable != null) {
       _hub.setSpanContext(_throwable, this, _tracer.name);
     }
-    _finishedCallback?.call();
-    await super.finish(status: status);
+    _finishedCallback?.call(endTimestamp: _endTimestamp);
+    await super.finish(status: status, endTimestamp: _endTimestamp);
   }
 
   @override
@@ -128,7 +139,7 @@ class SentrySpan extends ISentrySpan {
   DateTime get startTimestamp => _startTimestamp;
 
   @override
-  DateTime? get endTimestamp => _timestamp;
+  DateTime? get endTimestamp => _endTimestamp;
 
   @override
   SentrySpanContext get context => _context;
@@ -137,8 +148,9 @@ class SentrySpan extends ISentrySpan {
     final json = _context.toJson();
     json['start_timestamp'] =
         formatDateAsIso8601WithMillisPrecision(_startTimestamp);
-    if (_timestamp != null) {
-      json['timestamp'] = formatDateAsIso8601WithMillisPrecision(_timestamp!);
+    if (_endTimestamp != null) {
+      json['timestamp'] =
+          formatDateAsIso8601WithMillisPrecision(_endTimestamp!);
     }
     if (_data.isNotEmpty) {
       json['data'] = _data;
@@ -153,7 +165,7 @@ class SentrySpan extends ISentrySpan {
   }
 
   @override
-  bool get finished => _timestamp != null;
+  bool get finished => _endTimestamp != null;
 
   @override
   dynamic get throwable => _throwable;
--- a/dart/lib/src/protocol/sentry_span.dart
+++ b/dart/lib/src/protocol/sentry_span.dart
@@ -10,7 +10,7 @@ import '../utils.dart';
 class SentrySpan extends ISentrySpan {
   final SentrySpanContext _context;
   DateTime? _timestamp;
-  final DateTime _startTimestamp = getUtcDateTime();
+  late final DateTime _startTimestamp;
   final Hub _hub;
 
   final SentryTracer _tracer;
@@ -28,9 +28,11 @@ class SentrySpan extends ISentrySpan {
     this._tracer,
     this._context,
     this._hub, {
+    DateTime? startTimestamp,
     bool? sampled,
     Function()? finishedCallback,
   }) {
+    _startTimestamp = startTimestamp?.toUtc() ?? getUtcDateTime();
     this.sampled = sampled;
     _finishedCallback = finishedCallback;
   }
@@ -94,8 +96,17 @@ class SentrySpan extends ISentrySpan {
   ISentrySpan startChild(
     String operation, {
     String? description,
+    DateTime? startTimestamp,
   }) {
     if (finished) {
+      return NoOpSentrySpan();
+    }
+
+    if (startTimestamp?.isBefore(_startTimestamp) ?? false) {
+      _hub.options.logger(
+        SentryLevel.warning,
+        "Start timestamp ($startTimestamp) cannot be before parent span's start timestamp ($_startTimestamp). Returning NoOpSpan.",
+      );
       return NoOpSentrySpan();
     }
 
@@ -103,6 +114,7 @@ class SentrySpan extends ISentrySpan {
       _context.spanId,
       operation,
       description: description,
+      startTimestamp: startTimestamp,
     );
   }
 
--- a/dart/lib/src/sentry_tracer.dart
+++ b/dart/lib/src/sentry_tracer.dart
@@ -55,21 +55,27 @@ class SentryTracer extends ISentrySpan {
   }
 
   @override
-  Future<void> finish({SpanStatus? status}) async {
+  Future<void> finish({SpanStatus? status, DateTime? endTimestamp}) async {
+    final commonEndTimestamp = endTimestamp ?? getUtcDateTime();
     _autoFinishAfterTimer?.cancel();
     _finishStatus = SentryTracerFinishStatus.finishing(status);
     if (!_rootSpan.finished &&
         (!_waitForChildren || _haveAllChildrenFinished())) {
       _rootSpan.status ??= status;
 
-      // finish unfinished spans otherwise transaction gets dropped
-      for (final span in _children) {
-        if (!span.finished) {
-          await span.finish(status: SpanStatus.deadlineExceeded());
-        }
-      }
+      // remove span where its endTimestamp is before startTimestamp
+      _children.removeWhere(
+          (span) => !_hasSpanSuitableTimestamps(span, commonEndTimestamp));
 
-      var _rootEndTimestamp = getUtcDateTime();
+      // finish unfinished spans otherwise transaction gets dropped
+      final spansToBeFinished = _children.where((span) => !span.finished);
+      await Future.forEach(
+          spansToBeFinished,
+          (SentrySpan span) async => await span.finish(
+              status: SpanStatus.deadlineExceeded(),
+              endTimestamp: commonEndTimestamp));
+
+      var _rootEndTimestamp = commonEndTimestamp;
       if (_trimEnd && children.isNotEmpty) {
         final childEndTimestamps = children
             .where((child) => child.endTimestamp != null)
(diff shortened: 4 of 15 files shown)


vaind:
Any thoughts/issues with this @marandaneto? We need this (as a sentry.io Business subscriber, if that makes any difference here :)

dart/lib/src/protocol/sentry_span.dart
```diff
@@ -28,30 +28,39 @@ class SentrySpan extends ISentrySpan {
...
     }
 
     if (status != null) {
       _status = status;
     }
-    _timestamp = getUtcDateTime();
+
+    // if endTimestamp is before startTimestamp, set throwable
+    if (endTimestamp?.isBefore(_startTimestamp) ?? false) {
+      throwable = StateError('endTimestamp cannot be before startTimestamp');
```
bruno-garcia: This throwable exists to help us connect the transaction with the capture of the trowable. So in Sentry you can see in the issue details page the information about a transaction. And from the transaction, you can see an error occurred  and navigate to the issue detail. Here we better log out the issue so when the SDK is in debug mode we can see what's going on:


Whatever equivalent we have here:

```suggestion
      this._logger.logError('endTimestamp cannot be before startTimestamp');
```

ueman: I believe `sentry_span` doesn't have access to the logger, but yeah, logging would be the correct solution.

fatihergin: @ueman is right. `sentry_span` has no access to the logger. 
Is it acceptable to implement `_options`'s getter on the [Hub.dart](https://github.com/getsentry/sentry-dart/blob/4eb2218eefb9bf74ee5b80efa64f1f395c580cb8/dart/lib/src/hub.dart)? Then we can use it like
```dart
_hub.options.logger(SentryLevel.error, 'endTimestamp cannot be before startTimestamp');
```

ueman: This is a question for @marandaneto, @denrase or @brustolin . It's probably useful because there are a couple other places where logging isn't available but where it could be useful.

brustolin: For cocoa we can access options from the client property in the hub. I don't see any problem with having and option getter in the hub, but Im also not aware of all implications that this could cause for Dart SDK.

marandaneto: https://github.com/getsentry/sentry-dart/pull/716 is exposing the `options` via the `Hub` for internal-only usage.
The problem is that the options should be read-only, that's why it can't be public or misused.

fatihergin: The PR is rebased to `main` and updated to use `_hub.options.logger` for this conversation.
I guess it's ready to review again @marandaneto, @denrase , @brustolin.

dart/lib/src/protocol/sentry_span.dart
```diff
@@ -44,14 +46,23 @@ class SentrySpan extends ISentrySpan {
...
       _status = status;
     }
-    _timestamp = endTimestamp ?? getUtcDateTime();
+
+    if (endTimestamp?.isBefore(_startTimestamp) ?? false) {
+      _hub.options.logger(
+        SentryLevel.warning,
+        'End timestamp ($endTimestamp) cannot be before start timestamp ($_startTimestamp)',
+      );
+      _endTimestamp = getUtcDateTime();
```
marandaneto: This is a good idea, but `getUtcDateTime()` can still be technically before the given `_startTimestamp`.
This needs to be tried/tested and see if Sentry.io throws an ingestion error or just an UI warning, otherwise we have to fix on the SDK side.

marandaneto:
> Any thoughts/issues with this @marandaneto? We need this (as a sentry.io Business subscriber, if that makes any difference here :)

Sorry late reply, I was on vacay back then and missed the comment.

marandaneto:
@fatihergin since there were changes in the public API, please check the other packages within this repo, they might need to be fixed too, e.g. `Hub`.

marandaneto:
@brustolin would you mind having a look at it?

dart/lib/src/protocol/sentry_span.dart
```diff
@@ -44,14 +46,23 @@ class SentrySpan extends ISentrySpan {
     if (status != null) {
       _status = status;
     }
-    _timestamp = endTimestamp ?? getUtcDateTime();
+
+    if (endTimestamp?.isBefore(_startTimestamp) ?? false) {
```
marandaneto, line 52:
```suggestion
    if (endTimestamp?.toUtc()?.isBefore(_startTimestamp) ?? false) {
```

I guess this is needed?

dart/lib/src/protocol/sentry_span.dart
```diff
@@ -44,14 +46,23 @@ class SentrySpan extends ISentrySpan {
...
+      _hub.options.logger(
+        SentryLevel.warning,
+        'End timestamp ($endTimestamp) cannot be before start timestamp ($_startTimestamp)',
+      );
+      _endTimestamp = getUtcDateTime();
+    } else {
+      _endTimestamp = endTimestamp?.toUtc() ?? getUtcDateTime();
```
marandaneto: `final newTimeStamp = endTimestamp?.toUtc() ?? getUtcDateTime();`

This could be done before and here you just assign the value to the `_endTimestamp` var so we have the `endTimestamp` var computed as early as possible in the finish method call.

dart/lib/src/protocol/sentry_span.dart
marandaneto: `_startTimestamp` is already UTC at this point, right?

dart/lib/src/protocol/sentry_span.dart
```diff
@@ -94,15 +105,25 @@ class SentrySpan extends ISentrySpan {
   ISentrySpan startChild(
     String operation, {
     String? description,
+    DateTime? startTimestamp,
   }) {
     if (finished) {
       return NoOpSentrySpan();
     }
 
+    if (startTimestamp?.isBefore(_startTimestamp) ?? false) {
```
marandaneto, line 116: Same issue from https://github.com/getsentry/sentry-dart/pull/676/files#r800678030
are `startTimestamp` and `_startTimestamp` already or not?

dart/lib/src/protocol/sentry_span.dart
fatihergin: I wasn't sure about that and just checked it now, and date comparisons ([isAtSameMomentAs](https://api.dart.dev/stable/2.16.0/dart-core/DateTime/isAtSameMomentAs.html), [isBefore](https://api.dart.dev/stable/2.16.0/dart-core/DateTime/isBefore.html), [isAfter](https://api.dart.dev/stable/2.16.0/dart-core/DateTime/isAfter.html) etc.) look time zone independent.

For `isBefore`:
> Returns true if [this](https://api.dart.dev/stable/2.16.0/dart-core/DateTime-class.html) occurs before other.
> The comparison is independent of whether the time is in UTC or in the local time zone.

fatihergin: > 

Yes, all inputs are converted to UTC before assigning. See [sentry_span_test.dart](https://github.com/fatihergin/sentry-dart/blob/feat/allow-to-set-sentryspan-starttimestamp-explicitly/dart/test/sentry_span_test.dart#L11).

fatihergin:
> @fatihergin since there were changes in the public API, please check the other packages within this repo, they might need to be fixed too, e.g. `Hub`.

I think I did it for `hub`, `hub_adapter`, `noop_hub` and `hub_test`. It looks there is no other relation between SentrySpan's start & end timestamps and the other packages @marandaneto

dart/lib/src/protocol/sentry_span.dart
fatihergin: Replied on [the first occurrence](https://github.com/getsentry/sentry-dart/pull/676#discussion_r800695134)

marandaneto:
> > @fatihergin since there were changes in the public API, please check the other packages within this repo, they might need to be fixed too, e.g. `Hub`.
> 
> I think I did it for `hub`, `hub_adapter`, `noop_hub` and `hub_test`. It looks there is no other relation between SentrySpan's start & end timestamps and the other packages @marandaneto

There are untouched packages by this PR, e.g. [dio](https://github.com/getsentry/sentry-dart/tree/main/dio) and [logging](https://github.com/getsentry/sentry-dart/tree/main/logging)
Those also have mocks of the `Hub` class and that's why it fails.
See the [CI Checks](https://github.com/getsentry/sentry-dart/pull/676/checks) for reference.

dart/lib/src/noop_sentry_span.dart
```diff
@@ -26,7 +26,7 @@ class NoOpSentrySpan extends ISentrySpan {
   }
 
   @override
-  Future<void> finish({SpanStatus? status}) async {}
+  Future<void> finish({SpanStatus? status, DateTime? endTimestamp}) async {}
```
marandaneto, line 29: Wondering if it makes sense to be `finishTimestamp`, the method is called `finish` and not `end`.
Opinions @brustolin ?

dart/lib/src/protocol/sentry_span.dart
fatihergin:
``` dart
#50    if (endTimestamp?.isBefore(_startTimestamp) ?? false) {
#51      _hub.options.logger(some warning...);
#52      _endTimestamp = getUtcDateTime();
#53    } else {
#54      _endTimestamp = endTimestamp?.toUtc() ?? getUtcDateTime();
#55    }`
```
If it's assigned earlier than the if statement(#50) and if condition returns true, we can't know the exact reason behind the logged warning. Caller method may have passed an invalid date time, or our early assignment may have assigned [an invalid date time](https://github.com/getsentry/sentry-dart/pull/676#discussion_r800667788). Should I proceed with it?

ueman:
> > > @fatihergin since there were changes in the public API, please check the other packages within this repo, they might need to be fixed too, e.g. `Hub`.
> > 
> > 
> > I think I did it for `hub`, `hub_adapter`, `noop_hub` and `hub_test`. It looks there is no other relation between SentrySpan's start & end timestamps and the other packages @marandaneto
> 
> There are untouched packages by this PR, e.g. [dio](https://github.com/getsentry/sentry-dart/tree/main/dio) and [logging](https://github.com/getsentry/sentry-dart/tree/main/logging) Those also have mocks of the `Hub` class and that's why it fails. See the [CI Checks](https://github.com/getsentry/sentry-dart/pull/676/checks) for reference.

That makes me wonder if those mocks should only override what's important to them and then just override `noSuchMethod` in order to ignore all other methods.

marandaneto:
> > > > @fatihergin since there were changes in the public API, please check the other packages within this repo, they might need to be fixed too, e.g. `Hub`.
> > > 
> > > 
> > > I think I did it for `hub`, `hub_adapter`, `noop_hub` and `hub_test`. It looks there is no other relation between SentrySpan's start & end timestamps and the other packages @marandaneto
> > 
> > 
> > There are untouched packages by this PR, e.g. [dio](https://github.com/getsentry/sentry-dart/tree/main/dio?rgh-link-date=2022-02-07T14%3A53%3A54Z) and [logging](https://github.com/getsentry/sentry-dart/tree/main/logging?rgh-link-date=2022-02-07T14%3A53%3A54Z) Those also have mocks of the `Hub` class and that's why it fails. See the [CI Checks](https://github.com/getsentry/sentry-dart/pull/676/checks) for reference.
> 
> That makes me wonder if those mocks should only override what's important to them and then just override `noSuchMethod` in order to ignore all other methods.

yep, that makes a lot of sense.

dart/lib/src/protocol/sentry_span.dart
marandaneto: You'd not assign to `_endTimestamp`, rather to a new var e.g. `newTimeStamp` as my code snippet and only later assign to `_endTimestamp` depending on the `if` condition result.
The idea is just to compute `getUtcDateTime()` as early as possible.
Just an early `final utcDateTime = getUtcDateTime()` is already good enough instead of calling multiple times `getUtcDateTime()` later.

dart/lib/src/noop_sentry_span.dart
brustolin: Well, the property being manipulated by this parameter is called `timestamp`, maybe only timestamp would be enough. 

But I don't have strong opinions about parameter naming. Either option sounds good to me. The purpose of the parameter is clear (at least for me, it might be because I'm used to the SDK).

dart/lib/src/sentry_tracer.dart
```diff
@@ -63,11 +65,13 @@ class SentryTracer extends ISentrySpan {
       // finish unfinished spans otherwise transaction gets dropped
       for (final span in _children) {
         if (!span.finished) {
-          await span.finish(status: SpanStatus.deadlineExceeded());
+          await span.finish(
+              status: SpanStatus.deadlineExceeded(),
+              endTimestamp: endTimestamp);
```
brustolin: m: I believe we should have an auxiliary variable at the beginning of this function to hold the end value ( `endTimestamp ?? getUtcDateTime()` ).
This way, every unfinished span will have the same endTimestamp as well the `_rootEndTimestamp`

marandaneto:
Quick update https://github.com/getsentry/develop/pull/508
Our docs now mention that passing a custom `endTimestamp` is fine, so we can move forward with this PR.
@fatihergin would you like to address the comments? thanks.

dart/lib/src/noop_sentry_span.dart
fatihergin: Since there is now one more time stamp, `startTimestamp`, I renamed timestamp as `endTimestamp` in this PR. Just like PHP and JS repo did.
And endTimestamp is already in use in the [same class](https://github.com/getsentry/sentry-dart/blob/7a39898ffa64fd4b00b0e7e5f0337ce9e22ab382/dart/lib/src/protocol/sentry_span.dart#L119):
```dart
#118  @override
#119  DateTime? get endTimestamp => _timestamp;
```

I can revert it to timestamp and rename the method parameter to the same name if it make sense.

dart/lib/src/protocol/sentry_span.dart
fatihergin: AFAIU, Sentry JS simply [discards spans if end timestamp is before start timestamp](https://github.com/getsentry/sentry-javascript/blob/889024a8b9da0664447f157eacaaf8713d6cc7d4/packages/tracing/src/idletransaction.ts#L131). 
Should this approach be applicable here?

```dart
      this.spanRecorder.spans = this.spanRecorder.spans.filter((span: Span) => {
        // If we are dealing with the transaction itself, we just return it
        if (span.spanId === this.spanId) {
          return true;
        }
         ...
        const keepSpan = span.startTimestamp < endTimestamp;
        if (!keepSpan) {
          logger.log(
            '[Tracing] discarding Span since it happened after Transaction was finished',
            JSON.stringify(span, undefined, 2),
          );
        }
        return keepSpan;
      });

```

marandaneto: Yes, otherwise it'll be discarded in the event ingestion, so no need to send the data over.

dart/lib/src/noop_sentry_span.dart
marandaneto: `endTimestamp` is the correct one, it's also part of our Unified API https://develop.sentry.dev/sdk/performance/
The serialization name should not change tho, since it's part of our protocol during event ingestion.

fatihergin: `endTimestamp` is the current one, and the serialization name is left [as is](https://github.com/getsentry/sentry-dart/blob/bce00c31543b75ba9b017a6bad5df4d784af9ead/dart/lib/src/protocol/sentry_span.dart#L151).

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@
 
 * Fix: Disable log by default in debug mode (#753)
 * [Dio] Ref: Replace FailedRequestAdapter with FailedRequestInterceptor (#728)
+* Feat: Allow to set startTimestamp & endTimestamp manually to SentrySpan (#676)
```
marandaneto: This should go under the Unreleased entry

dart/lib/src/protocol/sentry_span.dart
fatihergin: I can't find a suitable place to filter invalid spans out. 
Possible places: 
1. in Hub.captureTransaction, before [item.client.captureTransaction(transaction)](https://github.com/getsentry/sentry-dart/blob/c25be502cb39fcc5699bdf241c20da4f14856519/dart/lib/src/hub.dart#L463): can't modify transaction
2. in SentryTracer.finish, before [_hub.captureTransaction(transaction)](https://github.com/getsentry/sentry-dart/blob/c25be502cb39fcc5699bdf241c20da4f14856519/dart/lib/src/sentry_tracer.dart#L95): can't modify transaction
3. in constructor of SentryTransaction, while initializing spans: [spans = _tracer.children](https://github.com/getsentry/sentry-dart/blob/c25be502cb39fcc5699bdf241c20da4f14856519/dart/lib/src/protocol/sentry_transaction.dart#L55).where(...): 
#(3) might work, but smells. Any recommendation?
+ @marandaneto

marandaneto:
@fatihergin only missing https://github.com/getsentry/sentry-dart/pull/676#discussion_r820451757 and fix the changelog entry, otherwise LGTM.

dart/lib/src/protocol/sentry_span.dart
marandaneto: @fatihergin I believe `SentryTracer#finish` is the way to go, within the `for` loop from L67-L73, just remove the child span from the `_children` list if the timestamp is out of bounds.

marandaneto (approved):

Issue #682, opened 2021-12-25
(PR created) Allow attachments to be bound to exception more easily


fzyzcjy:
## Why I need this (very much!)

My code is like:

```dart
void main_function() {
  try {
    inner_function(); // indeed, functions calling functions calling functions... so the calling stack is deep
  } catch (e,s) {
    Sentry.captureException(e, s);
  }
}

void inner_function() {
  var my_data = ...;
  try {
    do_some_work_that_can_throw_exception(my_data);
  } catch (e,s) {
     Sentry.pleaseBindAttachmentToException(e, my_data); // <-- !!
     rethrow;
  }
}
```

Notice the `Sentry.pleaseBindAttachmentToException`, where I want to bind attachment to exception.

However, as we know, we cannot do it. I can create some code to bind `extra` to exception, because `extra` can be modified in EventProcessors (so I can manually create a static map holding some extras and bind to exception in my own EventProcessor). However, EventProcessors cannot modify or add attachments. So I have no way to do it :(

## Proposed solution

Say, add a field to `SentryOptions`, e.g. `void Function(Scope)? withScopeWhenCapture`. Then, in `Hub.captureException` and similar places, call that callback. This is very simple and I will implement it if you like.

related: https://github.com/getsentry/sentry-dart/issues/646

Hi @marandaneto could you please have a look (since this thread is strongly related to #646)? Thanks! And @ueman could you please also have a look? Thanks!

fzyzcjy:
Hi @marandaneto @ueman could you please suggest whether this is acceptable? If so I will make a PR. This is a blocking issue on our app...

ueman:
If I'm not mistaken you can already do something like this:
```dart
Sentry.captureException(exception: ex, withScope:(scope) {
    scope.addAttachment(attachment);
});
```

I'm not sure if that already works, but you might try that. If not @marandaneto probably appreciates a PR which enable scoped attachments. I'm not a maintainer of this SDK anymore.

Currently is also Christmas season, so most people are probably out of office.

fzyzcjy:
@ueman No I cannot do that...Notice that I `captureException` in a outer function, while I `addAttachment` in a inner function. For example, outer_function() calls func1() which calls func2() which calls func3(), ..., which calls func100() which calls inner_function().

> Currently is also Christmas season, so most people are probably out of office.

Ah I did not realized that (I have been doing my project for 2 years without using holidays ;) ). Sorry for disturbing.

ueman:
[...]

PR #683, opened 2021-12-26
Allow attachments to be bound to exception more easily

fzyzcjy:
Fixs #682


marandaneto:
https://github.com/getsentry/sentry-dart/pull/1136 is going to enable this type of feature, which is passing or removing attachments from the hint.

marandaneto:
Closing, the new approach should use [hints](https://github.com/getsentry/sentry-dart/blob/b5f0119a9d344b3dc0217487bddbd3557a554784/dart/lib/src/hint.dart)

PR #684, opened 2021-12-27
Feature: Add ability to mutate navigational breadcrumbs

ueman:
## :scroll: Description
This adds the option to mutate `RouteSettings` via callback in the Navigation Observer and it also adds the ability to add some context to each navigational breadcrumb.

Flutter allowes you to nest `Navigator`s. In order to differentiate breadcrumbs, I added the ability to do

```dart
SentryNavigatorObserver(
    additionalInfoProvider: (from, to) {
        return <String, dynamic>{'navigator': 'root-navigator'};
    },
);
``` 

Sometimes you don't want specific routes to be included, for example if they come from third party dependencies or you want your routes to be renamed to match analytic events. So you're now able to:

```dart
SentryNavigatorObserver(
    routeNameExtractor: (routeSettings) {
        return routeSettings.copyWith(name: 'different name');
        // or reset name to `null`:
       return RouteSettings(arguments: routeSettings.arguments);
    },
);
``` 

New tests

## :pencil: Checklist

- [ ] All tests passing
- [x] No breaking changes


flutter/lib/src/navigation/sentry_navigator_observer.dart
```diff
@@ -8,6 +8,15 @@ import '../../sentry_flutter.dart';
 /// See https://develop.sentry.dev/sdk/event-payloads/breadcrumbs/
 const _navigationKey = 'navigation';
 
+typedef RouteNameExtractor = RouteSettings Function(RouteSettings settings);
+
+typedef AdditionalInfoProvider = Map<String, dynamic>? Function(
```
ueman: I'm not quite sure about these names, any suggestions?

brustolin: I would change it from Extractor and Provider to Callback.
'Extractor' is very odd, and Provider is usually associated with other things in Flutter.

ueman: Sounds good

marandaneto:
@ueman before going ahead, wouldn't be possible to do this using the before breadcrumb callback? https://docs.sentry.io/platforms/flutter/enriching-events/breadcrumbs/#customize-breadcrumbs

ueman:
> @ueman before going ahead, wouldn't be possible to do this using the before breadcrumb callback? https://docs.sentry.io/platforms/flutter/enriching-events/breadcrumbs/#customize-breadcrumbs

Unfortunately not, at that point you're not able to know which `Navigator` the breadcrumb comes from.


Some  more  information on why this is useful:
You're able to nest `Navigator`s. For example, you can have one navigator for the app, which has a Tab Bar UI. Each tab then again has its own navigator. Each Navigator could theoretically have the same routes as the other ones. 

So this makes it useful to add some information at the `NavigationObserver` level because you're now able to associate information about which navigator is used.

I hope that clarifies a little bit.

marandaneto:
@ueman looks legit then, a last question, what if we pass the `RouteSettings` to the `hint` param from `addBreadcrumb`? it's meant for this type of usage, we set automatically everything that would make sense, but the user can still mutate the breadcrumb using the before breadcrumb callback, the `hint` is actually where the data was set from.

ueman:
> @ueman looks legit then, a last question, what if we pass the `RouteSettings` to the `hint` param from `addBreadcrumb`? it's meant for this type of usage, we set automatically everything that would make sense, but the user can still mutate the breadcrumb using the before breadcrumb callback, the `hint` is actually where the data was set from.

RouteSettings still don't specify which Navigator is its origin :(

marandaneto (approved):

marandaneto:
thanks for doing this @ueman !

PR #685, opened 2021-12-27
Feature: SentryAssetBundle

ueman:
## :scroll: Description
The `SentryAssetBundle` can create transactions for loading assets like images and other data.

See the following example:
Bildschirmfoto 2022-01-12 um 16 54 05

Not yet

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes


ueman:
@denrase how do I test all of this transaction stuff?

flutter/lib/src/sentry_asset_bundle.dart
```diff
@@ -0,0 +1,93 @@
...
+    }
+  }
+
+  @override
+  Future<T> loadStructuredData<T>(
+    String key,
+    _Parser<T> parser,
+  ) async {
+    final span = _hub.getSpan()?.startChild(
+          'SentryAssetBundle.loadStructuredData',
```
ueman: I'm not sure what to use for name and description

ueman: According to https://develop.sentry.dev/sdk/performance/span-operations/ it should probably be `file.read` as a name and `AssetBundle.loadStructuredData($filename)` as description. Or should the filename be added as a tag?

flutter/lib/src/sentry_asset_bundle.dart
```diff
@@ -0,0 +1,93 @@
...
+}
+
+Future<T> _wrapParsing<T>(
+  _Parser<T> parser,
+  String value,
+  String key,
+  ISentrySpan? span,
+) async {
+  final innerSpan = span?.startChild(
+    'SentryAssetBundle.parseStructuredData',
```
ueman: This span name should probably be `deserialize`?

flutter/lib/src/sentry_asset_bundle.dart
```diff
@@ -0,0 +1,80 @@
...
+      await span?.finish(status: SpanStatus.internalError());
+      rethrow;
+    }
+  }
+
+  /// Does not create a span. Sometimes [CachingAssetBundle] can throw errors
+  /// which are outside the current zone. This is not easy to handle and can
+  /// result in corrupt spans.
+  @override
+  Future<T> loadStructuredData<T>(String key, _Parser<T> parser) =>
```
ueman: I couldn't get this to reliably work in the error case, so no support for now. Though, I believe this is the most interesting method because of the parsing :(

flutter/lib/src/sentry_asset_bundle.dart
```diff
@@ -0,0 +1,97 @@
...
+
+  @override
+  Future<String> loadString(String key, {bool cache = true}) async {
+    final span = _hub.getSpan()?.startChild(
+          'file.read',
+          description: 'AssetBundle.loadString(key=$key, cache=$cache)',
+        );
+
+    try {
+      final data = await _bundle.loadString(key, cache: true);
```
brustolin:
```suggestion
      final data = await _bundle.loadString(key, cache: cache);
```

Shouldn't be this way?

ueman: Yeah, absolutely. Good catch!

flutter/lib/src/sentry_asset_bundle.dart
```diff
@@ -0,0 +1,97 @@
...
+          'file.read',
+          description: 'AssetBundle.load(key=$key)',
+        );
+
+    try {
+      final data = await _bundle.load(key);
+      await span?.finish(status: SpanStatus.ok());
+      return data;
+    } catch (e) {
+      await span?.finish(status: SpanStatus.internalError());
```
marandaneto: We can assign the `span.throwable` here.

flutter/lib/src/sentry_asset_bundle.dart
```diff
@@ -0,0 +1,97 @@
...
+        _bundle = bundle ?? rootBundle;
+
+  final Hub _hub;
+  final AssetBundle _bundle;
+
+  @override
+  Future<ByteData> load(String key) async {
+    final span = _hub.getSpan()?.startChild(
+          'file.read',
+          description: 'AssetBundle.load(key=$key)',
```
marandaneto: let's align the description with the `file.read` from Java, https://github.com/getsentry/sentry-java/blob/main/sentry/src/main/java/io/sentry/instrumentation/file/SentryFileInputStream.java
@romtsn can give you more details like extra fields as well when available e.g. file name and size

flutter/lib/src/sentry_asset_bundle.dart
```diff
@@ -0,0 +1,97 @@
...
+    try {
+      (_bundle as dynamic).clear();
+    } on NoSuchMethodError catch (_) {
+      // The clear method exists as of Flutter 2.8
+      // Previous versions don't have it, but later versions do.
+      // We can't use `extends` in order to provide this method because this is
+      // a wrapper and thus the method call must be forwarded.
+      // On Flutter version before 2.8 we can't forward this call and
+      // just catch the error which is thrown. On later version the call gets
+      // correctly forwarded.
```
marandaneto, line 176: Nice comment, is it possible to check at runtime the Flutter version and only call it if >= 2.8?

flutter/pubspec.yaml
```diff
@@ -17,10 +17,10 @@ dependencies:
   package_info_plus: ^1.0.0
 
 dev_dependencies:
-  build_runner: ^2.1.5
+  #build_runner: ^2.1.5
```
marandaneto: Don't we need it at all? if so, let's remove it.

flutter/lib/src/sentry_asset_bundle.dart
marandaneto: @romtsn

flutter/lib/src/sentry_asset_bundle.dart
marandaneto: if we just wrap on a try-catch like the others, isn't that enough for now?

flutter/pubspec.yaml
ueman: That's still needed in order to create the mocks. So this line should be reverted. I'm currently not able to use the latest Flutter version, so I had to temporarily remove it. 
```suggestion
  build_runner: ^2.1.5
```

flutter/lib/src/sentry_asset_bundle.dart
ueman: No, it can throw errors which somehow escape the async context and thus never reaches the try-catch. It's very weird. The Sentry root zone integration catches it though, so that's nice.

flutter/lib/src/sentry_asset_bundle.dart
ueman: There's no tooling for that :(

flutter/lib/src/sentry_asset_bundle.dart
ueman: Okay, I've finally got it to work 🎉

flutter/lib/src/sentry_asset_bundle.dart
```diff
@@ -0,0 +1,176 @@
...
+    runZonedGuarded(() async {
+      final data = await _bundle.loadStructuredData(
+        key,
+        (value) async => await _wrapParsing(parser, value, key, span),
+      );
+      span?.status = SpanStatus.ok();
+      completer.complete(data);
+    }, (exception, stackTrace) {
+      completer.completeError(exception, stackTrace);
+    });
```
marandaneto, line 99: Why do we use its own `runZonedGuarded`? just a try/catch wouldn't be enough?

flutter/lib/src/sentry_asset_bundle.dart
```diff
@@ -0,0 +1,176 @@
...
+
+  @override
+  Future<T> loadStructuredData<T>(String key, _Parser<T> parser) {
+    if (_enableStructureDataTracing) {
+      return _loadStructuredDataWithTracing(key, parser);
+    }
+    return _bundle.loadStructuredData(key, parser);
+  }
+
+  Future<T> _loadStructuredDataWithTracing<T>(
```
marandaneto, line 77: @brustolin ideas on how to make it work without this hacky `runZonedGuarded`?

flutter/lib/src/sentry_asset_bundle.dart
ueman: Somehow it isn't enough. You can test this easily by removing the `runZoneGuarded`. The test for the error cases will fail after that.

flutter/lib/src/sentry_asset_bundle.dart
ueman: If you get this to work I would love to get an explanation

brustolin: I believe it's not possible.

This is a part of flutter's `AssetBundle.loadStruturedData` function: 

```dart
    ...
    Completer<T>? completer;
    Future<T>? result;
    loadString(key, cache: false).then<T>(parser).then<void>((T value) {
      result = SynchronousFuture<T>(value);
      _structuredDataCache[key] = result!;
      if (completer != null) {
        // We already returned from the loadStructuredData function, which means
        // we are in the asynchronous mode. Pass the value to the completer. The
        // completer's future is what we returned.
        completer.complete(value);
      }
    });
    ...
```

As you can see `loadString` is not awaited, so, it asynchronous execution is done in a different scope, and would not be handled by a try/catch that could be used instead of the `runZoneGuarded`.

In my understanding this is a dart bad approach.

But, that being said, we are changing the behavior of Flutter here. We are allowing and exception that should be thrown not matter what (by dart's design) to be handled.

ueman: > But, that being said, we are changing the behavior of Flutter here. We are allowing and exception that should be thrown not matter what (by dart's design) to be handled.

That's why I've put it behind a flag which is disabled by default.

I'm okay with removing it and with keeping it. It's your call.

marandaneto: @bruno-garcia opinions? I'd rather not do it at all (documenting this specific use case - telling people to do custom instrumentation) or keep this flag false by default, it still could lead to bugs anyway when people turn it on.

marandaneto:
@ueman it'd be nice to get the setup example and short description in the https://github.com/getsentry/sentry-dart/tree/main/flutter README as well.

flutter/lib/src/sentry_asset_bundle.dart
marandaneto: will keep discussion on https://github.com/getsentry/sentry-dart/pull/685/files#r784998311

flutter/lib/src/sentry_asset_bundle.dart
bruno-garcia: There's a lot to unpack here and I have no context, is there a summary of what's going on?

flutter/lib/src/sentry_asset_bundle.dart
romtsn: yep, described it [here](https://github.com/getsentry/develop/issues/463#issuecomment-990042948)

flutter/lib/src/sentry_asset_bundle.dart
brustolin: I would suggest to use try/catch instead of `runZonedGuarded` and let the app break as intended. 
The SDK for Dart does register unhandled exceptions, right?

marandaneto: I had a call with @bruno-garcia and agreed to keep the flag disabled by default for now, till we get some feedback during the beta tests, if all is good, we flip during the next major.
@ueman just to confirm, the behavior is still the same when using it's own `runZoneGuarded`, right? the question is if the `rethrow` exception works correctly.

ueman: > the behavior is still the same when using it's own runZoneGuarded, right?

I'm not quite sure what you mean by that, could you rephrase the question?

> the question is if the rethrow exception works correctly.

There's a test for that.

flutter/lib/src/sentry_asset_bundle.dart
ueman: The `key` doesn't need to guarded by `sendDefaultPii` because it's a path relative to the pubspec.yaml. The other points we're changed accordingly.

flutter/lib/src/sentry_asset_bundle.dart
```diff
@@ -0,0 +1,197 @@
...
+    Hub? hub,
+    AssetBundle? bundle,
+    bool enableStructuredDataTracing = false,
+  })  : _hub = hub ?? HubAdapter(),
+        _bundle = bundle ?? rootBundle,
+        _enableStructureDataTracing = enableStructuredDataTracing;
+
+  final Hub _hub;
+  final AssetBundle _bundle;
+  final bool _enableStructureDataTracing;
```
marandaneto:
```suggestion
  final bool _enableStructuredDataTracing;
```

flutter/lib/src/sentry_asset_bundle.dart
marandaneto: never mind, the test answers the question :D thanks.

flutter/lib/src/sentry_asset_bundle.dart
ueman: Dang it

flutter/lib/src/sentry_asset_bundle.dart
```diff
@@ -0,0 +1,197 @@
...
+  }
+
+  void _setDataLength(dynamic data, ISentrySpan? span) {
+    int? byteLength;
+    if (data is List<int>) {
+      byteLength = data.length;
+    }
+    if (data is ByteData) {
+      byteLength = data.lengthInBytes;
+    }
```
marandaneto: Good use case for `if ... else if` otherwise we test 2 times what could be only once

flutter/lib/src/sentry_asset_bundle.dart
marandaneto: @ueman descriptions are still not following @romtsn comment from the other thread: https://github.com/getsentry/sentry-dart/pull/685#discussion_r786022279

See here: https://github.com/getsentry/sentry-java/blob/4c1cf85df7af23b88495120bd9a089e601593eef/sentry/src/main/java/io/sentry/instrumentation/file/FileIOSpanManager.java#L86-L94
It's supposed to be something like, e.g. `file.txt (1 MB)`.
We could also do e.g. `loadStructuredData: file.txt (1 MB)` but it's Important the file name and file size, it's easier to recognize where it comes from.
If the file name isn't available, we can use either `$key` or `$T` or something that leads to what's being used.

flutter/lib/src/sentry_asset_bundle.dart
marandaneto: Left a comment about the `description` -> https://github.com/getsentry/sentry-dart/pull/685#discussion_r786228015

flutter/lib/src/sentry_asset_bundle.dart
ueman: We don't have any code to pretty print the file size in the SDK. Any suggestions for that? The key is a path to a file bundled with the application. Do you want the whole path or just the last path segment?

marandaneto: @ueman based on https://github.com/getsentry/develop/issues/463#issuecomment-990042948 just the last path segment in the description and the whole path in the `file.path` data field (alset set).
For the file size, can we do something like https://gist.github.com/zzpmaster/ec51afdbbfa5b2bf6ced13374ff891d9#file-formatbytes-dart ?

ueman: I've just noticed that we can't easily modify the spans description after it was started.
So I can add the file name, but I can't add the file size.

marandaneto: That sounds like a good start, we can iterate that on a different PR.

flutter/lib/src/sentry_asset_bundle.dart
```diff
@@ -0,0 +1,198 @@
...
+      byteLength = data.length;
+    } else if (data is ByteData) {
+      byteLength = data.lengthInBytes;
+    }
+    if (byteLength != null) {
+      span?.setData('file.size', byteLength);
+    }
+  }
+
+  String _fileName(String key) => Uri.parse(key).pathSegments.last;
```
marandaneto: That requires if `pathSegments.isNotEmpty` otherwise it throws, similar to `SentryStackTraceFactory` what does, if the `pathSegments` is empty, we can use the raw `key`

ueman: So this
```suggestion
  String _fileName(String key) => Uri.tryParse(key)?.pathSegments.last ?? key;
```
would be better.

Technically all keys must be a valid uri, otherwise there's no way to load anything, but yeah, there's nothing which would stop you from passing an empty string.

marandaneto: indeed, but `pathSegments` still needs to be checked if `isNotEmpty` otherwise `.last` throws, I guess

marandaneto (approved):
Left the last comment https://github.com/getsentry/sentry-dart/pull/685#discussion_r786266164
Happy to merge and approve after addressing this last one.
thanks once more @ueman

marandaneto:
@brustolin all good from my side, would you like to double-check?

brustolin:
> @brustolin all good from my side, would you like to double-check?

Sure!

flutter/lib/src/sentry_asset_bundle.dart
```diff
@@ -0,0 +1,204 @@
...
+          description: 'AssetBundle.loadString: ${_fileName(key)}',
+        );
+
+    span?.setData('file.path', key);
+    span?.setData('from-cache', cache);
+
+    String? data;
+    try {
+      data = await _bundle.loadString(key, cache: cache);
+      span?.status = SpanStatus.ok();
```
brustolin, line 129: l: I believe we should set data 'file.size' here as well.

brustolin (approved):
Just need to check if this [comment](https://github.com/getsentry/sentry-dart/pull/685/files#r787042390) needs to be done, otherwise is good to go.

flutter/lib/src/sentry_asset_bundle.dart
ueman: I thought about that, too, but I don't believe the cost of doing it is worth it. We would have to convert the string back to bytes which is potentially an expensive operation.

marandaneto: agreed, and if we do, we should consider what Flutter does too:

https://github.com/flutter/flutter/blob/5112f86c9416143a33de4ead42c8aff6153767f7/packages/flutter/lib/src/services/asset_bundle.dart#L75-L82

brustolin: Agreed

PR #698, opened 2022-01-16
Proof of Concept: Performance tracing for files

ueman:
This PR show how performance monitoring for files could work like.
It's overall very draft-y, but it already works.
I'm currently not able to finish this, so I would appreciate it, if someone would take it over. 

This obviously only works on dart:io platforms. Implementing this should be pretty straight forward, though it's a lot of work to test. Maybe the current implementation is already sufficient.


marandaneto:
@ueman thanks for that, we'll check that out.

ueman:
Closing this in favor of https://github.com/ueman/sentry-dart-tools/tree/main/sentry_plus

PR #699, opened 2022-01-16
Proof of Concept: Frame Tracking

ueman:
This PR shows how one could add the 'total frames' measurement for transactions. 
The current code needs some more polishing before it's ready for release.
I'm currently not able to finish this, so I would appreciate it, if someone else would take it over the finishing line.

This also seems to require some (bigger?) changes in the transaction/span API. It's also tricky to get the total frame number with auto closing navigation transaction, because there's no callback.

screenshot


marandaneto:
@denrase you probably want to check that out since we'd need something similar for the Mobile vitals.

flutter/lib/src/navigation/frame_tracker.dart
```diff
@@ -0,0 +1,70 @@
...
+
+  void _frameCallback(Duration _) {
+    final timeStamp = _timeStamp;
+    if (timeStamp == null) {
+      return;
+    }
+
+    // postFrameCallbacks are called just once,
+    // so we have to add it each frame.
+    _binding.addPostFrameCallback(_frameCallback);
```
ueman, line 33: Maybe a ticker is a better candidate than this? See https://api.flutter.dev/flutter/scheduler/Ticker-class.html

ueman:
This feature was implemented with a different approach.

PR #701, opened 2022-01-17
Fix: maxRequestBodySize should be never by default when using the FailedRequestClientAdapter directly

marandaneto:
## :scroll: Description
Fix: maxRequestBodySize should be never by default when using the FailedRequestClientAdapter directly

Not considered a breaking change because people don't use the `FailedRequest` adapter/client directly but rather the higher level `SentryHttp` client/adapter.

Tests don't fail because we're always passing the optional parameter in tests.

## :pencil: Checklist

- [X] All tests passing
- [X] No breaking changes


brustolin (approved):
LGTM

Issue #624, opened 2021-10-23
Support HTTP response body if sendDefaultPii and maxResponseBodySize are enabled


marcellodesales:
# Requirement

* We'd like to have full visibility on the logs from the app
* We need to print the full HTTP Request/Response submitted to Sentry
* For whitelisting dozens of apps, it's impossible to double-check the configs for all of them
  * We need ta way to verify how the apps are configured

# Design

* We have mobile apps: Flutter (Android, iOS)
* We have Web apps: Flutter (Progressive Web Apps)
* We have backend services: Java, Golang, Python, Node.js

We need a consistent way to turn HTTP requests `on/off`

marandaneto:
@marcellodesales thanks for reporting.

not sure if I follow the feature, can you describe the use case?
do you mean when an error happens (eg 500) or every single request even if 200?

See https://docs.sentry.io/platforms/dart/usage/advanced-usage/

The `SentryHttpClient` class has configurations like `captureFailedRequests`, can be used along with `failedRequestStatusCodes` and `maxRequestBodySize`, does that work?

we don't log `response` since it could contain PII and it could be quite a huge payload, events cannot be higher than 1MB.

in case none of them work, you can always implement your own HTTP Client calling the Sentry APIs manually eg `Sentry.captureEvent(x)`, `Sentry.captureException(x)`, etc...

marandaneto:
@marcellodesales any news here? thanks

marcellodesales:
@marandaneto Sorry I couldn't comment on this earlier... I think it's reasonable to implement it manually as you described since logging may contain PII info for the body, but I don't think it applies for the rest of the HTTP request (verb, headers) when connecting to the sentry servers.

Before I close the ticket, do you think you could support in the basic implementation the option to turn on logging the HTTP request/response from sentry without the body? Everything we are submitting to sentry would be captured... The reason why I'm asking is because we use HTTP Traces with a given specific TransactionID that actually gets to Sentry dashboard as a tag. However, in our systems should be able to track that the specific transactionID was submitted to Sentry... 

Anyway, let me know if that's possible... 
thank you

marandaneto:
We do collect some data if the `sendDefaultPii ` is enabled (including headers, verb, request body), see https://github.com/getsentry/sentry-dart/blob/main/dart/lib/src/http_client/failed_request_client.dart#L162-L173
[...]

PR #718, opened 2022-01-26
Feature: Improve Dio exception reports

### LINKED ISSUE #715: [dio] Improve fingerprinting
I tested the new `sentry_dio` package a little bit.

My issue is that due to how DIO throws `DioError`s, the stacktrace is always the same (that's a problem in itself and probably needs to be improved) - this results in all captured errors having the same fingerprint.

Additionally a `DioError` can wrap a `cause` error which may be the real reason why the request failed, this currently does not get reported.

@ueman What are you thoughts on this?

ueman:
## :scroll: Description
Adds an event processor, which add chained exceptions for the inner exception of `DioError`, 
see https://sentry.io/share/issue/638a3758a43d4f61bc2a0c62acdefd69/

This one should be merged after https://github.com/getsentry/sentry-dart/pull/728

Fixes https://github.com/getsentry/sentry-dart/issues/715

## :pencil: Checklist

- [x] All tests passing
- [ ] No breaking changes


flutter/example/lib/main.dart
```diff
@@ -22,6 +22,7 @@ Future<void> main() async {
       options.dsn = _exampleDsn;
       options.tracesSampleRate = 1.0;
       options.reportPackages = false;
+      options.addEventProcessor(DioEventProcessor(options));
```
ueman: After options is exposed on the hub, I can move this to rest of the Dio setup, so that it's still only one step to configure.

kuhnroyal:
I never get any `DioError` in the `FailedRequestClient` , the whole block here: https://github.com/flutterchina/dio/blob/develop/dio/lib/src/dio_mixin.dart#L620. is executed later.

dio/lib/src/dio_event_processor.dart
```diff
@@ -0,0 +1,54 @@
+// ignore_for_file: implementation_imports, public_member_api_docs
+
+import 'dart:async';
+
+import 'package:dio/dio.dart';
+import 'package:sentry/sentry.dart';
+import 'package:sentry/src/sentry_exception_factory.dart';
+import 'package:sentry/src/sentry_stack_trace_factory.dart';
```
ueman: Both of these aren't exposed, should I expose them or ignore the `implementation_imports` warning?

kuhnroyal:
This works, I have the correct in-app grouping.

dio/lib/src/dio_event_processor.dart
```diff
@@ -0,0 +1,54 @@
...
+      return event;
+    }
+
+    // Potential further improvements:
+    // Add dioError.requestOptions to event.extra
+    // Add dioError.response to event.extra
+
+    try {
+      final exception = sentryExceptionFactory.getSentryException(
+        dioError.error,
```
kuhnroyal: This can be `null`, not sure what happens then.
Also could wrap it in a `ThrowableMechanism` similar to the `FailedRequestClient`.

dio/lib/src/dio_event_processor.dart
```diff
@@ -0,0 +1,54 @@
...
+        stackTrace: dioError.stackTrace,
+      );
+
+      final exceptions = event.exceptions;
+
+      return event.copyWith(
+        exceptions: [
+          exception,
+          if (exceptions != null) ...exceptions,
```
kuhnroyal: `...?` not supported in min Dart version?

kuhnroyal (commented):
I basically copied this and and added some more code from `FailedRequestClient` in order to add a `SentryRequest` to the event. And I disabled the `FailedRequestClient`. This combination now gives me a pretty good result.
The only problem I see, is that the error is not tied to the transaction in any way.

dio/lib/src/dio_event_processor.dart
ueman: Not sure, I'll check that.

dio/lib/src/dio_event_processor.dart
ueman: Yeah, I know. But this is not yet tested at all, that's why it's a draft :D

ueman:
> I basically copied this and and added some more code from `FailedRequestClient` in order to add a `SentryRequest` to the event. And I disabled the `FailedRequestClient`. This combination now gives me a pretty good result.

That's actually a pretty good idea.

> The only problem I see, is that the error is not tied to the transaction in any way. 

As this integration doesn't start any transactions, you basically have to write code like the one below anyway
```dart
  final transaction = Sentry.getSpan() ??
      Sentry.startTransaction(
        'dio-web-request',
        'request',
        bindToScope: true,
      );
  try {
    final response = await dio.get<String>('https://flutter.dev/');
    transaction.spanStatus = SpanStatus.ok();
    transaction.throwable = exception;
    transaction.spanStatus = SpanStatus.internalError();
    Sentry.captureException(exception, stacktrace);
  } finally {
    await transaction.finish();
  }
```
and then it is tied together.

dio/lib/src/dio_event_processor.dart
```diff
@@ -0,0 +1,54 @@
...
+
+    // Potential further improvements:
+    // Add dioError.requestOptions to event.extra
+    // Add dioError.response to event.extra
+
+    try {
+      final exception = sentryExceptionFactory.getSentryException(
+        dioError.error,
+        stackTrace: dioError.stackTrace,
+      );
```
ueman: After doing this, I would like to set `dioError.stackTrace` (and maybe `dioError.error`) to `null`, so the report on Sentry looks much nicer.
Also, because we're adding an extra SentryException for those, it should be okay. Any opinions on that?

kuhnroyal:
> As this integration doesn't start any transactions, you basically have to write code like the one below anyway

True, it is all tied to my Flutter route transactions.

dio/lib/src/dio_event_processor.dart
kuhnroyal: Not sure, no opinion yet.

dio/lib/src/dio_event_processor.dart
```diff
@@ -0,0 +1,110 @@
...
+  FutureOr<SentryEvent?> apply(SentryEvent event, {dynamic hint}) {
+    final dynamic dioError = event.throwable;
+    if (dioError is! DioError) {
+      return event;
+    }
+
+    try {
+      final mechanism = Mechanism(
+        type: 'DioIntegration',
+        description: 'This is the inner exception of the DioError',
```
kuhnroyal: I found that setting the description screws up the UI in Sentry, the mechanism name is no longer showing. There is no description set for `runZonedGuarded` etc.

dart/lib/src/http_client/sentry_http_client.dart
```diff
@@ -148,3 +148,12 @@ class SentryStatusCode {
     return '$_min..$_max';
   }
 }
+
+extension SentryStatusCodeListExtension on List<SentryStatusCode> {
+  bool containsStatusCode(int? statusCode) {
```
kuhnroyal: Somewhat unrelated, but can we somehow use the `ValidateStatus` from the Dio options as default for `failedRequestStatusCodes`?

dio/lib/src/dio_event_processor.dart
```diff
@@ -0,0 +1,110 @@
...
+      );
+      final throwableMechanism = ThrowableMechanism(mechanism, dioError.error);
+      final exception = sentryExceptionFactory.getSentryException(
+        throwableMechanism,
+        stackTrace: dioError.stackTrace,
+      );
+
+      // Remove the StackTrace so the message on Sentry looks much better
+      dioError.stackTrace = null;
```
kuhnroyal: Hmm I don't see any difference.

dio/lib/src/dio_event_processor.dart
```diff
@@ -0,0 +1,110 @@
+// ignore_for_file: implementation_imports, public_member_api_docs
+
+import 'dart:async';
+
+import 'package:dio/dio.dart';
+import 'package:sentry/sentry.dart';
+import 'package:sentry/src/sentry_exception_factory.dart';
+import 'package:sentry/src/sentry_stack_trace_factory.dart';
+
+class DioEventProcessor extends EventProcessor {
```
kuhnroyal:
```suggestion
class DioEventProcessor implements EventProcessor {
```

dio/lib/src/dio_event_processor.dart
ueman: Yeah, I also noticed it. It could be because the original exception was already converted to a SentryException at this point. I'll have to look into it some more.

dio/lib/src/dio_event_processor.dart
ueman: But it should still show the original Mechanism for the DioError. 

@marandaneto How do order the exceptions, so that the DioError is the main exception and the the inner exception is the second?

dio/lib/src/dio_event_processor.dart
ueman: Does that make any difference here?

dart/lib/src/http_client/sentry_http_client.dart
ueman: I'm thinking of removing the `FailingRequestAdapter` all together because of this. If Dio throws on specific status codes anyway, there's no need for that functionality.

dio/lib/src/dio_event_processor.dart
kuhnroyal: Well it is an `interface`, but I am coming from the Java world so when I see `extends`, I expect some implementation to exist in the superclass. Not really important, up to you :)

kuhnroyal:
When a description is set on the mechanism, the type is not shown in Sentry UI. Not sure why.

Bildschirmfoto 2022-01-27 um 10 56 09
Bildschirmfoto 2022-01-27 um 10 56 18

dio/lib/src/dio_event_processor.dart
kuhnroyal: It always shows `runZonedGuarded` for the `DioError`.
But not for the cause when the description is set, see https://github.com/getsentry/sentry-dart/pull/718#issuecomment-1023036545

kuhnroyal: Might be something buggy in the UI.

ueman:
> When a description is set on the mechanism, the type is not shown in Sentry UI. Not sure why.
> 
> Bildschirmfoto 2022-01-27 um 10 56 09 Bildschirmfoto 2022-01-27 um 10 56 18

Yeah, that's also true for `FlutterError` which sometime have the description set. 
I don't know whether that's working as expected or not. Maybe @marandaneto know. 
Probably this is worth an issue at the Sentry repo itself.

dio/lib/src/dio_event_processor.dart
marandaneto: They are internal, what we can do is to add them to `SentryOptions` and use the getter from there, so we have also a single instance of it.

dio/lib/src/dio_event_processor.dart
marandaneto: the sorting matter, See `SentryStackTraceFactory#getStackFrames`, we `reversed` the list because of that.
`DioError` should be the last item on the exception list.

dio/lib/src/dio_event_processor.dart
```diff
@@ -0,0 +1,110 @@
...
+
+  @override
+  FutureOr<SentryEvent?> apply(SentryEvent event, {dynamic hint}) {
+    final dynamic dioError = event.throwable;
+    if (dioError is! DioError) {
+      return event;
+    }
+
+    try {
+      final mechanism = Mechanism(
```
marandaneto: You don't need a mechanism for the inner exception, only for the "main" one.

dio/lib/src/dio_event_processor.dart
```diff
@@ -0,0 +1,110 @@
...
+        type: 'DioIntegration',
+        description: 'This is the inner exception of the DioError',
+      );
+      final throwableMechanism = ThrowableMechanism(mechanism, dioError.error);
```
marandaneto: What if we always use the `dioError.error` for the `dio` integration, isn't that any better? is there info on `dioError` that's not on `dioError.error`?

marandaneto:
> > When a description is set on the mechanism, the type is not shown in Sentry UI. Not sure why.
> > Bildschirmfoto 2022-01-27 um 10 56 09 Bildschirmfoto 2022-01-27 um 10 56 18
> 
> Yeah, that's also true for `FlutterError` which sometime have the description set. I don't know whether that's working as expected or not. Maybe @marandaneto know. Probably this is worth an issue at the Sentry repo itself.

I agree, an issue there would be better, I don't know the answer either.

dio/lib/src/dio_event_processor.dart
ueman: Yep, there's info on `DioError` which is not available on `DioError.error`. Also, `DioError.error` is not always available.

marandaneto: Makes sense to report both whenever possible then.

I'm wondering if we can make it without the EventProcessor.
What if we pass a List of exceptions to `event.throwable` since its a `dynamic` field and in the `sentryExceptionFactory` we special case if it's a `List`, we traverse the list converting to `SentryException`, didn't try, but it could work.

The downside is if somebody does `throw randomList` and this List isn't this special list of exceptions, but rather an App's list, we need to consider that too.

kuhnroyal: We don't have the `DioError` available earlier. The `SentryDioClientAdapter` is not aware of these errors. The `DioError`s usually are manually caught and captured or come from the zones uncaught handler.

ueman: But then we would have to know how to create list from `DioError` inside the `sentry` sdk. But that doesn't know about Dio.

Is there a way to register a `SentryExceptionFactory` for specific exceptions?

marandaneto: I see, then an event processor makes sense, we just need to add it automatically, so fewer steps for the final user.
A thing to consider is, an App may have more than one `dio` instance, so when `addSentry` is called, we should not register the event processor twice.

ueman: > I see, then an event processor makes sense, we just need to add it automatically, so fewer steps for the final user.

See https://github.com/getsentry/sentry-dart/pull/718#discussion_r792821809

> A thing to consider is, an App may have more than one dio instance, so when addSentry is called, we should not register the event processor twice.

Good point.

dio/lib/src/dio_event_processor.dart
ueman: I would like to add some explanation to the inner exception to describe that it is an inner exception. How do I do that without a mechanism?

dio/lib/src/dio_event_processor.dart
```diff
@@ -0,0 +1,149 @@
...
+        // Don't override just parts of the original request.
+        // Keep the original one or if there's none create one.
+        request: event.request ?? _toRequest(dioError),
+      );
+    }
+
+    try {
+      final exception = sentryExceptionFactory.getSentryException(
+        dioError.error,
+        stackTrace: dioError.stackTrace,
```
kuhnroyal: I tested a lot today, there is even more to this I think.
When `dioError.error is Error` then we have to use  `dioError.error.stackTrace`. We may even have to add 2 chained exceptions here.
For example some form of serialization error or `SocketException`s have a distinct stackTrace in `dioError.error.stackTrace` but the caller information from your own code is in `dioError.stackTrace`. 
On the other hand for HTTP response errors, `dioError.error` is just a String so there is no additional stackTrace.

So there can actually be 3 separate stackTraces for each DioError.
1. The on you get with `catch(e, stackTrace)`- this one is mostly useless as it always points to some internal Dio code. But it is likely the one being captured with `captureException` by the user. This is the one you are removing now. (mostly useless)
2. The `dioError.stackTrace` - this one is generated with `StackTrace.current` inside Dio and most of the time contains the correct call stack from the users code (inApp). (important)
3. If `dioError.error is Error` then there is a `dioError.error.stackTrace` - this one contains detailed information about potential inner errors which happen during serialization/send/receive. (kinda important)

dio/lib/src/dio_event_processor.dart
```diff
@@ -0,0 +1,149 @@
...
+      );
+
+      final exceptions = _removeDioErrorStackTraceFromValue(event, dioError);
+
+      return event.copyWith(
+        exceptions: [
+          exception,
+          ...exceptions,
```
kuhnroyal, line 56: I think we can just drop these exceptions?

dio/lib/src/dio_event_processor.dart
ueman: Thank you for this. That's a great deep dive.
Also something to think about in https://github.com/flutterchina/dio/pull/1405

I wonder if this also applies to pure Dart code. Errors (in contrast to Exceptions) always have their own stacktrace. So if you try-catch an error, you have two stacktraces. Errors are kinda defined as mistakes a developer makes at the time of writing but then again people probably still try-catch errors 🤷‍♂️

dio/lib/src/dio_event_processor.dart
ueman: What makes you think so?

dio/lib/src/dio_event_processor.dart
kuhnroyal: Dart 2.16 added `Error.throwWithStackTrace(error, stackTrace)` which should probably be used in Dio 5.0.
This would solve some of the problems.

dio/lib/src/dio_event_processor.dart
kuhnroyal: Well for me, this is always just the `DioError` with the wrong stacktrace making it completely useless.
And we are adding the correct `DioError`.

dio/lib/src/dio_event_processor.dart
ueman: Yeah, or just to completely remove any catching of errors/excetptions in Dio itself

dio/lib/src/dio_event_processor.dart
ueman: Yeah, for me that's also the case. But there may be other event processers at work which add to it, and we shouldn't drop those. We should only replace the DioError and keep everything as is.

dio/lib/src/dio_event_processor.dart
marandaneto: Everything but the main exception is an inner exception (or chained exception), even the UI shows them below the main exception, why do you need a mechanism for that? just trying to understand the use case.

dio/lib/src/dio_event_processor.dart
kuhnroyal: Yea I agree

dio/lib/src/dio_event_processor.dart
kuhnroyal: But then they can't be wrapped into a `DioError` :)

ueman: Which isn't a problem if it doesn't add anything valuable to the original exception and you don't run into problems with masking or losing the original exception/stacktrace.

dart/lib/src/http_client/sentry_http_client.dart
ueman: https://github.com/getsentry/sentry-dart/pull/728

dio/lib/src/dio_event_processor.dart
ueman: https://github.com/getsentry/sentry-dart/pull/727

marandaneto:
There are a lot of trade-offs here and I'd need some time to review this and decide how we could improve the chained exceptions and grouping for `dio`, the discussions are insightful tho, thank you @ueman and @kuhnroyal

dio/lib/src/dio_event_processor.dart
```diff
@@ -0,0 +1,149 @@
...
+
+    return SentryRequest(
+      method: options.method,
+      headers: _options.sendDefaultPii ? headers : null,
+      url: urlWithoutQuery,
+      queryString: query,
+      cookies: _options.sendDefaultPii
+          ? options.headers['Cookie']?.toString()
+          : null,
+      data: _getRequestData(dioError.response?.data),
```
marandaneto, line 116: You're using request props configuration for the `response` object, which is wrong.
Looks like `DioError` does not have access to the `request` data, but only `response`.

marandaneto: Btw adding response data requires discussion yet, See https://github.com/getsentry/sentry-dart/issues/624#issuecomment-1012296140 due to PII concerns, etc...

kuhnroyal: Access to the request options is available.

marandaneto: Oh looks like it's a different name, `error.requestOptions.data` and friends would work, indeed.

marandaneto:
@ueman and @kuhnroyal when it throws e.g. `SocketException`, the stackTrace has only one frame and it is low level, those events are grouped together, I don't see a big problem for that since it does not matter where it comes from, there's no network or you can't do much about it anyway.

If it's about `failedRequestStatusCodes`, it uses the `StackTrace.current` and it does not contain the App's frames, fair, events may be grouped together since the frames are almost always the same, but the important bit here is, which URL is throwing? and this is answered by the events, does it matter if the same API is called from different places?

`DioError` seems to be richer in both cases.
If you have access to the `DioError` and it throws because of e.g. 404, you have access to the stack trace with App's frames.
If you have access to the `DioError` and it throws because of e.g. no network, you have access to `dio.error` which is the wrapped `SocketException`, also `dio.stackTrace` has Apps frames.

so `DioError` is always better, `DioError.stackTrace` is always better than the captured `stackTrace` via the `catch block`, `dio.error` could be a chained exception.

Do you know if using Dio Interceptors, do we have access to the DioError?
What was the trade-off between the Adapter vs the Interceptor?

kuhnroyal:
> so `DioError` is always better, `DioError.stackTrace` is always better than the captured `stackTrace` via the `catch block`, `dio.error` could be a chained exception.

Yes thats right.

Adapter was easier to implement and may be easier to maintain, as the implementation is similar to the one for the `http` package.
But an adapter doesn't know about the `DioError`, an interceptor does.
The interceptor is harder to configure correctly, as interceptors actually can prevent following interceptors from running or modify the response/error.

marandaneto:
> > so `DioError` is always better, `DioError.stackTrace` is always better than the captured `stackTrace` via the `catch block`, `dio.error` could be a chained exception.
> 
> Yes thats right.
> 
> Adapter was easier to implement and may be easier to maintain, as the implementation is similar to the one for the `http` package. But an adapter doesn't know about the `DioError`, an interceptor does. The interceptor is harder to configure correctly, as interceptors actually can prevent following interceptors from running or modify the response/error.

Indeed, we do something similar for Java-OkHttp and it was quite straightforward -> https://github.com/getsentry/sentry-java/blob/main/sentry-android-okhttp/src/main/java/io/sentry/android/okhttp/SentryOkHttpInterceptor.kt
I don't know the internals of `dio` or how it handles interceptors, and errors, but it should not be that different.

We could try since I don't see any other way to improve the current impl. without `DioError`.

@ueman solution with the EventProcessor also seems to be a good idea, I'm wondering which one is better.

ueman:
> The interceptor is harder to configure correctly, as interceptors actually can prevent following interceptors from running or modify the response/error.

That's the point I'm concerned about.

Depending on the configuration we might report errors even if they're later discarded and we could also get the data wrong as it might be modified after recording. Also we might not know about errors because a potential SentryDioInterceptor wasn't even called. 

I'm not quite sure how hard or error prone it is to get it right, so I'm gladly hear your thoughts on it.

We probably shouldn't promote the dio integration to stable until this discussion is solved tho 😄

kuhnroyal:
I have used an interceptor based approach so far and it worked pretty well, but the configuration problems are a point of concern.

Maybe it just needs to be documented, I mean now we have documented that `..configureSentry()` needs to be called last. With an interceptor we just leave that and default just add the Sentry interceptor it at index 0? Dunno.

On the other hand this "feature" may be what we want. Let's says you have an interceptor which handles OAuth refreshes when there is a 401/whatever response. I can then choose to not forward this response error because I am instead refreshing the token and re-execute the original request. I don't want to see this in Sentry.

marandaneto:
If we don't capture events because any other interceptor has discarded it, well, it really depends on the setup and what the other interceptors do, it's a best-effort approach, if the data comes in handy to our interceptor, we're able to enrich and capture errors, otherwise not.

For 401 errors, you can either disable not passing to the error code range or even use the `beforeSend` callback to drop it.

The interceptor should be configurable enough and fit most of the "normal" use cases, users are free to either configure accordingly or just not use it.

I don't think there's a solution for every edge case, but it should be possible for the majority that does simple REST calls, etc...

What do you think?

ueman:
> What do you think?

For performance tracing it would be very bad to have a span/transaction which is not closed. So performance tracing is going to stay as is. 
The FailedRequestAdapter should be converted to an interceptor. The DioEventProcessor is still useful for chained stack traces, so it's also going to stay.

Did I get that right?

kuhnroyal:
> For performance tracing it would be very bad to have a span/transaction which is not closed. So performance tracing is going to stay as is.

Is it that bad? Or is it the responsibility of the developer to adapt his code?

> The DioEventProcessor is still useful for chained stack traces, so it's also going to stay.

We can get chained exceptions from interceptor as well but then it doesn't work for try/catch... And do you need to capture with try/catch when it was already captured in an interceptor?


How does it work with the `http` package? Uncaught vs adapter/client?

ueman:
> How does it work with the `http` package? Uncaught vs adapter/client?

`http` only throws on network failures and it has no interceptors. So everything is done as adapter/client.

kuhnroyal:
We will end. up with duplicate events If we capture all errors in an interceptor and then automatically have the same errors captured in the zone...

kuhnroyal:
@ueman I didn't get around to testing again yet, but have a slightly related question.

What do you think about adding a way to pass a custom parent span to a request via Dio's `extra` map, that the `TracingClientAdapter` will use by default and fallback to `Sentry.getSpan()` if it is not set?

Like so:
```dart
  @override
  Future<ResponseBody> fetch(
    RequestOptions options,
    Stream<Uint8List>? requestStream,
    Future? cancelFuture,
  ) async {
    // see https://develop.sentry.dev/sdk/performance/#header-sentry-trace
    final currentSpan = options.extra['sentrySpan'] is SentrySpan ? options.extra['sentrySpan']! as SentrySpan : _hub.getSpan();
    final span = currentSpan?.startChild(
      'http.client',
      description: '${options.method} ${options.uri}',
    );
```

ueman:
> @ueman I didn't get around to testing again yet, but have a slightly related question.

Nothing much has changed regarding the functionality of this PR. Mostly clean up. There's another one in https://github.com/getsentry/sentry-dart/pull/728. Regarding https://github.com/getsentry/sentry-dart/pull/718#discussion_r794086467, I'll raise another PR which should enable that for all errors, even non DioErrors.

> What do you think about adding a way to pass a custom parent span to a request via Dio's `extra` map, that the `TracingClientAdapter` will use by default and fallback to `Sentry.getSpan()` if it is not set?

That should be doable, but better in another PR 😅

Edit: https://github.com/getsentry/sentry-dart/issues/751

dio/test/sentry_dio_extension_test.dart
```diff
@@ -1,18 +1,55 @@
...
+import 'mocks/mock_hub.dart';
 
 void main() {
   group('SentryDioExtension', () {
-    test('addSentry add client and transformer', () {
+    test('addSentry adds $SentryTransformer', () {
       final dio = Dio();
-      dio.addSentry();
-      expect(dio.httpClientAdapter, isA<SentryDioClientAdapter>());
+      final hub = MockHub();
```
marandaneto: We could make a Fixture here.

dio/lib/src/dio_event_processor.dart
```diff
@@ -0,0 +1,140 @@
...
+
+    try {
+      var innerException = _sentryExceptionFactory.getSentryException(
+        innerDioErrorException ?? 'DioError inner stacktrace',
+        stackTrace: innerDioStackTrace,
+      );
+      if (innerDioErrorException.runtimeType == String) {
+        innerException =
+            innerException.copyWith(type: 'DioError inner stacktrace');
+      }
```
marandaneto: What's the reason for not using it's type? `String` in this case.

marandaneto:
@ueman can you share a link of an issue after those changes or a screenshot? thanks.
Looks like this PR indeed makes it much better.
Logic wise LGTM, @brustolin please review :)

ueman:
`await dio.get<String>('https://flutter.dev/asdfasg');`, throws because 404 status code: https://sentry.io/organizations/sentry-sdks/issues/3007144862/events/31495690091f4d7684c339e6fd9a6ebc/?project=5428562

`await dio.post<String>('https://flutter.dev/', data: Exception());`, throws because `Exception` can't be serialized: https://sentry.io/organizations/sentry-sdks/issues/3009008663/events/fcb1d27e11be4af4b71b25b6c04128c6/?project=5428562

brustolin:
@marandaneto, as you said, code looks good. 
But I don't really know how EventProcessor works in the Dart SDK, need to get familiar with this.

ueman:
> @marandaneto, as you said, code looks good. But I don't really know how EventProcessor works in the Dart SDK, need to get familiar with this.

Feel free to ask (here or on Discord), I'll probably can answer at least some questions.

marandaneto:
@ueman sentry-dio / package-analysis is unhappy, otherwise LGTM.

marandaneto (approved):
@ueman thanks for making the `dio` integration better :)

marandaneto:
@kuhnroyal also thanks for the valuable tests and insights.

ueman:
I would guess it's because pana doesn't check the dependency overrides? The getter which pana complains about isn't available in the latest stable but on main. I don't think I can do anything about that.

Pana isn't happy because of:

------

### [x] 0/30 points: code has no errors, warnings, lints, or formatting issues

Found 3 issues. Showing the first 2:<details>
<summary>
ERROR: The getter 'exceptionFactory' isn't defined for the type 'SentryOptions'.
</summary>`lib/src/dio_event_processor.dart:20:16````
   ╷
20 │       _options.exceptionFactory;
   │                ^^^^^^^^^^^^^^^^
   ╵
```
To reproduce make sure you are using the [lints_core](https://pub.dev/packages/lints) and run `dart analyze lib/src/dio_event_processor.dart`
</details>
<details>
<summary>
ERROR: The method 'shouldAddBody' isn't defined for the type 'MaxRequestBodySize'.
</summary>
`lib/src/dio_event_processor.dart:127:31`
```
    ╷
127 │       if (_maxRequestBodySize.shouldAddBody(data.codeUnits.length)) {
    │                               ^^^^^^^^^^^^^
    ╵
```
To reproduce make sure you are using the [lints_core](https://pub.dev/packages/lints) and run `dart analyze lib/src/dio_event_processor.dart`
</details>

------

File: dio/lib/src/dio_event_processor.dart [hunk 17]
```diff
@@ -0,0 +1,137 @@
+import 'dart:async';
+
+import 'package:dio/dio.dart';
+import 'package:sentry/sentry.dart';
+// ignore: implementation_imports
```
marandaneto, line 5: @ueman is that because `sentry_exception_factory` isn't exported?

dio/lib/src/dio_event_processor.dart
```diff
@@ -0,0 +1,137 @@
...
+class DioEventProcessor implements EventProcessor {
+  /// This is an [EventProcessor], which improves crash reports of [DioError]s.
+  DioEventProcessor(this._options, this._maxRequestBodySize);
+
+  final SentryOptions _options;
+  final MaxRequestBodySize _maxRequestBodySize;
+
+  SentryExceptionFactory get _sentryExceptionFactory =>
+      // ignore: invalid_use_of_internal_member
+      _options.exceptionFactory;
```
marandaneto, line 20: Maybe if we use `_options.exceptionFactory` directly, we avoid this linter ignore.

dio/lib/src/sentry_dio_extension.dart
```diff
@@ -16,10 +17,20 @@ extension SentryDioExtension on Dio {
     bool recordBreadcrumbs = true,
     bool networkTracing = true,
     MaxRequestBodySize maxRequestBodySize = MaxRequestBodySize.never,
-    List<SentryStatusCode> failedRequestStatusCodes = const [],
```
marandaneto, line 19: The reason for that is because `dio` itself already throws in some cases, right?

dio/lib/src/dio_event_processor.dart
ueman: I don't think so. The error message says that the getter do not exist, which is true for 6.3.0 but at the current tip of main it exists. Also, locally I don't get any errors/warnings. Only if I run it against the current stable I get the same warnings.

dio/lib/src/sentry_dio_extension.dart
ueman: Yes, exactly

dio/lib/src/dio_event_processor.dart
marandaneto: Adding `export 'src/sentry_exception_factory.dart';` to `sentry.dart` fixes the issue, since its a property on SentryOptions, its fine to be there too (Tested it locally).

dio/lib/src/dio_event_processor.dart
marandaneto: Nvm, the problem would still exist, Since its `@internal`, the linter only likes it if used within its own package.

marandaneto:
> I would guess it's because pana doesn't check the dependency overrides? The getter which pana complains about isn't available in the latest stable but on main. I don't think I can do anything about that.
> 
> Pana isn't happy because of:
> 
> ### [x] 0/30 points: code has no errors, warnings, lints, or formatting issues
> Found 3 issues. Showing the first 2:
> 
> ERROR: The getter 'exceptionFactory' isn't defined for the type 'SentryOptions'.
> ERROR: The method 'shouldAddBody' isn't defined for the type 'MaxRequestBodySize'.

Indeed, I believe `pana` should respect `dependency_overrides` or at least offer an option to respect it since multi-module projects would run into this issue very often.
Maybe worth filing an issue on https://github.com/dart-lang/pana/ as a feature request.
Let's merge and not block on this then :)

marandaneto:
I'm fine if you want to address https://github.com/getsentry/sentry-dart/pull/718#discussion_r807145080 or keep as it is, your call.
I'm good to merge as it is.

dio/lib/src/dio_event_processor.dart
ueman: I would say we keep it as is.

PR #722, opened 2022-01-26
Dio extension flexible

Overman775:
## :scroll: Description

I made the extension more flexible and added the main objects to the export

Sometimes projects have custom transformers and adapters, for example, an http2 adapter or compute transformer for jsons.
For convenience, need to pass the `HttpClientAdapter` and `Transformer` ​​​​directly in function `addSentry`
And sometimes it's not convenient to use the constructor, you need the ability to directly set dio options

Check Dio extension

## :pencil: Checklist

- [ ] All tests passing
- [X] No breaking changes

### DIFF
--- a/dio/lib/sentry_dio.dart
+++ b/dio/lib/sentry_dio.dart
@@ -1,3 +1,5 @@
 library sentry_dio;
 
 export 'src/sentry_dio_extension.dart';
+export 'src/sentry_dio_client_adapter.dart';
+export 'src/sentry_transformer.dart';
--- a/dio/lib/src/sentry_dio_extension.dart
+++ b/dio/lib/src/sentry_dio_extension.dart
@@ -16,10 +16,12 @@ extension SentryDioExtension on Dio {
     List<SentryStatusCode> failedRequestStatusCodes = const [],
     bool captureFailedRequests = false,
     bool sendDefaultPii = false,
+    HttpClientAdapter? httpClientAdapter,
+    Transformer? transformer,
   }) {
     // intercept http requests
-    httpClientAdapter = SentryDioClientAdapter(
-      client: httpClientAdapter,
+    this.httpClientAdapter = SentryDioClientAdapter(
+      client: httpClientAdapter ?? this.httpClientAdapter,
       recordBreadcrumbs: recordBreadcrumbs,
       networkTracing: networkTracing,
       maxRequestBodySize: maxRequestBodySize,
@@ -29,6 +31,7 @@ extension SentryDioExtension on Dio {
     );
 
     // intercept transformations
-    transformer = SentryTransformer(transformer: transformer);
+    this.transformer =
+        SentryTransformer(transformer: transformer ?? this.transformer);
   }
 }


ueman:
I'm wondering what your goal is with this PR. You don't need to pass the adapter and transformer as they're not overridden. Both, the current adapter and transformer are used even after adding the Sentry integration. They don't get changed.

Do you mind explaining in more detail?

brustolin (commented):
I kind like this flexibility, but we need some tests.

ueman:
> I kind like this flexibility

The thing is, it already works like this.

So (as suggested in this PR)
```dart
final dio = Dio();
dio.addSentry(adapter: SomeAdapter());
```

is equivalent to the following code, which is already possible 

```dart
final dio = Dio();
dio.adapter = SomeAdapter();
dio.addSentry();
```
This example is also closer to what Dio itself suggests doing, if you want your own custom adapter.

I'm kinda against this change because it's not that intuitive that

```dart
final dio = Dio();
dio.adapter = SomeAdapter();
dio.addSentry(adapter: FooAdapter());
```
overrides the adapter which was previously set.

If you want this change anyway, I recommend adding a warning if the adapter which was previously set is replaced by another adapter.

Same thing is also true for the transformer.

brustolin:
@ueman Thanks for the explanation. 
The override part is really a problem.

marandaneto:
> I'm wondering what your goal is with this PR. You don't need to pass the adapter and transformer as they're not overridden. Both, the current adapter and transformer are used even after adding the Sentry integration. They don't get changed.
> 
> Do you mind explaining in more detail?

Yep, mind clarifying @Overman775 ?

Overman775:
> I'm wondering what your goal is with this PR. You don't need to pass the adapter and transformer as they're not overridden. Both, the current adapter and transformer are used even after adding the Sentry integration. They don't get changed.
> 
> Do you mind explaining in more detail?

They are overridden implicitly, which is a bit confusing.
If we first add via the Sentry extension, and then install the adapter, we will get a bug. For example:

```
final dio = Dio();
dio.addSentry();
dio.adapter = SomeAdapter();
```

Maybe remove the extension and make it like Wrapper pattern? This approach is more explicit
```
dio.adapter = SentryAdapterWrapper(adapter: dio.adapter);
dio.transformer = SentryTransformerWrapper(dio.transformer)
```

@ueman @brustolin That do you think?

ueman:
> ```
> final dio = Dio();
> dio.addSentry();
> dio.adapter = SomeAdapter();
> ```

The docs state that `dio.addSentry()` should be the last step of the Dio configuration.
See https://github.com/getsentry/sentry-dart/blob/17894ef526b677f59b7f7e43bae61146a6653b39/dio/lib/src/sentry_dio_extension.dart#L8-L11

Maybe we should add that part to the readme and make it more explicit.

Overman775:
> Maybe we should add that part to the readme and make it more explicit.

In my experience, not everyone reads the documentation 😃

brustolin:
For me, this example is clearly replacing the adapter, therefore removing sentry.
> ```
> final dio = Dio();
> dio.addSentry();
> dio.adapter = SomeAdapter();
> ```

I understand that almost no one reads the documentation, but as @ueman pointed out before, the proposal can also lead to some confusion.

We need to choose the less confusing option.

Overman775:
Thanks for your time guys 😃 , closing PR

PR #726, opened 2022-01-29
Add `Sentry.runWithClone` and `Sentry.runWithSpan`

### LINKED ISSUE #724: Support for binding hubs and scopes to zones
I would like to be able to create a `Hub` and make it replace the root `Hub` within a Dart `Zone`, and similarly make a span the current span within a `Zone`.

Creating composable functions, that create child spans to trace themselves, requires manually passing the current parent span around. Binding a span to the scope of the root `Hub` does not work for asynchronous code. Multiple asynchronous transactions can be ongoing at the same time and interleave each other. 

Here is a bit of code to demonstrate what I would like to be able to do:

```dart
Future<T> runAppTransaction<T>(String name, Future<T> Function() fn) async {
  final span = Sentry.startTransaction(name, 'task');

  try {
    return await Sentry.runWithSpan(span, fn);
  } catch (e) {
    span.throwable = e;
    span.status = const SpanStatus.internalError();
    rethrow;
  } finally {
    await span.finish();
  }
}

Future<T> runAppOperation<T>(String name, Future<T> Function() fn) async {
  final span = Sentry.getSpan()!.startChild(name);

  try {
    return await Sentry.runWithSpan(span, fn);
  } catch (e) {
    span.throwable = e;
    span.status = const SpanStatus.internalError();
    rethrow;
  } finally {
    await span.finish();
  }
}

final httpClient = SentryHttpClient(networkTracing: true);

Future<String> fetchFooModel() =>
    // We can transparently trace this function, without having to pass in the parent span.
    runAppOperation('get.foo', () async {
      final response =
          // This HTTP request will be transparently traced as a child span of the `get.foo` span.
          await httpClient.get(Uri.parse('https://example.com/foo'));
      if (response.statusCode != 200) {
        throw Exception('Failed to fetch foo.');
      }
      return Foo.fromJson(jsonDecode(response.body));
    });

Future<void> printFooName() async =>
    await runAppTransaction('printFooName', () async {
      final foo = await fetchFooModel();
      print(foo.name);
    });

Future<void> main() => Sentry.runWithClone(() async {
      Sentry.configureScope((scope) {
        scope.setTag('printing', 'foo.name');
      });

      await printFooName();
    });
```

I'd be happy to work on a PR to implement this.

**Update**: Modified example to show how functions from PR #726 would be used.

blaugold:
## :scroll: Description

This change adds support for using Sentry with Dart `Zone`s.

`Sentry.runWithClone` allows users to run code in a new `Zone`, in which a clone of the current `Hub` becomes the current `Hub`. This is similar to how [Kotlin coroutines](https://docs.sentry.io/platforms/java/enriching-events/scopes/#kotlin-coroutines) have their own `Hub`s.

`Sentry.runWithSpan` allows users to run code in new `Zone`, in which a provided
span become the current span. In this `Zone` calls to `Sentry.getSpan` and `Hub.getSpan` return this span.

Let me know what you think!

Closes #724

I added unit tests.

## :pencil: Checklist

- [ ] All tests passing
- [x] No breaking changes

### DIFF
--- a/dart/lib/src/hub.dart
+++ b/dart/lib/src/hub.dart
@@ -433,7 +433,7 @@ class Hub {
   }
 
   /// Runs a function in a new [Zone] in which the given [span] becomes the
-  /// current [ISentrySpan].
+  /// current span.
   ///
   /// In the new [Zone], calls to [getSpan] will return the given [span].
   T runWithSpan<T>(ISentrySpan span, T Function() fn) =>
--- a/dart/lib/src/hub.dart
+++ b/dart/lib/src/hub.dart
@@ -420,13 +420,24 @@ class Hub {
         "Tracing is disabled and this 'getSpan' returns null.",
       );
     } else {
-      final item = _peek();
+      span = Zone.current[#_span] as ISentrySpan?;
 
-      span = item.scope.span;
+      if (span == null) {
+        final item = _peek();
+
+        span = item.scope.span;
+      }
     }
 
     return span;
   }
+
+  /// Runs a function in a new [Zone] in which the given [span] becomes the
+  /// current [ISentrySpan].
+  ///
+  /// In the new [Zone], calls to [getSpan] will return the given [span].
+  T runWithSpan<T>(ISentrySpan span, T Function() fn) =>
+      runZoned(fn, zoneValues: {#_span: span});
 
   @internal
   Future<SentryId> captureTransaction(SentryTransaction transaction) async {
--- a/dart/lib/src/hub_adapter.dart
+++ b/dart/lib/src/hub_adapter.dart
@@ -94,6 +94,10 @@ class HubAdapter implements Hub {
   ISentrySpan? getSpan() => Sentry.currentHub.getSpan();
 
   @override
+  T runWithSpan<T>(ISentrySpan span, T Function() fn) =>
+      Sentry.runWithSpan(span, fn);
+
+  @override
   Future captureUserFeedback(SentryUserFeedback userFeedback) =>
       Sentry.captureUserFeedback(userFeedback);
 
--- a/dart/lib/src/noop_hub.dart
+++ b/dart/lib/src/noop_hub.dart
@@ -1,4 +1,5 @@
 import 'dart:async';
+
 import 'hub.dart';
 import 'protocol.dart';
 import 'sentry_client.dart';
@@ -98,6 +99,9 @@ class NoOpHub implements Hub {
 
   @override
   ISentrySpan? getSpan() => null;
+
+  @override
+  T runWithSpan<T>(ISentrySpan span, T Function() fn) => fn();
 
   @override
   void setSpanContext(throwable, ISentrySpan span, String transaction) {}
--- a/dart/lib/src/sentry.dart
+++ b/dart/lib/src/sentry.dart
@@ -8,13 +8,13 @@ import 'environment/environment_variables.dart';
 import 'event_processor/deduplication_event_processor.dart';
 import 'hub.dart';
 import 'hub_adapter.dart';
+import 'integration.dart';
+import 'noop_hub.dart';
 import 'noop_isolate_error_integration.dart'
     if (dart.library.io) 'isolate_error_integration.dart';
-import 'noop_hub.dart';
 import 'protocol.dart';
 import 'sentry_client.dart';
 import 'sentry_options.dart';
-import 'integration.dart';
 import 'sentry_user_feedback.dart';
 import 'tracing.dart';
 
@@ -142,7 +142,7 @@ class Sentry {
     dynamic hint,
     ScopeCallback? withScope,
   }) =>
-      _hub.captureEvent(
+      currentHub.captureEvent(
         event,
         stackTrace: stackTrace,
         hint: hint,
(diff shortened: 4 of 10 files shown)


dart/lib/src/hub.dart
```diff
@@ -420,14 +420,25 @@ class Hub {
         "Tracing is disabled and this 'getSpan' returns null.",
       );
     } else {
-      final item = _peek();
+      span = Zone.current[#_span] as ISentrySpan?;
```
brustolin, line 423: l: I wouldn't use `Symbol ` as key since `Symbol` are more related to reflection. Also I don't believe I've seen it being used this way on the SDK.

ueman: This is not reflection, that's just how you add something to a zone. https://dart.dev/articles/archive/zones#storing-zone-local-values

But the key should probably have a `sentry` prefix, as the keys are global for a zone.

blaugold: Prefixing a [symbol literal with `_`](https://api.dart.dev/stable/2.15.1/dart-core/Symbol/Symbol.html) makes them file private, just like other identifiers.

brustolin (commented):
From the code point of view, it looks ok to me.

I would like @marandaneto or @bruno-garcia  to review the implications on the SDK.

dart/lib/src/hub.dart
ueman: Huh, TIL. I thought that if any other code also adds a `_span` identifier, this one is overridden because it's a key after all?

blaugold: Private symbol literals create unique symbols in each file in which they are used, so `#_span` in file `a.dart` is not the same symbol as `#_span` in file `b.dart`. If those symbols are used as keys in a map, they won't override each other since they have unique identities.

ueman: Thanks for the explanation

brustolin: @ueman I understand that there is no reflection involved. But the key could be any object that implements `==`  and `hashCode`.

In the `Symbol` [documentation](https://api.dart.dev/stable/2.15.1/dart-core/Symbol-class.html) you can see that they are related to reflections. Not that they could not be used in other places, but that is their purpose. But I like the uniqueness by file that @blaugold mentioned.

kuhnroyal:
This is perfect, was looking for exactly this. Gonna link my issues to this PR.

kuhnroyal:
I will try to rebase this and see if I can test it. I pretty much implemented the same thing last night :(

One thing I notice is that the zones create with `runWithClone` don't use the correct hub and therefore the wrong context for uncaught errors.

In my tests I also registered a new uncaught error handler which uses the cloned/current hub. I assume this is in most cases what you want because otherwise uncaught errors don't have the correct scope.

When keeping the naming convention of `Zone`, we could go with something like this:
`Sentry.runZoned()` - creates a new zone without a new uncaught error handler- uncaught errors will bubble up to the next parent with an uncaught handler
`Sentry.runZonedGuarded()` - creates a new uncaught error handler with the cloned/current Hub (basically what is done in `RunZonedGuardedIntegration`)

A couple more thoughts:
* it should be possible to pass an custom `ZoneSpecification` to these functions
* this is currently non-breaking which is good but I would suggest removing the `RunZonedGuardedIntegration` in the future - it is hard to customize and work with, using something like `Sentry.runZonedGuarded()` with smart defaults is a lot easier to use and understand

ueman:
> When keeping the naming convention of `Zone`, we could go with something like this: `Sentry.runZoned()` - creates a new zone without a new uncaught error handler - uncaught errors will bubble up to the next parent with an uncaught handler `Sentry.runZonedGuarded()` - creates a new uncaught error handler with the cloned/current Hub (basically what is done in `RunZonedGuardedIntegration`)

I like this idea. That also makes clear that it uses `Zone`s for it to work and it let's people avoid it if they don't want to bother with Zones

blaugold:
Closing this since it's probably not the right approach.

I think something like `Hub::with_scope(callback)` and maybe also `Hub::run(hub, callback)` as described in the [Unified API](https://develop.sentry.dev/sdk/unified-api/) would be more appropriate.

kuhnroyal:
But that is effectively what the implementation does, the rest is just naming.

ueman:
Especially, since there's basically no other way to isolate (Not Darts Isolates) Sentrys scope.

kuhnroyal:
> Especially, since there's basically no other way to isolate (Not Darts Isolates) Sentrys scope.

Yea I think Zones are pretty much the only way to go.

blaugold:
I still think Sentry should have support for `Zone`s, but with a different API that is better aligned with the Unified API. This PR is getting stale, and I would rather start fresh after settling on an API with the maintainers.

`Sentry.runWithClone` forces the user to clone the current `Hub`. The Unified API suggests taking a `Hub` from the users, which makes for a more flexible API. Users can already create `Hub`s with `Sentry.clone`. It would also make sense for `Sentry.currentHub` to be public and return the `Hub` bound to the current `Zone`.

`Sentry.runWithSpan` is unnecessarily limiting by only binding a `Span` to a `Zone`. Instead, Sentry could support pushing and binding a `Scope` to a `Zone`. That would be in line with the Unified API.

Implementing something like `Hub::with_scope(callback)` looks more complicated than the `Sentry.runWithSpan` implementation in this PR. It probably makes sense to implement support for binding `Hub`s and `Scope`s to `Zone`s in two separate PRs.

#724 would be a good place to discuss a concrete API.

PR #727, opened 2022-01-30
ref: Change how factories are accessed

ueman:
## :scroll: Description
This changes how the exception and stacktrace factories are accessed.

__#skip-changelog__

Needed for https://github.com/getsentry/sentry-dart/pull/718

This is covered by the existing tests

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/dart/lib/src/sentry_client.dart
+++ b/dart/lib/src/sentry_client.dart
@@ -28,9 +28,9 @@ class SentryClient {
 
   static final _sentryId = Future.value(SentryId.empty());
 
-  late SentryExceptionFactory _exceptionFactory;
+  SentryExceptionFactory get _exceptionFactory => _options.exceptionFactory;
 
-  late SentryStackTraceFactory _stackTraceFactory;
+  SentryStackTraceFactory get _stackTraceFactory => _options.stackTraceFactory;
 
   /// Instantiates a client using [SentryOptions]
   factory SentryClient(SentryOptions options) {
@@ -43,10 +43,7 @@ class SentryClient {
 
   /// Instantiates a client using [SentryOptions]
   SentryClient._(this._options)
-      : _random = _options.sampleRate == null ? null : Random() {
-    _stackTraceFactory = _options.stackTraceFactory;
-    _exceptionFactory = _options.exceptionFactory;
-  }
+      : _random = _options.sampleRate == null ? null : Random();
 
   /// Reports an [event] to Sentry.io.
   Future<SentryId> captureEvent(
--- a/dart/lib/src/sentry_client.dart
+++ b/dart/lib/src/sentry_client.dart
@@ -44,11 +44,8 @@ class SentryClient {
   /// Instantiates a client using [SentryOptions]
   SentryClient._(this._options)
       : _random = _options.sampleRate == null ? null : Random() {
-    _stackTraceFactory = SentryStackTraceFactory(_options);
-    _exceptionFactory = SentryExceptionFactory(
-      _options,
-      _stackTraceFactory,
-    );
+    _stackTraceFactory = _options.stackTraceFactory;
+    _exceptionFactory = _options.exceptionFactory;
   }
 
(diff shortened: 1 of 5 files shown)


dart/lib/src/sentry_stack_trace_factory.dart
```diff
@@ -13,6 +13,13 @@ class SentryStackTraceFactory {
   static const _stackTraceViolateDartStandard =
       'This VM has been configured to produce stack traces that violate the Dart standard.';
 
+  static const _sentryPackagesIdentifier = <String>[
+    'sentry',
```
ueman, line 17: I've noticed that the other packages aren't checked, so I've also added this.

kuhnroyal: Could this use registered `options.sdk.packages`? (`pub:sentry_logging`)

ueman: Yeah, probably. @marandaneto how do you handle this in other SDKs?

brustolin (commented):
Hey @ueman. Could you elaborate why do you believe removing the `_stackTraceFactory`and `_exceptionFactory` from SentryClient and putting it on SentryOptions is a better approach? 
I'm not a fan of SentryOptions being this kind of objects hub for the SDK.

Thanks!!

ueman:
Sure, this came out of a discussion at https://github.com/getsentry/sentry-dart/pull/718#discussion_r792822787 . We need those factories for [this PR](https://github.com/getsentry/sentry-dart/pull/718) in an event processor and it was suggested to go with this approach. I don't have preference on which approach to take, I'll  need to know which one I should take.

brustolin:
@ueman, my bad, I should have read #718 first. 

I believe @marandaneto is on board with it. If this is the simplest solution I'm ok.

Thanks!!

dart/lib/src/sentry_stack_trace_factory.dart
marandaneto: Not really, the `sdk.packages` are the identifier that connects with https://github.com/getsentry/sentry-release-registry, Our docs read this repo to get the latest versions, Sentry UI also to see if you're not using outdated packages

dart/lib/src/sentry_options.dart
```diff
@@ -297,6 +300,14 @@ class SentryOptions {
   bool isTracingEnabled() {
     return tracesSampleRate != null || tracesSampler != null;
   }
+
+  @internal
+  late SentryExceptionFactory exceptionFactory =
+      SentryExceptionFactory(this, stackTraceFactory);
```
marandaneto: Since `stackTraceFactory` now can be accessed via `options.stackTraceFactory`, we don't need to pass `stackTraceFactory` anymore.

dart/lib/src/sentry_client.dart
```diff
@@ -44,11 +44,8 @@ class SentryClient {
   /// Instantiates a client using [SentryOptions]
   SentryClient._(this._options)
       : _random = _options.sampleRate == null ? null : Random() {
-    _stackTraceFactory = SentryStackTraceFactory(_options);
-    _exceptionFactory = SentryExceptionFactory(
-      _options,
-      _stackTraceFactory,
-    );
+    _stackTraceFactory = _options.stackTraceFactory;
+    _exceptionFactory = _options.exceptionFactory;
```
marandaneto: You can remove `_stackTraceFactory` and `_exceptionFactory` and read them directly from the `_options` right now, no need to have references around.

marandaneto: Or make the properties a `get` pointing to `_options`

marandaneto (approved):
@ueman left 2 comments but I am pre-approving it.
@brustolin please double check it before merging it.

Thanks again @ueman :)

brustolin (approved):
LGTM

PR #728, opened 2022-01-30
[Dio] Replace failed http client adapter with interceptor 

ueman:
## :scroll: Description
This PR removes the failed http client adapter from the Dio integrations.
There's no need for this functionality, as Dio can be configured to throw on certain status codes. The users are probably using Dios way, which would cause having to reports for a bad status code. For reference, see [`ValidateStatus`](https://pub.dev/packages/dio#request-options) here. So instead we're adding an interceptor which can listen for exceptions.
We automatically listen for those exceptions and report them.

The reporting of DioErrors gets improved in https://github.com/getsentry/sentry-dart/pull/718

This came out of the discussion over at https://github.com/getsentry/sentry-dart/pull/718#discussion_r793087381

New tests

## :pencil: Checklist

- [x] All tests passing
- [ ] No breaking changes


brustolin (commented):
I agree with this decision.

marandaneto:
@ueman @brustolin 

> Dio can be configured to throw on certain status codes

That's fine, they are free to disable `captureFailedRequests` or to configure `failedRequestStatusCodes` properly.

> The users are probably using Dios way

How is `dio` way? That still requires wrapping on try/catch blocks and calling `Sentry.captureException(...)` manually, or letting uncaught exceptions bubble up on the uncaught error handler, which would not be personalized with headers, request data, etc...

> which would cause having to reports for a bad status code.

Only if they are using both features at the same time, and not configuring one or the other to avoid duplicate events.

We're pretty much removing a feature that captures errors automatically based on the user's configuration, without any other alternative, unless I'm missing something here.

ueman:
> How is `dio` way? That still requires wrapping on try/catch blocks and calling `Sentry.captureException(...)` manually, or letting uncaught exceptions bubble up on the uncaught error handler, which would not be personalized with headers, request data, etc...

That's what https://github.com/getsentry/sentry-dart/pull/718 fixes.

marandaneto:
> > How is `dio` way? That still requires wrapping on try/catch blocks and calling `Sentry.captureException(...)` manually, or letting uncaught exceptions bubble up on the uncaught error handler, which would not be personalized with headers, request data, etc...
> 
> That's what #718 fixes.

Yep, got that, but event processors only run if there's a captured event, and with the removal of  `FailedRequestClientAdapter`, users would need to do it manually.

kuhnroyal:
Wondering again if we should consider moving to an interceptor approach, which probably would solve most cases.

brustolin:
> > > How is `dio` way? That still requires wrapping on try/catch blocks and calling `Sentry.captureException(...)` manually, or letting uncaught exceptions bubble up on the uncaught error handler, which would not be personalized with headers, request data, etc...
> > 
> > 
> > That's what #718 fixes.
> 
> Yep, got that, but event processors only run if there's a captured event, and with the removal of `FailedRequestClientAdapter`, users would need to do it manually.

Didn't thought that. Capturing crash manually would be bad.

I like @kuhnroyal suggestion. Using interceptors we could capture the error but let the request continue its flow.

brustolin (approved):
LGTM

dio/lib/src/failed_request_interceptor.dart
```diff
@@ -0,0 +1,21 @@
...
+
+  final Hub _hub;
+
+  @override
+  void onError(
+    DioError err,
+    ErrorInterceptorHandler handler,
+  ) {
+    _hub.captureException(err);
+    _hub.getSpan()?.throwable = err;
```
marandaneto:
```suggestion
    _hub.getSpan()?.throwable = err;
    _hub.captureException(err);
```

dio/lib/src/failed_request_interceptor.dart
```diff
@@ -0,0 +1,21 @@
...
+  FailedRequestInterceptor({Hub? hub}) : _hub = hub ?? HubAdapter();
+
+  final Hub _hub;
+
+  @override
+  void onError(
+    DioError err,
+    ErrorInterceptorHandler handler,
+  ) {
+    _hub.captureException(err);
```
marandaneto: Should we rather convert the `DioError` into SentryEvent with its meta data? e.g. `SentryRequest`, `Mechanism`, etc, similar to `FailedRequestClientAdapter`, I believe it still makes sense to offer the `maxRequestBodySize`, `failedRequestStatusCodes`, and `sendDefaultPii`

dio/test/failed_request_interceptor_test.dart
```diff
@@ -0,0 +1,46 @@
...
+class MockedErrorInterceptorHandler implements ErrorInterceptorHandler {
+  bool nextWasCalled = false;
+
+  @override
+  void next(DioError err) {
+    nextWasCalled = true;
+  }
+
+  @override
+  void noSuchMethod(Invocation invocation) {}
```
marandaneto, line 45: What's about raising a new issue to overwrite every "empty" method from our mocks with `noSuchMethod`?

dio/lib/src/failed_request_interceptor.dart
ueman: What's the reason for calling one before the other?

dio/lib/src/failed_request_interceptor.dart
ueman: That's what #718 is for. So even manually captured DioErrors get that benefit.

ueman: `failedRequestStatusCodes` is done by `ValidateStatus` from dio. If the status code is bad, Dio throws an exception which the interceptor reports.

marandaneto: oh yeah that makes sense.

dio/lib/src/failed_request_interceptor.dart
marandaneto: The event captured by `captureException` will contain the `event.contexts.trace` data from the span associated with the previously assigned `throwable`.

ueman: Oh, yeah, in that case it should definitely be changed.

dio/test/failed_request_interceptor_test.dart
ueman: https://github.com/getsentry/sentry-dart/issues/747

marandaneto (approved):

ueman:
Can this be merged, so I can integrate it to the other Dio PR?

Issue #691, opened 2022-01-06
Reduce logging to console.


robertflesch:
Specifically Flutter Android or iOS
I love Sentry, it has helped me a number of times in the last two weeks, BUT

Currently Sentry spews TONs of data to my console, so much so that it makes seeing my own data very difficult.
While I understand some of this info might be useful, I would prefer not seeing intermediate steps, really I just want to see problems that a plugin has encountered, not internal data used by developers.

ueman:
You should be able to set `options.debug` to false. I'm not sure if it disables logs completely, but it should at least reduce logs significantly.

kuhnroyal:
Should seriously consider making this default :)

robertflesch:
I agree that debug = false should be default.

ueman:
> I agree that debug = false should be default.

@robertflesch do you mind opening a PR for that?

AliAkberAakash:
Is this available? I'd like to make a PR

marandaneto:
@AliAkberAakash please do it :)

AliAkberAakash:
I can't seem to find it, It is already set to false in SentryOptions. Could you help me locate this please? @marandaneto

marandaneto:
@AliAkberAakash We'd need to remove this -> https://github.com/getsentry/sentry-dart/blob/main/dart/lib/src/sentry_options.dart#L270-L275

PR #753, opened 2022-02-10
fix: disable log by default in debug mode

AliAkberAakash:
## :scroll: Description
Reduce logging to console

Fix of issue #691 

## :pencil: Checklist

- [ ] All tests passing
- [x] No breaking changes


marandaneto:
@AliAkberAakash thanks for doing this.
Please add an entry to the changelog and fix possible broken tests.

AliAkberAakash:
I have added entry to change log and fixed broken tests. @marandaneto

CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 # Unreleased
 
+* Fix: Disable log by default in debug mode
```
```suggestion
* Fix: Disable log by default in debug mode (#753)
```

marandaneto:
@AliAkberAakash 

>    info • Unused import: 'package:flutter/foundation.dart' • test/sentry_flutter_test.dart:1:8 • unused_import

AliAkberAakash:
@marandaneto Hi, I have pushed the suggested changes, did u have a chance to look at it?

marandaneto:
> @marandaneto Hi, I have pushed the suggested changes, did u have a chance to look at it?

Hi, There are still failing tests, see https://github.com/getsentry/sentry-dart/runs/5179416915?check_suite_focus=true
Is it related to your PR?

brustolin:
My problem with this PR is that it changes a behavior already known to anyone using the SDK. And the solution for those who don't want logs is simple, just disable it in the options.

I would prefer to add a `debug` parameter to the SentryOptions constructor. Even though in flutter it is easy to initialize properties like this:

```dart
SentryOptions()..debug = false
```

AliAkberAakash:
I'm not sure if this test is required anymore? or should this now check if the logger is noOpLogger right from the start? @marandaneto 
https://github.com/getsentry/sentry-dart/blob/a3d70071172e450a7afe5f21e4215493e02035e3/dart/test/sentry_test.dart#L320

marandaneto:
> I'm not sure if this test is required anymore? or should this now check if the logger is noOpLogger right from the start? @marandaneto
> 
> https://github.com/getsentry/sentry-dart/blob/a3d70071172e450a7afe5f21e4215493e02035e3/dart/test/sentry_test.dart#L320

Since `debug` is false now, you have to add this `options.debug = true;` between L325 and L326

marandaneto:
> My problem with this PR is that it changes a behavior already known to anyone using the SDK. And the solution for those who don't want logs is simple, just disable it in the options.
> 
> I would prefer to add a `debug` parameter to the SentryOptions constructor. Even though in flutter it is easy to initialize properties like this:
> 
> ```dart
> SentryOptions()..debug = false
> ```

Indeed, but that was the only SDK doing so, enabling `debug` by default, and it got too noisy, users can still `debug = true` manually, this won't affect production Apps but rather debugging only.

marandaneto (approved):

PR #758, opened 2022-02-17
ci: Add release workflow for Craft automation

kamilogorek:
There is 1 pending PR that need to be merged before letting this one in:

https://github.com/getsentry/craft/pull/353

#skip-changelog (this change is not relevant for users)


scripts/bump-version.sh
```diff
@@ -0,0 +1,21 @@
...
+
+OLD_VERSION="${1}"
+NEW_VERSION="${2}"
+
+echo "Current version: ${OLD_VERSION}"
+echo "Bumping version: ${NEW_VERSION}"
+
+for pkg in {dart,flutter,logging,dio}; do
+  # Bump version in pubspec.yaml
+  perl -pi -e "s/^version: .*/version: $NEW_VERSION/" $pkg/pubspec.yaml
```
marandaneto, line 15: Can we do the very same, outside of the `for` loop, for the `flutter/example/pubspec.yaml` file as well?

marandaneto:
Missing `release/**` on/push/branches from GH action files

marandaneto:
Since `dio` and `logging` dont exist on https://github.com/getsentry/sentry-release-registry/tree/master/packages/pub
It's likely going to fail for GA releases.

scripts/bump-version.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+set -euo pipefail
```
marandaneto, line 2: Should this script also address the `dependency_overrides` problem?
To be able to release a version, These lines https://github.com/getsentry/sentry-dart/blob/main/flutter/pubspec.yaml#L29-L31 should be commented out, and after release (Maybe via post-release script?) should be back again.

kamilogorek:
> Missing release/** on/push/branches from GH action files

Added

> Since dio and logging dont exist on https://github.com/getsentry/sentry-release-registry/tree/master/packages/pub
It's likely going to fail for GA releases.

We only release `sentry` and `sentry_flutter` to the registry, or am I missing something?

scripts/bump-version.sh
kamilogorek: Added. Although its "meh" solution, as it will always append removed `dependencies_override` to the end of the file. Shouldn't be too bad though.

marandaneto:
> > Missing release/** on/push/branches from GH action files
> 
> Added
> 
> > Since dio and logging dont exist on [getsentry/sentry-release-registry@`master`/packages/pub](https://github.com/getsentry/sentry-release-registry/tree/master/packages/pub?rgh-link-date=2022-02-18T14%3A42%3A49Z)
> > It's likely going to fail for GA releases.
> 
> We only release `sentry` and `sentry_flutter` to the registry, or am I missing something?

Yes, but only because they were contributed by the community (dio and logging) and there are no docs yet, but they should be in the release registry anyway, we can do that later tho, if Craft does not fail because of that.

scripts/bump-version.sh
marandaneto: yeah unfortunately I didn't find any way around `dependencies_override`, either we comment it out and comment it back before/after release, or we lose the ability to debug all the packages together.

marandaneto (approved):

kamilogorek:
Note that craft `0.29.0` is not published yet, but we can merge it, as its not run automatically.

PR #761, opened 2022-02-20
Fix: Add missing iOS contexts

### LINKED ISSUE #759: missing fallback userId
According to the document, the SDK uses a random `installationId` as a fallback for the user ID.

https://github.com/getsentry/sentry-docs/pull/4201
https://docs.sentry.io/platforms/flutter/enriching-events/identify-user/

However, I could not find an implementation of this and the reported issue seems to be tagged as `Unknown User`.

image

Is this behavior unimplemented? Or if I've missed something, please let me know.

marandaneto:
## :scroll: Description

- Add additional event data taken from native iOS bridge.

Closes https://github.com/getsentry/sentry-dart/issues/759

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes


flutter/ios/Classes/SentryFlutterPluginApple.swift
```diff
@@ -73,17 +73,51 @@ public class SentryFlutterPluginApple: NSObject, FlutterPlugin {
...
 
+            if let tags = serializedScope["tags"] as? [String: String] {
+                infos["tags"] = tags
+            }
+            if let extra = serializedScope["extra"] as? [String: Any] {
+                infos["extra"] = extra
+            }
+            if let user = serializedScope["user"] as? [String: Any] {
+                infos["user"] = user
+            }
```
marandaneto, line 99: @brustolin the `user` here isn't set, so where should I read the installation id from sentry-cocoa?

marandaneto: https://github.com/getsentry/sentry-cocoa/pull/1680

flutter/lib/src/default_integrations.dart
```diff
@@ -195,15 +196,95 @@ class _LoadContextsIntegrationEventProcessor extends EventProcessor {
         event = event.copyWith(contexts: eventContexts);
       }
 
-      if (infos['integrations'] != null) {
-        final integrations = List<String>.from(infos['integrations'] as List);
+      final tagsMap = infos['tags'];
```
marandaneto: tags, extra and fingerprint set on Dart have precedence over the ones set in the native layer if the key exists on both sides

flutter/lib/src/default_integrations.dart
```diff
@@ -195,15 +196,95 @@ class _LoadContextsIntegrationEventProcessor extends EventProcessor {
...
+        }
+        event = event.copyWith(fingerprint: eventFingerprints);
+      }
+
+      final levelString = infos['level'];
+      if (event.level == null && levelString != null) {
+        event = event.copyWith(level: SentryLevel.fromName(levelString));
+      }
+
+      final breadcrumbsList = infos['breadcrumbs'];
```
marandaneto: We need to be sure that we don't duplicate the crumbs, and the timeline of the crumbs should make sense.

dart/lib/src/protocol/sentry_user.dart
```diff
@@ -63,12 +63,18 @@ class SentryUser {
 
   /// Deserializes a [SentryUser] from JSON [Map].
   factory SentryUser.fromJson(Map<String, dynamic> json) {
+    Map<String, dynamic>? extras;
+    if (json['extras'] != null) {
+      extras = Map<String, dynamic>.from(json['extras'] as Map);
+    } else {
+      extras = null;
+    }
```
marandaneto:
```suggestion
    var extras = json['extras'];
    if (extras != null) {
      extras = Map<String, dynamic>.from(extras as Map);
    }
```

I think that would be cleaner

flutter/lib/src/default_integrations.dart
```diff
@@ -195,14 +196,87 @@ class _LoadContextsIntegrationEventProcessor extends EventProcessor {
         event = event.copyWith(contexts: eventContexts);
       }
 
+      final tagsMap = infos['tags'];
+      if (tagsMap != null) {
```
```suggestion
      if (tagsMap?.isNotEmpty ?? false) {
```

That would spare some code to be run, same for other fields.

flutter/lib/src/default_integrations.dart
```diff
@@ -195,14 +196,87 @@ class _LoadContextsIntegrationEventProcessor extends EventProcessor {
...
       final userMap = infos['user'];
       if (event.user == null && userMap != null) {
         final user = Map<String, dynamic>.from(userMap as Map);
         event = event.copyWith(user: SentryUser.fromJson(user));
       }
 
-      if (infos['integrations'] != null) {
-        final integrations = List<String>.from(infos['integrations'] as List);
+      final distString = infos['dist'];
+      if (event.dist == null && distString != null) {
```
marandaneto, line 232:
```suggestion
      if (event.dist == null && distString?.isNotEmpty ?? false) {
```

Same for other fields

flutter/lib/src/default_integrations.dart
denrase: Right now the breadcrumbs form the native bridge are merged with the hybrid ones and sorted by date, is there anything else we need to check?

marandaneto:
@denrase Some tests are failing on CI.

marandaneto:
Screenshot 2022-04-26 at 14 08 39

The navigation and started breadcrumb are not sorted out correctly based on timestamp

denrase:
@marandaneto Hmm, we just do a simple sort on the timestamps. Just tested this and it looks ok to me. 🤔 

Bildschirmfoto 2022-04-27 um 14 29 53

marandaneto:
> @marandaneto Hmm, we just do a simple sort on the timestamps. Just tested this and it looks ok to me. 🤔
> 
> Bildschirmfoto 2022-04-27 um 14 29 53

I could not reproduce it again, so let's ship it.

marandaneto:
@denrase approve it please so I can merge it, since I opened the PR, I cannot approve myself.

denrase (approved):

denrase:
@marandaneto Sry about the delay!

PR #767, opened 2022-02-23
ci: Make sure that CI builds are always run using local sentry dep

kamilogorek:
Crafts pre-publish step removes `dependencies_override` from the `pubspec.yaml` files, which prevents release builds from every succeeding.
This makes sure that we always use local version of sentry, even when it's not overridden in the `yaml`.
Note that it has to be removed first before adding back.

#skip-changelog


kamilogorek:
Eh, this won't do, as `axel-op/dart-package-analyzer` is running its own dart container with its own `dart pub upgrade` that breaks.

PR #770, opened 2022-02-24
ci: Disable dart-package-analyzer on release branches

kamilogorek:
`axel-op/dart-package-analyzer` is using `flutter pub upgrade` instead of `get`, which ignores `pubspec.yaml` `dependency_overrides`.
Because of that, all `release/*` branches are failing, because the package cannot find the "about to be released" version of our sentry-dart package that it depends on.

#skip-changelog

### DIFF
--- a/.github/workflows/dio.yml
+++ b/.github/workflows/dio.yml
@@ -69,6 +69,10 @@ jobs:
           dart format --set-exit-if-changed ./
 
   package-analysis:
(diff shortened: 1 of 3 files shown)


kamilogorek:
@marandaneto we can restore those once we find out how to properly do that.

romtsn (approved):

Issue #636, opened 2021-11-15
e2e tests should run on min. supported sdk versions


marandaneto:
running on the min. supported SDK version, we'd be able to catch back-compatibility during CI run and avoid #634 similar issues

ueman:
Currently you can't run the tests in the Flutter SDK on Flutter 2.0.x up until at least 2.2.3

marandaneto:
> Currently you can't run the tests in the Flutter SDK on Flutter 2.0.x up until at least 2.2.3

why so? do we use features on tests that are not available under the min. supported Dart/Flutter SDK? if so, we'd need to adjust the tests then.

ueman:
It's because of the dev_dependencies having higher requirements than the minimal supported Flutter version.

marandaneto:
> It's because of the dev_dependencies having higher requirements than the minimal supported Flutter version.

Gotcha, we'd need to downgrade then to make it work, it's better to detect issues on CI in advance rather than using the latest tooling on dev deps.

PR #778, opened 2022-03-02
Additional Flutter Context information

kuhnroyal:
## :scroll: Description
Add additional Flutter Context information:
* (compile) mode
* renderView availability 

Could be useful information to some.

Bildschirmfoto 2022-03-02 um 17 30 27

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps

CC @ueman

### DIFF
--- a/dart/lib/src/enricher/io_enricher_event_processor.dart
+++ b/dart/lib/src/enricher/io_enricher_event_processor.dart
@@ -37,7 +37,7 @@ class IoEnricherEventProcessor extends EventProcessor {
       runtimes: _getRuntimes(event.contexts.runtimes),
     );
 
-    contexts['dart_context'] = _getDartContext(_options.sendDefaultPii);
+    contexts['dart_context'] = _getDartContext();
 
     return event.copyWith(
       contexts: contexts,
@@ -61,13 +61,18 @@ class IoEnricherEventProcessor extends EventProcessor {
     ];
   }
 
-  Map<String, dynamic> _getDartContext(bool includePii) {
+  Map<String, dynamic> _getDartContext() {
     final args = Platform.executableArguments;
     final packageConfig = Platform.packageConfig;
     final isolate = Isolate.current.debugName;
+    final compile_mode = _options.platformChecker.isReleaseMode()
+        ? 'release'
+        : _options.platformChecker.isDebugMode()
+            ? 'debug'
+            : 'profile';
 
     String? executable;
-    if (includePii) {
+    if (_options.sendDefaultPii) {
       try {
         // This throws sometimes for some reason
         // https://github.com/flutter/flutter/issues/83921
@@ -83,11 +88,12 @@ class IoEnricherEventProcessor extends EventProcessor {
     }
 
     return <String, dynamic>{
+      'compile_mode': compile_mode,
       if (packageConfig != null) 'package_config': packageConfig,
       'number_of_processors': Platform.numberOfProcessors,
       if (isolate != null) 'isolate': isolate,
       // The following information could potentially contain PII
-      if (includePii) ...{
+      if (_options.sendDefaultPii) ...{
         'executable': executable,
         'resolved_executable': Platform.resolvedExecutable,
         'script': Platform.script.toString(),
--- a/dart/lib/src/enricher/io_enricher_event_processor.dart
+++ b/dart/lib/src/enricher/io_enricher_event_processor.dart
@@ -65,11 +65,6 @@ class IoEnricherEventProcessor extends EventProcessor {
     final args = Platform.executableArguments;
     final packageConfig = Platform.packageConfig;
     final isolate = Isolate.current.debugName;
-    final compile_mode = _options.platformChecker.isReleaseMode()
-        ? 'release'
-        : _options.platformChecker.isDebugMode()
-            ? 'debug'
-            : 'profile';
 
     String? executable;
     if (_options.sendDefaultPii) {
@@ -88,7 +83,7 @@ class IoEnricherEventProcessor extends EventProcessor {
     }
 
     return <String, dynamic>{
-      'compile_mode': compile_mode,
+      'compile_mode': _options.platformChecker.compileMode,
       if (packageConfig != null) 'package_config': packageConfig,
       'number_of_processors': Platform.numberOfProcessors,
       if (isolate != null) 'isolate': isolate,
--- a/dart/lib/src/enricher/web_enricher_event_processor.dart
+++ b/dart/lib/src/enricher/web_enricher_event_processor.dart
@@ -2,8 +2,8 @@ import 'dart:async';
 import 'dart:html' as html show window, Window;
 
 import '../event_processor.dart';
-import '../sentry_options.dart';
 import '../protocol.dart';
+import '../sentry_options.dart';
 
 EventProcessor enricherEventProcessor(SentryOptions options) {
   return WebEnricherEventProcessor(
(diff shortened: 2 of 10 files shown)


flutter/lib/src/flutter_enricher_event_processor.dart
```diff
@@ -130,8 +129,18 @@ class FlutterEnricherEventProcessor extends EventProcessor {
     final tempDebugBrightnessOverride = debugBrightnessOverride;
     final initialLifecycleState = _window?.initialLifecycleState;
     final defaultRouteName = _window?.defaultRouteName;
+    final mode = kDebugMode
+        ? 'debug'
+        : kReleaseMode
+            ? 'release'
+            : 'profile';
```
marandaneto: That's more useful if you set it as an `options#environment` during SDK init.
It should already be set by the SDK automatically.

kuhnroyal: Well environment can be something custom, this is the actual Flutter compile mode.
For example I can have `environment: staging` with a release mode or debug mode compile.

And yes you are right, it is being set in the Dart package into the `options.environment` but I can not find it anywhere in the Sentry UI.

kuhnroyal: > And yes you are right, it is being set in the Dart package into the `options.environment` but I can not find it anywhere in the Sentry UI.

My own stupidity, I am overriding it with `staging`.

kuhnroyal: Doesn't change the fact, that I need both bits in the event.
But this part could move into the pure `dart_context`.

flutter/lib/src/flutter_enricher_event_processor.dart
```diff
@@ -130,8 +129,18 @@ class FlutterEnricherEventProcessor extends EventProcessor {
...
+    // A FlutterEngine has no renderViewElement if it was started or is
+    // accessed from an isolate different to the main isolate.
+    final hasRenderView = _widgetsBinding?.renderViewElement != null;
 
     return <String, String>{
+      'mode': mode,
+      'has_render_view': hasRenderView.toString(),
```
ueman, line 139: What does that mean?

flutter/lib/src/flutter_enricher_event_processor.dart
```diff
@@ -111,8 +111,7 @@ class FlutterEnricherEventProcessor extends EventProcessor {
   SentryCulture _getCulture(SentryCulture? culture) {
     // The editor says it's fine without a `?` but the compiler complains
     // if it's missing
-    // ignore: invalid_null_aware_operator
-    final languageTag = _window?.locale?.toLanguageTag();
+    final languageTag = _window?.locale.toLanguageTag();
```
ueman: I believe this is because of an API change in some of the early 2.x Flutter versions. It should be kept as is, otherwise Sentry does not work with Flutter 2.0 anymore.

kuhnroyal: Ok, Flutter 1.17 is currently in the pubspec.yaml - this should probably be run in CI then, might be a couple of things failing.

flutter/lib/src/flutter_enricher_event_processor.dart
kuhnroyal: See above at the initialization site.

```dart
// A FlutterEngine has no renderViewElement if it was started or is
// accessed from an isolate different to the main isolate.
```

flutter/lib/src/flutter_enricher_event_processor.dart
marandaneto: Right now we run on different channels, but not on the minimal version, yes, See https://github.com/getsentry/sentry-dart/issues/636

marandaneto: I'd rather keep as it is for compatibility.

flutter/lib/src/flutter_enricher_event_processor.dart
marandaneto: Since the `mode` is already set via the `environment` flag and you've changed yourself.
I don't see much benefit in adding it to the event by default, it'd be duplicated data for most of the users.

Also, I believe the majority of the events would come from `release` anyway, sounds like this bit of data you could add yourself using `Scope#setExtra` and friends.

@brustolin opinions?

marandaneto: Take a lot at `_getRuntimes` as well, if the matter is release x debug, you can also see it from there, because of the Dart AOT or VM.

ueman: The runtime is not available on Flutter for web, so that won't always work.

kuhnroyal: I am using (and I assume others as well) environments across different projects for things like `alpha`, `beta`, `staging`, `production`.
I think that inferring an environment from the compile mode is an ok default. But the information should be visible somewhere else as well, especially since the default naming does not always imply a direct connection.

`kReleaseMode`, `--release` -> `production` -  I understand this now because I looked at the code but other users of the library may not dive this deep.

flutter/lib/src/flutter_enricher_event_processor.dart
kuhnroyal: Yea I'll revert this.

flutter/lib/src/flutter_enricher_event_processor.dart
marandaneto: Ok, let's keep this one then, it's a good point.

marandaneto:
```suggestion
    final mode = _checker.isDebugMode()
        ? 'debug'
        : _checker.isReleaseMode()
            ? 'release'
            : 'profile';
```

So it can be tested.

kuhnroyal: Did this but moved it to the Dart enricher.

flutter/lib/src/flutter_enricher_event_processor.dart
kuhnroyal: Reverted

dart/lib/src/enricher/io_enricher_event_processor.dart
```diff
@@ -61,13 +61,18 @@ class IoEnricherEventProcessor extends EventProcessor {
     ];
   }
 
-  Map<String, dynamic> _getDartContext(bool includePii) {
+  Map<String, dynamic> _getDartContext() {
     final args = Platform.executableArguments;
     final packageConfig = Platform.packageConfig;
     final isolate = Isolate.current.debugName;
+    final compile_mode = _options.platformChecker.isReleaseMode()
```
marandaneto:
```suggestion
    final compileMode = _options.platformChecker.isReleaseMode()
```

dart/lib/src/enricher/web_enricher_event_processor.dart
```diff
@@ -84,4 +90,16 @@ class WebEnricherEventProcessor extends EventProcessor {
     }
     return null;
   }
+
+  Map<String, dynamic> _getDartContext() {
+    final compile_mode = _options.platformChecker.isReleaseMode()
+        ? 'release'
+        : _options.platformChecker.isDebugMode()
+            ? 'debug'
+            : 'profile';
```
marandaneto: We could extract that within the `platformChecker` then so we don't duplicate this part of the code

kuhnroyal: Done

marandaneto (approved):
Thanks @kuhnroyal

brustolin (approved):
LGTM

PR #785, opened 2022-03-18
Update workflow files

AliAkberAakash:
## :scroll: Description
   _#skip-changelog_ 
 - Add condition to run e2e tests only when `SENTRY_AUTH_TOKEN` is not present
 - Remove unused env variable from `flutter.yml`

## :bulb: Motivation and Context
Fix to issue #733 

## :green_heart: How did you test it?
Merged with my main branch and ran workflows. Some tests are failing but I don't think they are for my changes.

## :pencil: Checklist
- [x] I reviewed submitted code
- [ ] I added tests to verify changes
- [ ] I updated the docs if needed
- [x] All tests passing
- [x] No breaking changes

PR #786, opened 2022-03-18
Fix bindings nullability on Flutter beta channel

marandaneto:
## :scroll: Description
_#skip-changelog_

Flutter beta upgrade found by https://github.com/getsentry/sentry-dart/pull/785

Running tests

## :pencil: Checklist

- [ ] All tests passing
- [ ] No breaking changes

### DIFF
--- a/flutter/lib/src/flutter_enricher_event_processor.dart
+++ b/flutter/lib/src/flutter_enricher_event_processor.dart
@@ -5,6 +5,7 @@ import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
 import 'package:sentry/sentry.dart';
 
+import 'binding_utils.dart';
 import 'sentry_flutter_options.dart';
 
 typedef WidgetBindingGetter = WidgetsBinding? Function();
@@ -23,7 +24,7 @@ class FlutterEnricherEventProcessor extends EventProcessor {
   }) {
     return FlutterEnricherEventProcessor(
       options,
-      () => WidgetsBinding.instance,
+      () => BindingUtils.getWidgetsBindingInstance(),
     );
   }
 
--- a/flutter/lib/src/flutter_enricher_event_processor.dart
+++ b/flutter/lib/src/flutter_enricher_event_processor.dart
@@ -24,7 +24,7 @@ class FlutterEnricherEventProcessor extends EventProcessor {
   }) {
     return FlutterEnricherEventProcessor(
       options,
-      () => BindingUtils.getWidgetsBindingInstance(),
+      BindingUtils.getWidgetsBindingInstance,
     );
   }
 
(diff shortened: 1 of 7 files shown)


ueman:
The solution of the flutter team itself is to [create a method like](https://github.com/flutter/plugins/blob/10a59af20f143d52d5a2c471bb9e7e3e79367a1e/packages/video_player/video_player_avfoundation/test/avfoundation_video_player_test.dart#L336-L341):

```dart
/// This allows a value of type T or T? to be treated as a value of type T?.
///
/// We use this so that APIs that have become non-nullable can still be used
/// with `!` and `?` on the stable branch.
// TODO(ianh): Remove this once we roll stable in late 2021.
T? _ambiguate<T>(T? value) => value;

/// ...

final instance = _ambiguate(WidgetsFlutterBinding.instance)
```

This is because of https://github.com/flutter/flutter/pull/89451.

Maybe it's a good opportunity, to make `WidgetsFlutterBinding.instance` non-nullable by default, since it's safe to use once `WidgetsFlutterBinding.ensureInitialized()` is called.

marandaneto:
> The solution of the flutter team itself is to [create a method like](https://github.com/flutter/plugins/blob/10a59af20f143d52d5a2c471bb9e7e3e79367a1e/packages/video_player/video_player_avfoundation/test/avfoundation_video_player_test.dart#L336-L341):
> 
> ```dart
> /// This allows a value of type T or T? to be treated as a value of type T?.
> ///
> /// We use this so that APIs that have become non-nullable can still be used
> /// with `!` and `?` on the stable branch.
> // TODO(ianh): Remove this once we roll stable in late 2021.
> T? _ambiguate<T>(T? value) => value;
> 
> /// ...
> 
> final instance = _ambiguate(WidgetsFlutterBinding.instance)
> ```
> 
> This is because of [flutter/flutter#89451](https://github.com/flutter/flutter/pull/89451).
> 
> Maybe it's a good opportunity, to make `WidgetsFlutterBinding.instance` non-nullable by default, since it's safe to use once `WidgetsFlutterBinding.ensureInitialized()` is called.

That won't prevent if `WidgetsBinding.instance` throws tho which is the case right now on Beta.

https://github.com/flutter/flutter/blob/874f5e5793433e75777282cd2dfca30c0cf67168/packages/flutter/lib/src/widgets/binding.dart#L304
https://github.com/flutter/flutter/blob/852bfe2a73e2ca059c1a0031879101231b946f5f/packages/flutter/lib/src/foundation/binding.dart#L278

flutter/lib/src/flutter_enricher_event_processor.dart
```diff
@@ -23,7 +24,7 @@ class FlutterEnricherEventProcessor extends EventProcessor {
   }) {
     return FlutterEnricherEventProcessor(
       options,
-      () => WidgetsBinding.instance,
+      () => BindingUtils.getWidgetsBindingInstance(),
```
ueman:
```suggestion
      BindingUtils.getWidgetsBindingInstance,
```

bruno-garcia (approved):

PR #798, opened 2022-03-23
Fix env vars for e2e tests

marandaneto:
_#skip-changelog_


.github/workflows/e2e_dart.yml
```diff
@@ -14,7 +14,6 @@ jobs:
   build:
     name: E2E
     runs-on: 'ubuntu-latest'
-    if: env.SENTRY_AUTH_TOKEN != null
```
marandaneto, line 17: `env.` isn't available for the whole job, it'd be nice tho :(

brustolin (approved):

Issue #500, opened 2021-06-15
Tests seems to be somewhat flaky


ueman:
The following tests run for 12 minutes until they timeout:

```
sentry-dart/flutter/test/load_android_image_list_test.dart
sentry-dart/flutter/test/native_sdk_integration_test.dart
sentry-dart/flutter/test/sentry_flutter_options_test.dart
sentry-dart/flutter/test/sentry_flutter_options_test.dart
sentry-dart/flutter/test/not_initialized_widgets_binding_test.dart
sentry-dart/flutter/test/sentry_flutter_test.dart
```

We should fix them, though I'm not sure yet what they have in common.
If we can't fix them, we should reduce the timeout.

marandaneto:
@ueman did we figure out the issue?

ueman:
No, but it seems better recently.

ueman:
I'm still unable to reproduce this locally.

The only thing they have in common, as far as I can tell, is that they are Flutter tests and mostly (always?) fail on macOS due to timeouts in the test framework.

Links to a example failing builds:
https://github.com/getsentry/sentry-dart/runs/3348163366?check_suite_focus=true

Looking through the past workflow runs, it seems to have improved recently.

marandaneto:
The issue isn't only macOS anymore, it happens on Windows and Linux too.
So the fix https://github.com/getsentry/sentry-dart/pull/743 didn't help.
I don't really know what's going on, and I can't reproduce it locally.

marandaneto:
Runnings Chrome tests only on Linux seems to help.

vaind:
No tests currently for `dart` and `flutter` dirs currently run on macOS. Having this long-term doesn't look like a good idea. Maybe reopen and re-scope this issue to try & bring them back?

PR #799, opened 2022-03-23
Await futures on testing

marandaneto:
_#skip-changelog_

Might help with https://github.com/getsentry/sentry-dart/issues/500

Fixed a few `async-await` problems.
Added `--test-randomize-ordering-seed=random --chain-stack-traces` to CI tests, so in case it fails again, maybe we can find out.
Increased timeout to 30 in a few job steps.


dio/lib/src/failed_request_interceptor.dart
```diff
@@ -9,12 +9,12 @@ class FailedRequestInterceptor extends Interceptor {
...
   @override
-  void onError(
+  Future<void> onError(
     DioError err,
     ErrorInterceptorHandler handler,
-  ) {
+  ) async {
     _hub.getSpan()?.throwable = err;
-    _hub.captureException(err);
+    await _hub.captureException(err);
```
marandaneto, line 17: @ueman what do you think about this change?
That was the only integration not awaiting for the capture call

ueman: Hm, not sure. I could imagine that the overrides could make some problems.

flutter/lib/src/default_integrations.dart
```diff
@@ -124,14 +124,14 @@ class FlutterErrorIntegration extends Integration<SentryFlutterOptions> {
...
-  FutureOr<void> close() {
+  FutureOr<void> close() async {
     /// Restore default if the integration error is still set.
     if (FlutterError.onError == _integrationOnError) {
       FlutterError.onError = _defaultOnError;
       _defaultOnError = null;
       _integrationOnError = null;
     }
-    super.close();
+    await super.close();
```
ueman: Instead of awaiting, you could remove the await and just `return super.close();`

marandaneto: fair point

dio/lib/src/failed_request_interceptor.dart
brustolin: I don't think this will cause any trouble.
Since DIO is waiting for `handler.next` being called, it doesn't matter if this function is synchronous or not.

brustolin (approved):
I'm not familiar with the dart test tool configuration changes, but the code looks good to me.

dio/lib/src/failed_request_interceptor.dart
ueman: Yeah, that's not my worry. It's that we're changing the method signature from the overriden method. I would keep all the async in the method but keep just `void` as return type, so that it still matches the super method.

I'm also pretty sure that this is not the culprit for https://github.com/getsentry/sentry-dart/issues/500 since https://github.com/getsentry/sentry-dart/issues/500 is older than the Dio integration.

But I'm happy if it fixes the errors 🤷

dart/lib/src/enricher/web_enricher_event_processor.dart
```diff
@@ -55,7 +54,7 @@ class WebEnricherEventProcessor extends EventProcessor {
     );
   }
 
-  Future<SentryDevice> _getDevice(SentryDevice? device) async {
```
marandaneto, line 58: @ueman do you remember why this was a Future? apparently, there's no reason.

dio/lib/src/failed_request_interceptor.dart
marandaneto: It's not since this error exists before the `dio` integration, but I'm looking for all possibilities of improvements for async stuff.

dart/lib/src/enricher/web_enricher_event_processor.dart
ueman: Nope, no idea. I probably tried to get more information out of the navigator. There are a few things on the navigator which need to be awaited.

dio/lib/src/failed_request_interceptor.dart
marandaneto: It works btw, didn't have any issues running the code that is now async.

ueman:
> widgets_binding_observer_test.dart

Seems to fail once again 😅

Maybe 

https://github.com/getsentry/sentry-dart/blob/main/flutter/test/widgets_binding_observer_test.dart#L17

should be changed to TestWidgetsBinding

PR #801, opened 2022-03-24
Use TestWidgetsFlutterBinding instead of WidgetsFlutterBinding for tests

marandaneto:
_#skip-changelog_

### DIFF
--- a/.github/workflows/flutter.yml
+++ b/.github/workflows/flutter.yml
@@ -68,12 +68,21 @@ jobs:
 
         - run: flutter upgrade
 
-        - name: Test
-          if: runner.os != 'macOS'
+        - name: Pub Get
           run: |
             cd flutter
             flutter pub get
+
+        - name: Test chrome
+          if: runner.os != 'macOS'
+          run: |
+            cd flutter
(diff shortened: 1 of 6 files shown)


ueman:
I wonder if this code snippet at the end of the file is still needed after this change.
```dart
/// Flutter >= 2.12 throws if ServicesBinding.instance isn't initialized.
ServicesBinding? getServicesBindingInstance() {
  try {
    return ServicesBinding.instance; // Maybe this should also return TestWidgetsFlutterBinding.ensureInitialized(); ?
  } catch (_) {}
  return null;
}
```

marandaneto:
> ```dart
> .instance
> ```

This is for Flutter Stable/Beta compatibility.
How `SchedulerBinding` instance would allow to return `TestWidgetsFlutterBinding.ensureInitialized()`?
Copy paste error?

ueman:
`TestWidgetsFlutterBinding` implements `SchedulerBinding`, so that replacement should be possible.

marandaneto:
> `TestWidgetsFlutterBinding` implements `SchedulerBinding`, so that replacement should be possible.

Oh good catch, thanks

brustolin (approved):

PR #802, opened 2022-03-24
Make scheduler binding return null instead of throwing if not initialised

marandaneto:
_#skip-changelog_

### DIFF
--- a/flutter/lib/src/sentry_flutter.dart
+++ b/flutter/lib/src/sentry_flutter.dart
@@ -131,7 +131,11 @@ mixin SentryFlutter {
       integrations.add(NativeAppStartIntegration(
         SentryNative(),
         () {
-          return SchedulerBinding.instance;
+          try {
+            /// Flutter >= 2.12 throws if SchedulerBinding.instance isn't initialized.
+            return SchedulerBinding.instance;
+          } catch (_) {}
+          return null;
         },
       ));
     }


flutter/lib/src/sentry_flutter.dart
```diff
@@ -131,7 +131,11 @@ mixin SentryFlutter {
       integrations.add(NativeAppStartIntegration(
         SentryNative(),
         () {
-          return SchedulerBinding.instance;
+          try {
+            /// Flutter >= 2.12 throws if SchedulerBinding.instance isn't initialized.
+            return SchedulerBinding.instance;
```
ueman, line 136: Same as in https://github.com/getsentry/sentry-dart/pull/801, `WidgetsFlutterBinding` implements `SchedulerBinding`. Therefor, you could use [`WidgetsFlutterBinding.ensureInitialized()`](https://api.flutter.dev/flutter/widgets/WidgetsFlutterBinding/ensureInitialized.html) or `WidgetsFlutterBinding.instance`. 

But at this point, the instance should already be initialized, so in theory this try catch shouldn't be needed?

marandaneto: This method should be called in a specific order, See `WidgetsFlutterBindingIntegration`, so I'd rather only use the `instance` if already there otherwise this might trigger some other problems.

brustolin (approved):

PR #808, opened 2022-03-28
add concurrency CI

marandaneto:
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#concurrency=

Didn't work well with matrixes, but examples are not good for matrixes, might be possible.

`styfle/cancel-workflow-action` works well.

Review only adding the `styfle/cancel-workflow-action`, the rest is code format.

_#skip-changelog_

### DIFF
--- a/.github/workflows/dart.yml
+++ b/.github/workflows/dart.yml
@@ -6,8 +6,8 @@ on:
       - release/**
   pull_request:
     paths-ignore:
-      - 'logging/**'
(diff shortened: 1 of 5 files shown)


romtsn (approved):

marandaneto:
Cancel in favor of https://github.com/getsentry/sentry-dart/pull/1119

PR #823, opened 2022-04-07
feat: iOS upload DebugMeta with exceptions

### LINKED ISSUE #444: Support for `--obfuscate` and `--split-debug-info` for iOS/macOS apps
This issue is to track the support for `--obfuscate` and `--split-debug-info` on iOS apps.

The original request was done on the [flutter/flutter issue tracker](https://github.com/flutter/flutter/issues/59321) and since then, Android support was added to Sentry.

For this to happen, Sentry expect symbols on iOS to be in a 'standard' format for that platform. Namely Mach-O files.
The [request for the Dart team for this support is tracked here](https://github.com/dart-lang/sdk/issues/43612#issuecomment-734438901).

Support for desktop is tracked separately, [here](https://github.com/getsentry/sentry-dart/issues/433).

vaind:
## :scroll: Description

This enables support for loaded-libary-list (a.k.a `DebugMeta`) upload so that Sentry has all the info needed to symbolicate stacktraces. Also fixes StackTrace parsing issues that will come up with the next Flutter version after v3.3.

closes #444 
To actually get any benefit, depends on https://github.com/flutter/flutter/pull/101586. Can be merged before that of course, just won't show anything useful on sentry.io

Necessary to get symbolication working for #444, but doesn't close the issue - that depends on changes in the [`flutter` tool - PR](https://github.com/flutter/flutter/pull/101586) - which means it will only work with a new flutter version after the PR is eventually merged.

Manually on device (requires a local build of the patched `flutter` tool version):
* `cd flutter/example; run.sh ios`
* check the event on Sentry: https://sentry.io/organizations/sentry-sdks/issues/3170309318/events/154a2060d1234899aa217d9254cf0660/?project=5428562

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps

macOS/iOS symbols don't get replaced by their real name on sentry.io - works locally with `flutter symbolize` though... https://sentry.io/organizations/sentry-sdks/issues/3565123989/events/1f1b96a0563443fcbf4b4a4653e24548/?project=5428562

### DIFF
--- a/dart/lib/src/sentry_stack_trace_factory.dart
+++ b/dart/lib/src/sentry_stack_trace_factory.dart
@@ -9,9 +9,8 @@ import 'sentry_options.dart';
 class SentryStackTraceFactory {
   final SentryOptions _options;
 
-  final _absRegex = RegExp('abs +([A-Fa-f0-9]+)');
-  static const _stackTraceViolateDartStandard =
-      'This VM has been configured to produce stack traces that violate the Dart standard.';
+  final _absRegex = RegExp(r'^\s*#[0-9]+ +abs +([A-Fa-f0-9]+)');
+  final _frameRegex = RegExp(r'^\s*#', multiLine: true);
 
   static const _sentryPackagesIdentifier = <String>[
     'sentry',
@@ -24,39 +23,42 @@ class SentryStackTraceFactory {
 
   /// returns the [SentryStackFrame] list from a stackTrace ([StackTrace] or [String])
   List<SentryStackFrame> getStackFrames(dynamic stackTrace) {
-    final chain = (stackTrace is StackTrace)
-        ? Chain.forTrace(stackTrace)
-        : (stackTrace is String)
-            ? Chain.parse(stackTrace)
-            : Chain.parse('');
+    Chain? chain;
+    if (stackTrace is StackTrace) {
+      chain = Chain.forTrace(stackTrace);
+    } else if (stackTrace is String) {
+      // Remove headers (everything before the first line starting with '#').
+      // *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***
+      // pid: 19226, tid: 6103134208, name io.flutter.ui
+      // os: macos arch: arm64 comp: no sim: no
+      // isolate_dso_base: 10fa20000, vm_dso_base: 10fa20000
+      // isolate_instructions: 10fa27070, vm_instructions: 10fa21e20
+      //     #00 abs 000000723d6346d7 _kDartIsolateSnapshotInstructions+0x1e26d7
+      //     #01 abs 000000723d637527 _kDartIsolateSnapshotInstructions+0x1e5527
+
+      final match = _frameRegex.firstMatch(stackTrace);
+      if (match != null) {
+        chain = Chain.parse(
+            match.start == 0 ? stackTrace : stackTrace.substring(match.start));
+      }
+    }
+    chain ??= Chain([]);
 
     final frames = <SentryStackFrame>[];
-    var symbolicated = true;
-
     for (var t = 0; t < chain.traces.length; t += 1) {
       final trace = chain.traces[t];
 
       for (final frame in trace.frames) {
         // we don't want to add our own frames
-        if (_sentryPackagesIdentifier.contains(frame.package)) {
+        if (frame.package != null &&
+            _sentryPackagesIdentifier.contains(frame.package)) {
           continue;
         }
 
-        final member = frame.member;
-        // ideally the language would offer us a native way of parsing it.
-        if (member != null && member.contains(_stackTraceViolateDartStandard)) {
-          symbolicated = false;
+        final stackTraceFrame = encodeStackTraceFrame(frame);
+        if (stackTraceFrame != null) {
+          frames.add(stackTraceFrame);
         }
-
-        final stackTraceFrame = encodeStackTraceFrame(
-          frame,
-          symbolicated: symbolicated,
-        );
-
-        if (stackTraceFrame == null) {
-          continue;
-        }
-        frames.add(stackTraceFrame);
       }
 
       // fill asynchronous gap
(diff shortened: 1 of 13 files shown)


flutter/ios/Classes/SentryFlutterPluginApple.swift
```diff
@@ -11,6 +11,7 @@ import AppKit
 public class SentryFlutterPluginApple: NSObject, FlutterPlugin {
 
     private var sentryOptions: Options?
+    let sentryIOSPrivate = PrivateSentrySDKOnly()
```
marandaneto: We can either replace all the calls with `sentryIOSPrivate` now or just use `PrivateSentrySDKOnly` since it's a static class.

marandaneto: We can either replace all the calls with `sentryIOSPrivate` now or just use `PrivateSentrySDKOnly` since it's a static class.

flutter/ios/Classes/SentryFlutterPluginApple.swift
```diff
@@ -56,7 +57,7 @@ public class SentryFlutterPluginApple: NSObject, FlutterPlugin {
     public func handle(_ call: FlutterMethodCall, result: @escaping FlutterResult) {
         switch call.method as String {
         case "loadContexts":
-            loadContexts(result: result)
+            loadContexts(call, result: result)
```
marandaneto: On Android we have a specific method call for loading image list, called `loadImageList`, can we reuse the same integration instead of returning it along with the `loadContexts`?

marandaneto: On Android we have a specific method call for loading image list, called `loadImageList`, can we reuse the same integration instead of returning it along with the `loadContexts`?

flutter/lib/src/default_integrations.dart
```diff
@@ -444,6 +410,30 @@ class _LoadAndroidImageListIntegrationEventProcessor extends EventProcessor {
 
     return event;
   }
+
+  static bool needsSymbolication(SentryEvent event) {
+    final frames = event.exceptions?.first.stackTrace?.frames;
+    if (frames == null) return false;
+    return Platform.isAndroid
```
marandaneto: You can use `options.platformChecker.platform.isAndroid` instead so it's possible to unit test it.

marandaneto: You can use `options.platformChecker.platform.isAndroid` instead so it's possible to unit test it.

flutter/lib/src/default_integrations.dart
```diff
@@ -444,6 +410,30 @@ class _LoadAndroidImageListIntegrationEventProcessor extends EventProcessor {
 
     return event;
   }
+
+  static bool needsSymbolication(SentryEvent event) {
```
marandaneto: This method could be e.g. an extension method of `SentryEvent` so you could easily do `event.needsSymbolication()`

marandaneto: This method could be e.g. an extension method of `SentryEvent` so you could easily do `event.needsSymbolication()`

flutter/lib/src/default_integrations.dart
```diff
@@ -444,6 +410,30 @@ class _LoadAndroidImageListIntegrationEventProcessor extends EventProcessor {
...
+    final frames = event.exceptions?.first.stackTrace?.frames;
+    if (frames == null) return false;
+    return Platform.isAndroid
+        ? frames.any((frame) => 'native' == frame.platform)
+        : frames.isNotEmpty;
+  }
+
+  static SentryEvent copyWithDebugImages(
```
marandaneto, line 500: If we reuse the `loadImageList` method channel call, we don't need to make this static and non-private, but rather being used only within the `_LoadImageListIntegrationEventProcessor` integration.

marandaneto, line 500: If we reuse the `loadImageList` method channel call, we don't need to make this static and non-private, but rather being used only within the `_LoadImageListIntegrationEventProcessor` integration.

flutter/ios/Classes/SentryFlutterPluginApple.swift
vaind: the method we need is not static: 

https://github.com/getsentry/sentry-cocoa/blob/17969fa1c80d643df62aa92e8a5fa724c9af0d8c/Sources/Sentry/Public/PrivateSentrySDKOnly.h#L42

flutter/ios/Classes/SentryFlutterPluginApple.swift
vaind: We could, I actually started that and that realized you specifically mentioned to use "loadContexts" during our call so went that route. I'll make the update (this also applies to the other related comments)

flutter/ios/Classes/SentryFlutterPluginApple.swift
marandaneto: Oh I see, maybe it was a mistake on the Cocoa SDK, let's double-check this.
@brustolin or @philipphofmann is there a reason why `getDebugImages` isn't static as every other method?

flutter/ios/Classes/SentryFlutterPluginApple.swift
marandaneto: Ok, it was likely a communication issue, I explained that we already use `loadContexts` only on iOS, and `loadImageList` only on Android, but we could actually make the integration (loadImageList) for both and reuse it, sorry about that.

flutter/ios/Classes/SentryFlutterPluginApple.swift
philipphofmann: No, there is no real reason for that. We can also change this if you want.

brustolin: I believe is because, `getDebugImages` need a `SentryDebugImageProvider` instance and we don't want to create a new one every call. But I guess we could move the provider to `SentryDependencyContainer` and make the Hybrid class full static.

marandaneto: https://github.com/getsentry/sentry-cocoa/issues/1940

vaind: already resolved in https://github.com/getsentry/sentry-cocoa/pull/1769

flutter/test/load_image_list_test.dart
```diff
@@ -0,0 +1,176 @@
...
+
+  tearDown(() {
+    fixture.channel.setMockMethodCallHandler(null);
+  });
+
+  for (var platform in [
+    MockPlatform.android(),
+    MockPlatform.iOs(),
+    MockPlatform.macOs()
+  ]) {
```
vaind, line 22: This file was renamed (`git mv`) from `load_android_image_list_test.dart` but looks like it didn't go well with github - sorry about that. Basically the tests are (about) the same, just run for three platforms.

flutter/lib/src/default_integrations.dart
vaind: changed

flutter/ios/Classes/SentryFlutterPluginApple.swift
vaind: changed

vaind: done

flutter/lib/src/default_integrations.dart
vaind: done

flutter/lib/src/default_integrations.dart
vaind: done

flutter/lib/src/default_integrations.dart
```diff
@@ -5,6 +5,10 @@ import 'package:flutter/services.dart';
 import 'package:flutter/widgets.dart';
 import 'package:package_info_plus/package_info_plus.dart';
 import 'package:sentry/sentry.dart';
+// TODO is there a "semi-internal" export we can use instead?
+//      sentry_io.dart looks like that but doesn't have any comments.
+// ignore: implementation_imports
+import 'package:sentry/src/platform/platform.dart';
```
marandaneto: You can probably look into:

`options.platformChecker.platform.isAndroid`, etc...
Since you have the instance of `SentryFlutterOptions`, you don't need to take the `platform` as a param.

vaind: removed

marandaneto:
@vaind besides the changelog, is there something pending or is it ready to be reviewed/merged?
If people target the `master` branch in `flutter`, it works already, right?

marandaneto:
@vaind so if I look at the linked event, `anonymous_closure_13965` and `YOd_13966`, they should be `in_app=true` but they are not, apparently that does not work for the `inApp` logic, right, is there a way to know at runtime?
Besides that, the file names are correctly but the function names are still "minified", what do we need to do to map from these 2 anonymous names (anonymous_closure_13965 and YOd_13966 and all the others) to the real names?

vaind:
> @vaind besides the changelog, is there something pending or is it ready to be reviewed/merged? If people target the `master` branch in `flutter`, it works already, right?

I've kept the changelog update for after the review. Yep works with the previous flutter version fine, but won't be useful because of the lack of dSYM upload

vaind:
> @vaind so if I look at the linked event, `anonymous_closure_13965` and `YOd_13966`, they should be `in_app=true` but they are not, apparently that does not work for the `inApp` logic, right, is there a way to know at runtime? 

Not available at runtime, there are only absolute addresses available, nothing more. This would need to happen on the server and apparently also doesn't work for Android.

> Besides that, the file names are correctly but the function names are still "minified", what do we need to do to map from these 2 anonymous names (anonymous_closure_13965 and YOd_13966 and all the others) to the real names?

I'm aware, I've written it in "Next Steps" in the description. Seems something's missing during symbolication on the server. All the info is available in the uploaded dSYM, but it's not used fully. See my question on Discord and feel free to ping if you're aware of anyone who could help. https://discord.com/channels/621778831602221064/621802340449779720/1016782208033370234

marandaneto:
@vaind on Android it used to work unless something changed from this version on, all the function names were properly symbolicated.

vaind:
> @vaind on Android it used to work unless something changed from this version on, all the function names were properly symbolicated.

And it still does (the function names are translated properly). My comment about Android was related to the part about `inApp` detection

dart/lib/src/sentry_stack_trace_factory.dart
```diff
@@ -68,67 +50,84 @@ class SentryStackTraceFactory {
...
-  }) {
-    final member = frame.member;
-
-    SentryStackFrame? sentryStackFrame;
-
-    if (symbolicated) {
-      final fileName = frame.uri.pathSegments.isNotEmpty
-          ? frame.uri.pathSegments.last
-          : null;
+  Chain parseStackTrace(dynamic stackTrace) {
```
marandaneto:
```suggestion
  Chain _parseStackTrace(dynamic stackTrace) {
```

marandaneto: Can it be private?

flutter/example/lib/main.dart
```diff
@@ -31,6 +31,18 @@ Future<void> main() async {
...
+      // going to log too much for your app, but can be useful when figuring out
+      // configuration issues, e.g. finding out why your events are not uploaded.
+      options.debug = true;
+      // Override the logger in release mode in this exmaple. Sentry normally
+      // uses developer.log which doesn't print anything in release mode.
+      if (kReleaseMode) {
+        options.logger = (SentryLevel level, String message,
+            {String? logger, Object? exception, StackTrace? stackTrace}) {
+          print('${DateTime.now().toUtc()} Sentry [${level.name}]: $message');
+        };
```
marandaneto: This would add a breadcrumb for each `print` call, was that intended? They are SDK/internal logs.
The problem I see is that many people ctrlc+ctrl v and that would not be a good thing.

vaind: no, it having breadcrumbs wasn't intended... I just needed to get the output in release mode _somehow_...

I guess I'll just remove this then

dart/lib/src/sentry_stack_trace_factory.dart
vaind: yes, changed

marandaneto:
> @vaind so if I look at the linked event, `anonymous_closure_13965` and `YOd_13966`, they should be `in_app=true` but they are not, apparently that does not work for the `inApp` logic, right, is there a way to know at runtime? Besides that, the file names are correctly but the function names are still "minified", what do we need to do to map from these 2 anonymous names (anonymous_closure_13965 and YOd_13966 and all the others) to the real names?

@fzyzcjy I recall you were converting the `instruction_addr` (from Sentry) to a way that `flutter symbolize` understands, I'm wondering if this could be the reason why it does not work completely for iOS (function names still anonymous), right now they look like `0x000000723d637527`, what did you change to make it work?

vaind:
> > @vaind so if I look at the linked event, `anonymous_closure_13965` and `YOd_13966`, they should be `in_app=true` but they are not, apparently that does not work for the `inApp` logic, right, is there a way to know at runtime? Besides that, the file names are correctly but the function names are still "minified", what do we need to do to map from these 2 anonymous names (anonymous_closure_13965 and YOd_13966 and all the others) to the real names?
> 
> @fzyzcjy I recall you were converting the `instruction_addr` (from Sentry) to a way that `flutter symbolize` understands, I'm wondering if this could be the reason why it does not work completely for iOS (function names still anonymous), right now they look like `0x000000723d637527`, what did you change to make it work?

FYI `flutter symbolize` (with the latest flutter master) does understand those stack traces and converts the symbol names properly. 
But as far as I understood, we **don't** use `flutter symbolize` on the server but rather use [symbolic](https://github.com/getsentry/symbolic)?

marandaneto:
> > > @vaind so if I look at the linked event, `anonymous_closure_13965` and `YOd_13966`, they should be `in_app=true` but they are not, apparently that does not work for the `inApp` logic, right, is there a way to know at runtime? Besides that, the file names are correctly but the function names are still "minified", what do we need to do to map from these 2 anonymous names (anonymous_closure_13965 and YOd_13966 and all the others) to the real names?
> > 
> > 
> > @fzyzcjy I recall you were converting the `instruction_addr` (from Sentry) to a way that `flutter symbolize` understands, I'm wondering if this could be the reason why it does not work completely for iOS (function names still anonymous), right now they look like `0x000000723d637527`, what did you change to make it work?
> 
> FYI `flutter symbolize` (with the latest flutter master) does understand those stack traces and converts the symbol names properly. But as far as I understood, we **don't** use `flutter symbolize` on the server but rather use [symbolic](https://github.com/getsentry/symbolic)?

Yes, that's correct, I'm just wondering if the format of our `instruction_addr` is correct since I recall @fzyzcjy having issues with it.
It does work on Android tho so it should not be an issue but I'd like to confirm anyway.

fzyzcjy:
@marandaneto Could you please point my comments about it? It has been really a long time and I have completely forgot what experiments I did... If the comment link is given, I can try to see whether I have some internal documentations linking to that particular comment.

marandaneto:
> @marandaneto Could you please point my comments about it? It has been really a long time and I have completely forgot what experiments I did... If the comment link is given, I can try to see whether I have some internal documentations linking to that particular comment.

Nope, I don't recall which issue was it, will try to find it.

vaind:
> > @marandaneto Could you please point my comments about it? It has been really a long time and I have completely forgot what experiments I did... If the comment link is given, I can try to see whether I have some internal documentations linking to that particular comment.
> 
> Nope, I don't recall which issue was it, will try to find it.

 Maybe let's wait how the investigation with symbolic goes and get back to this if there's still an issue? 
 
 Also, I don't think this is a blocker for this PR. Do you?

marandaneto (approved):
@vaind Thanks for doing this, this is HUGE.

marandaneto:
> > > @marandaneto Could you please point my comments about it? It has been really a long time and I have completely forgot what experiments I did... If the comment link is given, I can try to see whether I have some internal documentations linking to that particular comment.
> > 
> > 
> > Nope, I don't recall which issue was it, will try to find it.
> 
> Maybe let's wait how the investigation with symbolic goes and get back to this if there's still an issue?
> 
> Also, I don't think this is a blocker for this PR. Do you?

We can release once https://github.com/getsentry/symbolic/issues/675 is merged/released as well, so we have a fully working version when people use the `master` branch from Flutter at least.

PR #837, opened 2022-04-20
Fix mac architecture flutter

marandaneto:
_#skip-changelog_

Running into https://stackoverflow.com/questions/71924098/flutter-github-actions-bad-cpu-type-in-executable


marandaneto:
I've tried both architecture and it didn't work out: https://stackoverflow.com/questions/71924098/flutter-github-actions-bad-cpu-type-in-executable

bruno-garcia (approved):

PR #844, opened 2022-04-22
Fixes serialization of threads in an event

### LINKED ISSUE #841: `Thread`s are not correctly attached to events
_Platform:_
- [x]  Dart
- [x]  Flutter Android or iOS
- [x]  Flutter Web

**6.5.x**

---
I have the following issue:

I want to attach a `Thread` for an `Exception`

The description goes here ...

**Steps to reproduce:**
1. Add exception to event with an id of 1
2. Add thread to event with an id of 1
3. Thread does not get serialized

**Actual result:**
- Thread is not attached

**Expected result:**
- Thread should get attached

---

I suppose the culprit is [this code](https://github.com/getsentry/sentry-dart/blob/e93430abb2ac24b30bb175ad3a54cb80d7671262/dart/lib/src/protocol/sentry_event.dart#L368-L385) and the fix could be:

```diff
-    final threadIds =
-        exceptions?.map((element) => element.threadId).toList(growable: false);
+    // create a set of every thread id of all exceptions
+    final threadIds = exceptions
+        ?.map((element) => element.threadId)
+        .where((element) => element != null)
+        .toSet();

    final threadJson = threads
-        ?.where((element) => !(threadIds?.contains(element.id) ?? false))
+        // only send threads wich has a matching exception
+        ?.where((element) => threadIds?.contains(element.id) ?? false)
        .map((e) => e.toJson())
        .where((e) => e.isNotEmpty)
        .toList(growable: false);
```

ueman:
## :scroll: Description

Fixes https://github.com/getsentry/sentry-dart/issues/841


ueman:
Huh, it seems to me that a few jobs are wrongly marked as being successful.

dart/test/sentry_event_test.dart
```diff
@@ -297,16 +297,46 @@ void main() {
...
+        ],
+        exceptions: [
+          SentryException(
+            type: 'foo',
+            value: 'bar',
+            threadId: 0,
+          )
+        ],
+      ).toJson();
+      expect(serialized['threads']['values'], isNotEmpty);
```
brustolin: Im curios, how is this two tests different? As far as I know dart doesn't care with parameter order when they are labeled.

ueman: They aren't different, I just made a mistake and duplicated it somehow by accident 😑

ueman: fixed this mistake

dart/test/sentry_event_test.dart
```diff
@@ -309,15 +309,41 @@ void main() {
       expect(serialized['threads']['values'], isNotNull);
     });
 
+    test('should serialize $SentryThread when id matches exception id', () {
```
ueman, line 312: This is the test that demonstrates #841

dart/test/sentry_event_test.dart
```diff
@@ -309,15 +309,41 @@ void main() {
...
+            crashed: true,
+            current: true,
+            name: 'Isolate',
+          )
+        ],
+      ).toJson();
+      expect(serialized['threads']?['values'], isNotEmpty);
+    });
+
     test(
```
ueman, line 333: I think this test also had wrong expectations. After adding some more expectations, it also fails.

marandaneto:
Making it a draft since it's just an example.
I will look into it late this week.

brustolin (approved):
LGTM

ueman:
@brustolin Can you already merge this, so I can use it in #853 and #847?

PR #847, opened 2022-04-25
Report isolate as thread

### LINKED ISSUE #839: Report isolate name and Android thread name in `Thread` context
_Platform:_
- [x]  Dart
- [x]  Flutter Android or iOS
- [ ]  Flutter Web

**6.5.0**

---
I have the following issue:

Currently the `Isolate.debugName` is added to the [`dart_context`](https://github.com/getsentry/sentry-dart/blob/e93430abb2ac24b30bb175ad3a54cb80d7671262/dart/lib/src/enricher/io_enricher_event_processor.dart#L89). This should however go into the [`Thread`](https://develop.sentry.dev/sdk/event-payloads/threads/) context, so it is linked to the exception. 

A caveat is, that throwing code executed in the [`compute`](https://api.flutter.dev/flutter/foundation/compute-constant.html) function is typically moved from the executing isolate to the main isolate. This also applies to manually spawned isolates. The reported isolate name for the exception is wrong in this case. However, this behavior can already observed in the current code.

---- 

PlatformExceptions on Android with a Java/Kotlin stack trace also do not currently report the thread name. However, if the exception includes the [thread name](https://github.com/getsentry/sentry-dart/blob/e93430abb2ac24b30bb175ad3a54cb80d7671262/flutter/lib/src/jvm/jvm_exception.dart#L30), it should also be reported via the `Thread` context.

ueman:
## :scroll: Description

This PR introduces reporting of the current isolate as thread.

See for example events:
- https://sentry.io/organizations/sentry-sdks/issues/3228921456/events/acd6e57ceafc47c098c79c6997db0f0d/?project=5428562
- https://sentry.io/organizations/sentry-sdks/issues/3229349658/events/863eaa662813446796a5c255a2ef899e/?project=5428562

Event with isolate as thread

Closes https://github.com/getsentry/sentry-dart/issues/839

Added a new test case.

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps

- I'm waiting until #841 is fixed, then I'll look at this again to see if everything works and fix it if needed.
- Create follow up ticket for https://github.com/getsentry/sentry-dart/pull/847#discussion_r858518214


dart/test/sentry_client_test.dart
```diff
@@ -102,6 +102,28 @@ void main() {
       expect(capturedEvent.exceptions?.first.stackTrace, isNotNull);
     });
 
+    test('should attach isolate info in thread', () async {
```
ueman: This test only runs on dart:io. Web stays as it was before this PR, because it doesn't has isolates.

dart/lib/src/sentry_client.dart
```diff
@@ -131,18 +132,42 @@ class SentryClient {
...
     }
 
-    if (event.exceptions?.isNotEmpty ?? false) return event;
+    if (event.exceptions?.isNotEmpty ?? false) {
+      return event;
+    }
+
+    final isolateName = getIsolateName();
+    // isolates have no id, so we use the hashCode as id
+    final isolateId = isolateName?.hashCode;
```
ueman: Isolates have no id, so I'm using the hashcode of the name, in order to get a unique id in order to link thread and exception.

I believe the same approach has to be taken for the Java/Kotlin exception reporting from the PlatformException as the thread id is not reported in a stringified stracktrace.

That could lead to having the android and Dart exception being linked to the same thread, if both are named `main`, but I guess that is ok, since it would still be semantically correct.

dart/lib/src/sentry_client.dart
```diff
@@ -131,18 +132,42 @@ class SentryClient {
...
+          ],
+        );
+      }
+
+      sentryException = sentryException.copyWith(threadId: isolateId);
+
+      final thread = SentryThread(
+        id: isolateId,
+        name: isolateName,
+        crashed: false,
```
ueman: I'm setting this to `false` since an exception doesn't crash the app, but not sure if that's a good default.

marandaneto: The docs is off here https://develop.sentry.dev/sdk/event-payloads/threads/#attributes
That means this is the thread that actually caused the error/exception, it does not need to be a hard crash.

If the App crashed or not, look at https://develop.sentry.dev/sdk/event-payloads/exception/#attributes-1 `handled`

dart/lib/src/sentry_client.dart
```diff
@@ -131,18 +132,42 @@ class SentryClient {
       return event;
     }
 
-    if (event.exceptions?.isNotEmpty ?? false) return event;
+    if (event.exceptions?.isNotEmpty ?? false) {
+      return event;
+    }
+
+    final isolateName = getIsolateName();
```
marandaneto, line 148: The problem here is that `Isolate.current` returns the current Isolate, but if the error was thrown by another Isolate and bubbled up on this one, it'd use the wrong Isolate data.
I feel that the event itself (the integration that created the exception) should already have set this info (isolateName) in the event, so we can reuse to create the thread metadata.

marandaneto:
Made it a draft since its blocked by https://github.com/getsentry/sentry-dart/issues/841

dart/lib/src/sentry_client.dart
ueman: Okay, I'll change it

dart/lib/src/sentry_client.dart
ueman: How should that work then for `Sentry.captureException()`? There's no integration, which could add it.

dart/lib/src/sentry_client.dart
ueman: In that case, the Dart isolate and the Android thread probably shouldn't be connected to the same thread as suggested in https://github.com/getsentry/sentry-dart/pull/847#discussion_r857608217

dart/lib/src/sentry_client.dart
ueman: One solution I can think of is that integrations supply the isolate info as a hint and then do a `hint['isolate'] ?? Isolate.current.debugName` to figure out the correct isolate.

marandaneto: `Sentry.captureException` and `captureMessage` if called manually would use the current Isolate.
Integrations usually call `Sentry.captureEvent` and in this case, we could set the Isolate data to the event.
Using `hints` is another option, but we'd need to move to a `Map` first.

ueman: Then I would say we keep it like it is, because it most likely covers most use cases?

dart/lib/src/sentry_client.dart
```diff
@@ -140,18 +141,42 @@ class SentryClient {
...
+      }
+
+      sentryException = sentryException.copyWith(threadId: isolateId);
+
+      final thread = SentryThread(
+        id: isolateId,
+        name: isolateName,
+        crashed: true,
+        current: true,
+      );
```
marandaneto: I'd check if `isolateId` or `isolateName` are not null at least, otherwise it does not make sense I guess? 

Threads are only added if `options.attachThreads` is enabled as well, See the logic: https://github.com/getsentry/sentry-java/blob/896333a176c8d58cb85e33f8a10bf5c50ccce5c3/sentry/src/main/java/io/sentry/MainEventProcessor.java#L235-L243

I believe we should replicate it here.

ueman: Seems reasonable

dart/lib/src/sentry_client.dart
```diff
@@ -163,8 +188,10 @@ class SentryClient {
 
       if (frames.isNotEmpty) {
         event = event.copyWith(threads: [
-          ...(event.threads ?? []),
+          ...?event.threads,
```
marandaneto, line 197: We need to be sure that we are not duplicating the threads here, See comment https://github.com/getsentry/sentry-dart/pull/847/files#r866090263
Could be that we are adding on L169 and L192?

dart/lib/src/sentry_exception_factory.dart
```diff
@@ -45,13 +45,11 @@ class SentryExceptionFactory {
 
     // if --obfuscate feature is enabled, 'type' won't be human readable.
     // https://flutter.dev/docs/deployment/obfuscate#caveat
-    final sentryException = SentryException(
-      type: '${throwable.runtimeType}',
-      value: '$throwable',
+    return SentryException(
+      type: (throwable.runtimeType).toString(),
+      value: throwable.toString(),
```
marandaneto, line 50: Just out of curiosity, is `.toString` preferred over String interpolation?

ueman: I would say so, if it's just a single variable. Though, I can revert it, if you want me to.

dart/lib/src/sentry_client.dart
ueman: That shouldn't happen because the event will get returned in the if branch before that.

dart/lib/src/sentry_exception_factory.dart
marandaneto: No, it actually makes sense, I was just curious if there's some sort of linting for that.

ueman: Actually there is https://dart-lang.github.io/linter/lints/unnecessary_string_interpolations.html

brustolin (approved):
LGTM

dart/lib/src/sentry_options.dart
```diff
@@ -218,7 +218,14 @@ class SentryOptions {
...
 
-  /// When enabled, all the threads are automatically attached to all logged events (Android).
+  /// When enabled, the current isolate will be attached to the event.
+  /// This only applies to Dart:io platforms and only the current isolate.
+  /// Other isolates can't be read.
+  ///
+  /// When running on web, this option has no effect at all.
+  ///
+  /// When running in the Flutter context, this enables attaching of threads
+  /// for native events, if supported for the native platform.
```
marandaneto, line 228: I'd add the note that it's only Android at the moment tho.

ueman: Are there any docs I could link to? Then no one has to remember to update this doc comment if that changes.

ueman:
```suggestion
  /// When enabled, the current isolate will be attached to the event.
  /// This only applies to Dart:io platforms and only the current isolate.
  /// The Dart runtime doesn't provide information about other active isolates.
  ///
  /// When running on web, this option has no effect at all.
  ///
  /// When running in the Flutter context, this enables attaching of threads
  /// for native events, if supported for the native platform.
  /// Currently, this is only supported on Android.
```

ueman: Done.

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## 6.6.0-alpha.2
 
+- Attach Isolate name to thread context (#847)
```
```suggestion
## Unreleased

- Attach Isolate name to thread context (#847)

## 6.6.0-alpha.2
```

marandaneto:
@ueman please merge main into this branch, so I can test both changes together :) thanks.

ueman:
> @ueman please merge main into this branch, so I can test both changes together :) thanks.

@marandaneto done

ueman:
@marandaneto When testing with the Flutter example app, you may need to enable the `attachThreads` option.

marandaneto:
@ueman not sure if its a bug on the Android SDK or this SDK, but if I capture an exception on the Dart side:

Screenshot 2022-05-13 at 12 40 37

`current=false` should be `true`.
I'm using `options.attachThreads = true`

ueman:
> @ueman not sure if its a bug on the Android SDK or this SDK, but if I capture an exception on the Dart side:
> 
> Screenshot 2022-05-13 at 12 40 37
> 
> `current=false` should be `true`. I'm using `options.attachThreads = true`

I think this might be an error on the Android side. I even wrote tests, that make sure that `current = true`. 
`current` is also never set to false as seen here:
https://github.com/ueman/sentry-dart/blob/40072f42a35c41deab011e25a54cbc642774c505/dart/lib/src/sentry_client.dart#L169-L176

and here 

https://github.com/ueman/sentry-dart/blob/40072f42a35c41deab011e25a54cbc642774c505/dart/lib/src/sentry_client.dart#L198-L204

marandaneto:
@ueman I can confirm, since it's correct on iOS.
https://github.com/getsentry/sentry-java/issues/2047

ueman:
> @ueman I can confirm, since it's correct on iOS. [getsentry/sentry-java#2047](https://github.com/getsentry/sentry-java/issues/2047)

Soooo, this can be merged then, right?

ueman:
Since the Sentry Android SDK was updated, this should now also work as expected on Android. The failing CI test is unrelated to my changes, since I didn't touch any iOS code.

ueman:
I've attached two example events to the PR description. It works really well now.
Also enabled that option for the example app.

For convenience:

- https://sentry.io/organizations/sentry-sdks/issues/3228921456/events/acd6e57ceafc47c098c79c6997db0f0d/?project=5428562
- https://sentry.io/organizations/sentry-sdks/issues/3229349658/events/863eaa662813446796a5c255a2ef899e/?project=5428562

marandaneto:
> I've attached two example events to the PR description. It works really well now. Also enabled that option for the example app.
> 
> For convenience:
> 
> * [sentry.io/organizations/sentry-sdks/issues/3228921456/events/acd6e57ceafc47c098c79c6997db0f0d/?project=5428562](https://sentry.io/organizations/sentry-sdks/issues/3228921456/events/acd6e57ceafc47c098c79c6997db0f0d/?project=5428562)
> * [sentry.io/organizations/sentry-sdks/issues/3229349658/events/863eaa662813446796a5c255a2ef899e/?project=5428562](https://sentry.io/organizations/sentry-sdks/issues/3229349658/events/863eaa662813446796a5c255a2ef899e/?project=5428562)

Thanks @ueman 
We are releasing the Android SDK v6 right now and will test it soon.

marandaneto (approved):

PR #853, opened 2022-05-02
Feature: Add threads for Android platform exceptions

### LINKED ISSUE #841: `Thread`s are not correctly attached to events

ueman:
## :scroll: Description
Adds thread for android platform exception.

Example event: https://sentry.io/organizations/sentry-sdks/issues/3228921456/events/749fddaffc164a34a6efd82aba0f7f80/?project=5428562

Fixes #841 

- New tests

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps

- Test this together with https://github.com/getsentry/sentry-dart/pull/847

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,10 @@
 # Changelog
 
-## 6.6.0-alpha.2
+## Unreleased
 
 * Feat: Add Android thread to platform stacktraces (#853)
+
+## 6.6.0-alpha.2
 - Fix serialization of threads (#844)
 - Feat: Allow manual init of the Native SDK (#765)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,10 @@
 # Changelog
 
-## Unreleased
+## 6.6.0-alpha.2
 
 * Feat: Add Android thread to platform stacktraces (#853)
 - Fix serialization of threads (#844)
+- Feat: Allow manual init of the Native SDK (#765)
 
 ## 6.6.0-alpha.1
 
@@ -68,7 +69,7 @@
 * Fix: Disable log by default in debug mode (#753)
 * [Dio] Ref: Replace FailedRequestAdapter with FailedRequestInterceptor (#728)
 * Fix: Add missing return values - dart analyzer (#742)
-* Feature: Add `DioEventProcessor` which improves DioError crash reports (#718)
+* Feat: Add `DioEventProcessor` which improves DioError crash reports (#718)
 * Fix: Do not report duplicated packages and integrations (#760)
 * Feat: Allow manual init of the Native SDK or no Native SDK at all (#765)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Unreleased
 
-* Feat: Add Android thread to platform stacktraces
+* Feat: Add Android thread to platform stacktraces (#853)
 
 ## 6.6.0-alpha.1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+* Feat: Add Android thread to platform stacktraces
+
 ## 6.6.0-alpha.1
 
 * Feat: Client Reports (#829)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 * Feat: Add Android thread to platform stacktraces (#853)
 
 ## 6.6.0-alpha.2
+
 - Fix serialization of threads (#844)
 - Feat: Allow manual init of the Native SDK (#765)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 * Feat: Add Android thread to platform stacktraces (#853)
+- Fix serialization of threads (#844)
 
 ## 6.6.0-alpha.1
 
--- a/flutter/lib/src/event_processor/android_platform_exception_event_processor.dart
+++ b/flutter/lib/src/event_processor/android_platform_exception_event_processor.dart
@@ -64,20 +64,17 @@ class AndroidPlatformExceptionEventProcessor implements EventProcessor {
       exception,
     );
 
-    final threads = _markThreadsAsNonCrashed(event.threads);
+    final threads = _markDartThreadsAsNonCrashed(event.threads);
 
     final jvmExceptions = jvmException.map((e) => e.key);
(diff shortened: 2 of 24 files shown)


flutter/lib/src/event_processor/android_platform_exception_event_processor.dart
```diff
@@ -152,14 +185,36 @@ extension on JvmException {
...
+    if (thread != null) {
+      // Needs to be prefixed with 'Android', otherwise this thread id might
+      // clash with isolate names from the Dart side.
+      threadName = 'Android: $thread';
+    } else {
+      // If there's no thread in the exception, we just indicate that it's
+      // from Android
+      threadName = 'Android';
+    }
+    final threadId = threadName.hashCode;
```
ueman, line 210: - If an exception has a thread, it's prefixed with Android, so it doesn't clash with possible Dart isolate names (which is likely for `main`).
- If no thread name is given, I marked it as `Android` thread, because it's not from the Dart runtime.
- Also, once again, the thread id is based on the hashcode, because we just don't know the id.

This is a kinda opinionated workaround, because of the given limitations. I hope it's okay like that.

brustolin (approved):
I'm not sure about the thread naming conventions, @marandaneto  probably have better understanding of this. 

The rest LGTM.

flutter/lib/src/event_processor/android_platform_exception_event_processor.dart
```diff
@@ -152,14 +185,36 @@ extension on JvmException {
...
+
+    final sentryThread = SentryThread(
+      crashed: true,
+      current: true,
```
ueman: This is also up for debate: The Dart isolate is active, so the Android thread isn't active (`current`). So maybe that should be as follows? 
```suggestion
      current: false,
```

flutter/lib/src/event_processor/android_platform_exception_event_processor.dart
```diff
@@ -62,12 +64,41 @@ class AndroidPlatformExceptionEventProcessor implements EventProcessor {
...
+
+  /// If the crash originated on Android, the Dart side didn't crash.
+  /// Mark it accordingly.
+  List<SentryThread>? _markDartThreadsAsNonCrashed(
+    List<SentryThread>? threads,
+  ) {
+    return threads
+        ?.map((e) => e.copyWith(
+              crashed: false,
+              current: false,
```
ueman: Should be matched accordingly to https://github.com/getsentry/sentry-dart/pull/853/files#r863674267

ueman: This should also just mark the actual current isolate as current = true, not others if there are any present

ueman:
After thinking a little bit more about it: I think it makes the most sense if the Android thread is marked as `current = false, crashed = true` and the Dart isolate is marked as `current = true, crashed = false`. What do you think?

flutter/lib/src/event_processor/android_platform_exception_event_processor.dart
```diff
@@ -62,12 +64,41 @@ class AndroidPlatformExceptionEventProcessor implements EventProcessor {
...
+      // filter potential duplicated threads
+      final first = jvmThreads.first;
+      jvmThreads = jvmThreads
+          .skip(1)
+          .where((element) => element.id != first.id)
+          .toList(growable: true);
+      jvmThreads.add(first);
+    }
+
+    return event.copyWith(exceptions: [
```
marandaneto, line 83: If you enable `options.attachThreads`, how would that behave? Do we duplicate threads?

flutter/lib/src/event_processor/android_platform_exception_event_processor.dart
marandaneto: Yes, `current` means this thread was being executed or is still alive, and false otherwise, not sure if its even possible to know at runtime, sometimes its better to not set at all rather than assuming a value.
`crashed` means, this is the thread that actually captured the event/exception.

flutter/lib/src/event_processor/android_platform_exception_event_processor.dart
ueman: I'll check

ueman: Nope, at this point we're override the old threads with the new mutated threads.

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## 6.6.0-alpha.2
 
+* Feat: Add Android thread to platform stacktraces (#853)
```
```suggestion
## Unreleased

* Feat: Add Android thread to platform stacktraces (#853)

## 6.6.0-alpha.2
```

CHANGELOG.md
```diff
@@ -1,7 +1,10 @@
 # Changelog
 
-## 6.6.0-alpha.2
+## Unreleased
+
+* Feat: Add Android thread to platform stacktraces (#853)
 
+## 6.6.0-alpha.2
 - Fix serialization of threads (#844)
```
```suggestion
## 6.6.0-alpha.2

- Fix serialization of threads (#844)
```

flutter/lib/src/event_processor/android_platform_exception_event_processor.dart
marandaneto: Sentry requires the `id` to link the thread and the exception, so it's a good workaround.

marandaneto (approved):

PR #882, opened 2022-06-08
Fix: clear method sets all properties synchronously

marandaneto:
## :scroll: Description
Fix: clear method sets all properties synchronously

`clear` method had a side effect if not awaiting its future.

There are tests in place for it already

## :pencil: Checklist

- [X] All tests passing
- [X] No breaking changes

### DIFF
--- a/dart/lib/src/scope.dart
+++ b/dart/lib/src/scope.dart
@@ -53,9 +53,13 @@ class Scope {
   /// Get the current user.
   SentryUser? get user => _user;
 
+  void _setUserSync(SentryUser? user) {
+    _user = user;
+  }
+
   /// Set the current user.
   Future<void> setUser(SentryUser? user) async {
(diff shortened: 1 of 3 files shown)


dart/lib/src/scope.dart
```diff
@@ -199,16 +206,20 @@ class Scope {
...
     _span = null;
     _transaction = null;
-    await setUser(null);
     _fingerprint = [];
     _tags.clear();
     _extra.clear();
     _eventProcessors.clear();
+
+    _clearBreadcrumbsSync();
+    _setUserSync(null);
```
marandaneto, line 219: Now, this is executed synchronously, and it's not a breaking of behavior even if you don't `await`.

bruno-garcia (approved):
LGTM!

ueman (approved):

Issue #400, opened 2021-04-06
Debouncing of SentryWidgetsBindingObserver.didChangeMetrics.


ueman:
_Platform:_
- [ ]  Dart
- [x]  Flutter Android or iOS
- [ ]  Flutter Web

_IDE:_
- [ ]  VSCode
- [ ]  IntelliJ/AS
- [ ]  XCode
- [ ]  Other, which one?

The version of the SDK (See pubspec.lock):
**5.0.0**

---

SentryWidgetsBindingObserver.didChangeMetrics gets called way too often. As seen in [this crash report](https://sentry.io/organizations/sentry-sdks/issues/2149575550/?query=is%3Aunresolved) it gets called way too much.

marandaneto:
@ueman the linked issue does not have any breadcrumb with `device.screen` category, could you double-check, please?

marandaneto:
I'd chose between `bug` or `enhancement` btw, it can't be both though, sounds like it does its job so maybe debouncing would be an enhancement to not make it too noisy

ueman:
It is an older event of the same exception
https://sentry.io/organizations/sentry-sdks/issues/2149575550/events/0fb9c693eea84858b8b0a6986a7e16e6/?project=5428562

marandaneto:
indeed, thanks, let's debouce it and take the last then, what would be a reasonable timeout for deboucing? 1-3s?

ueman:
Yeah, that time span seems reasonable

ueman:
On Linux this also seems to be a problem even without resizing the window:
https://sentry.io/share/issue/014e554f41e449298005bc120e1641da/

PR #888, opened 2022-06-16
Fix duplicative `Screen size changed` breadcrumbs

lavinov-mercury:
## :scroll: Description

This PR adds private var to save last sent metrics breadcrumb & prevent sending new breadcrumbs if data didn't change.
I'm not sure if mutable var is fine here, however, it works & passes all tests (including new ones). Any proposals on better ways to implement this are welcome 😄
I've also updated link to `onMetricsChanged` docs, as current one seems to be outdated

There are tons of repetitive & absolutely useless `Screen size changed` events in our project. Funniest thing here is that size didn't change actually. One of the reasons for that is that `didChangeMetrics` also reacts on `viewInsets` changes, which means that it fires on keyboard opening and closing. And it's 30+ events on my iOS device for each keyboard appearance. So, it would be nice to add new breadcrumbs only when reporting values really changes.

I've changed related test & added new ones, checked that tests are still passing. Also i've checked it with the `example` app with new Sentry project. I changed app orientation several times, then caused error and checked appeared issue in Sentry. Then i've added `TextField` to open & close keyboard, caused one more error & checked new Sentry event. In both cases `Screen size changed` only when it was expected.

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes


ueman:
This sounds like potential fix for a variant of this issue https://github.com/getsentry/sentry-dart/issues/400

marandaneto:
I understand that `Screen size changed` all the time is annoying and noisy, but it's what the OS/framework triggers anyway.
I'd be fine adding a `enableWindowMetricBreadcrumbs` flag but I'd rather prefer the `debouncing` in another PR, because it can be reused across different sources of duplication.
I'd rather prefer a debouncing algorithm that works generically instead of a local var which requires the implementation everywhere, how does that sound?

lavinov-mercury:
@marandaneto I'm not sure that debouncing is right solution here. It's also incorrect, that each time client sends event to Sentry means that framework reports about screen size change. 
According to the Flutter docs, `onMetricsChanged` method can be called `whenever the devicePixelRatio, physicalSize, padding, viewInsets, PlatformDispatcher.views, or systemGestureInsets values change` ([docs](https://api.flutter.dev/flutter/dart-ui/SingletonFlutterWindow/onMetricsChanged.html)). As you can see, there are few more values that has no relation to screen size and can trigger event on update. As i mentioned in the description, i have 30+ events each time user opens a keyboard, while screen size, obviously, persists the same (it's viewInsets changes when keyboard opens). 

However, i agree, that variable + `if` statement doesn't look like an enterprise solution, i'm going to spend some time to refactor it to something more generic & easy to reuse, but i still believe that this solution should aim to send only distinct events, not debounced.

lavinov-mercury:
@marandaneto @brustolin I've refactored this to use stream controller with `distinct` to avoid duplicates

flutter/lib/src/widgets_binding_observer.dart
```diff
@@ -20,11 +24,31 @@ class SentryWidgetsBindingObserver with WidgetsBindingObserver {
...
+
+      final window = BindingUtils.getWidgetsBindingInstance()?.window;
+      _screenSizeStreamController.add(window);
+    }
+  }
 
   final Hub _hub;
   final SentryFlutterOptions _options;
 
+  final StreamController<SingletonFlutterWindow?> _screenSizeStreamController;
```
ueman, line 50: Instead of using the SingletonFlutterWindow, you could add window.viewConfiguration. That way you can use the equals method of `ViewConfiguration`  instead of a deepEquals for maps. That might be a cheaper operation.

flutter/lib/src/widgets_binding_observer.dart
```diff
@@ -20,11 +24,31 @@ class SentryWidgetsBindingObserver with WidgetsBindingObserver {
...
+      _screenSizeStreamController.stream
+          .map(
+            (window) => {
+              'new_pixel_ratio': window?.devicePixelRatio,
+              'new_height': window?.physicalSize.height,
+              'new_width': window?.physicalSize.width,
+            },
+          )
+          .distinct(mapEquals)
+          .skip(1) // Skip initial event added below in constructor
```
ueman, line 39: Why does it need an initial event?

flutter/lib/src/widgets_binding_observer.dart
lavinov-mercury: I don't use `deepEquals`, there is actually a `mapEquals` function used instead, which provides shallow check with top level keys comparison only, but i like the idea with ViewConfiguration, i'll try it 😄

flutter/lib/src/widgets_binding_observer.dart
lavinov-mercury: It's needed to make it emit events only when data actually changed. I mean, if at the beginning screen size was X, and then user opened a keyboard, screen size is still X, but if distinct stream is empty, it will emit `Screen size changed` event, though it actually didn't, because it had nothing to compare with. So i've added initial configuration, prevented it from emitting due to `.skip(1)`, and it works as expected: new events will come only when screen size or pixel ratio actually changes

It also fits to the actual test suite

marandaneto:
@lavinov-mercury thanks, I'll review it early next week.

flutter/lib/src/widgets_binding_observer.dart
lavinov-mercury: @ueman I've just checked, it's not worth to rely on `ViewConfiguration` and his equals method, since it also contains `viewInsets` & other fields besides screen size & pixel ratio

flutter/lib/src/widgets_binding_observer.dart
```diff
@@ -20,11 +24,31 @@ class SentryWidgetsBindingObserver with WidgetsBindingObserver {
     Hub? hub,
     required SentryFlutterOptions options,
   })  : _hub = hub ?? HubAdapter(),
-        _options = options;
+        _options = options,
+        _screenSizeStreamController = StreamController(sync: true) {
```
marandaneto, line 28: Maybe use the `SynchronousStreamController` directly then or better the `StreamController` interface?

marandaneto:
@lavinov-mercury missing changelog entry.
Thanks for doing this.

flutter/lib/src/widgets_binding_observer.dart
lavinov-mercury: @marandaneto it's a matter of taste, both will work. I prefer to leave StreamController interface, but since it's not my codebase, it's totally up to you. Would you like to change type to `SynchronousStreamController`?

marandaneto: Lets keep as it is, thanks :)

lavinov-mercury:
@marandaneto I've updated changelog, could you please check, if it's done right?

marandaneto:
> @marandaneto I've updated changelog, could you please check, if it's done right?

All good, thanks.
CI is going to fail due to https://github.com/getsentry/sentry-dart/pull/938 but its a different issue.

marandaneto (approved):

lavinov-mercury:
@marandaneto Hi! It failed to merge as branch became outdated again, could you please start workflows?

PR #915, opened 2022-07-01
Feature: PlatformDispatcher.onError integration

### LINKED ISSUE #833: Support new `PlatformDispatcher.onError`
https://github.com/flutter/engine/pull/32078
https://docs.google.com/document/d/1HdGLKoVw_KWLrVzDNZ8Lk5nN-biHQzi_CS5mCkpiDwA/edit#heading=h.pub7jnop54q0

ueman:
## :scroll: Description

This adds an integration for `PlatformDispatcher.onError` which was introduced in the Flutter 3.1 beta.

The interesting thing about the addition of this error handler is that the `runZonedGuarded` integration isn't needed anymore. This needs some more testing, though. 
Ideally, the `runZonedGuarded` integration shouldn't be applied anymore if the `PlatformDispatcher.onError` is available. The removal of `runZonedGuarded` also removes the ability to log `print()` statements. In Flutter that's not that bad, as Flutter uses `debugPrint()` which is still logged as breadcrumbs.
This could also replace the [`IsolateErrorIntegration`](https://github.com/getsentry/sentry-dart/blob/main/dart/lib/src/isolate_error_integration.dart) (at least on Flutter), but that needs to be tested, too.

The removal of the `runZonedGuarded` integration leads to much less complicated manual initialization code, because Sentry can now be initialized anywhere in the application lifecycle. There's no need anymore for Sentry to be the first call the user does in his code.

This is especially important for the use case, where crash reporting should only be enabled after user consent.

Example event: https://sentry.io/organizations/sentry-sdks/issues/3229407573/events/c912a632f2b147099657544fa4f22c60/?project=5428562

The error handler was introduced in https://github.com/flutter/engine/pull/32078

Closes #833

New test + example application

#### Test scenarios
- Passing error directly to the error handler
- Remove the `runZoneGuarded` integration and throw and uncaught exception

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps
- Release it.
- Add functionality to remove the need for the `runZoneGuarded` integration if this error handler is available.


ueman:
Not sure why the tests fail on the beta channel. The type errors which are shown are just wrong: https://github.com/flutter/engine/blob/28a5d3c05d61b25f82080260bae4b6c3b77d03b5/lib/ui/platform_dispatcher.dart#L64

Edit: Web has a different signature: https://github.com/flutter/engine/blob/28a5d3c05d61b25f82080260bae4b6c3b77d03b5/lib/web_ui/lib/platform_dispatcher.dart#L17

Edit 2: I guess the tests have to be disabled until https://github.com/flutter/engine/pull/34428 hits the beta/stable channel or this should only be merged when the fixed error handler hits the beta/stable channel.

Edit 3: Fix has landed

Edit 4: Latest Flutter beta includes the fix, so this is ready

flutter/lib/src/integrations/on_error_integration.dart
```diff
@@ -0,0 +1,120 @@
...
+    /// Restore default if the integration error is still set.
+    if (dispatchWrapper?.onError == _integrationOnError) {
+      dispatchWrapper?.onError = _defaultOnError;
+      _defaultOnError = null;
+      _integrationOnError = null;
+    }
+  }
+}
+
+// Should not become part of public API.
```
ueman, line 86:
```suggestion
/// This class helps to make this a backwards compatible feature.
/// It wraps the `this as dynamic` hack to call methods only available in newer Flutter versions in a type safe way.
// Should not become part of public API.
```

flutter/lib/src/integrations/on_error_integration.dart
```diff
@@ -0,0 +1,123 @@
...
+        // WidgetsBinding works with WidgetsFlutterBinding and other custom bindings
+        PlatformDispatcherWrapper(WidgetsBinding.instance.platformDispatcher);
+
+    if (!wrapper.isOnErrorSupported(options)) {
+      return;
+    }
+    _defaultOnError = wrapper.onError;
+
+    _integrationOnError = (Object exception, StackTrace stackTrace) {
+      final handled = _defaultOnError?.call(exception, stackTrace) ?? false;
```
ueman: The default value could also be `true`? What do you guys think?

marandaneto: Hard to say, I'd need to test it out a bit, but I guess by default should be `true` for now due to the app never crashing anyways.

marandaneto:
@ueman Right now the `OnErrorIntegration` is installed automatically.
We also have the `IsolateErrorIntegration`, `RunZonedGuardedIntegration`, `FlutterErrorIntegration`.
I'm wondering how all of these integrations play together, maybe not adding the `OnErrorIntegration` by default right now is the right way to do so? We could add code snippets on readme to actually remove such integrations on SDK init and add the `OnErrorIntegration` manually until this is GA? what do you think?

ueman:
> @ueman Right now the `OnErrorIntegration` is installed automatically. We also have the `IsolateErrorIntegration`, `RunZonedGuardedIntegration`, `FlutterErrorIntegration`. I'm wondering how all of these integrations play together, maybe not adding the `OnErrorIntegration` by default right now is the right way to do so? We could add code snippets on readme to actually remove such integrations on SDK init and add the `OnErrorIntegration` manually until this is GA? what do you think?

Sounds reasonable, since it's only available in the beta anyway. 

The FlutterErrorIntegration and this one play nicely together. The IsolateIntegration probably too. Just the RunZoneGuardedIntegration will become obsolete. 

That's however quite nice, because it enables us (speaking as a user) to be much more flexible in enabling and disabling Sentry. That's important for GDPR, for example.

----

The goal is to use this integration when available and the RunZoneGuardedIntegration on older Flutter versions.

flutter/test/integrations/on_error_integration_test.dart
```diff
@@ -0,0 +1,163 @@
+@TestOn('vm')
+// Run only on vm for now because of
```
ueman: Can ne removed

flutter/lib/src/integrations/on_error_integration.dart
```diff
@@ -0,0 +1,123 @@
...
+import '../sentry_flutter_options.dart';
+
+typedef ErrorCallback = bool Function(Object exception, StackTrace stackTrace);
+
+/// Integration which captures `PlatformDispatcher.onError`
+/// See:
+/// - https://master-api.flutter.dev/flutter/dart-ui/PlatformDispatcher/onError.html
+///
+/// Remarks:
+/// - Only usable on Flutter >= 3.1.0
```
ueman: It's 3.3 because earlier are buggy

flutter/example/lib/main.dart
```diff
@@ -180,6 +180,17 @@ class MainScaffold extends StatelessWidget {
               },
               child: const Text('Capture from FlutterError.onError'),
             ),
+            ElevatedButton(
```
ueman, line 183: Register Integration manually

CHANGELOG.md
```diff
@@ -30,6 +30,7 @@
 
 ### Features
 
+- Add integration for `PlatformDispatcher.onError` ([#915](https://github.com/getsentry/sentry-dart/pull/915))
```
ueman: Needs to be at the top

flutter/example/lib/main.dart
```diff
@@ -28,6 +28,8 @@ Future<void> main() async {
       options.considerInAppFramesByDefault = false;
       options.attachThreads = true;
       options.enableWindowMetricBreadcrumbs = true;
+      // Only usable on Flutter >= 3.3
+      options.addIntegration(OnErrorIntegration());
```
marandaneto: We could maybe add a boolean in the options that control if we should use the zoned guard or the on error integration, by default, the zoned guard integration, wdyt?

ueman: Yeah, I'll do that in a follow up. If that's fine with you?

marandaneto: Sure, I'd remove this line from the sample because we're running on the latest stable version for now.

flutter/example/lib/main.dart
```diff
@@ -180,6 +182,17 @@ class MainScaffold extends StatelessWidget {
...
               child: const Text('Capture from FlutterError.onError'),
             ),
+            ElevatedButton(
+              onPressed: () {
+                (WidgetsBinding.instance.platformDispatcher as dynamic)
+                    .onError
+                    ?.call(
+                      Exception('PlatformDispatcher.onError'),
+                      StackTrace.current,
+                    );
```
marandaneto, line 193: Maybe add a comment that to make this work, you need to `options.addIntegration(OnErrorIntegration());` and run on Flutter 3.3

marandaneto (approved):
Thanks @ueman

PR #933, opened 2022-07-05
Add request instead of response data in DioEventProcessor

kuhnroyal:
## :scroll: Description

Add request instead of response data to `SentryRequest` in `DioEventProcessor`

Response data may contain PII and should not be logged by the framework.
The `SentryRequest.data` is meant to contain request data. 
See discussion in https://github.com/getsentry/sentry-dart/issues/624 about PII.

Ran and adapted the tests

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps

CC @ueman


ueman:
This looks good to me.
I still whish, that there's not only a `SentryRequest` but also a `SentryResponse`.

On mobile it makes a lot of sense to record the response since, at least in my experience, that's a lot of the times also a source of errors (https://github.com/getsentry/develop/issues/401).

Regarding PII, why does it matter? I mean response and request can contain PII?

ueman (approved):

kuhnroyal:
> This looks good to me. I still whish, that there's not only a `SentryRequest` but also a `SentryResponse`.
> 
> On mobile it makes a lot of sense to record the response since, at least in my experience, that's a lot of the times also a source of errors.
> 
> Regarding PII, why does it matter? I mean response and request can contain PII?

I absolutely agree but it should be consistent in the SDK.
For me usually requests contain PII, responses contain weird validation errors which I am currently trying to manually log into `SentryRequest.other` map. But the order of `DioEventProcessor` and my custom processor can not be controlled. I need to add mine after `DioEventProcessor` somehow ...

marandaneto:
Makes sense, since the request data is guarded by `maxRequestBodySize`, wondering if this was an oversight or if we've decided to add the response in there for some reason, @ueman ?
@ueman the problem with the response is payload size, events can get easily dropped if we don't introduce a flag to limit that, such as https://github.com/getsentry/sentry-dart/issues/624 but that's a different issue.

marandaneto (approved):

ueman:
I think adding the response instead of the request was an oversight. The SentryHttpClient also adds the request and not the response.

PR #940, opened 2022-07-08
Revert "Breadcrumbs should not be duplicated if there is no mechaism on Android"

marandaneto:
Figure out if tests are broken because of this PR
https://github.com/getsentry/sentry-dart/pull/938

_#skip-changelog_


brustolin (approved):
LGTM

PR #943, opened 2022-07-12
fix: windows native method need default result

wangbax:
## :scroll: Description

When flutter is natively connected, the unimplemented logic needs to call back `NotImplemented` by default, otherwise it will be stuck in `await`.

Pending line: https://github.com/getsentry/sentry-dart/blob/main/dart/lib/src/sentry_tracer.dart#L102
Link code: [sentry_native_channel](https://github.com/getsentry/sentry-dart/blob/main/flutter/lib/src/sentry_native_channel.dart)

- Init set `SentryFlutter.init.options.debug = true`
- Check log print

image

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps
no op

_#skip-changelog_

### DIFF
--- a/flutter/windows/sentry_flutter_plugin.cpp
+++ b/flutter/windows/sentry_flutter_plugin.cpp
@@ -57,6 +57,7 @@ void SentryFlutterPlugin::HandleMethodCall(
     const flutter::MethodCall<flutter::EncodableValue> &method_call,
     std::unique_ptr<flutter::MethodResult<flutter::EncodableValue>> result) {
   // Native features will be added in a next release
+  result->NotImplemented();
 }
 
 }  // namespace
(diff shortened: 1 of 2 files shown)


flutter/windows/sentry_flutter_plugin.cpp
```diff
@@ -57,6 +57,7 @@ void SentryFlutterPlugin::HandleMethodCall(
     const flutter::MethodCall<flutter::EncodableValue> &method_call,
     std::unique_ptr<flutter::MethodResult<flutter::EncodableValue>> result) {
   // Native features will be added in a next release
+  result->NotImplemented();
```
marandaneto, line 60: Do we need that for Linux too?

See `sentry_flutter_plugin_handle_method_call`

wangbax: Actually, already processed on linux. ☺️

[fl_method_not_implemented_response_new](https://github.com/getsentry/sentry-dart/blob/main/flutter/linux/sentry_flutter_plugin.cc#L25)

marandaneto:
@wangbax please add a changelog entry :) it's a user facing fix.

wangbax:
@marandaneto has done. 😊

marandaneto (approved):

PR #958, opened 2022-07-24
Fix missing OS context for iOS events

### LINKED ISSUE #942: Issues reported as "Unknown OS"
Looks like iOS/macOS issues are missing the OS info. Examples:

* https://sentry.io/organizations/sentry-sdks/issues/3410159820/?project=5428562&query=is%3Aunresolved
* https://stackoverflow.com/questions/72853304/sentry-doesnt-provide-operating-system-on-events

marandaneto:
## :scroll: Description
Fix missing OS context for iOS events

Fix https://github.com/getsentry/sentry-dart/issues/942

## :pencil: Checklist

- [X] All tests passing
- [X] No breaking changes

## :crystal_ball: Next steps
https://github.com/getsentry/team-mobile/issues/11


flutter/lib/src/default_integrations.dart
```diff
@@ -185,10 +185,19 @@ class _LoadContextsIntegrationEventProcessor extends EventProcessor {
...
-              } else if (eventContexts[key] == null) {
+              } else if (currentValue == null) {
                 eventContexts[key] = value;
+              } else {
+                if (key == SentryOperatingSystem.type &&
+                    currentValue is SentryOperatingSystem &&
+                    value is SentryOperatingSystem) {
+                  final osMap = {...value.toJson(), ...currentValue.toJson()};
+                  final os = SentryOperatingSystem.fromJson(osMap);
+                  eventContexts[key] = os;
```
marandaneto, line 199: It's a quick workaround for now, the whole process of merging contexts from the Native and Dart layer should be done differently.

ueman: What's the issue?

marandaneto: https://github.com/getsentry/sentry-dart/issues/942
Flutter sets the `os.theme` by default, so when merging the contexts, it does not take the `os` from the iOS layer since we don't merge field per field recursively, but rather if the `key` exists or not.

ueman: Ahh, I see. Yeah, that's a stupid behavior.

brustolin (approved):
LGTM

PR #994, opened 2022-09-05
Fix flutter analyze imports

marandaneto:
_#skip-changelog_


ueman:
I think in Flutter 3.3 they exported more types, so that might be why those are now not needed anymore. Definitely make sure that this still works on lower Flutter versions.

marandaneto:
> I think in Flutter 3.3 they exported more types, so that might be why those are now not needed anymore. Definitely make sure that this still works on lower Flutter versions.

Yep, that was my goal.
https://github.com/getsentry/sentry-dart/runs/8183518774?check_suite_focus=true

I will have to ignore the linter instead :D

brustolin (approved):
LGTM

PR #995, opened 2022-09-05
Fix pubspec version test

marandaneto:
_#skip-changelog_

### DIFF
--- a/logging/test/version_test.dart
+++ b/logging/test/version_test.dart
@@ -14,6 +14,5 @@ void main() {
           yaml.loadYaml(File('pubspec.yaml').readAsStringSync());
       expect(sdkVersion, pubspec['version']);
     },
-    skip: 'works locally but not on CI',
   );
 }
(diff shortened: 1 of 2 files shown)


logging/test/version_test.dart
```diff
@@ -14,6 +14,5 @@ void main() {
           yaml.loadYaml(File('pubspec.yaml').readAsStringSync());
       expect(sdkVersion, pubspec['version']);
     },
-    skip: 'works locally but not on CI',
```
vaind, line 17: maybe remove this whole file completely? It's just a duplicate of the other one.

brustolin (approved):
LGTM.

logging/test/version_test.dart
marandaneto: They are under different packages, and each package has its own version, so if https://github.com/getsentry/sentry-dart/blob/main/scripts/bump-version.sh fails, this test would fail.

PR #998, opened 2022-09-07
ci: fix windows CI not failing on failed tests

vaind:
#skip-changelog

This PR ensures that when a test fails, the CI fails, same as on other platforms. 

The default powershell runner on Windows doesn't fail if a command fails and should be avoided. The global `defaults` section in a workflow yaml is overwritten by a job-specific `defaults` section, not merged.

We go from [this](https://github.com/getsentry/sentry-dart/runs/8223542693?check_suite_focus=true)
image

 to [this](https://github.com/getsentry/sentry-dart/runs/8223783749?check_suite_focus=true)
image

### DIFF
--- a/.github/workflows/dart.yml
+++ b/.github/workflows/dart.yml
@@ -8,9 +8,7 @@ on:
     paths-ignore:
       - 'logging/**'
       - 'dio/**'
-defaults:
-  run:
-    shell: bash
+
 jobs:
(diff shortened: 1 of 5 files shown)


brustolin (approved):
LGTM

marandaneto:
Good catch @vaind 
Thanks.

PR #1005, opened 2022-09-13
feat: Allow `routeNameExtractor` to set transaction names

JaspervanRiet:
## :scroll: Description
This PR adds the functionality to have the routeNameExtractor override the route names logged in Sentry, which I personally think makes more sense. It does not make sense to me that this extractor would only affect breadcrumbs, which is the current state, see Motivation. 

Note that this is currently technically a breaking change. I'm open to changing the PR to avoid that, but I'd say the decision on how to proceed there is best left to the maintainers of this project. :)

Setting the `routeNameExtractor` of the `SentryNavigatorObserver` currently only affects breadcrumbs. For our app, this does not work. We use named routes with IDs to navigate to our pages, e.g. `/movies/1500` to navigate to the detail page of a movie. We want to measure the performance of this whole group of pages instead of the unique instance. We thought we would be able to use the extractor to do this, but sadly not.

As for why I think this makes more sense: `routeNameExtractor` to me says that this functionality will allow you to mutate the route. Intuitively, that means this data will be used throughout the `SentryNavigatorObserver`; nothing about this signifies that this would only affect breadcrumbs. I see no discussion about this in the introducing PR (#684), so hard to say if this was thought about.

I used the current tests available for the observer, and added an extra test specific to this functionality since one was missing for the `routeNameExtractor`.

## :pencil: Checklist

- [x] All tests passing
- [ ] No breaking changes

### DIFF
--- a/flutter/lib/src/navigation/sentry_navigator_observer.dart
+++ b/flutter/lib/src/navigation/sentry_navigator_observer.dart
@@ -86,9 +86,7 @@ class SentryNavigatorObserver extends RouteObserver<PageRoute<dynamic>> {
   void didPush(Route<dynamic> route, Route<dynamic>? previousRoute) {
     super.didPush(route, previousRoute);
 
-    final routeName =
-        (_routeNameExtractor?.call(route.settings) ?? route.settings).name;
-    _setCurrentRoute(routeName);
+    _setCurrentRoute(route);
 
     _addBreadcrumb(
       type: 'didPush',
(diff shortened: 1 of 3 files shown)


flutter/lib/src/navigation/sentry_navigator_observer.dart
```diff
@@ -86,7 +86,9 @@ class SentryNavigatorObserver extends RouteObserver<PageRoute<dynamic>> {
   void didPush(Route<dynamic> route, Route<dynamic>? previousRoute) {
     super.didPush(route, previousRoute);
 
-    _setCurrentRoute(route.settings.name);
+    final routeName =
+        (_routeNameExtractor?.call(route.settings) ?? route.settings).name;
+    _setCurrentRoute(routeName);
```
marandaneto: We can extract to a method and avoid duplication since `didReplace` does the same, Also `didPop`.

marandaneto: `_startTransaction` should also take `routeName` then.

marandaneto:
@JaspervanRiet thanks for doing this.
I agree with the approach, makes sense.
I'd generify the code to call the `_routeNameExtractor` only once tho, right now it's being called multiple times in multiple places, better to just pass the `routeName` as a param to the methods.
It's a breaking change but I believe this was an oversight and it should be fixed, right now route names in crumbs and transactions might not be aligned if using the extractor.
What do you think @ueman ?

ueman:
Yes, sounds reasonable to me, too.

flutter/lib/src/navigation/sentry_navigator_observer.dart
JaspervanRiet: Done.

marandaneto:
@JaspervanRiet please add a changelog entry and we are ready to go :)

JaspervanRiet:
@marandaneto Should be done!

marandaneto (approved):
Thanks @JaspervanRiet

JaspervanRiet:
Thanks for quick handling of this PR :)

PR #1011, opened 2022-09-21
Set custom measurements on transactions

### LINKED ISSUE #789: Allow custom instrumentation to collect slow and frozen frames
I would like to be able to use the frames integration on macOS. Currently that's not possible because it depends on native code, which is also not able to provide frame information. 

However I could gather frame informations in pure Flutter code, based on https://github.com/getsentry/sentry-dart/pull/699 . That would also enable that feature on Linux and Windows and it's compatible from Flutter 2.0 onwards.

Therefore, I would like to have a way to somehow achieve this.

marandaneto:
## :scroll: Description
Users are allowed to set custom measurements

Fixes https://github.com/getsentry/sentry-dart/issues/789
https://github.com/getsentry/team-mobile/issues/31
And https://github.com/getsentry/team-mobile/issues/51

## :pencil: Checklist

- [X] All tests passing
- [ ] No breaking changes


CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 ### Features
 
 - Dynamic sampling ([#1004](https://github.com/getsentry/sentry-dart/pull/1004))
+- Allow users to set custom measurements ([#1011](https://github.com/getsentry/sentry-dart/pull/1011))
```
philipphofmann: `l`: Maybe a bit clearer
```suggestion
- Set custom measurements on transactions ([#1011](https://github.com/getsentry/sentry-dart/pull/1011))
```

dart/lib/src/sentry_measurement.dart
```diff
@@ -1,34 +1,52 @@
...
+    this.name,
+    this.value, {
+    this.unit,
+  });
 
   /// Amount of frames drawn during a transaction
-  SentryMeasurement.totalFrames(this.value) : name = 'frames_total';
+  SentryMeasurement.totalFrames(this.value)
+      : name = 'frames_total',
+        unit = SentryMeasurementUnit.none;
```
philipphofmann, line 13: `m`: I'm a bit confused that this is `none`. Why not `int`?

dart/lib/src/sentry_measurement_unit.dart
```diff
@@ -0,0 +1,53 @@
...
+  minute,
+
+  /// Hour (`"hour"`), 3600 seconds.
+  hour,
+
+  /// Day (`"day"`), 86,400 seconds.
+  day,
+
+  /// Week (`"week"`), 604,800 seconds.
+  week,
```
philipphofmann, line 24: `h`: Why do we have so many different units for time? Wouldn't it make sense to use `DateTime`?

dart/lib/src/protocol/sentry_span.dart
```diff
@@ -183,6 +183,15 @@ class SentrySpan extends ISentrySpan {
         sampled: samplingDecision?.sampled,
       );
 
+  @override
+  void setMeasurement(
+    String name,
+    num value, {
+    SentryMeasurementUnit? unit,
+  }) {
```
philipphofmann, line 191: `h`: I think a nicer API would be to have multiple methods for different units. One for num, one for int, one for DateTime, etc. I know there is no method overloading in Dart, but there are optional types or something. If that's the dart way of doing it, ignore me, please.

philipphofmann (commented):
Gave this a quick pass and added some comments on the API.

dart/lib/src/protocol/sentry_span.dart
ueman: One way of doing that would be to just take a `SentryMeasurement` object and have a lot of different constructors on `SentryMeasurement`. Then the span isn't full of different methods and at the same time it's easy to pick the right approach for the measurement

```dart
// pseudo code
class SentryMeasurement {
  factory SentryMeasurement.fromTimeSpan(Duration duration) {
    return SentryMeasurement(duration.milliseconds); // or whatever makes sense
  }
  // more constructors
}

class SentrySpan {
  void setMeasurement(String name, SentryMeasurement measurement) {
    // implementation ...
  }
}
```

dart/lib/src/sentry_measurement.dart
marandaneto: Its part of the protocol linked in the [meta issue](https://github.com/getsentry/team-mobile/issues/31) description.
https://github.com/getsentry/relay/blob/0d825c0e259ff6759e5c49f6412178f4768c8516/relay-server/src/metrics_extraction/transactions.rs#L248-L258

dart/lib/src/sentry_measurement_unit.dart
marandaneto: Its part of the protocol linked in the [meta issue](https://github.com/getsentry/team-mobile/issues/31) description.
https://github.com/getsentry/relay/blob/ed3a521f773ecd4bf9a2aefd7af80080d56d0841/relay-common/src/constants.rs#L353-L374

dart/lib/src/protocol/sentry_span.dart
marandaneto: The units are defined in the protocol.
This is the Dart overload already, which optionally takes a `unit` or defaults to `none` as mandated.
Otherwise, we'd need to add 8 different methods which is worse IMO.

brustolin (approved):
LGTM

dart/lib/src/protocol/sentry_span.dart
marandaneto: Yeah using named ctor helps, we already have some with the pre-defined measurements.
I have kept it as it is since its a Unified API across other SDKs.
Users can still make extension methods by converting a `Duration` to their preferred units.

PR #1026, opened 2022-09-28
Do not allow setting measurement if transaction is finished

marandaneto:
## :scroll: Description

Related to https://github.com/getsentry/sentry-dart/pull/1011.

Does not allow setting measurement if transaction is finished

## :pencil: Checklist

- [X] All tests passing
- [X] No breaking changes

### DIFF
--- a/dart/lib/src/sentry_tracer.dart
+++ b/dart/lib/src/sentry_tracer.dart
@@ -106,8 +106,10 @@ class SentryTracer extends ISentrySpan {
         }
       }
 
-      await _rootSpan.finish(endTimestamp: _rootEndTimestamp);
+      // the callback should run before because if the span is finished,
+      // we cannot attach data, its immutable after being finished.
       await _onFinish?.call(this);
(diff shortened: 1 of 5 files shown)


romtsn (approved):

dart/lib/src/sentry_tracer.dart
```diff
@@ -284,6 +284,9 @@ class SentryTracer extends ISentrySpan {
 
   @override
   void setMeasurement(String name, num value, {SentryMeasurementUnit? unit}) {
+    if (finished) {
+      return;
```
philipphofmann, line 292: `l`: Maybe log a debug message.

philipphofmann (approved):
LGTM

dart/lib/src/sentry_tracer.dart
marandaneto: There's a separate issue to make the tracer loggable because we don't have the instance of the logger, but we'll do it soon.

PR #1033, opened 2022-09-30
Add missing measurements units

marandaneto:
## :scroll: Description
Not as nice as `SentryMeasurementUnit.duration.milliSecond` as some other languages, but it's a limitation of the `enum` type in Dart in the current min. supported version, we can refactor that in the future.

Changing it to a plain `String` would be a breaking change and would have a poor UX because we'd need to document all the possible values.

The current limitation is that you can only set the unit with one of the enum values, which is well covered for most of the cases.
In case somebody wants to use their own metric unit, it's currently not supported, we can refactor this and bump the Dart version if needed.

Reason: https://github.com/getsentry/sentry-java/pull/2260#discussion_r984387110

## :pencil: Checklist

- [X] All tests passing
- [X] No breaking changes

### DIFF
--- a/dart/lib/src/sentry_measurement_unit.dart
+++ b/dart/lib/src/sentry_measurement_unit.dart
@@ -1,4 +1,12 @@
+/// The unit of measurement of a metric value.
+/// Units augment metric values by giving them a magnitude and semantics.
+/// Units and their precisions are uniquely represented by a string identifier.
+///
+/// This is a singe enum because Enhanced Enums in Dart is only available
+/// in newer versions.
 enum SentryMeasurementUnit {
+  /// Duration units
+
   /// Nanosecond (`"nanosecond"`), 10^-9 seconds.
   nanoSecond,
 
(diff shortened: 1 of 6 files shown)


dart/lib/src/sentry_measurement_unit.dart
```diff
@@ -1,4 +1,12 @@
+/// The unit of measurement of a metric value.
+/// Units augment metric values by giving them a magnitude and semantics.
+/// Units and their precisions are uniquely represented by a string identifier.
+///
+/// This is a singe enum because Enhanced Enums in Dart is only available
+/// in newer versions.
```
marandaneto, line 6: Limitation before Dart 2.17 and we cannot do breaking changes just yet.
It's either this or an open String, but this has better UX, so users don't need to find out the units by themselves via docs.

dart/lib/src/sentry_measurement_unit.dart
```diff
@@ -23,13 +31,66 @@ enum SentryMeasurementUnit {
   /// Week (`"week"`), 604,800 seconds.
   week,
 
+  /// Information units
+
+  /// Bit (`"bit"`), corresponding to 1/8 of a byte.
+  bit,
```
philipphofmann, line 37: `m`: Can you do nested enums somehow similar to Java? If yes, we could duplicate the existing ones into an inner class or interface enum, mark the old ones as deprecated, and do information and fraction units properly.

```dart
// Pseudo code 
enum SentryMeasurementUnit {

@deprecated
nanoSecond,

...

  enum Duration {
    nanoSecond, // ...
  }

  enum Information {
   // ...
   }

}
```

philipphofmann (commented):
Maybe there is a different way of fixing it.

dart/lib/src/sentry_measurement_unit.dart
marandaneto: Nope, the description states the limitation of the enum type.

marandaneto: See https://dart.dev/guides/language/language-tour#enumerated-types
Before and after `Declaring enhanced enums` section.

romtsn (approved):

dart/lib/src/sentry_measurement_unit.dart
philipphofmann: What about splitting it up into `SentryMeasurementDurationUnit`, `SentryMeasurementInformationUnit`, etc. and marking `SentryMeasurementUnit` as deprecated?

marandaneto: Discussed offline with @philipphofmann and agreed with the current solution after explaining the limitations of `enum` in the current Dart version.

philipphofmann (approved):
LGTM, as discussed on a call with @marandaneto, that Dart doesn't offer inheritance for enums or method overloads, and that's why we put all types into one enum.

PR #1035, opened 2022-10-01
Add renderer info to context

### LINKED ISSUE #687: Add Flutter web renderer information to event
One can detect wether Flutter for Web runs with the HTML renderer or the CanvasKit renderer like this:
```dart
import 'dart:js' as js;

final dynamic flutterCanvasKit = js.context['flutterCanvasKit'];
if(flutterCanvasKit != null){
// runs with CanvasKit renderer
}
``` 
This is especially helpful for bugs where the UI isn't renderer right.

Docs on the web renderer: https://docs.flutter.dev/development/tools/web-renderers

Since this issue is already quite old, the proposed solution above needs to be double-checked to see if it's still valid.

ueman:
## :scroll: Description

Adds the renderer info to the Flutter context.
Currently, it just distinguishes between skia, html and webkit.

I've also moved the `FlutterEnricherEventProcessor` class into the `event_processor` folder.

Fixes https://github.com/getsentry/sentry-dart/issues/687

New test

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes


brustolin (approved):
LGTM

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@
 - Bump Cocoa SDK from v7.26.0 to v7.27.0 ([#1030](https://github.com/getsentry/sentry-dart/pull/1030))
   - [changelog](https://github.com/getsentry/sentry-cocoa/blob/master/CHANGELOG.md#7270)
   - [diff](https://github.com/getsentry/sentry-cocoa/compare/7.26.0...7.27.0)
+- Added [Flutter renderer](https://docs.flutter.dev/development/platform-integration/web/renderers) information to events ([#1035](https://github.com/getsentry/sentry-dart/pull/1035))
```
marandaneto: Move to unreleased section.

flutter/lib/src/renderer/renderer.dart
```diff
@@ -0,0 +1,25 @@
+import 'unknown_renderer.dart'
+    if (dart.library.html) 'html_renderer.dart'
+    if (dart.library.io) 'io_renderer.dart' as implementation;
```
marandaneto, line 3: Should we add a `noop` in case there's no `html` nor `io`, is that even possible?

flutter/lib/src/renderer/renderer.dart
```diff
@@ -0,0 +1,25 @@
...
+String getRendererAsString() {
+  switch (getRenderer()) {
+    case FlutterRenderer.skia:
+      return 'Skia';
+    case FlutterRenderer.canvasKit:
+      return 'CanvasKit';
+    case FlutterRenderer.html:
+      return 'HTML';
+    case FlutterRenderer.unknown:
+      return 'unknown';
```
marandaneto:
```suggestion
      return 'Unknown';
```

flutter/lib/src/renderer/renderer.dart
ueman: I don't think so, I just thought it would be better to be safe than sorry

CHANGELOG.md
ueman: Done

PR #1036, opened 2022-10-01
Remove deprecated fields for v7

### LINKED ISSUE #842: Remove deprecated device context fields
There are a couple of fields which are deprecated in the device context. Those fields should be removed in v7.0. Follow up from #838

ueman:
## :scroll: Description

This PR removes some deprecated fields. This PR goes against the v7 branch, so it doesn't affect current v6 versions.

Fixes #842

## :pencil: Checklist

- [x] All tests passing
- [ ] No breaking changes


CHANGELOG.md
```diff
@@ -2,6 +2,16 @@
 
 ## Unreleased
 - Fix: Remove `SentryOptions` related parameters from classes which also take `Hub` as a parameter (#816)
+
+## Breaking
```
```suggestion
### Breaking Changes

```

CHANGELOG.md
```diff
@@ -2,6 +2,16 @@
 
 ## Unreleased
 - Fix: Remove `SentryOptions` related parameters from classes which also take `Hub` as a parameter (#816)
```
```suggestion
## 7.0.0

- Fix: Remove `SentryOptions` related parameters from classes which also take `Hub` as a parameter (#816)
```

dart/lib/src/enricher/io_enricher_event_processor.dart
```diff
@@ -55,7 +54,6 @@ class IoEnricherEventProcessor extends EventProcessor {
     // Pure Dart doesn't have specific runtimes per build mode
     // like Flutter: https://flutter.dev/docs/testing/build-modes
     final dartRuntime = SentryRuntime(
-      key: 'sentry_dart_runtime',
```
marandaneto, line 58: Is that right? the runtime needs a `key`.

dart/lib/src/enricher/io_enricher_event_processor.dart
```diff
@@ -71,7 +69,6 @@ class IoEnricherEventProcessor extends EventProcessor {
   Map<String, dynamic> _getDartContext() {
     final args = Platform.executableArguments;
     final packageConfig = Platform.packageConfig;
-    final isolate = Isolate.current.debugName;
```
marandaneto, line 74: This is only set on the `AndroidPlatformExceptionEventProcessor`.
So I'd not remove it, otherwise, the iOS metadata won't have the `Isolate.current.debugName` at all.

dart/lib/src/enricher/io_enricher_event_processor.dart
ueman: No, a runtime doesn't need a key. It's just required if there are multiple runtimes with the same name, but that should never be the case.

dart/lib/src/enricher/io_enricher_event_processor.dart
ueman: No, this is also set when enabling the `attachThreads` option

https://github.com/getsentry/sentry-dart/blob/9835e1cc20cfb03785448ea1d2dd2e3297c2e544/dart/lib/src/sentry_client.dart#L185-L193

dart/lib/src/enricher/io_enricher_event_processor.dart
marandaneto: Yeah indeed, just wondering why this is removed.

ueman: It looks better in the UI without the key :D

dart/lib/src/enricher/io_enricher_event_processor.dart
marandaneto: Indeed, for some reason, VS Code only returned 1 result for `Isolate.current.debugName` search, all good then, thanks.

dart/lib/src/enricher/io_enricher_event_processor.dart
marandaneto: What's about `sentry_flutter_runtime` then?

marandaneto: Oh also removed, ok, closing it.

PR #873, opened 2022-05-28
Send DSN with SentryEnvelopeHeader

vishnukvmd:
## :scroll: Description
This change injects DSN into the `SentryEnvelopeHeader` class and serializes it if available.

## :bulb: Motivation and Context
This change ensures that the DSN is passed to the server along with events, bringing in consistency with the JavaScript SDK. This gets us one step closer to passing a `tunnel` that can be used to proxy data to Sentry servers (to fixe #872). Once this change is merged, will create a separate PR that accepts `tunnel` as a parameter within the `SentryConfig` and configures the transport accordingly (again, similar to the JS SDK).

## :green_heart: How did you test it?
Verified that the `dsn` was passed within the envelope to the remote server.

## :pencil: Checklist

- [x] I reviewed submitted code
- [x] I added tests to verify changes
- [x] I updated the docs if needed
- [x] All tests passing
- [x] No breaking changes

Issue #872, opened 2022-05-28
Support for tunnel parameter


vishnukvmd:
We are protecting our Sentry instance behind Cloudflare Access, and it makes it impossible to report events without a tunnel.

So similar to the Javascript SDK, it'd be great if the Dart SDK could expose a [`tunnel`](https://docs.sentry.io/platforms/javascript/troubleshooting/#using-the-tunnel-option) parameter.

marandaneto:
@vishnukvmd makes sense, are you using Flutter Web or the Pure Dart SDK? thanks.

vishnukvmd:
We're using the Flutter SDK, but for mobile (not web).

marandaneto:
@vishnukvmd thanks, just out of curiosity, which Mobile Ad blockers are blocking calls to sentry.io? I'm aware of that only in the browser.

marandaneto:
@vishnukvmd Is Cloudflare Access a sort of VPN/Proxy? `tunnel` was designed to be a workaround for AD Blockers, maybe what you want is to be able to set a proxy when calling Sentry's API?

vishnukvmd:
It is [kind of like a VPN](https://ente.io/blog/tech/secure-internal-tools/).

From my limited understanding of Sentry, `tunnel`s seemed like the best way to workaround this. Please let me know if there's a better solution!

marandaneto:
@vishnukvmd are you using the `sentry` or `sentry_flutter` package? if `sentry`, we'd not be blocked by the Native SDK changes.

vishnukvmd:
We have a dependency on `sentry_flutter` 😔

marandaneto:
> We have a dependency on `sentry_flutter` 😔

Thanks, I will discuss the priority of this issue today.

smeubank:
potentially relevant for the proposed Project to Improve Tunneling https://github.com/getsentry/projects/issues/385

Angelodaniel:
Hi @buenaflor -> a customer has reported to need this as well! 
Is there a temporary workaround or could you let us know where this is on the backlog?

buenaflor:
@Angelodanie Since we dont supprt the tunnel option yet l'd suggest exploring altering the domain part of the dsn and setting up the reverse proxy properly.

For example, if your original DSN is:
`https://<public-key>@oXXXXXX.ingest.sentry.io/123456`

Alter the host to your own domain:
`https://<public-key>@yourdomain.com/123456`

[...]

PR #1050, opened 2022-10-07
Added missing DSN field into the SentryEnvelopeHeade

marandaneto:
## :scroll: Description
Cont. https://github.com/getsentry/sentry-dart/pull/873

Part of https://github.com/getsentry/sentry-dart/issues/872

## :pencil: Checklist

- [X] All tests passing
- [X] No breaking changes


dart/test/sentry_envelope_vm_test.dart
```diff
@@ -143,6 +145,55 @@ final envelopeBinaryData = [
   34,
   34,
   125,
+  44,
```
krystofwoldrich: I know it's not the point of this PR, but I feel like having the envelope bin data in a file could be nicer.

marandaneto: Good catch, https://github.com/getsentry/sentry-dart/pull/1050/commits/614334e97c58b69347f69ae0f95f9a4ca8eb6c99

krystofwoldrich (approved):
Looks good to me.

dart/test/sentry_envelope_vm_test.dart
```diff
@@ -50,248 +53,21 @@ void main() {
...
 
       final data = (await envelope.envelopeStream(SentryOptions()).toList())
           .reduce((a, b) => a + b);
 
-      expect(data, envelopeBinaryData);
+      final file = File('test_resources/envelope-no-attachment.envelope');
+      final jsonStr = await file.readAsString();
+      final dataStr = utf8.decode(data);
+
+      expect(dataStr, jsonStr);
```
marandaneto, line 70: For some reason, this fails on Windows:

> Expected: ... 1234567"}\r\n{"conte ...
> Actual: ... 1234567"}\n{"content ...
> ---------------------^

Differ at offset 128

Comparing the bytes directly also fail, only on Windows.

ueman: The 
> some reason

seems to be the different new line characters: 

> \r\n

> \n

marandaneto: @ueman @vaind ideas why this is failing?
Don't wanna lose more time on this =D

marandaneto: At some places we do `utf8.encode('\n')`
Maybe we don't need to do that? or maybe this has different behavior on Windows?

marandaneto: > The
> 
> > some reason
> 
> seems to be the different new line characters:
> 
> > \r\n
> 
> > \n

yeah I know, but the question is how the `\r` gets added in there only on Windows.

dart/test/sentry_envelope_vm_test.dart
```diff
@@ -50,248 +53,21 @@ void main() {
...
         attachments: [attachment],
       );
 
       final data = (await envelope.envelopeStream(SentryOptions()).toList())
           .reduce((a, b) => a + b);
 
-      expect(data, envelopeBinaryData);
+      final file = File('test_resources/envelope-no-attachment.envelope');
+      final jsonStr = await file.readAsString();
+      final dataStr = utf8.decode(data);
```
ueman, line 68: This seems pretty hacky, but maybe it solves the failure?
```suggestion
      final file = File('test_resources/envelope-no-attachment.envelope');
      final jsonStr = await file.readAsString();
      final dataStr = utf8.decode(data.replaceAll('\r\n', '\n'));
```

dart/test/sentry_envelope_vm_test.dart
vaind: from a superquick look just at these lines: if the file is checked out, I assume it's related to automatic CRLF line endings conversion during checkout

ueman: > At some places we do utf8.encode('\n')

Because the Sentry spec demands it.

> Maybe we don't need to do that? or maybe this has different behavior on Windows?

I'm not aware of such behavior.

One solution could be 
https://github.com/getsentry/sentry-dart/pull/1050/files#r990065322

vaind: see https://github.com/actions/checkout/issues/135

marandaneto: Lets see https://github.com/getsentry/sentry-dart/pull/1050/commits/a0ef68c6f473c45a4bb7947b04e48374353c8bd3

dart/test/sentry_envelope_vm_test.dart
marandaneto: yeah I did that locally, but I dont like it.

dart/test/sentry_envelope_vm_test.dart
marandaneto: And its green, thanks a lot @vaind

vaind: my pleasure

PR #986, opened 2022-08-26
feat: SentryUserFeedback.name verification on captureUserFeedback()

NazarenoCavazzon:
## :scroll: Description
I added some documentation regardless to the `Sentry.captureUserFeedback();` method and error handling for it.

## :bulb: Motivation and Context
I was trying to implement the `Sentry.captureUserFeedback();` method and it wasn't adding the feedback into my sentry feedback, and every time the method was called I received an error in the debug console saying that the envelope header was null, and I wasn't really sure what was the problem, and after adding the name to the `SentryUserFeedback()` class it worked, and it was confusing cause the name parameter is not required and nullable.

## :green_heart: How did you test it?
Because this only occurs when we don't pass a name to the  `SentryUserFeedback()` and the method per se it's working, this PR is testless, the only thing I added was an extra if statement verifying that the name wasn't null nor empty, and if it is we log the error and return.

image

This would really save me a couple hours of debugging.

## :pencil: Checklist

- [x] I reviewed submitted code
- [ ] I added tests to verify changes
- [x] I updated the docs if needed
- [x] All tests passing
- [x] No breaking changes

PR #1051, opened 2022-10-07
Improve docs static methods

marandaneto:
## :scroll: Description
_#skip-changelog_

Cont. https://github.com/getsentry/sentry-dart/pull/986/files

## :pencil: Checklist

- [ ] All tests passing
- [ ] No breaking changes

### DIFF
--- a/dart/lib/src/sentry.dart
+++ b/dart/lib/src/sentry.dart
@@ -168,6 +168,7 @@ class Sentry {
         withScope: withScope,
       );
 
+  /// Reports a [message] to Sentry.io.
   static Future<SentryId> captureMessage(
     String? message, {
     SentryLevel? level = SentryLevel.info,
@@ -185,6 +186,9 @@ class Sentry {
         withScope: withScope,
       );
 
+  /// Reports a [userFeedback] to to Sentry.io.
+  /// 
+  /// First capture an event and use the [SentryId] to create a [SentryUserFeedback]
   static Future<void> captureUserFeedback(SentryUserFeedback userFeedback) =>
       _hub.captureUserFeedback(userFeedback);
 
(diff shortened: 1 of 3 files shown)


dart/lib/src/sentry.dart
```diff
@@ -185,6 +186,9 @@ class Sentry {
         withScope: withScope,
       );
 
+  /// Reports a [userFeedback] to to Sentry.io.
```
```suggestion
  /// Reports a [userFeedback] to Sentry.io.
```

dart/lib/src/sentry_user_feedback.dart
```diff
@@ -6,7 +6,10 @@ class SentryUserFeedback {
     this.name,
     this.email,
     this.comments,
-  }) : assert(eventId != SentryId.empty());
+  }) : assert(eventId != SentryId.empty() &&
+            (name?.isNotEmpty == true ||
+                email?.isNotEmpty == true ||
+                comments?.isNotEmpty == true));
```
marandaneto, line 12: @ueman I guess we missed that at least.

krystofwoldrich (approved):
I've checked how is it JS/RN and there all the fields are mandatory on the type.

dart/lib/src/sentry_user_feedback.dart
ueman: Oh well, but now it's fixed :D

marandaneto:
> I've checked how is it JS/RN and there all the fields are mandatory on the type.

IIRC only one of them is mandatory, but not all of them, @philipphofmann designed this API so let's double-check with him.

philipphofmann:
> > I've checked how is it JS/RN and there all the fields are mandatory on the type.
> 
> IIRC only one of them is mandatory, but not all of them, @philipphofmann designed this API so let's double-check with him.

Yes, only one of them is mandatory. Sentry can handle that. On Cocoa, we made the fields `nonnullable` and the default value is an empty string. We don't validate though if one of the three fields is nonempty or nil. If you want to do that, I think it's good.

PR #1077, opened 2022-10-12
ref(ci): fix set-output / set-state deprecation

asottile-sentry:
https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

Committed via https://github.com/asottile/all-repos

#skip-changelog

### DIFF
--- a/.github/workflows/metrics.yml
+++ b/.github/workflows/metrics.yml
@@ -40,7 +40,7 @@ jobs:
         id: conf
         run: |
           version=$(grep "version" metrics/flutter.properties | cut -d'=' -f2 | xargs)
-          echo "::set-output name=flutter::$version"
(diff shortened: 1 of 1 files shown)


.github/workflows/metrics.yml
```diff
@@ -38,7 +38,7 @@ jobs:
         id: conf
         run: |
           version=$(grep "version" metrics/flutter.properties | cut -d'=' -f2 | xargs)
-          echo "::set-output name=flutter::$version"
+          echo "flutter=$version" >> "$GITHUB_OUTPUT"
```
marandaneto, line 43: @vaind is this fine?

vaind: yes

vaind (approved):

PR #1081, opened 2022-10-15
Removes dead code and unnecessary nullables from tests

### LINKED ISSUE #949: Run dart_code_metrics on CI and fix current warnings
https://pub.dev/packages/dart_code_metrics

Similar to https://github.com/flutter/flutter/pull/104550

hugobrancowb:
## :scroll: Description
Removes dead code, dead files and unecessary nullable parameters from `dart/test` and `dart/lib` with the help of [dart_code_metrics](https://pub.dev/packages/dart_code_metrics)

Fixes #949

`dart test`

## :pencil: Checklist
- [x] All tests passing
- [x] No breaking changes


dart/test/http_client/breadcrumb_client_test.dart
```diff
@@ -202,8 +202,8 @@ void main() {
 class CloseableMockClient extends Mock implements BaseClient {}
 
 class Fixture {
-  BreadcrumbClient getSut([MockClient? client]) {
-    final mc = client ?? getClient();
+  BreadcrumbClient getSut(MockClient client) {
+    final mc = client;
```
marandaneto: No need of defining another var `mc` in this case, you can use `client` directly, same for the other use cases.

dart/test/mocks/mock_environment_variables.dart
```diff
@@ -5,29 +5,29 @@ import 'no_such_method_provider.dart';
...
-  String? get dsn => _dsn;
+  String get dsn => _dsn;
 
   @override
-  String? get environment => _environment;
+  String get environment => _environment;
 
   @override
-  String? get release => _release;
+  String get release => _release;
```
marandaneto: I'd not change this otherwise you break the contract of the interface EnvironmentVariables

marandaneto:
The cases here https://github.com/getsentry/sentry-dart/issues/949#issuecomment-1257566705 are not addressed

hugobrancowb:
@marandaneto I've added some `ignore: unused-code` to the parts that were breaking the tests if removed

dart/lib/src/enricher/web_enricher_event_processor.dart
```diff
@@ -1,17 +1,11 @@
...
 import '../event_processor.dart';
 import '../protocol.dart';
 import '../sentry_options.dart';
 
-EventProcessor enricherEventProcessor(SentryOptions options) {
-  return WebEnricherEventProcessor(
-    html.window,
-    options,
-  );
-}
```
marandaneto, line 13: @ueman do you remember why this isn't used?

dart/lib/src/environment/_web_environment_variables.dart
```diff
@@ -1,8 +1,6 @@
 import 'environment_variables.dart';
 import 'keys.dart';
 
-final EnvironmentVariables envs = WebEnvironmentVariables();
```
marandaneto, line 4: @ueman same here

dart/lib/src/enricher/web_enricher_event_processor.dart
```diff
@@ -1,17 +1,11 @@
...
-
+// ignore: unused-code
```
marandaneto, line 8: This is not needed if we solve https://github.com/getsentry/sentry-dart/pull/1081/files#r1000232135
Either we remove this integration or it should be added in the integration list

dart/lib/src/http_client/failed_request_client.dart
```diff
@@ -147,7 +148,7 @@ class FailedRequestClient extends BaseClient {
 
   // See https://develop.sentry.dev/sdk/event-payloads/request/
   Future<void> _captureEvent({
-    required Object? exception,
+    required Object exception,
```
marandaneto, line 151: Since its an internal method, we can actually change to
`(Object exception, {...}`

dart/lib/src/environment/_web_environment_variables.dart
ueman: Same thing here, used via conditional import:
https://github.com/getsentry/sentry-dart/blob/eecbbcabbbe00a5dcc7d42cd6334f771a20b547f/dart/lib/src/environment/environment_variables.dart#L2-L3

dart/lib/src/enricher/web_enricher_event_processor.dart
ueman: It is used, but imported via conditional import, maybe that's the problem?

https://github.com/getsentry/sentry-dart/blob/eecbbcabbbe00a5dcc7d42cd6334f771a20b547f/dart/lib/src/enricher/enricher_event_processor.dart#L1-L8

dart/lib/src/isolate_error_integration.dart
```diff
@@ -9,6 +9,7 @@ import 'protocol.dart';
 import 'sentry_options.dart';
 import 'throwable_mechanism.dart';
 
+// ignore: unused-code
```
marandaneto, line 12: Is this a false positive?

dart/lib/src/transport/noop_encode.dart
```diff
@@ -1,6 +1,3 @@
-/// gzip compression is not available on browser
-List<int> compressBody(List<int> body, Map<String, String> headers) => body;
```
marandaneto, line 2: I believe this should not be removed

marandaneto:
@hugobrancowb there are some classes or methods that are removed but it's likely false positive.
there are conditional imports if it's Web or not, and in some cases, it's only used if it's `web` or `io`, and those classes or methods are found as not used but they are.
You'd need to double-check all of that, I suspect that you should delete any class or method most likely.

dart/lib/src/enricher/web_enricher_event_processor.dart
marandaneto: yep, thats my hunch, its a false positive on https://github.com/dart-code-checker/dart-code-metrics/issues

marandaneto:
Closing the issue as a part of large repository cleanup, due to it being inactive and/or outdated.
Please do not hesitate to ping me if it is still relevant, and I will happily reopen and work on it.
Cheers!

PR #1095, opened 2022-10-27
feat: capture Response context in HTTP Client

### LINKED ISSUE #969: Support `response` field for the `HttpClient` integration
Its been added to `Dio` already -> https://github.com/getsentry/sentry-dart/pull/934
We have to add for the `HttpClient` integration now.

Relates to https://github.com/getsentry/team-mobile/issues/38

This includes deprecating the fields from `response` that are not part of the Response RFC and removing the experimental annotations from the Response class as well.

vaind:

closes #969

## :pencil: Checklist

- [x] All tests passing
- [ ] No breaking changes


dart/lib/src/http_client/failed_request_client.dart
```diff
@@ -165,6 +167,7 @@ class FailedRequestClient extends BaseClient {
       url: urlWithoutQuery,
       queryString: query,
       cookies: sendDefaultPii ? request.headers['Cookie'] : null,
+      // TODO FIXME - this should check [sendDefaultPii], same as DIO integration.
       data: _getDataFromRequest(request),
```
vaind: This is an open issue. I'd say body is PII and the DIO integration treats it the same. It's not in the scope of this PR per-se, just something I've noticed here.

dart/lib/src/http_client/failed_request_client.dart
```diff
@@ -165,6 +167,7 @@ class FailedRequestClient extends BaseClient {
       url: urlWithoutQuery,
       queryString: query,
       cookies: sendDefaultPii ? request.headers['Cookie'] : null,
+      // TODO FIXME - this should check [sendDefaultPii], same as DIO integration.
```
marandaneto: No need, the default here is `MaxRequestBodySize.never`
So if the user has set something else, this is opt-in for whatever the data is, even if PII.

dart/lib/src/http_client/failed_request_client.dart
```diff
@@ -183,6 +186,19 @@ class FailedRequestClient extends BaseClient {
...
+      event.contexts.response = SentryResponse(
+        headers: sendDefaultPii ? response.headers : null,
+        redirected: response.isRedirect,
+        // Body not captured yet.
+        // See https://github.com/getsentry/sentry-dart/pull/1093#issuecomment-1293056244
+        body: null,
+        statusCode: response.statusCode,
+        status: response.reasonPhrase,
+      );
+    }
```
marandaneto: Let's only set the values that are defined here https://github.com/getsentry/rfcs/blob/main/text/0022-response-context.md#proposal-namely-option-2
Let's add `@Deprecated` to the other fields, such as `url`, `redirected`, `body`, `status`, and `other`.
Let's add `body_size` here and also `SentryRequest`.

dart/lib/src/protocol/sentry_response.dart
```diff
@@ -54,7 +54,10 @@ class SentryResponse {
   factory SentryResponse.fromJson(Map<String, dynamic> json) {
     return SentryResponse(
       url: json['url'],
-      headers: json['headers'],
+      headers: json.containsKey('headers')
+          ? (json['headers'] as Map<String, dynamic>)
+              .map((key, value) => MapEntry(key, value as String))
+          : null,
```
marandaneto: We just do `headers: json['headers']`, any reason why the defensive approach? only the SDK calls `SentryResponse.fromJson` from anyway, `value as String` cal also fails if `value` isn't a `String`.

dart/lib/src/http_client/failed_request_client.dart
marandaneto: We can also remove `@experimental` from the usage of SentryRequest, maybe we can even remove such fields (the ones mentioned above) since its still marked as experimental, wdyt?

dart/lib/src/http_client/failed_request_client.dart
vaind: OK, according to that logic, the following should be removed too? https://github.com/getsentry/sentry-dart/blob/main/dio/lib/src/dio_event_processor.dart#L143

dart/lib/src/protocol/sentry_response.dart
vaind: assigning json['headers'] would fail because it's `Map<String, dynamic>` and the target is `Map<String, string>`. I've changed to `Map.from()` though - should work just fine here as well.

The same logic for other similar cases [was historically implemented in this test code](https://github.com/getsentry/sentry-dart/pull/1095/files#diff-3825cfe5b084ce396d606318e82b63a3f7b736e7924b34836fe115adeb1bbb7dR49-R73), which, AFAICT, is also currently the only place executing `fromJSON` for these types.

dart/lib/src/http_client/failed_request_client.dart
marandaneto: Mmm good catch, I've missed that, I guess its an oversight on the Dio integration them.
https://docs.sentry.io/platforms/php/configuration/options/#max-request-body-size
Does not mention the`sendDefaultPii` flag , let's keep this TODO here and I will double-check internally.

vaind: actually, with the following comment that I should remove `body` and replace it with `bodySize`, it's OK anyway, since body size itself is not PII.

marandaneto: Yep but `body` still lives in the `Request`, it's not only there in the `Response`.

dart/lib/src/protocol/sentry_response.dart
marandaneto: Yeah that makes sense, the reason was that the test deserializes from a file, but Dart won't do it, its always created by the SDK or the user do it manually, in this case, they don't use the `fromJson` but rather the normal ctor respecting the data structure type, `Map.from()` should do.

dart/lib/src/http_client/failed_request_client.dart
```diff
@@ -183,6 +186,15 @@ class FailedRequestClient extends BaseClient {
       throwable: throwableMechanism,
       request: sentryRequest,
     );
+
+    if (response != null) {
```
marandaneto, line 188: Missing the `cookies` field

dio/lib/src/dio_event_processor.dart
```diff
@@ -164,29 +164,8 @@ class DioEventProcessor implements EventProcessor {
 
     return SentryResponse(
       headers: _options.sendDefaultPii ? headers : null,
-      url: response?.realUri.toString(),
-      redirected: response?.isRedirect,
-      body: _getResponseData(dioError.response?.data),
+      bodySize: dioError.response?.data?.length as int?,
```
marandaneto, line 164: Missing field `cookies`

dart/lib/src/http_client/failed_request_client.dart
marandaneto: Let's add the `sendDefaultPii` check here for the `body`

dart/lib/src/http_client/failed_request_client.dart
vaind: Not available in the response object. BTW cookies are not really a standalone part of an http response - they're sent as "set-cookie" headers

marandaneto: You can read from the headers, https://github.com/getsentry/sentry-dart/blob/d7758e8c1a1884d46734ccb197dbcde683bc4189/dart/lib/src/http_client/failed_request_client.dart#L167
If the response one is called `Set-Cookie`, just use this one for the response.

vaind: Yep, that's what I meant in the BTW part of my comment. 

Wouldnt it be better to do this on the server instead of each client lib?

marandaneto: The server already does that actually, but it's a back compatibility issue with older versions of the server.

vaind: I've checked the http response and it's providing headers as a map in which case it gives all the cookies as a single string, something like: `AWSALB=AWSALB_TEST; Expires=Tue, 26 Apr 2022 00:26:55 GMT; Path=/,AWSALBCORS=AWSALBCORS_TEST; Expires=Tue, 26 Apr 2022 00:26:55 GMT; Path=/; SameSite=None; Secure,jwt_token=JWT_TEST; Domain=.test.com; Max-Age=31536000; Path=/; expires=Wed, 19-Apr-2023 00:26:55 GMT; SameSite=lax; Secure,csrf_token=CSRF_TOKEN_TEST_1; Domain=.test.com; Max-Age=31536000; Path=/; expires=Wed, 19-Apr-2023 00:26:55 GMT,csrf_token=CSRF_TOKEN_TEST_2; Domain=.test.com; Max-Age=31536000; Path=/; expires=Wed, 19-Apr-2023 00:26:55 GMT,wuuid=WUUID_TEST`

Parsing that correctly is non-trivial and I'd move it out to a separate issue, prioritized according to the actual user needs (might be pretty low since you're saying it's just for backward compatibility).

Also see https://github.com/dart-lang/http/issues/24 - though any change in that area is unlikely to happen any time soon. I'm just flabbergasted how anyone could have implemented headers as a `map<string,string>` and called it a day, resulting in a package that can't move away from it due to compatibility and being one of the core ones basically.

marandaneto: We don't need to parse it, `cookies` is a String.
The server actually accepts multiple types, See https://develop.sentry.dev/sdk/event-payloads/types/#request for `cookies`, same for the response object.
For strongly typed languages, we have to choose one and it's being String for some reason, so I'd not care about the format, the server will accept and display it nicely anyway.
`response.headers['Set-Cookie']` will do.

dio/lib/src/dio_event_processor.dart
vaind: added in https://github.com/getsentry/sentry-dart/pull/1095/commits/3cf0e6f95d6265a4edf12286812758db95cb1df4

dart/lib/src/http_client/failed_request_client.dart
vaind: added in https://github.com/getsentry/sentry-dart/pull/1095/commits/3cf0e6f95d6265a4edf12286812758db95cb1df4

dart/lib/src/http_client/failed_request_client.dart
vaind: ok, added the check

dart/lib/src/protocol/sentry_response.dart
```diff
@@ -35,73 +25,57 @@ class SentryResponse {
...
+  SentryResponse(
+      {this.bodySize,
+      this.statusCode,
+      Map<String, String>? headers,
+      String? cookies})
+      : _headers = headers != null ? Map.from(headers) : null,
+        // Look for a 'Set-Cookie' header (case insensitive) if not given.
+        cookies = cookies ??
+            headers?.entries
+                .firstWhereOrNull((e) => e.key.toLowerCase() == 'set-cookie')
```
marandaneto, line 40: Why don't we just read directly from the headers as suggested before when creating the `SentryResponse`?
The Request relies on `cookies: sendDefaultPii ? request.headers['Cookie'] : null,`, but the response now does in the ctor and is case insensitive, I feel like the mix of both approaches just adds more complexity.

vaind: Well http headers are case insensitive, so that's why the access is.
As for why here - to avoid code duplication (Dio & this integration)

dart/lib/src/protocol/sentry_response.dart
```diff
@@ -35,73 +25,57 @@ class SentryResponse {
...
     return SentryResponse(
-      url: json['url'],
-      headers: json['headers'],
-      other: json['other'],
-      body: json['body'],
-      statusCode: json['status_code'],
-      status: json['status'],
-      redirected: json['redirected'],
-    );
+        headers: json.containsKey('headers') ? Map.from(json['headers']) : null,
```
marandaneto, line 46: `headers` is nullable, can we just `headers: json['headers']`?
If there's a problem, `SentryRequest.fromJson` has to be changed too.

dart/test/mocks/mock_transport.dart
```diff
@@ -31,6 +46,10 @@ class MockTransport implements Transport {
     final envelopeMap = envelopeItemJson as Map<String, dynamic>;
     final requestJson = envelopeMap['request'] as Map<String, dynamic>?;
 
+    // TODO the following code should really be part of fromJson() that handle those keys.
+    // JSON being Map<String, dynamic> is nothing out of ordinary.
+    // See [SentryResponse.fromJson()] as an example.
```
marandaneto: Is this still necessary? We already do `Map.from(...)` in the ctor.

dart/lib/src/protocol/sentry_response.dart
marandaneto: > http headers are case insensitive

Let's change for the request `Cookie` then as well?
Good point.

dart/test/mocks/mock_transport.dart
vaind: The TODO is here for the remainder of the code in the function, i.e. the old code not touched by this PR. I mean I could just do Map.from() for the other two affected places - Request[other] & Request[headers]

dart/lib/src/protocol/sentry_response.dart
vaind: > `headers` is nullable, can we just `headers: json['headers']`? If there's a problem, `SentryRequest.fromJson` has to be changed too.

json['headers'] is a different type so you can't pass it to an argument expecting `Map<String,String>`, regardless of nullability, see the following stacktrace for the error that occurs when I do that. It's a standard problem with the data coming from json (string) parsed through jsonDecode()
```
'type \'_InternalLinkedHashMap<String, dynamic>\' is not a subtype of type \'Map<String, String>?\'\n'
              '#0      new SentryResponse.fromJson (package:sentry/src/protocol/sentry_response.dart:46:22)\n'
              '#1      new Contexts.fromJson (package:sentry/src/protocol/contexts.dart:65:28)\n'
              '#2      new SentryEvent.fromJson (package:sentry/src/protocol/sentry_event.dart:313:22)\n'
              '#3      MockTransport._eventFromEnvelope (file:///C:/dev/sentry-dart/dart/test/mocks/mock_transport.dart:46:24)\n'
              '<asynchronous suspension>\n'
              '#4      MockTransport.send (file:///C:/dev/sentry-dart/dart/test/mocks/mock_transport.dart:30:21)\n'
              '<asynchronous suspension>\n'
              '#5      SentryClient.captureEvent (package:sentry/src/sentry_client.dart:141:16)\n'
              '<asynchronous suspension>\n'
              '#6      Hub.captureEvent (package:sentry/src/hub.dart:91:20)\n'
              '<asynchronous suspension>\n'
              '#7      FailedRequestClient._captureEvent (package:sentry/src/http_client/failed_request_client.dart:196:5)\n'
              '<asynchronous suspension>\n'
              '#8      FailedRequestClient.send (package:sentry/src/http_client/failed_request_client.dart:134:9)\n'
              '<asynchronous suspension>\n'
              '#9      BaseClient._sendUnstreamed (package:http/src/base_client.dart:93:32)\n'
              '<asynchronous suspension>\n'
              '#10     main.<anonymous closure>.<anonymous closure> (file:///C:/dev/sentry-dart/dart/test/http_client/failed_request_client_test.dart:92:7)\n'
              '<asynchronous suspension>\n'
              '#11     Declarer.test.<anonymous closure>.<anonymous closure> (package:test_api/src/backend/declarer.dart:215:9)\n'
              '<asynchronous suspension>\n'
              '#12     Declarer.test.<anonymous closure> (package:test_api/src/backend/declarer.dart:213:7)\n'
              '<asynchronous suspension>\n'
              '#13     Invoker._waitForOutstandingCallbacks.<anonymous closure> (package:test_api/src/backend/invoker.dart:257:7)\n'
              '<asynchronous suspension>\n'
```

> If there's a problem, `SentryRequest.fromJson` has to be changed too.

yes, that's what the TODO in the MockTransport is about. I've resolved it now & changed SentryRequest in my latest commit.

dart/lib/src/protocol/sentry_response.dart
vaind: changed

dart/test/mocks/mock_transport.dart
vaind: done

dart/test/mocks/mock_transport.dart
```diff
@@ -28,35 +43,13 @@ class MockTransport implements Transport {
 
     final envelopeItem = utf8.decode(envelopeItemData);
     final envelopeItemJson = jsonDecode(envelopeItem.split('\n').last);
-    final envelopeMap = envelopeItemJson as Map<String, dynamic>;
-    final requestJson = envelopeMap['request'] as Map<String, dynamic>?;
-
-    // '_InternalLinkedHashMap<dynamic, dynamic>' is not a subtype of type 'Map<String, String>'
```
marandaneto, line 34: We have the same issue on dio/test/mocks/mock_transport as well, Want to remove it from there as well? since the ctor is safe now.

vaind: OK, synced the mock_transport.dart copy in DIO

marandaneto:
@vaind there are some linters:

  error - test/mocks/mock_transport.dart:45:11 - Missing variable type for 'envelopeItemJson'. Try adding an explicit type, or remove implicit-dynamic from your analysis options file. - implicit_dynamic_variable
  error - test/mocks/mock_transport.dart:46:33 - The argument type 'dynamic' can't be assigned to the parameter type 'Map<String, dynamic>'. - argument_type_not_assignable

vaind:
> @vaind there are some linters:

Fixed now. BTW any reason why do the various packages use different linter rules?

marandaneto:
> > @vaind there are some linters:
> 
> Fixed now. BTW any reason why do the various packages use different linter rules?

nope, either an oversight or because it's a dep depending on other dep, and other packages, might differ a bit, but in general, they should be very similar if not the same.

marandaneto (approved):
Thanks @vaind

PR #1099, opened 2022-11-02
feat: move plugin from sentry_dart_plugin (experimental)

vaind:
## :scroll: Description

Testing whether moving plugin directly to the sentry package has any impact on app size or usability.

The way it's added in this PR, users would be able to run plugin by `dart run sentry:plugin`

TODO:
* [ ] I don't like the dependency updates - try if it can somehow be integrated as a dev_dependency instead - maybe if the actual source is kept as a separate package?

See discussions here: https://github.com/getsentry/sentry-dart-plugin/issues/64

## :pencil: Checklist

- [ ] All tests passing
- [ ] No breaking changes


marandaneto:
@vaind I meant to raise an issue only, I still have to discuss with folks if this is the way to go.
The Android Gradle plugin, for example, was a monorepo, and this only caused problems.
So we've to point out the trade-offs before going ahead, let's pause this before we are sure that we want to do it.

vaind:
> @vaind I meant to raise an issue only, I still have to discuss with folks if this is the way to go. The Android Gradle plugin, for example, was a monorepo, and this only caused problems. So we've to point out the trade-offs before going ahead, let's pause this before we are sure that we want to do it.

I know, I've only done this in order to see the impact on the size, as measured by the SDK-metrics CI - it was a quick thing to do because I just moved the code from the other repo. 

This way we can actually know what happens rather than trying to guess whether the app size is affected or not (and how much), which may save a lot of talking with the actual data in hand :)

PR #1109, opened 2022-11-07
ci: analyzer action changes

### LINKED ISSUE #840: Package Analysis step in CI does not upload results for contributors
From https://github.com/getsentry/sentry-dart/runs/6113791214?check_suite_focus=true

> Warning:  It seems that this action doesn't have the required permissions to call the GitHub API with the token you gave. This can occur if this repository is a fork, as in that case GitHub reduces the GITHUB_TOKEN's permissions for security reasons. Consequently, no report will be made on GitHub. Check this issue for more information: https://github.com/axel-op/dart-package-analyzer/issues/2

### LINKED ISSUE #957: Use Invertase's Dart analyzer action
See https://github.com/invertase/github-action-dart-analyzer

It can be used for all projects in this repository.

vaind:
## :scroll: Description

* deduplicate analyzer action (use an in-repo reusable workflow)
* add [invertase/github-action-dart-analyzer](https://github.com/invertase/github-action-dart-analyzer) - closes #957 

CI

## :pencil: Checklist

- [ ] All tests passing
- [ ] No breaking changes

### DIFF
--- a/.github/workflows/analyze.yml
+++ b/.github/workflows/analyze.yml
@@ -8,6 +8,11 @@ on:
         required: false
         type: string
         default: dart
+      panaThreshold:
+        description: Minumum percentage of Dart Package Analyzer score that must be achieved.
+        required: false
+        type: number
+        default: 100
 
 jobs:
   analyze:
@@ -54,7 +59,7 @@ jobs:
           TOTAL_MAX: ${{ steps.analysis.outputs.total_max }}
         run: |
           PERCENTAGE=$(( $TOTAL * 100 / $TOTAL_MAX ))
-          if (( $PERCENTAGE < 90 ))
+          if (( $PERCENTAGE < ${{ inputs.panaThreshold }} ))
           then
             echo Score too low!
             exit 1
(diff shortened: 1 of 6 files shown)


.github/workflows/analyze.yml
```diff
@@ -0,0 +1,61 @@
...
+        with:
+          githubToken: ${{ secrets.GITHUB_TOKEN }}
+          relativePath: ${{ inputs.package }}
+      - name: Check scores
+        env:
+          TOTAL: ${{ steps.analysis.outputs.total }}
+          TOTAL_MAX: ${{ steps.analysis.outputs.total_max }}
+        run: |
+          PERCENTAGE=$(( $TOTAL * 100 / $TOTAL_MAX ))
+          if (( $PERCENTAGE < 90 ))
```
marandaneto: This should be configurable, if you look at `if (( $PERCENTAGE < 90 ))`, some packages have a lower or higher value because of ... reasons.

vaind: OK, I've added an input with default 100 and let's see what fails. I'll then update the inputs in the packages that don't pass -in the worst case to original values.

vaind: no issues anywhere, keeping at 100 % for now

marandaneto (approved):
I like the refactoring and using workflows.
Thanks for doing this @vaind

PR #1110, opened 2022-11-07
refactor: use options.clock wherever possible

### LINKED ISSUE #811: Use `SentryOptions#clock` whenever possible instead of `getUtcDateTime` or `DateTime.now()` directly.
It makes it testable and it reuses the function.

vaind:
## :scroll: Description

Closes #811

#skip-changelog

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/dart/lib/src/protocol/sentry_span.dart
+++ b/dart/lib/src/protocol/sentry_span.dart
@@ -32,7 +32,7 @@ class SentrySpan extends ISentrySpan {
     this.samplingDecision,
     Function({DateTime? endTimestamp})? finishedCallback,
   }) {
-    _startTimestamp = (startTimestamp ?? _hub.options.clock()).toUtc();
+    _startTimestamp = startTimestamp?.toUtc() ?? _hub.options.clock();
     _finishedCallback = finishedCallback;
   }
 
@@ -47,13 +47,13 @@ class SentrySpan extends ISentrySpan {
     }
 
     if (endTimestamp == null) {
-      _endTimestamp = _hub.options.clock().toUtc();
+      _endTimestamp = _hub.options.clock();
     } else if (endTimestamp.isBefore(_startTimestamp)) {
       _hub.options.logger(
         SentryLevel.warning,
         'End timestamp ($endTimestamp) cannot be before start timestamp ($_startTimestamp)',
       );
-      _endTimestamp = _hub.options.clock().toUtc();
+      _endTimestamp = _hub.options.clock();
     } else {
       _endTimestamp = endTimestamp.toUtc();
     }
--- a/dart/lib/src/protocol/sentry_span.dart
+++ b/dart/lib/src/protocol/sentry_span.dart
@@ -32,7 +32,7 @@ class SentrySpan extends ISentrySpan {
     this.samplingDecision,
     Function({DateTime? endTimestamp})? finishedCallback,
   }) {
-    _startTimestamp = startTimestamp?.toUtc() ?? getUtcDateTime();
+    _startTimestamp = (startTimestamp ?? _hub.options.clock()).toUtc();
     _finishedCallback = finishedCallback;
   }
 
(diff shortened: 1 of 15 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Fixes
+
+- Use `Options.clock` where possible instead of current UTC DateTime ([#1110](https://github.com/getsentry/sentry-dart/pull/1110))
```
marandaneto: Not really needed since its a non user facing change.

vaind:
~Outstanding~ - [`dartLogger()`](https://github.com/getsentry/sentry-dart/blob/fa7e619df10918bff2062545c9746df886c878bf/dart/lib/src/sentry_options.dart#L409) - should we also use `options.clock` there, somehow?

```dart
/// A Logger that prints out the level and message
void dartLogger(
  SentryLevel level,
  String message, {
  String? logger,
  Object? exception,
  StackTrace? stackTrace,
}) {
  log(
    '[${level.name}] $message',
    level: level.toDartLogLevel(),
    name: logger ?? 'sentry',
    time: getUtcDateTime(),
    error: exception,
    stackTrace: stackTrace,
  );
}
```

dart/lib/src/protocol/sentry_span.dart
```diff
@@ -32,7 +32,7 @@ class SentrySpan extends ISentrySpan {
     this.samplingDecision,
     Function({DateTime? endTimestamp})? finishedCallback,
   }) {
-    _startTimestamp = startTimestamp?.toUtc() ?? getUtcDateTime();
+    _startTimestamp = (startTimestamp ?? _hub.options.clock()).toUtc();
```
marandaneto: `_hub.options.clock()` already returns UTC, you are calling it again.

dart/lib/src/protocol/sentry_span.dart
```diff
@@ -42,20 +42,20 @@ class SentrySpan extends ISentrySpan {
...
 
-    final utcDateTime = getUtcDateTime();
-
     if (status != null) {
       _status = status;
     }
 
-    if (endTimestamp?.isBefore(_startTimestamp) ?? false) {
+    if (endTimestamp == null) {
+      _endTimestamp = _hub.options.clock().toUtc();
```
marandaneto: Same as https://github.com/getsentry/sentry-dart/pull/1110/files#r1016301034
Double check all the calls to `_hub.options.clock()` since its already UTC

dart/lib/src/protocol/sentry_span.dart
vaind: The original code was also converting `toUtc()` from an unknown source, which `_hub.options.clock` is, being a (potentially) user-provided option

dart/lib/src/protocol/sentry_span.dart
vaind: same as above:

> since its already UTC

it is provided by the user, so there's no guarantee it is. Nor even option docs say it should be.

I can drop it of course, just doing the same that was already in this piece of the code.

marandaneto:
@vaind lets add @internal to `SentryOptions#clock` since it should not be used by people.

dart/lib/src/protocol/sentry_span.dart
marandaneto: Yep that makes sense, but this can be:

`startTimestamp?.toUtc() ?? _hub.options.clock()`
Just a micro-optimization

vaind:
> @vaind lets add @internal to `SentryOptions#clock` since it should not be used by people.

OK, with the final comment about making the option internal, I agree it is fine to remove toUTC(). BTW when you do a review as a whole (as opposed to individual comments), the final comment would show above code-review comments and would better clarify the intent.

marandaneto:
> ```dart
> 
> ```

If possible, I'd say so, the reason to use `options.clock` everywhere is testability and reusability.
Plus single source of truth for date and time.

dart/lib/src/protocol/sentry_span.dart
marandaneto: My point is not about making it UTC or not.
`_hub.options.clock()` already returns an UTC date, so there's no need to call `.toUtc()` again.

vaind:
Implemented the requested changes, including the default debug logger change. I've kept public `dartLogger` untouched - not sure it should really be public but there's nothing for it now as it may be used by client code. At best, it could be made deprecated.

marandaneto:
> Implemented the requested changes, including the default debug logger change. I've kept public `dartLogger` untouched - not sure it should really be public but there's nothing for it now as it may be used by client code. At best, it could be made deprecated.

Let's add @deprecated then, so we remove in the next bump.

marandaneto (approved):
Left a last comment, besides that, LGTM

PR #1112, opened 2022-11-07
VendorId should be a String

marandaneto:
https://github.com/getsentry/team-mobile/issues/67

## :pencil: Checklist

- [X] All tests passing
- [ ] No breaking changes

It's technically a breaking change but nobody should be using it directly.
Also its a bug fixing since the previous type does not work.


marandaneto:
@romtsn or @markushi can we fix this on Android? otherwise the serialization would fail on Android anyway.

krystofwoldrich:
I guess this need to wait for Android SDK.

marandaneto:
> I guess this need to wait for Android SDK.

its in the making, no need to block the PR tho.

brustolin (approved):

marandaneto:
@markushi can we cut a release on Android so we can unblock this?

markushi:
@marandaneto the required `sentry-java` changes will be shipped with `6.7.1`, [which is about to be published.](https://github.com/getsentry/publish/issues/1573)

PR #1118, opened 2022-11-09
Add SentryRequest context for HttpException and SocketException

ueman:
## :scroll: Description

Add SentryRequest context for HttpException and SocketException.
Also SocketException and FileSystemException have an underlying OSError, which is now reported as underlying exception.

I've also done some small refactoring of moving the enricher event processors into the event processor folder. I've also added a new constructor to `SentryRequest` and removed duplicated code.

HttpExceptions and SocketException can now be filtered via the request context.

Added new tests.

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,14 @@
 # Changelog
 
-## 6.15.0
+## Unreleased
+
+### Features
 
 - Add request context to `HttpException` and `SocketException` ([#1118](https://github.com/getsentry/sentry-dart/pull/1118))
 - `SocketException` and `FileSystemException` with `OSError`s report the `OSError` as root exception ([#1118](https://github.com/getsentry/sentry-dart/pull/1118))
+
+## 6.15.0
+
 ### Features
 
 - Feat: Screenshot Attachment ([#1088](https://github.com/getsentry/sentry-dart/pull/1088))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 - Add request context to `HttpException` and `SocketException` ([#1118](https://github.com/getsentry/sentry-dart/pull/1118))
 - `SocketException` and `FileSystemException` with `OSError`s report the `OSError` as root exception ([#1118](https://github.com/getsentry/sentry-dart/pull/1118))
+
 ## 6.15.1
 
 ### Various fixes & improvements
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,8 @@
 
 ## Unreleased
 
-- Add request context to `HttpException` and `SocketException`
-- `SocketException` and `FileSystemException` with `OSError`s report the `OSError` as root exception
+- Add request context to `HttpException` and `SocketException` ([#1118](https://github.com/getsentry/sentry-dart/pull/1118))
+- `SocketException` and `FileSystemException` with `OSError`s report the `OSError` as root exception ([#1118](https://github.com/getsentry/sentry-dart/pull/1118))
 ### Features
 
 - Feat: Screenshot Attachment ([#1088](https://github.com/getsentry/sentry-dart/pull/1088))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+- Add request context to `HttpException` and `SocketException`
+- `SocketException` and `FileSystemException` with `OSError`s report the `OSError` as root exception
+
 ### Fixes
 
 - Merging of integrations and packages ([#1111](https://github.com/getsentry/sentry-dart/pull/1111))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Features
 
-- Add request context to `HttpException` and `SocketException` ([#1118](https://github.com/getsentry/sentry-dart/pull/1118))
+- Add request context to `HttpException`, `SocketException` and `NetworkImageLoadException` ([#1118](https://github.com/getsentry/sentry-dart/pull/1118))
 - `SocketException` and `FileSystemException` with `OSError`s report the `OSError` as root exception ([#1118](https://github.com/getsentry/sentry-dart/pull/1118))
 
 ### Fixes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 - Add request context to `HttpException` and `SocketException` ([#1118](https://github.com/getsentry/sentry-dart/pull/1118))
 - `SocketException` and `FileSystemException` with `OSError`s report the `OSError` as root exception ([#1118](https://github.com/getsentry/sentry-dart/pull/1118))
+
 ### Fixes
 
 - Disable `enableUserInteractionBreadcrumbs` on Android when `enableAutoNativeBreadcrumbs` is disabled ([#1131](https://github.com/getsentry/sentry-dart/pull/1131))
--- a/dart/lib/src/event_processor/exception/io_exception_event_processor.dart
+++ b/dart/lib/src/event_processor/exception/io_exception_event_processor.dart
@@ -1,30 +1,35 @@
 import 'dart:io';
 
 import '../../protocol.dart';
+import '../../sentry_options.dart';
 import 'exception_event_processor.dart';
 
-ExceptionEventProcessor exceptionEventProcessor() =>
-    IoExceptionEventProcessor();
+ExceptionEventProcessor exceptionEventProcessor(SentryOptions options) =>
+    IoExceptionEventProcessor(options);
 
 class IoExceptionEventProcessor implements ExceptionEventProcessor {
+  IoExceptionEventProcessor(this._options);
+
+  final SentryOptions _options;
+
   @override
   SentryEvent apply(SentryEvent event, {dynamic hint}) {
     final throwable = event.throwable;
     if (throwable is HttpException) {
-      return applyHttpException(throwable, event);
+      return _applyHttpException(throwable, event);
     }
     if (throwable is SocketException) {
-      return applySocketException(throwable, event);
+      return _applySocketException(throwable, event);
     }
     if (throwable is FileSystemException) {
-      return applyFileSystemException(throwable, event);
+      return _applyFileSystemException(throwable, event);
     }
 
     return event;
   }
 
   // https://api.dart.dev/stable/dart-io/HttpException-class.html
-  SentryEvent applyHttpException(HttpException exception, SentryEvent event) {
+  SentryEvent _applyHttpException(HttpException exception, SentryEvent event) {
     final uri = exception.uri;
     if (uri == null) {
       return event;
@@ -35,7 +40,7 @@ class IoExceptionEventProcessor implements ExceptionEventProcessor {
   }
 
   // https://api.dart.dev/stable/dart-io/SocketException-class.html
-  SentryEvent applySocketException(
+  SentryEvent _applySocketException(
     SocketException exception,
     SentryEvent event,
   ) {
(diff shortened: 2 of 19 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Add request context to `HttpException` and `SocketException`
+- `SocketException` and `FileSystemException` with `OSError`s report the `OSError` as root exception
```
```suggestion
- Add request context to `HttpException` and `SocketException` ([#1118](https://github.com/getsentry/sentry-dart/pull/1118))
- `SocketException` and `FileSystemException` with `OSError`s report the `OSError` as root exception ([#1118](https://github.com/getsentry/sentry-dart/pull/1118))
```

dart/lib/src/event_processor/exception/io_exception_event_processor.dart
```diff
@@ -0,0 +1,94 @@
...
+SentryException _sentryExceptionfromOsError(OSError osError) {
+  return SentryException(
+    type: osError.runtimeType.toString(),
+    value: osError.toString(),
+    mechanism: Mechanism(
+      type: 'OSError',
+      meta: {
+        'code': osError.errorCode,
+      },
+    ),
```
ueman: I'm wondering if one of the listed mechanism metas from [here](https://develop.sentry.dev/sdk/event-payloads/exception/#meta-information) applies here since the OSError surfaces the error code from the OS.

I'd like some guidance here if someone has some more insights.

marandaneto: I guess it makes sense to be set as `meta`, right now meta is only used for iOS, but that's alright, we just never had any other use case on Android/Flutter/RN.
[See Relay spec](https://develop.sentry.dev/sdk/event-payloads/types/#mechanismmeta)
So the `code` should be under the `signal`.
Or even under `errno`, we'd need to understand where the `osError.errorCode` comes from and if it respects a spec.
Last case, we can also set this metadata to the generic `data` field.

dart/lib/src/event_processor/exception/io_exception_event_processor.dart
```diff
@@ -0,0 +1,94 @@
...
+    );
+  }
+
+  // https://api.dart.dev/stable/dart-io/FileSystemException-class.html
+  SentryEvent applyFileSystemException(
+    FileSystemException exception,
+    SentryEvent event,
+  ) {
+    final osError = exception.osError;
+    return event.copyWith(
```
marandaneto, line 91: I'd append to the event the `exception.path` info, it might be useful.

dart/lib/src/event_processor/exception/io_exception_event_processor.dart
```diff
@@ -0,0 +1,94 @@
...
+
+import '../../protocol.dart';
+import 'exception_event_processor.dart';
+
+ExceptionEventProcessor exceptionEventProcessor() =>
+    IoExceptionEventProcessor();
+
+class IoExceptionEventProcessor implements ExceptionEventProcessor {
+  @override
+  SentryEvent apply(SentryEvent event, {dynamic hint}) {
```
marandaneto, line 16: Wondering if there's more exception types that we can enrich the event with.
Good idea.

dart/lib/src/event_processor/exception/io_exception_event_processor.dart
ueman: It's part of the exception message, so I don't think that's needed. See https://api.dart.dev/stable/dart-io/FileSystemException/toString.html

dart/lib/src/event_processor/exception/io_exception_event_processor.dart
ueman: > So the code should be under the signal.
Or even under errno, we'd need to understand where the osError.errorCode comes from and if it respects a spec.

It depends on the OS, as per [OSError.errorCode](https://api.dart.dev/stable/2.18.4/dart-io/OSError/errorCode.html) documentation:

> Error code supplied by the operating system.

Maybe I'll do a follow-up PR, if I have a better understanding of it.

> Last case, we can also set this metadata to the generic data field.

That isn't really surfaced in the Sentry UI though, and that really bugs me because I want to use that for https://github.com/ueman/sentry-dart-tools/tree/main/sentry_link

dart/lib/src/event_processor/exception/io_exception_event_processor.dart
ueman: There are quite a few from the `dart:io` world, for which we could do something like this:
- https://api.dart.dev/stable/2.18.4/dart-io/ProcessException-class.html
- https://api.dart.dev/stable/2.18.4/dart-io/SignalException-class.html
- https://api.dart.dev/stable/2.18.4/dart-io/StdinException-class.html
- https://api.dart.dev/stable/2.18.4/dart-io/StdoutException-class.html
- https://api.dart.dev/stable/2.18.4/dart-io/TlsException-class.html
- https://api.dart.dev/stable/2.18.4/dart-io/CertificateException-class.html
- https://api.dart.dev/stable/2.18.4/dart-io/HandshakeException-class.html
- https://api.dart.dev/stable/2.18.4/dart-io/WebSocketException-class.html
- https://api.dart.dev/stable/2.18.4/dart-isolate/RemoteError-class.html
- ...
Most of those have an underlying `OSError`

I haven't looked into the web side of things.

Exceptions which are valid for both:
- https://api.dart.dev/stable/2.18.4/dart-core/FormatException-class.html - Prettier messages
- Subclasses of `Error`: https://api.dart.dev/stable/2.18.4/dart-core/Error-class.html

ueman: I'm not sure though how much it's worth to try to improve reports for most of those exceptions, though.

dart/lib/src/event_processor/exception/io_exception_event_processor.dart
ueman: I'm pretty sure that `OSError.errorCode` is a `POSIXSignal` error code. I've googled for a bit and checked exceptions in Sentry, and the code matches the posix list of error codes.

ueman: So it should be the following correct?
```dart
Mechanism(
  type: 'OSError',
  meta: {
    'signal': {
      'number': osError.errorCode,
    },
  },
),
```

marandaneto: Yeah, are you testing on which platform? per docs, this is iOS only.
https://develop.sentry.dev/sdk/event-payloads/types/#typedef-POSIXSignal

dart/lib/src/event_processor/exception/io_exception_event_processor.dart
marandaneto: Yeah, I guess we can leverage this event processor over time depending on its needs.

dart/lib/src/event_processor/exception/io_exception_event_processor.dart
marandaneto: Maybe @brustolin or @philipphofmann can help here?

ueman: I've tested only on macOS but also looked at crash reports from Android and iOS. But now I'm confused, maybe the error codes on Android are the `CError` and just look very similar?

marandaneto: We don't have this `meta` field on Android events, so I'm unsure as well, I'll check with the native team, they likely use this more, let's gather feedback from the iOS folks first.

philipphofmann: I'm unsure how I can help here. What exactly is your question, @marandaneto, and @ueman?

ueman: I would like to know which platforms have which mechanism. I.e. is the [`POSIXSignal`](https://develop.sentry.dev/sdk/event-payloads/types/#posixsignal) exclusively for iOS/macOS and other platforms use [`CError`](https://develop.sentry.dev/sdk/event-payloads/types/#cerror)? Also, what's the difference between those two, since they're both talking about posix in their documentation. Or does that not make any difference at all, because I only set the `number` property.

marandaneto: @philipphofmann sharing your knowhow about the https://github.com/getsentry/sentry-cocoa/blob/master/Sources/Sentry/SentryMechanismMeta.m and why it's used for is a good start.
What type of info and where do you get from to set the `meta`?

marandaneto: IIRC this is also considered for grouping events together on iOS.

marandaneto: @ueman let's use the `meta.errno`, but I am unsure if this is displayed on the UI, is it?

ueman: It should be.

ueman: Done

dart/lib/src/event_processor/exception/io_exception_event_processor.dart
```diff
@@ -0,0 +1,104 @@
...
+    }
+    if (throwable is FileSystemException) {
+      return applyFileSystemException(throwable, event);
+    }
+
+    return event;
+  }
+
+  // https://api.dart.dev/stable/dart-io/HttpException-class.html
+  SentryEvent applyHttpException(HttpException exception, SentryEvent event) {
```
marandaneto:
```suggestion
  SentryEvent _applyHttpException(HttpException exception, SentryEvent event) {
```

Also for the other methods, or is there a reason to be public?

dart/lib/src/event_processor/exception/io_exception_event_processor.dart
```diff
@@ -0,0 +1,104 @@
...
+          if (osError != null) _sentryExceptionfromOsError(osError),
+          ...?event.exceptions,
+        ],
+      );
+    }
+    SentryRequest? request;
+    try {
+      var uri = Uri.parse(address.host);
+      request = SentryRequest.fromUri(uri: uri);
+    } catch (_) {}
```
marandaneto: Let's add a log statement instead.

dart/lib/src/event_processor/exception/io_exception_event_processor.dart
```diff
@@ -0,0 +1,104 @@
...
+SentryException _sentryExceptionfromOsError(OSError osError) {
+  return SentryException(
+    type: osError.runtimeType.toString(),
+    value: osError.toString(),
+    // osError.errorCode is likely a posix signal
+    // https://develop.sentry.dev/sdk/event-payloads/types/#mechanismmeta
+    mechanism: Mechanism(
+      type: 'OSError',
+      meta: {
+        'errno': osError.errorCode,
```
marandaneto: `errno` is an object, so we need another map that sets it into the `number`
https://develop.sentry.dev/sdk/event-payloads/types/#cerror

dart/lib/src/protocol/sentry_request.dart
```diff
@@ -81,6 +81,40 @@ class SentryRequest {
...
+    // As far as I can tell there's no way to get the uri without the query part
+    // so we replace it with an empty string.
+    final urlWithoutQuery = uri
+        .replace(query: '', fragment: '')
+        .toString()
+        .replaceAll('?', '')
+        .replaceAll('#', '');
+
+    final query = uri.query.isEmpty ? null : uri.query;
+    final fragment = uri.fragment.isEmpty ? null : uri.fragment;
```
marandaneto, line 104: Let's add the comment back about Dio, either here or in the Dio integration.

dart/lib/src/event_processor/exception/io_exception_event_processor.dart
ueman: Nope, just an oversight from me.

marandaneto (approved):
Thanks @ueman

ueman:
The failures come again because of the newly introduced `SentryRequest.fromUri` constructor which the dio and flutter libraries don't know when pana runs :(

marandaneto:
> The failures come again because of the newly introduced `SentryRequest.fromUri` constructor which the dio and flutter libraries don't know when pana runs :(

yeah, it's pana, after publishing it, it will be green.

PR #1125, opened 2022-11-10
fix: Reset transaction in scope if span is reset

lColinDl:
## :scroll: Description

The private `_transaction` is not set anymore when setting a `span` on `Scope`. Instead the code relies on the public getter for `transaction` that already had logic in using the span name as the `transaction` if available. By doing so the `transaction` name is set back to null after a transaction finishes.

Previously events had a transaction string set of transactions that already finished earlier. Because after a transaction finished, the `_transaction` field in `Scope`was never reset.

Added new new tests. Existing tests still work.

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/dart/lib/src/scope.dart
+++ b/dart/lib/src/scope.dart
@@ -19,28 +19,22 @@ class Scope {
   /// The name of the transaction which generated this event,
   /// for example, the route name: `"/users/<username>/"`.
   String? get transaction {
-    return ((_span is SentryTracer) ? (_span as SentryTracer?)?.name : null) ??
+    return ((span is SentryTracer) ? (span as SentryTracer?)?.name : null) ??
         _transaction;
   }
 
   set transaction(String? transaction) {
     _transaction = transaction;
 
-    if (_transaction != null && _span != null) {
+    if (_transaction != null && span != null) {
       final currentTransaction =
-          (_span is SentryTracer) ? (_span as SentryTracer?) : null;
+          (span is SentryTracer) ? (span as SentryTracer?) : null;
       currentTransaction?.name = _transaction!;
     }
   }
 
-  ISentrySpan? _span;
-
   /// Returns active transaction or null if there is no active transaction.
-  ISentrySpan? get span => _span;
-
-  set span(ISentrySpan? span) {
-    _span = span;
-  }
+  ISentrySpan? span;
 
   SentryUser? _user;
 
@@ -236,7 +230,7 @@ class Scope {
   Future<void> clear() async {
     clearAttachments();
     level = null;
-    _span = null;
+    span = null;
     _transaction = null;
     _fingerprint = [];
     _tags.clear();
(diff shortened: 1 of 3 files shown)


marandaneto:
@lColinDl I don't really follow the issue, the transaction name is bound to the span object, so it is doing what it should do.
If there's a span bound to the scope, return the transaction name otherwise not.
There should not be a transaction name if there's no span in the scope.

ueman:
> There should not be a transaction name if there's no span in the scope.

That's exactly what this PR does, because that's currently not the case.

When a transaction finishes, the transaction name is still added to events from after the transaction.

marandaneto:
> > There should not be a transaction name if there's no span in the scope.
> 
> That's exactly what this PR does, because that's currently not the case.
> 
> When a transaction finishes, the transaction name is still added to events from after the transaction.

That's right, but then this code is breaking the `transaction` feature anyway.
Maybe we should clean the transaction name from the scope here -> https://github.com/getsentry/sentry-dart/blob/25e9b5900ecb2ddc682fe68375964943fe087595/dart/lib/src/sentry_tracer.dart#L113-L118

`scope.transaction` (String) can be used standalone without the APM feature (spans), the difference is that when using the APM feature, we fallback to the span name instead of using the `transaction` String, and both should work as well, the current PR breaks the other feature, or I am missing something.

lColinDl:
> the current PR breaks the other feature, or I am missing something.

I don't see how this PR is breaking existing desired behavior. I intentionally tried to stay as close to the original behavior as possible. Setting the `transaction` name separately should still work as before. Notice also how all existing tests stayed untouched in their behavior.

Unfortunately I can't quite follow the rest of your argument. If you have a better solution to fixing the problem, I obviously don't mind it.

marandaneto:
> > the current PR breaks the other feature, or I am missing something.
> 
> I don't see how this PR is breaking existing desired behavior. I intentionally tried to stay as close to the original behavior as possible. Setting the `transaction` name separately should still work as before. Notice also how all existing tests stayed untouched in their behavior.
> 
> Unfortunately I can't quite follow the rest of your argument. If you have a better solution to fixing the problem, I obviously don't mind it.

You're right, sorry for the confusion, I was misreading the [spec](https://develop.sentry.dev/sdk/performance/#scope-changes).
Can you add a changelog entry? so we can merge it :)
Thanks for doing this.

lColinDl:
> Can you add a changelog entry? so we can merge it :)

Done: 8e21c1c

marandaneto (approved):

dart/lib/src/scope.dart
```diff
@@ -40,12 +40,6 @@ class Scope {
 
   set span(ISentrySpan? span) {
     _span = span;
-
```
marandaneto, line 43: @lColinDl 

> /home/runner/work/sentry-dart/sentry-dart/dart/lib/src/scope.dart:39:20 Avoid wrapping fields in getters and setters just to be "safe".
>   Notice: See https://dart-lang.github.io/linter/lints/unnecessary_getters_setters.html to learn more about this problem.

We can ignore the linter otherwise analyzer will fail.

PR #1130, opened 2022-11-13
Chore: consider using let instead of var type

sgsm74:
because variable 'sdkPackages' was never mutated

_#skip-changelog_

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes


marandaneto:
Makes sense @sgsm74 thanks

marandaneto (approved):

PR #1137, opened 2022-11-16
Auto performance monitoring for widgets

### LINKED ISSUE #706: Auto performance monitoring for widgets
https://api.flutter.dev/flutter/widgets/GestureDetector-class.html

### LINKED ISSUE #723: Breadcrumbs for widgets
https://api.flutter.dev/flutter/widgets/GestureDetector-class.html

Related to https://github.com/getsentry/sentry-dart/issues/706

marandaneto:
## :scroll: Description
Adds a breadcrumb for taps
Starts an Idle transaction for taps

Closes https://github.com/getsentry/sentry-dart/issues/706
Closes https://github.com/getsentry/sentry-dart/issues/723

## :pencil: Checklist

- [X] All tests passing
- [X] No breaking changes

### DIFF
--- a/dart/lib/src/sentry_tracer.dart
+++ b/dart/lib/src/sentry_tracer.dart
@@ -18,6 +18,7 @@ class SentryTracer extends ISentrySpan {
   final Map<String, SentryMeasurement> _measurements = {};
 
   Timer? _autoFinishAfterTimer;
+  Duration? _autoFinishAfter;
   Function(SentryTracer)? _onFinish;
   var _finishStatus = SentryTracerFinishStatus.notFinishing();
   late final bool _trimEnd;
@@ -56,11 +57,9 @@ class SentryTracer extends ISentrySpan {
       startTimestamp: startTimestamp,
     );
     _waitForChildren = waitForChildren;
-    if (autoFinishAfter != null) {
-      _autoFinishAfterTimer = Timer(autoFinishAfter, () async {
-        await finish(status: status ?? SpanStatus.ok());
-      });
-    }
+    _autoFinishAfter = autoFinishAfter;
+
+    _scheduleTimer();
     name = transactionContext.name;
     // always default to custom if not provided
     transactionNameSource = transactionContext.transactionNameSource ??
@@ -116,6 +115,11 @@ class SentryTracer extends ISentrySpan {
           scope.span = null;
         }
       });
+
+      // if it's an idle transaction which has no children, we drop it to save user's quota
+      if (children.isEmpty && _autoFinishAfter != null) {
+        return;
+      }
 
       final transaction = SentryTransaction(this);
       transaction.measurements.addAll(_measurements);
@@ -346,4 +350,24 @@ class SentryTracer extends ISentrySpan {
   @override
   SentryTracesSamplingDecision? get samplingDecision =>
       _rootSpan.samplingDecision;
+
+  @override
+  void scheduleFinish() {
+    if (finished) {
+      return;
+    }
+    if (_autoFinishAfterTimer != null) {
+      _autoFinishAfterTimer?.cancel();
+      _scheduleTimer();
+    }
+  }
+
+  void _scheduleTimer() {
+    final autoFinishAfter = _autoFinishAfter;
+    if (autoFinishAfter != null) {
+      _autoFinishAfterTimer = Timer(autoFinishAfter, () async {
+        await finish(status: status ?? SpanStatus.ok());
+      });
+    }
+  }
 }
--- a/dart/lib/src/sentry_tracer.dart
+++ b/dart/lib/src/sentry_tracer.dart
@@ -201,6 +201,9 @@ class SentryTracer extends ISentrySpan {
       return NoOpSentrySpan();
     }
 
+    // reset the timer if a new child is added
+    _scheduleTimer();
+
     if (children.length >= _hub.options.maxSpans) {
       _hub.options.logger(
         SentryLevel.warning,
@@ -357,7 +360,6 @@ class SentryTracer extends ISentrySpan {
       return;
     }
     if (_autoFinishAfterTimer != null) {
-      _autoFinishAfterTimer?.cancel();
       _scheduleTimer();
     }
   }
@@ -365,6 +367,7 @@ class SentryTracer extends ISentrySpan {
   void _scheduleTimer() {
     final autoFinishAfter = _autoFinishAfter;
     if (autoFinishAfter != null) {
+      _autoFinishAfterTimer?.cancel();
       _autoFinishAfterTimer = Timer(autoFinishAfter, () async {
         await finish(status: status ?? SpanStatus.ok());
       });
--- a/dart/lib/src/sentry_tracer.dart
+++ b/dart/lib/src/sentry_tracer.dart
@@ -19,6 +19,10 @@ class SentryTracer extends ISentrySpan {
 
   Timer? _autoFinishAfterTimer;
   Duration? _autoFinishAfter;
+
+  @visibleForTesting
+  Timer? get autoFinishAfterTimer => _autoFinishAfterTimer;
+
   Function(SentryTracer)? _onFinish;
   var _finishStatus = SentryTracerFinishStatus.notFinishing();
   late final bool _trimEnd;
--- a/flutter/example/lib/main.dart
+++ b/flutter/example/lib/main.dart
@@ -48,7 +48,6 @@ Future<void> setupSentry(AppRunner appRunner) async {
     options.enableNdkScopeSync = true;
     options.enableUserInteractionTracing = true;
     options.attachScreenshot = true;
-
     // We can enable Sentry debug logging during development. This is likely
     // going to log too much for your app, but can be useful when figuring out
     // configuration issues, e.g. finding out why your events are not uploaded.
@@ -214,7 +213,6 @@ class MainScaffold extends StatelessWidget {
               child: const Text('Dio: Web request'),
             ),
             ElevatedButton(
-              key: Key('print_breadcrumb'),
               onPressed: () {
                 print('A print breadcrumb');
                 Sentry.captureMessage('A message with a print() Breadcrumb');
--- a/flutter/example/lib/main.dart
+++ b/flutter/example/lib/main.dart
@@ -31,6 +31,10 @@ Future<void> main() async {
       options.attachThreads = true;
       options.enableWindowMetricBreadcrumbs = true;
       options.addIntegration(LoggingIntegration());
+      options.sendDefaultPii = true;
+      options.reportSilentFlutterErrors = true;
+      options.enableNdkScopeSync = true;
+
       // options.attachScreenshot = true;
       // We can enable Sentry debug logging during development. This is likely
       // going to log too much for your app, but can be useful when figuring out
@@ -200,10 +204,17 @@ class MainScaffold extends StatelessWidget {
               child: const Text('Flutter: Load assets'),
             ),
             ElevatedButton(
-              onPressed: () => makeWebRequestWithDio(context),
+              key: Key('dio_web_request'),
+              onPressed: () async => await makeWebRequestWithDio(context),
               child: const Text('Dio: Web request'),
             ),
             ElevatedButton(
+              key: Key('dio_web_request_2'),
+              onPressed: () async => await makeWebRequestWithDio(context),
+              child: const Text('Dio: Web request 2'),
+            ),
+            ElevatedButton(
+              key: Key('print_breadcrumb'),
               onPressed: () {
                 print('A print breadcrumb');
                 Sentry.captureMessage('A message with a print() Breadcrumb');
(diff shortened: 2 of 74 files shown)


flutter/example/lib/main.dart
```diff
@@ -31,15 +31,19 @@ Future<void> main() async {
...
+      // options.attachScreenshot = true;
       // We can enable Sentry debug logging during development. This is likely
       // going to log too much for your app, but can be useful when figuring out
       // configuration issues, e.g. finding out why your events are not uploaded.
       options.debug = true;
     },
     // Init your App.
     appRunner: () => runApp(
-      SentryScreenshotWidget(
+      SentryUserInteractionWidget(
```
marandaneto: TODO: undo, just testing

flutter/lib/src/sentry_user_interaction_widget.dart
```diff
@@ -0,0 +1,288 @@
...
+  // ignore: no_logic_in_create_state
+  _SentryUserInteractionWidgetState createState() =>
+      _SentryUserInteractionWidgetState(_hub);
+}
+
+class _SentryUserInteractionWidgetState
+    extends State<SentryUserInteractionWidget> {
+  int? _lastPointerId;
+  Offset? _lastPointerDownLocation;
+  final Hub _hub;
```
ueman: This shouldn't be here. In a state you can get properties of the widget via `widget.hub`. No need to explicitly pass it to the state.

ueman: Passing parameters via constructor to the state can even lead to weird behavior, because it gets out of date when the widget itself has changes to what gets passed.

ueman (commented):
This should be tested with platform views. I.e. it shouldn't prohibit interactions with platform views.

marandaneto:
> This should be tested with platform views. I.e. it shouldn't prohibit interactions with platform views.

What do you mean by platform view, [this](https://api.flutter.dev/javadoc/io/flutter/plugin/platform/PlatformView.html)?

ueman:
> > This should be tested with platform views. I.e. it shouldn't prohibit interactions with platform views.
> 
> What do you mean by platform view, [this](https://api.flutter.dev/javadoc/io/flutter/plugin/platform/PlatformView.html)?

Yeah, it can be tested by using one of the webview plugins, google maps or the camera plugin. The reason is that I've seen global click listeners prohibiting interactions with those.

marandaneto:
> > > This should be tested with platform views. I.e. it shouldn't prohibit interactions with platform views.
> > 
> > 
> > What do you mean by platform view, [this](https://api.flutter.dev/javadoc/io/flutter/plugin/platform/PlatformView.html)?
> 
> Yeah, it can be tested by using one of the webview plugins, google maps or the camera plugin. The reason is that I've seen global click listeners prohibiting interactions with those.

gotcha, I recall this from here https://github.com/getsentry/sentry-dart/issues/706#issuecomment-1031382401
Will do, thanks for the reminder.

flutter/lib/src/sentry_user_interaction_widget.dart
```diff
@@ -0,0 +1,287 @@
...
+    );
+    widget._hub.addBreadcrumb(crumb, hint: tappedWidget.element.widget);
+
+    // TODO: options
+    // ignore: invalid_use_of_internal_member
+    if (!widget._hub.options.isTracingEnabled()) {
+      return;
+    }
+
+    // TODO: name should be screenName.widgetName, maybe get from router?
```
marandaneto: @ueman is it possible to find out which screen (maybe the root `Scaffold`?) a widget belongs to?
For example, `MainScaffold.btn_login`, I know that the `btn_login` comes from the `MainScaffold` screen.
I could not find any way that really makes sense.

ueman: If you have the context of the widget, you can do `ModalRoute.of(context).settings.name;`.

You have to track which `StatefulElement` is the closest to the widget which gets tracked. Via that `StatefulElement`, you can do `ModalRoute.of(statefulElement.state.context).settings.name`.

So in `_getDescriptionFrom` you can do something like 
```dart
TappedWidget? _getDescriptionFrom(Element element) {
    String? currentRouteName;
    // Add the following lines
    final widget = element.widget;
    if(element is StatefulElement) {
      currentRouteName = ModalRoute.of(statefulElement.state.context).settings.name;
    }
    // Used by ElevatedButton, TextButton, OutlinedButton.
    if (widget is ButtonStyleButton) {
      if (widget.enabled) {
        return TappedWidget(
          element: element,
          description: _findDescriptionOf(element, true),
          type: 'ButtonStyleButton',
        );
      }
    } else if (widget is MaterialButton) {
      if (widget.enabled) {
        return TappedWidget(
          element: element,
          description: _findDescriptionOf(element, true),
          type: 'MaterialButton',
        );
      }
    } else if (widget is CupertinoButton) {
      if (widget.enabled) {
        return TappedWidget(
          element: element,
          description: _findDescriptionOf(element, true),
          type: 'CupertinoButton',
        );
      }
    } else if (widget is InkWell) {
      if (widget.onTap != null) {
        return TappedWidget(
          element: element,
          description: _findDescriptionOf(element, false),
          type: 'InkWell',
        );
      }
    } else if (widget is IconButton) {
      if (widget.onPressed != null) {
        return TappedWidget(
          element: element,
          description: _findDescriptionOf(element, false),
          type: 'IconButton',
        );
      }
    } else if (widget is GestureDetector) {
      if (widget.onTap != null) {
        return TappedWidget(
          element: element,
          description: '',
          type: 'GestureDetector',
        );
      }
    }

    return null;
  }
```

No clue about the performance impact of calling `ModalRoute.of` for each `StatefulElement` but it could be a bit too much. Better to optimize it :D

marandaneto: Yeah and then it works only for `StatefulElement` as well, `StatelessElement` elements don't have a `state`.
Maybe I will mention in the docs for now that the `key` must be unique across screens as well.
The other option is if I make it work along with `SentryNavigatorObserver`, because `SentryNavigatorObserver` always knows what's the current `route` if the observer is installed.

ueman: True, but I would bet that it's basically guaranteed that there's one, though.

I wouldn't use SentryNavigatorObserver tbh, because it's common to have multiple navigators and then you have to figure out which one is the navigator to get the current route from

dart/lib/src/sentry_tracer.dart
```diff
@@ -197,6 +205,9 @@ class SentryTracer extends ISentrySpan {
       return NoOpSentrySpan();
     }
 
+    // reset the timer if a new child is added
+    _scheduleTimer();
```
brustolin, line 209: m: Cocoa uses I different approach that I believe makes more sense.
Instead o reseting the timer, we cancel it all, and every time all child spans are finished we start the timer again to see if something else will happen.

Some operations may take more than the default timeout, together with `waitForChildren` what we really need to do is give a little room to something happen after an operation is finished.

flutter/lib/src/user_interaction/sentry_user_interaction_widget.dart
```diff
@@ -0,0 +1,335 @@
...
+      transactionNameSource: SentryTransactionNameSource.component,
+    );
+
+    final activeTransaction = _activeTransaction;
+    if (activeTransaction != null) {
+      if (_lastTappedWidget?.element.widget == element.widget &&
+          _lastTappedWidget?.eventType == tappedWidget.eventType &&
+          !activeTransaction.finished) {
+        // ignore: invalid_use_of_internal_member
+        activeTransaction.scheduleFinish();
```
brustolin, line 157: m: If you're gonna create a new one you dont need to reset the old one timer. Just call `finish` and let `waitForChildren` take care of it.

marandaneto: @brustolin not sure I get it, here the user clicked on the same button twice, before the timer was even finished, so we just reset the timer, Android and iOS do the same, its also part of the spec, can you clarify?

dart/lib/src/sentry_tracer.dart
marandaneto: @brustolin https://develop.sentry.dev/sdk/performance/ui-event-transactions/

> The SDK schedule an idle timeout for the transaction when starting it. When the transaction or any of its spans starts a new span, it resets the timeout. The SDK finishes the transaction when the timeout succeeds.

I believe this implementation meets the spec, Android is also implemented like that. I'm not 100% sure about iOS.

flutter/lib/src/user_interaction/sentry_user_interaction_widget.dart
marandaneto: https://develop.sentry.dev/sdk/performance/ui-event-transactions/

> Scenario: Same UI element with same event
>     Given an ongoing UI event transaction
>     When the user triggers the same UI event with the same type for the same
>         UI element
>     Then the SDK reschedule the idle timeout
>     And doesn't create a new transaction

flutter/lib/src/user_interaction/sentry_user_interaction_widget.dart
```diff
@@ -0,0 +1,335 @@
...
+    // WidgetsBinding.instance.renderViewElement does not work, so using
+    // the element from createElement
+    final rootElement = _clickTrackerElement;
+    if (rootElement == null || rootElement.widget != widget) {
+      return null;
+    }
+
+    UserInteractionWidget? tappedWidget;
+
+    void elementFinder(Element element) {
```
brustolin, line 250: h: Consider this:

```dart
Card(
                child: GestureDetector(
                  onTap: () => {
                    //open Card
                  },
                  child: Stack(
                    children: [
                      //fancy card layout
                      ElevatedButton(
                        onPressed: () => {
                          //mark card as favorite
                        },
                        child: const Text('Favorite'),
                      ),
                    ],
                  ),
                ),
              )
```

If a user press the "Favorite" `ElevatedButton`, the element found will always be the `GestureDetector`.
I guess this will be a problem to really detect a potential problem.

dart/lib/src/sentry_tracer.dart
brustolin: This is not a blocker. 
Im just suggestion something I believe make more sense. 
We can keep the current implementation.

flutter/lib/src/user_interaction/sentry_user_interaction_widget.dart
brustolin: My bad, is the same widget. Gotcha!

flutter/lib/src/user_interaction/sentry_user_interaction_widget.dart
brustolin: It's kind ok to keep this as it is. Since UI Transactions are meant to a bag for spans. Maybe we're ok to not have the perfect origin of a transaction as long we have a transaction.

But if you can figure it out how to solve this would be nice.

brustolin (approved):
There is this thing with the `elementFinder` that is worth taking a look at.
Apart from that, LGTM.

flutter/lib/src/user_interaction/sentry_user_interaction_widget.dart
marandaneto: Very good catch, the `elementFinder` always goes to the `GestureDetector`.
@ueman any ideas what to do here? since we don't know the instance of the `onTap` or `onPressed` to compare, I don't know if we can solve it.
@brustolin  The question is, does this even make sense? Having a parent `onTap` and an `onPressed` child? Is this a real use case? Flutter always executes the `onPressed` from the `ElevatedButton`, always ignores the `onTap` from `GestureDetector`, maybe we should just go till the very end of the tree and use the last found child?

brustolin: Yes, this is a real use case.
I had used this a lot.

ueman: > @ueman any ideas what to do here? since we don't know the instance of the onTap or onPressed to compare, I don't know if we can solve it.

> maybe we should just go till the very end of the tree and use the last found child?

Yeah, maybe don't stop searching on the first hit, but for the smallest hit?

Maybe there's a way to get some information from the gesture arena (which is the mechanism which decides which widget was tapped)? There's some information here https://docs.flutter.dev/development/ui/advanced/gestures#gesture-disambiguation

marandaneto: It turns out that going to the latest level of `child` is also wrong, because it ends up with an `InkWell` which is a child of the desired `ElevatedButton`.
Will do a bit of research.

marandaneto: For now, I dropped support for the nested `GestureDetector` so the PR is unblocked, I will keep searching on how to apply the gesture disambiguation
https://docs.flutter.dev/development/ui/advanced/gestures#gesture-disambiguation

PR #1148, opened 2022-11-19
PoC: Show UserFeedback dialog on events

ueman:
## :scroll: Description

This enables the automatic presenting of the user feedback dialog after events.
I consider this a proof of concept, and it's largely untested. However, it works quite well.

- Feedback dialog has no loading and "thank you" UIs, so its UX could be improved. But that's beyond the scope of this PR
- The dialog has just basic validation
- Maybe there should be an individual sampler for the dialog? It's quite annoying if it shows for each captured event
- The dialog is only shown if `beforeSend` does not return null
- Maybe this can be landed mostly as is, if marked as experimental?
- Dialog is modelled after https://docs.sentry.io/platforms/javascript/enriching-events/user-feedback/
- setting `options.showUserFeedbackDialog = true` enables the default behavior.
- For more advanced usages, `options.userFeedbackBuilder` can be used. This enables customizations, like custom translations and custom dialogs.

# :green_heart: How did you test it?

Manually :D

## :pencil: Checklist

- [ ] All tests passing
- [ ] No breaking changes


marandaneto:
The problem is that each app has its own UI/design system and isn't scalable to maintain a certain level of UI configuration for the custom user feedback screen.
We provide an API to capture user feedback, also an API to retrieve the `lastEventId` even if captured automatically, we believe that this is all one needs.
The JS SDK has its own "Sentry screen for user feedback", react native could technically reuse that but we've decided to not do it and instead, offer the public API for user feedback, similar to this one.
So we encourage users to use the public API, but not having a built-in user feedback screen in the SDK.

PR #1158, opened 2022-11-30
Update dev deps

marandaneto:
## :scroll: Description
_#skip-changelog_

## :pencil: Checklist

- [ ] All tests passing
- [ ] No breaking changes

### DIFF
--- a/.github/workflows/flutter.yml
+++ b/.github/workflows/flutter.yml
@@ -99,11 +99,11 @@ jobs:
           flutter test --coverage --test-randomize-ordering-seed=random
 
       #   remove comment after https://github.com/codecov/codecov-action/issues/600
-      # - uses: codecov/codecov-action@v2
-      #   if: runner.os == 'Linux' && matrix.sdk == 'stable' && matrix.target == 'linux'
-      #   with:
-      #     name: sentry_flutter
-      #     file: ./flutter/coverage/lcov.info
+      - uses: codecov/codecov-action@d9f34f8cd5cb3b3eb79b3e4b5dae3a16df499a70 # pin@v3
+        if: runner.os == 'Linux' && matrix.sdk == 'stable' && matrix.target == 'linux'
+        with:
+          name: sentry_flutter
+          file: ./flutter/coverage/lcov.info
 
       - uses: VeryGoodOpenSource/very_good_coverage@84e5b54ab888644554e5573dca87d7f76dec9fb3 # pin@v2.0.0
         if: runner.os == 'Linux' && matrix.sdk == 'stable' && matrix.target == 'linux'
--- a/.github/workflows/flutter.yml
+++ b/.github/workflows/flutter.yml
@@ -104,7 +104,6 @@ jobs:
       #   with:
       #     name: sentry_flutter
       #     file: ./flutter/coverage/lcov.info
-      # or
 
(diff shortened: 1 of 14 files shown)


dio/pubspec.yaml
```diff
@@ -18,4 +18,4 @@ dev_dependencies:
   test: ^1.21.1
   coverage: ^1.3.0
   mockito: ^5.1.0
-  http: ^0.13.4
+  http: ^0.13.0
```
marandaneto: @ueman do you remember if we used min `.4` for some reason or is it fine to align with the `sentry` dep which is `.0`?

.github/workflows/flutter.yml
```diff
@@ -99,11 +99,11 @@ jobs:
           flutter test --coverage --test-randomize-ordering-seed=random
 
       #   remove comment after https://github.com/codecov/codecov-action/issues/600
-      # - uses: codecov/codecov-action@v2
-      #   if: runner.os == 'Linux' && matrix.sdk == 'stable' && matrix.target == 'linux'
-      #   with:
-      #     name: sentry_flutter
-      #     file: ./flutter/coverage/lcov.info
+      - uses: codecov/codecov-action@d9f34f8cd5cb3b3eb79b3e4b5dae3a16df499a70 # pin@v3
```
marandaneto: Let's see if we still run into this issue

dio/pubspec.yaml
ueman: The http dep shouldn't be needed in the dio package at all. After all, dio is a replacement for http.

ueman: Oh, it's a dev dependency in which case it shouldn't matter, since it doesn't affect users. Still, I wonder why it's there

marandaneto: will check if its used in tests, otherwise will remove it, it'd be nice to have a linter that warn us something isn't being used :D

marandaneto: yeah it was a mistake, removing it.

.github/workflows/flutter.yml
```diff
@@ -98,12 +101,13 @@ jobs:
...
-      #   with:
-      #     name: sentry_flutter
-      #     file: ./flutter/coverage/lcov.info
+        # use codecov/codecov-action after this https://github.com/codecov/codecov-action/issues/600
+      - name: Upload coverage to Codecov
+        if: runner.os == 'Linux' && matrix.sdk == 'stable' && matrix.target == 'linux'
+        run: |
+          curl -Os https://uploader.codecov.io/linux/latest/codecov
+          chmod +x codecov
+          ./codecov -t ${CODECOV_TOKEN} -n sentry_flutter -f ./flutter/coverage/lcov.info
```
marandaneto: @kahest can you add the `CODECOV_TOKEN` to our repo?

> This assumes that you've set your Codecov token inside Settings > Secrets as CODECOV_TOKEN. If not, you can [get an upload token](https://docs.codecov.io/docs/frequently-asked-questions#section-where-is-the-repository-upload-token-found-) for your specific repo on [codecov.io](https://www.codecov.io/). Keep in mind that secrets are not available to forks of repositories.

kahest: done

brustolin (approved):
LGTM

PR #1160, opened 2022-12-02
Tracing for File IO integration

### LINKED ISSUE #835: Official support for File I/O auto performance
@ueman did it already and told us we could borrow the idea if we want to officially support those integrations https://github.com/ueman/sentry-dart-tools/tree/main/sentry_plus/lib/src/file
There's more than File IO stuff, it requires adapting to Sentry's API design and most likely tests.

marandaneto:
## :scroll: Description
Tracing for File IO integration

Closes https://github.com/getsentry/sentry-dart/issues/835

screenshot_2022-12-05_at_13 35 47

## :pencil: Checklist

- [X] I updated the docs if needed -> https://github.com/getsentry/sentry-docs/pull/5895
- [X] All tests passing
- [X] No breaking changes

## :crystal_ball: Next steps


.craft.yml
```diff
@@ -9,10 +9,12 @@ targets:
...
       dio:
+      file:
   - name: github
   - name: registry
     sdks:
       pub:sentry:
       pub:sentry_flutter:
       pub:sentry_logging:
       pub:sentry_dio:
+      #pub:sentry_dio:
```
marandaneto: Double check if craft is already able to auto create the first entry in release registry

file/analysis_options.yaml
```diff
@@ -0,0 +1,24 @@
+include: package:lints/recommended.yaml
+
+analyzer:
```
ueman, line 3: It might be worth to enable stricter analyzer checks, since this is a new package with significantly less code than the other packages.
https://dart.dev/guides/language/analysis-options#enabling-additional-type-checks
```suggestion
analyzer:
  language:
    strict-casts: true
    strict-inference: true
    strict-raw-types: true
```

file/lib/src/sentry_file_extension.dart
```diff
@@ -0,0 +1,42 @@
...
+  /// final file = File('test.txt');
+  /// final sentryFile = SentryFile(file);
+  /// // span starts
+  /// await sentryFile.writeAsString('Hello World');
+  /// // span finishes
+  /// ```
+  ///
+  /// All the copy, create, delete, open, rename, read, and write operations are
+  /// supported.
+  File sentryTrace({
```
ueman, line 31: What's the reason to not just automatically enable it for all file io?

You can do that with the following code:
```dart
IOOverrides.global = SentryIoOverrides(HubAdapter(), options);

File f = File('path/to/file'); // file is SentryFile by default

class SentryIoOverrides extends IOOverrides {
  final Hub _hub;
  final SentryOptions _options;

  SentryIoOverrides(this._hub, this._options);

  @override
  File createFile(String path) {
    if (!_options.isTracingEnabled()) {
      return super.createFile(path);
    }
    return SentryFile(
      super.createFile(path),
      _hub,
      _options,
    );
  }
}
```

file/analysis_options.yaml
marandaneto: Good catch, done.

file/lib/src/sentry_file_extension.dart
marandaneto: [I did see that](https://api.dart.dev/stable/2.18.5/dart-io/IOOverrides-class.html) and it is indeed awesome, but its also misusing the mocks API, as an SDK, we need to be a bit more conservative about that, I guess.
I'm afraid that this could break the app's behavior that relies on the `File` type or something like that, or, that some files take more time than usual and you don't want to trace them for some reason (overhead, long transactions, etc), an all in solution could be a deal breaker for some users.
I'd prefer to do this next after gathering some feedback on the current implementation.
What do you think?

.craft.yml
marandaneto: @adinauer @kamilogorek what's the outcome of that problem, do we still need to do the release registry dance manually or is it fine now?
https://github.com/getsentry/sentry-release-registry/pull/85

marandaneto:
@krystofwoldrich @brustolin this is ready for review

file/lib/src/sentry_file_extension.dart
ueman: Yeah, that sounds reasonable

ueman: I was mostly asking because `IOOverrides` allow you to also measure the performance, if you don't otherwise control the file io yourself. Like for example with [Hive](https://pub.dev/packages/hive), [sembast](https://pub.dev/packages/sembast) or similar.

.craft.yml
kamilogorek: Yes, it's still "broken"

file/lib/src/sentry_file_extension.dart
marandaneto: Yep, I believe it still makes sense to have it, its a hook that makes it possible since Dart does not have reflection nor monkey patching, not sure if it'd be ever enabled by default tho.
At least not during the first version.

.craft.yml
krystofwoldrich: This will be `pub:sentry_file`?

marandaneto:
```suggestion
      #pub:sentry_file:
```

marandaneto: yes, good catch.

file/lib/src/sentry_file.dart
```diff
@@ -0,0 +1,387 @@
...
+    return _wrap(
+      () async => _file.create(recursive: recursive),
+      'file.write',
+    );
+  }
+
+  @override
+  void createSync({bool recursive = false, bool exclusive = false}) {
+    return _wrapSync(
+      () => _file.createSync(recursive: recursive),
```
krystofwoldrich, line 66:
```suggestion
      () => _file.createSync(recursive: recursive, exclusive: exclusive),
```

Should we pass also the `exclusive`?

marandaneto: Nope, that will come in the next version due to being a breaking change in the Dart language.

brustolin (approved):
LGTM

file/lib/src/sentry_file.dart
```diff
@@ -0,0 +1,387 @@
...
+      try {
+        length = _file.lengthSync();
+        hasLength = true;
+      } catch (_) {
+        // ignore in case something goes wrong
+      }
+
+      data = callback() as T;
+
+      if (!hasLength) {
```
krystofwoldrich, line 291: L: What if we are overriding the content of the file do we want to know the length before or after?
If I'm looking correctly at the moment we would get the length before.

krystofwoldrich (approved):

file/lib/src/sentry_file.dart
marandaneto: The data can only be read before, because after, the file can be deleted, renamed, moved, so it's a trade-off we made.
Maybe I should point that out in the docs.

PR #1163, opened 2022-12-06
Enable strict mode for all SDKs but dart

marandaneto:
## :scroll: Description
_#skip-changelog_

https://dart.dev/guides/language/analysis-options#enabling-additional-type-checks

## :pencil: Checklist

- [ ] All tests passing
- [ ] No breaking changes

## :crystal_ball: Next steps
`dart` package but that would be another PR since its much bigger and more fragile


dio/analysis_options.yaml
```diff
@@ -1,10 +1,9 @@
 include: package:lints/recommended.yaml
 
 analyzer:
-  strong-mode:
-    implicit-casts: false
-    implicit-dynamic: false
   language:
+    strict-casts: true
+    strict-inference: true
```
marandaneto, line 6: @ueman IIRC these 3 options replace `strong-mode`, right?

ueman: That's correct

brustolin (approved):

PR #1184, opened 2022-12-12
Refactor `BindingUtils` to enable the use of custom bindings

ueman:
## :scroll: Description
This change enables the use of a custom WidgetsBinding. It doesn't introduce a custom binding though.

This change is done for the v7 branch, since this could be considered a breaking change.

I'm experimenting with initializing a custom widgets binding to add even more observability. Currently, that's not really supported. That's why I propose the following refactoring.

The new interface is modelled after the methods of the `WidgetsBinding` Sentry needs. By using a custom implementation of the `BindingWrapper` and assigning it to the options, you can make use of a custom binding. The custom binding then gets correctly initialized and used by Sentry.

This is covered by existing tests.

## :pencil: Checklist

- [x] All tests passing
- [ ] No breaking changes


flutter/lib/src/sentry_flutter_options.dart
```diff
@@ -251,4 +252,7 @@ class SentryFlutterOptions extends SentryOptions {
       useFlutterBreadcrumbTracking();
     }
   }
+
+  @internal
```
ueman, line 256: Maybe also `@experimental`?

flutter/lib/src/sentry_flutter_options.dart
```diff
@@ -251,4 +252,7 @@ class SentryFlutterOptions extends SentryOptions {
...
+  BindingWrapper bindingUtils = BindingWrapper();
```
ueman:
```suggestion
  /// Setting this to a custom [BindingWrapper] allows you to use a custom [WidgetsBinding].
  @experimental
  @internal
  BindingWrapper bindingUtils = BindingWrapper();
```

flutter/lib/src/binding_wrapper.dart
```diff
@@ -0,0 +1,14 @@
...
+  /// Flutter >= 2.12 throws if WidgetsBinding.instance isn't initialized.
+  // When this method is called, it is guaranteed that a binding was already
+  // initialized.
+  WidgetsBinding get instance => _ambiguate(WidgetsBinding.instance);
+
+  /// Initializes the Binding.
+  void ensureInitialized() => WidgetsFlutterBinding.ensureInitialized();
+}
+
+WidgetsBinding _ambiguate(WidgetsBinding? binding) => binding!;
```
marandaneto: The problem here is that `BindingBase.checkInstance` can throw a `FlutterError`.
The utils had a try/catch and this one does not.

marandaneto: I don't know if this would ever happen tho, since the SDK calls `ensureInitialized` automatically, but better safe than sorry.

ueman: `WidgetsBinding.instance` throwing indicates that we're not using it correctly. That's also why I introduced both, `ensureInitialized()` and `instance`. Having both makes it obvious when it fails why it failed. (Which is not initializing the binding before using it)

After the first call to `ensureInitialized()`, it's always safe to call `instance`.

I can reintroduce those safeguards, but I don't think it's necessary. The current Flutter version also has the same API like this PR does. If the try-catch is added back, the try catch should not swallow the exception anymore, because that makes it basically impossible to notice or to fix, if it happens to go wrong.

marandaneto: Yeah I totally get it, we can always log out in case this happens, but the SDK should never be a reason to degrade people's apps, if we call it automatically and we cause the prob, we should log out and swallow the exception rather than making the App or the SDK unusable.
https://develop.sentry.dev/sdk/philosophy/#degrade-gracefully
So I rather prefer a functional SDK + events without the contexts that depend on it rather than not working at all.
In general, this would never happen, so it's not a biggie, but the wrapper should still allow returning `null`.
Does that make sense?

marandaneto:
@ueman I will review that after the holidays, thank you for doing this btw.

flutter/lib/src/integrations/on_error_integration.dart
```diff
@@ -29,15 +28,11 @@ class OnErrorIntegration implements Integration<SentryFlutterOptions> {
   @override
   void call(Hub hub, SentryFlutterOptions options) {
     _options = options;
-    final binding = BindingUtils.getWidgetsBindingInstance();
-
-    if (binding == null) {
-      return;
-    }
```
marandaneto: I believe this should stay, otherwise, it won't work anyway, so bail out early.

flutter/lib/src/integrations/widgets_binding_integration.dart
```diff
@@ -10,10 +9,12 @@ import '../widgets_binding_observer.dart';
 ///   - [SentryWidgetsBindingObserver]
 ///   - [WidgetsBindingObserver](https://api.flutter.dev/flutter/widgets/WidgetsBindingObserver-class.html)
 class WidgetsBindingIntegration extends Integration<SentryFlutterOptions> {
-  SentryWidgetsBindingObserver? _observer;
+  late SentryWidgetsBindingObserver _observer;
+  late SentryFlutterOptions _options;
```
marandaneto: Better to keep it nullable, so if `close` is called before `call` which is wrong, wouldn't throw, it's defensive programming against API misusing.

flutter/lib/src/sentry_flutter_options.dart
```diff
@@ -251,4 +252,9 @@ class SentryFlutterOptions extends SentryOptions {
       useFlutterBreadcrumbTracking();
     }
   }
+
+  /// Setting this to a custom [BindingWrapper] allows you to use a custom [WidgetsBinding].
+  @experimental
```
marandaneto, line 258: If its `internal`, we dont need to use `experimental`

CHANGELOG.md
```diff
@@ -26,9 +26,11 @@
     - theme
   - Removed isolate name from Dart context. It's now reported via the threads interface. It can be enabled via `options.attachThreads`
 - Use `sentryClientName` instead of `sdk.identifier` ([#1135](https://github.com/getsentry/sentry-dart/pull/1135))
+- Refactor `BindingUtils` to `BindingWrapper` to enable the use of custom bindings ([#1184](https://github.com/getsentry/sentry-dart/pull/1184))
```
marandaneto, line 29: If this is a refactor, we don't need an entry here, it's more about user-facing changes.

flutter/lib/src/binding_wrapper.dart
```diff
@@ -0,0 +1,37 @@
+import 'dart:developer';
+
+import 'package:flutter/widgets.dart';
+
+import '../sentry_flutter.dart';
+
+// The methods and properties are modelled after the the real binding class.
```
marandaneto:
```suggestion
/// The methods and properties are modelled after the the real binding class.
@internal
```

flutter/lib/src/binding_wrapper.dart
```diff
@@ -0,0 +1,37 @@
...
+    try {
+      return _ambiguate(WidgetsBinding.instance);
+    } catch (e, s) {
+      log(
+        'WidgetsBinding.instance was not yet initialized',
+        level: SentryLevel.error.toDartLogLevel(),
+        error: e,
+        stackTrace: s,
+        name: 'Sentry',
+      );
```
marandaneto: What if `BindingWrapper` takes a `HubAdapter` so we can `hub.options.logger(...)`.
Otherwise, we just remove this call, the SDK should use its own logger.

flutter/test/integrations/not_initialized_widgets_binding_test.dart
```diff
@@ -1,29 +0,0 @@
-import 'package:flutter_test/flutter_test.dart';
-import 'package:sentry_flutter/sentry_flutter.dart';
-import 'package:sentry_flutter/src/integrations/widgets_binding_integration.dart';
-
-import '../mocks.mocks.dart';
-
-/// Tests that require `WidgetsFlutterBinding.ensureInitialized();` not
-/// being called at all.
-void main() {
```
marandaneto: Shouldn't just adapt the tests to use the new wrapper?

flutter/test/integrations/on_error_integration_no_binding_test.dart
```diff
@@ -1,59 +0,0 @@
...
-import 'package:mockito/mockito.dart';
-import 'package:sentry/sentry.dart';
-import 'package:sentry_flutter/src/integrations/on_error_integration.dart';
-import 'package:sentry_flutter/src/sentry_flutter_options.dart';
-
-import '../mocks.dart';
-import '../mocks.mocks.dart';
-import 'mock_platform_dispatcher.dart';
-
-void main() {
```
marandaneto: Same as https://github.com/getsentry/sentry-dart/pull/1184/files#r1060373844

flutter/lib/src/sentry_flutter_options.dart
ueman: It should be experimental, but not internal. After all it's designed to enable the use of custom WidgetsBindings.

CHANGELOG.md
ueman: I intend this to be user facing, although maybe it shouldn't be advertised.

flutter/test/integrations/not_initialized_widgets_binding_test.dart
ueman: I re-added it

flutter/test/integrations/on_error_integration_no_binding_test.dart
ueman: I re-added it

marandaneto (approved):
Thanks @ueman

PR #1189, opened 2022-12-14
Feat View hierarchy

marandaneto:
## :scroll: Description
View hierarchy as an attachment

https://github.com/getsentry/team-mobile/issues/64

Screenshot 2022-12-15 at 08 30 33

## :pencil: Checklist

- [X] All tests passing
- [X] No breaking changes


flutter/lib/src/view_hierarchy/sentry_tree_walker.dart
```diff
@@ -0,0 +1,75 @@
...
+      node.width = size?.width;
+      node.height = size?.height;
+    }
+    if (widget is Visibility) {
+      node.visible = widget.visible;
+    }
+    // TODO: not sure how to get Color#alpha direcly?
+    if (widget is Opacity) {
+      node.alpha = widget.opacity;
+    }
```
marandaneto: @ueman is that how people do visibility and opacity?
Opacity could be in specific properties, for example, the text color of the borders or something, so I believe this is not scalable.

flutter/lib/src/view_hierarchy/sentry_tree_walker.dart
```diff
@@ -0,0 +1,75 @@
...
+    // TODO: missing x, y, z, and extra if any?
```
marandaneto: @ueman do you know if its possible to get the `x` and `y` thru the `element` or `widget`? Could not find a way or I am checking the wrong properties, not sure.

ueman: I've updated my code at https://github.com/ueman/sentry-dart-tools/blob/main/sentry_flutter_plus/lib/src/integrations/tree_walker_integration.dart 

It includes support for the position now

flutter/lib/src/view_hierarchy/sentry_tree_walker.dart
```diff
@@ -0,0 +1,75 @@
...
+    return (Element element) {
+      final node = _toSentryViewHierarchyElement(element);
+      parentNode.children.add(node);
+      element.visitChildElements(_visitor(node));
+    };
+  }
+
+  SentryViewHierarchy? toSentryViewHierarchy() {
+    final rootNode = _toSentryViewHierarchyElement(rootElement);
+    rootElement.visitChildElements(_visitor(rootNode));
```
marandaneto: @ueman so the result of this for simple screens is quite deep in-depth, and it looks way more deeper in depth than the view hierarchy of the dev tooling, have you noticed that, or did I mess up something?

flutter/lib/src/view_hierarchy/sentry_tree_walker.dart
ueman: Yeah, I would guess that's correct. To be honest, I don't think the properties as outlined in the RFC make much sense here. Everything, which is a child of `Opacity` will have the opacity applied. So it's not like a property of the children.

I wouldn't try to get the opacity out of the color of the widgets. There's just way too many widgets for that, and you would have to special-case each of them. I mean theoretically a `ColoredBox` can have opacity when the chosen color has it, same for `Container`, `DecoratedBox`, and so on.

flutter/lib/src/view_hierarchy/sentry_tree_walker.dart
ueman: Unfortunately yes. See https://github.com/getsentry/team-mobile/issues/64#issuecomment-1284547394 for some ideas on how to tackle that.

Edit: Copied it over:

> The view hierarchy in Flutter is way deeper than those of Android and iOS. Classes of views (views are called widgets in Flutter) prefixed with a `_` are private (that's how it works in Dart). It might be worth to ignore private classes since you most probably can't look them up in docs and so on. Public children of private classes should be shown again, though. Private classes are often used to improve the structuring of a widget, so they are kinda an implementation detail. Classes not prefixed with an underscore are public and one should be able to look them up in docs. Ignoring private classes is currently not done in my code. Also, you don't know whether a class is from user code, 3-party code or framework code, since there's no reflection/introspection in Dart. When enabling obfuscation in Flutter, the class names are of course not really human readable, but I guess that's similar for Android/iOS.
> 
> In addition to the names of the widgets, you could also add the content of well known widgets like Text and *Buttons, as well as the on screen size and coordinates.

flutter/lib/src/view_hierarchy/sentry_tree_walker.dart
marandaneto: Yeah I'd just focus on both of them and highlight that on the docs, its fine, thanks for the input.

flutter/lib/src/view_hierarchy/sentry_tree_walker.dart
marandaneto: Oh indeed, missed that, thank you for sharing your thoughts, let me try this out.

ueman: I wonder if the `_` check for private classes also works in obfuscated code

marandaneto: I will check, also found this https://docs.flutter.dev/development/tools/devtools/inspector#track-widget-creation

marandaneto: There are also a few methods that print the tree, but only on debug mode https://docs.flutter.dev/testing/code-debugging#debug-flags-performance

dart/lib/src/protocol/sentry_view_hierarchy_element.dart
```diff
@@ -0,0 +1,74 @@
...
+    if (identifier != null) {
+      json['identifier'] = identifier;
+    }
+    if (children.isNotEmpty) {
+      json['children'] =
+          children.map((e) => e.toJson()).toList(growable: false);
+    }
+    if (width != null) {
+      json['width'] = width;
+    }
```
ueman: You can do this inline:

```dart
return {
  'type': type,
  'depth': depth,
  if (identifier != null) 'identifier': identifier,
  // ...
};
```

flutter/lib/src/view_hierarchy/sentry_tree_walker.dart
marandaneto: @ueman That does not work when obfuscated is enabled because everything is renamed at compile time.
Still worth keeping this quick workaround; it improves at least when it's disabled.

dart/lib/src/protocol/sentry_view_hierarchy_element.dart
marandaneto: yep, did all of them, thanks!

marandaneto:
@markushi @brustolin @krystofwoldrich  early feedback is welcomed since it's a working solution already.
Views are really nested, unfortunately.
Unclear what the `extra` params could be, we have to figure out what's useful besides the default ones.
It's not working on Android due to an issue on the bridge most likely, trying to figure out what's happening during de/serialization on Android.
Edit, [issue](https://github.com/getsentry/relay/issues/1704).

flutter/lib/src/view_hierarchy/sentry_tree_walker.dart
```diff
@@ -0,0 +1,103 @@
...
+      // when obfuscation is enabled, this won't work because all the types
+      // are renamed
+      if (sentryElement.type.startsWith(_privateDelimiter) ||
+          (sentryElement.identifier?.startsWith(_privateDelimiter) ?? false)) {
+        privateElement = true;
+      } else {
+        parentSentryElement.children.add(sentryElement);
+      }
+
+      // we don't want to add private children but we still want to walk the tree
```
markushi, line 30: Just out of curiosity: What would be a private children in Flutter?

ueman: Dart has the concept of file private classes (similar to Swift). So a private child is a file private (widget) class, which is used to have better internal structure for widgets classes. See also https://github.com/getsentry/sentry-dart/pull/1189#discussion_r1048223801 for a bit more detail

flutter/lib/src/view_hierarchy/view_hierarchy_event_processor.dart
```diff
@@ -0,0 +1,29 @@
...
+  FutureOr<SentryEvent?> apply(SentryEvent event, {Hint? hint}) async {
+    if (event.exceptions == null && event.throwable == null) {
+      return event;
+    }
+
+    final instance = BindingUtils.getWidgetsBindingInstance();
+    if (instance == null) {
+      return event;
+    }
+    final sentryViewHierarchy = walkWidgetTree(instance);
```
markushi, line 24: Being not too familiar with Flutter, it would be interesting to know:
* how fast walking the tree is
* if there are any multi-threading concerns (e.g. walking the tree whilst it get's modified)

markushi:
Looks good to me, nice one!

Regarding the deep nesting: I'm wondering if it could make sense to aggregate certain "known" widgets.
E.g.
````json
{
  "opacity": 0.5,
  "children": [{
    "visible": true,
    "children": [{
      "identifier": "button_login"
    }]
  }]
}
````
turns into
````json
{
  "identifier": "button_login",
  "opacity": 0.5,
  "visible": true,
}
````

File: flutter/lib/src/view_hierarchy/view_hierarchy_event_processor.dart (hunk 6 above)
Comment (marandaneto) [reply flutter/lib/src/view_hierarchy/view_hierarchy_event_processor.dart:24]: Dart is single-threaded, so no multi-threading concern, its also a sync call.
Running on debug mode (Profile mode isn't supported on iOS simulator, Debug is much slower) takes ~`30ms`, for our sample app/main screen, this is dynamic because it depends on the app's view, so hard to measure.
Final JSON around ~50-60kb

**Comment by marandaneto** [comment]
> Looks good to me, nice one!
> 
> Regarding the deep nesting: I'm wondering if it could make sense to aggregate certain "known" widgets. E.g.
> 
> ```json
> {
>   "opacity": 0.5,
>   "children": [{
>     "visible": true,
>     "children": [{
>       "identifier": "button_login"
>     }]
>   }]
> }
> ```
> 
> turns into
> 
> ```json
> {
>   "identifier": "button_login",
>   "opacity": 0.5,
>   "visible": true,
> }
> ```

I believe this should be an optimization for the FE, where we can fix/improve over time rather than changing on the SDK level and waiting for people to upgrade their SDKs before getting a meaningful view, but I agree whenever possible, the super nested view should be improved when debugging issues, but the SDK should report as it is.

**Comment by marandaneto** [comment]
@krystofwoldrich its a Draft because there's a bug on Relay, but its ready to review, test are failing locally, working on them.

**Comment by markushi** [comment]
> I believe this should be an optimization for the FE

I agree, that's something which could be done on FE! Looping in @narsaynorath

File: flutter/lib/src/view_hierarchy/view_hierarchy_event_processor.dart (hunk 6 above)
Comment (ueman) [reply flutter/lib/src/view_hierarchy/view_hierarchy_event_processor.dart:24]: > if there are any multi-threading concerns (e.g. walking the tree whilst it get's modified)

I actually think that's a valid concern because that's also a common problem when creating screenshots. Sometime you'll get an error which says it's not possible to create a screenshot because the current on screen content was already invalidated. But I also think it's something not a lot of people will run into, since it should be fast enough.

File: dart/lib/src/sentry_attachment/sentry_attachment.dart [hunk 7]
```diff
@@ -88,6 +93,17 @@ class SentryAttachment {
             contentType: 'image/png',
             attachmentType: SentryAttachment.typeAttachmentDefault);
 
+  SentryAttachment.fromViewHierrchy(SentryViewHierarchy sentryViewHierarchy)
```
krystofwoldrich:
```suggestion
  SentryAttachment.fromViewHierarchy(SentryViewHierarchy sentryViewHierarchy)
```

Just a typo.

flutter/lib/src/view_hierarchy/view_hierarchy_event_processor.dart
```diff
@@ -0,0 +1,29 @@
...
+
+    if (sentryViewHierarchy == null) {
+      return event;
+    }
+
+    final viewHierarchy =
+        SentryAttachment.fromViewHierrchy(sentryViewHierarchy);
```
krystofwoldrich:
```suggestion
        SentryAttachment.fromViewHierarchy(sentryViewHierarchy);
```

Just a typo.

krystofwoldrich:
The CI is failing but otherwise looks good to me.

flutter/lib/src/view_hierarchy/view_hierarchy_event_processor.dart
marandaneto: The feature is best effort, if for some reason you can't read the tree or take the screenshot, that's fine, there will be always a state that the framework won't allow you to do it, and we don't have control over that.

dart/lib/src/sentry_attachment/sentry_attachment.dart
marandaneto: Good catch

marandaneto:
> The CI is failing but otherwise looks good to me.

CI should fail, that's expected (just the analyze) step.

marandaneto:
Docs: https://github.com/getsentry/sentry-docs/pull/5986

narsaynorath:
Is the event from your sample screenshot available in prod? I'd be interested in seeing how deep it actually ends up being. Seeing it would also help me understand the aggregated optimization mentioned above.

When this example was provided:
> ```json
> {
>   "opacity": 0.5,
>   "children": [{
>     "visible": true,
>     "children": [{
>       "identifier": "button_login"
>     }]
>   }]
> }
> ```
is this form the result of ignoring the private views? And the idea to aggregate the children is because the private views will show up in the data as nodes, but without anything meaningful since we're ignoring them? Is there a chance that a private view contains two (or more) children or is it always a 1:1 relationship in those cases?

marandaneto:
> Is the event from your sample screenshot available in prod? I'd be interested in seeing how deep it actually ends up being. Seeing it would also help me understand the aggregated optimization mentioned above.
> 
> When this example was provided:
> 
> > ```json
> > {
> >   "opacity": 0.5,
> >   "children": [{
> >     "visible": true,
> >     "children": [{
> >       "identifier": "button_login"
> >     }]
> >   }]
> > }
> > ```
> 
> is this form the result of ignoring the private views? And the idea to aggregate the children is because the private views will show up in the data as nodes, but without anything meaningful since we're ignoring them? Is there a chance that a private view contains two (or more) children or is it always a 1:1 relationship in those cases?

Shared the JSON in the channel a few weeks ago.
Yes, private views are ignored, but it only works if minification isn't enabled.
I can't tell if it's always a 1:1 relationship, views are super nested, so I'd assume it's not.

marandaneto:
@krystofwoldrich @brustolin this is ready for final review.

dart/lib/src/sentry_attachment/sentry_attachment.dart
```diff
@@ -88,6 +93,17 @@ class SentryAttachment {
...
+  SentryAttachment.fromViewHierarchy(SentryViewHierarchy sentryViewHierarchy)
+      : this.fromLoader(
+          loader: () => Uint8List.fromList(
+              utf8JsonEncoder.convert(sentryViewHierarchy.toJson())),
+          filename: 'view-hierarchy.json',
+          contentType: 'application/json',
+        );
+  // TODO: rollback this change, relay is not accepting this type
+  // if sentt with a single item in the envelope
+  // attachmentType: SentryAttachment.typeViewHierarchy);
```
krystofwoldrich: Is this comment still up to date? Will the constructor be removed in the future?

marandaneto: Good catch, forgot to remove it

flutter/test/view_hierarchy/sentry_tree_walker_test.dart
```diff
@@ -0,0 +1,215 @@
...
+            false,
+            _findWidget(first, (element) {
+              return element.identifier == '_btn_3';
+            }));
+      });
+    });
+  });
+}
+
+SentryViewHierarchyElement _getFirstsentryViewHierarchy(WidgetsBinding instance) {
```
krystofwoldrich: L: _getFirstsentryViewHierarchy -> _getFirstSentryViewHierarchy
I'm not sure if the 's' is lower case on purpose or not.

krystofwoldrich (approved):
LGTM 🎉🚀🚀


I know I asked about the failing analyzer before, but why does it fail/why is it expected?

marandaneto:
> LGTM 🎉🚀🚀
> 
> I know I asked about the failing analyzer before, but why does it fail/why is it expected?

You can ignore this, it will be automatically solved once it's published.

flutter/test/view_hierarchy/sentry_tree_walker_test.dart
marandaneto: typo, fixed, thanks.

flutter/lib/src/view_hierarchy/sentry_tree_walker.dart
```diff
@@ -0,0 +1,104 @@
...
+
+    double? width;
+    double? height;
+    double? x;
+    double? y;
+    bool? visible;
+    double? alpha;
+
+    // Widget has to be RenderBox to have a size
+    if (widget is RenderBox) {
```
markushi, line 56: Hey @marandaneto, is it possible that this should be 
`if (element.renderObject is RenderBox) {` 
instead? 
@narsaynorath is seeing elements in the View Hierarchy which have a `x/y` coordinate but no `width/height`.

marandaneto: You're right, https://github.com/getsentry/sentry-dart/pull/1258

PR #1202, opened 2022-12-21
Update dependency `intl` to ^0.18.0

zeshuaro:
## :scroll: Description

Update dependency `intl` to ^0.18.0 - https://pub.dev/packages/intl.

My project uses both `sentry` and `intl`. Right now, I'm unable to upgrade `intl` to `^0.18.0` in my project because `sentry` depends on `intl: ^0.17.0` and the version resolution would fail.

Hoping GitHub Actions will verify that the new version of `intl` is still compatible.

## :pencil: Checklist

- [ ] All tests passing
- [ ] No breaking changes

### DIFF
--- a/dart/pubspec.yaml
+++ b/dart/pubspec.yaml
@@ -15,7 +15,7 @@ dependencies:
   meta: ^1.3.0
   stack_trace: ^1.10.0
   uuid: ^3.0.0
-  intl: ^0.17.0
+  intl: ^0.18.0
 
 dev_dependencies:
   mockito: ^5.1.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 - Bump Cocoa SDK from v7.31.3 to v7.31.4 ([#1190](https://github.com/getsentry/sentry-dart/pull/1190))
   - [changelog](https://github.com/getsentry/sentry-cocoa/blob/8.0.0/CHANGELOG.md#7314)
   - [diff](https://github.com/getsentry/sentry-cocoa/compare/7.31.3...7.31.4)
+- Bump dependency `intl` to ^0.18.0 ([#1202](https://github.com/getsentry/sentry-dart/pull/1202))
 
 ## 6.18.1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Dependencies
+
+- Bump dependency `intl` to ^0.18.0 ([#1202](https://github.com/getsentry/sentry-dart/pull/1202))
+
 ## 6.18.2
 
 ### Fixes
@@ -15,7 +21,6 @@
 - Bump Cocoa SDK from v7.31.3 to v7.31.5 ([#1190](https://github.com/getsentry/sentry-dart/pull/1190), [#1207](https://github.com/getsentry/sentry-dart/pull/1207))
   - [changelog](https://github.com/getsentry/sentry-cocoa/blob/8.0.0/CHANGELOG.md#7315)
   - [diff](https://github.com/getsentry/sentry-cocoa/compare/7.31.3...7.31.5)
-- Bump dependency `intl` to ^0.18.0 ([#1202](https://github.com/getsentry/sentry-dart/pull/1202))
 
 ## 6.18.1
 


marandaneto:
@zeshuaro why is it not possible? Sentry uses the `^` operator, so it should be fine to use the latest version as well.
Can you link us the error and why?

zeshuaro:
This is the error I'm getting, you can also see the build [here](https://github.com/zeshuaro/appainter/actions/runs/3748827578/jobs/6366607206).

```
Because no versions of sentry_flutter match >6.18.1 <7.0.0 and sentry_flutter >=6.18.0 <6.18.1 depends on sentry 6.18.0, sentry_flutter >=6.18.0 <6.18.1-∞ or >6.18.1 <7.0.0 requires sentry 6.18.0.
And because sentry >=6.11.0 depends on intl ^0.17.0, sentry_flutter >=6.18.0 <6.18.1-∞ or >6.18.1 <7.0.0 requires intl ^0.17.0.
And because sentry_flutter 6.18.1 depends on sentry 6.18.1 which depends on intl ^0.17.0, sentry_flutter ^6.18.0 requires intl ^0.17.0.
So, because appainter depends on both intl ^0.18.0 and sentry_flutter ^6.18.0, version solving failed.
```

My guess is that Pub classifies minor bumps in versions `0.x.x` as breaking change, so it's not able to resolve the latest version of `intl`.

dart/pubspec.yaml
```diff
@@ -15,7 +15,7 @@ dependencies:
   meta: ^1.3.0
   stack_trace: ^1.10.0
   uuid: ^3.0.0
-  intl: ^0.17.0
+  intl: ^0.18.0
```
ueman, line 18: There doesn't seem to be a breaking change: https://pub.dev/packages/intl/changelog
```suggestion
  intl: '>=0.17.0 <0.19.0'
```

zeshuaro: I'm not suggesting the package has breaking changes, I was guessing that Pub treats minor bumps in versions `0.x.x` as breaking, which does seem to be the case from [here](https://dart.dev/tools/pub/versioning#semantic-versions):

> Although semantic versioning doesn’t promise any compatibility between versions prior to 1.0.0, the Dart community convention is to treat those versions semantically as well. The interpretation of each number is just shifted down one slot: going from 0.1.2 to 0.2.0 indicates a breaking change, going to 0.1.3 indicates a new feature, and going to 0.1.2+1 indicates a change that doesn’t affect the public API.

marandaneto: That's strange because `intl 0.17.0 (0.18.0 available)`
it's a minor version but requires `dart pub upgrade --major-versions`.
Likely a bug in `pub`?
The suggestion won't work either because sentry won't be compatible with `0.19.0` once is released.

marandaneto: Likely related to https://github.com/flutter/flutter/issues/117163

marandaneto: Tag 0.18.0 is not created yet in the intl repo, so I can't check it for now either.
https://github.com/dart-lang/intl/tags

marandaneto: @zeshuaro is this still necessary? I guess this was fixed in the tooling.

zeshuaro: Sorry, what was fixed in the tooling? I'm still getting the following error:

```
Because sentry_flutter >=6.18.2 depends on sentry 6.18.2 which depends on intl ^0.17.0, sentry_flutter >=6.18.2 requires intl ^0.17.0.
So, because appainter depends on both intl 0.18.0 and sentry_flutter 6.18.2, version solving failed.
pub get failed (1; So, because appainter depends on both intl 0.18.0 and sentry_flutter 6.18.2, version solving failed.)
```

marandaneto: @zeshuaro I can't repro this since it works, its a conflict with the `appainter` lib, I don't find it on pub.dev, can you give us a minimal reproducible example?

zeshuaro: `appainter` is my project, so the conflict is not with it. See below for a minimal reproducible example. It would give the same error:

```
Because sentry_flutter >=6.18.2 <7.0.0-alpha.1 depends on sentry 6.18.2 which depends on intl ^0.17.0, sentry_flutter >=6.18.2 <7.0.0-alpha.1 requires intl ^0.17.0.
So, because hello_world depends on both intl ^0.18.0 and sentry_flutter ^6.18.2, version solving failed.
pub get failed (1; So, because hello_world depends on both intl ^0.18.0 and sentry_flutter ^6.18.2, version solving failed.)
```

<details>
  <summary>pubspec.yaml</summary>
  
  ```yml
  name: hello_world
  description: A new Flutter project.
  publish_to: "none"
  
  version: 1.0.0+1
  
  environment:
    sdk: ">=2.18.4 <3.0.0"
  
  dependencies:
    flutter:
      sdk: flutter
    intl: ^0.18.0
    sentry_flutter: ^6.18.2
  
  dev_dependencies:
    flutter_test:
      sdk: flutter
  
  flutter:
    uses-material-design: true
  ```
</details>

marandaneto: I could reproduce it now, plenty of packages are broken because of that.
See another issue https://github.com/dart-lang/intl/issues/522
I didn't see any breaking change so lets merge this.

marandaneto:
@zeshuaro can you add an entry to the changelog? Thanks!

zeshuaro:
I've added an entry to the changelog, but I think there might still be issues with `flutter_localization` from the issues you linked.

dart/pubspec.yaml
marandaneto: > I've added an entry to the changelog, but I think there might still be issues with `flutter_localization` from the issues you linked.

Indeed, https://github.com/flutter/flutter/blob/07c47dcc3d55be89cb3dbb3d10d5d65ed455c4bd/packages/flutter_localizations/pubspec.yaml#L11
They pinned the version, so if I bump, its incompatible as well.
Not sure what's the best approach here now.
@ueman ideas?
Instead of using `^0.18.0`, we could do `any`, but that could lead to new problems in the future as well due to breaking changes, but it solves the problem right now.

zeshuaro: I don't think setting `any` would be the solution here, would definitely cause more issues later on. Maybe the solution is to simply wait for Flutter to fix things on their sides.

ueman: As said, I would suggest
```suggestion
  intl: '>=0.17.0 <0.19.0'
```
It's better than the current situation and even `^0.18.0` wouldn't be compatible with `0.19.0` as demonstrated by this PR. So, I think the suggestion is the way to go. It's a bit better than any, but, granted, only a bit.

In the future, intl should probably be vendored or just removed to avoid this problem. There's just one usage anyway, I believe.

Does disallowing 0.17.0 break usages on older Flutter versions?

zeshuaro: > Does disallowing 0.17.0 break usages on older Flutter versions?

I believe so based on the fact that we're required to make this current change.

marandaneto: Yeah IMO the bug is not here, let's see if they act quickly on https://github.com/dart-lang/pub/issues/3747 otherwise we push @ueman suggested fix for now.
Vendoring the class would work too but I'd like to avoid it for now.

marandaneto:
Closed in favor of https://github.com/getsentry/sentry-dart/pull/1247
Thank you all!

Issue #1173, opened 2022-12-07
Enable APM features by default


marandaneto:
### Description

`enableUserInteractionTracing`
`captureFailedRequests`
`networkTracing`
`enableStructureDataTracing`
Probably a few more

PR #1217, opened 2023-01-10
Enable APM features by default

marandaneto:

Closes https://github.com/getsentry/sentry-dart/issues/1173

## :pencil: Checklist

- [X] All tests passing
- [ ] No breaking changes

### DIFF
--- a/flutter/README.md
+++ b/flutter/README.md
@@ -102,7 +102,7 @@ runApp(
 This adds performance tracing for all `AssetBundle` usages, where the `AssetBundle` is accessed with `DefaultAssetBunlde.of(context)`.
 This includes all of Flutters internal access of `AssetBundle`s, like `Image.asset` for example.
 Tracing for `AssetBundle.loadStructuredData()` is currently disabled.
-It's hidden by the `enableStructureDataTracing` flag and considered experimental. Using it could lead to bugs. We recognize the irony.
+It's hidden by the `enableStructuredDataTracing` flag and considered experimental. Using it could lead to bugs. We recognize the irony.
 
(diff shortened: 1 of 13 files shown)


flutter/README.md
```diff
@@ -102,7 +102,7 @@ runApp(
 This adds performance tracing for all `AssetBundle` usages, where the `AssetBundle` is accessed with `DefaultAssetBunlde.of(context)`.
 This includes all of Flutters internal access of `AssetBundle`s, like `Image.asset` for example.
 Tracing for `AssetBundle.loadStructuredData()` is currently disabled.
-It's hidden by the `enableStructureDataTracing` flag and considered experimental. Using it could lead to bugs. We recognize the irony.
+It's hidden by the `enableStructuredDataTracing` flag and considered experimental. Using it could lead to bugs. We recognize the irony.
```
ueman, line 105: Is it still experimental even though it's enabled by default?

marandaneto: I removed the experimental notes, and never got any complaints.

ueman: I guess the flag can then be removed altogether?

marandaneto: We can remove it on v8 maybe, it's always nice having an opt-out for whatever reason, at least for a bit, maybe enabled by default uncover bugs, so people can opt-out quickly.

brustolin (approved):
Looks like CHANGELOG is wrong, but other than that LGTM.

marandaneto:
> Looks like CHANGELOG is wrong, but other than that LGTM.

Good catch, fixed.

PR #1259, opened 2023-02-01
Try to avoid ConcurrentModificationError by not using a Future.forEach

### LINKED ISSUE #1256: ConcurrentModificationError when doing performance tracing
_Platform:_
- [ ]  Dart
- [x]  Flutter Android or iOS
- [ ]  Flutter Web

_IDE:_
- [x]  VSCode
- [ ]  IntelliJ/AS
- [ ]  XCode
- [ ]  Other, which one?

- [ ]  Enabled
- [x]  Disabled

_Platform installed with:_
- [x] pub.dev
- [ ] GitHub

**6.18.1** (but it's also happening on 6.19.0)

---
I have the following issue:

Sometimes when tracing performance, I see a `ConcurrentModificationError`. It is thrown by the following code:

https://github.com/getsentry/sentry-dart/blob/8fa393466a5d478f281e46cbc11e66c2b3db12d8/dart/lib/src/sentry_tracer.dart#L90-L94

Unfortunately, I failed to create a minimal reproducible example. Therefore, I'm attaching some screenshots. The first one is the issue on Sentry.io itself, but unfortunately it's not helpful at all.

Screenshot 2023-01-31 at 12 34 30

The second one come from debugging the issue, to show that it's indeed coming from Sentry. Unfortunately, the stacktrace of that error is the same as the one above.
Screenshot 2023-01-31 at 12 34 56

I'm assuming that it's caused by unfinished spans somehow. Roughly, the code in the (closed source) app is the following:

```
methodOne() {
 transaction{
   awaited spans (3 times)
   one unawaited span
   awaited spans (4 times)
 }
}

during the same time as methodOne is a web request running, which also attaches a span to the above mentioned transaction

After the above transaction is finished, another one is started. I don't know whether that's related or not.
``` 

I would expect Sentry to not throw a ConcurrentModificationError.

marandaneto:
## :scroll: Description
Try to avoid ConcurrentModificationError by not using a `Future.forEach`.
I believe that the event loop has a sort of priority queue and maybe the `future` of finishing a late span is actually executed within the items of `forEach` (more priority in the scheduling) hence causing the `ConcurrentModificationError`, this is a hunch, I can't reproduce it either, let's go with a simple for loop for now.

Closes https://github.com/getsentry/sentry-dart/issues/1256

## :pencil: Checklist

- [X] All tests passing
- [X] No breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Fixes
+
+- Try to avoid ConcurrentModificationError by not using a Future.forEach ([#1247](https://github.com/getsentry/sentry-dart/pull/1247))
+
 ### Dependencies
 
 - Bump Android SDK from v6.12.1 to v6.13.0 ([#1250](https://github.com/getsentry/sentry-dart/pull/1250))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Fixes
 
-- Try to avoid ConcurrentModificationError by not using a Future.forEach ([#1247](https://github.com/getsentry/sentry-dart/pull/1247))
+- Try to avoid ConcurrentModificationError by not using a Future.forEach ([#1259](https://github.com/getsentry/sentry-dart/pull/1259))
 
 ### Dependencies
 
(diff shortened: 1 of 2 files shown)


marandaneto:
@ueman I added on the description what I think, but I can't reproduce or confirm it.

ueman:
> @ueman I added on the description what I think, but I can't reproduce or confirm it.

Thanks! The explanation seems plausible to me. Unfortunately, I also don't have any further insights. For me, it seems like there can happen some weird kind of race condition. The fix in this PR is something which solved it for, but I can't really explain why :/

It's really nagging me, that I can't figure this bug out.

Even if this doesn't fix the issue, this will improve the stacktrace of the thrown error and thus helpfully get us closer to figuring it out for real.

ueman (approved):

krystofwoldrich (approved):
LGTM 🚀

PR #1260, opened 2023-02-01
Add In Foreground to App context

marandaneto:
## :scroll: Description
Add In Foreground to App context

Relates to https://github.com/getsentry/team-mobile/issues/36

## :pencil: Checklist

- [X] All tests passing
- [X] No breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Features
 
-- Add In Foreground to App context ([#1258](https://github.com/getsentry/sentry-dart/pull/1258))
+- Add In Foreground to App context ([#1260](https://github.com/getsentry/sentry-dart/pull/1260))
 
 ### Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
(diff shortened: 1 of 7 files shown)


marandaneto:
I'd like to add the `event.thread.main=true|false` as well but since `Isolate.current` always returns the main Isolate during unhandled callbacks, we can't really tell which Isolate threw the error, if it was the main or not.

Ideally, unhandled callbacks such as [onError](https://master-api.flutter.dev/flutter/dart-ui/ErrorCallback.html) would return the faulty Isolate in the `exception` or even an extra parameter.

So its better to not report at all instead of reporting everything as the `main` Isolate.
@ueman ideas around this?

ueman:
With Flutter 3.7 you can actually initialize Sentry Flutter on a background isolate, where `Isolate.current != 'main'`. When the exceptions are passed from one isolate to another, I believe there's unfortunately no automatic way to detect it. For a lot of cases that's not a problem, though, I think. A lot of background computing is done with the `compute()` method from Flutter where the exception just gets rethrown. While the exception is technically from a background isolate, it still affects the main isolate.

brustolin (approved):
LGTM

PR #1265, opened 2023-02-06
View hierarchy sets to local instead of global position

marandaneto:
## :scroll: Description
View hierarchy sets to local instead of global position

The position was global but it should be local considering the parent.

Debugging and checking the view hierarchy preview

## :pencil: Checklist

- [X] All tests passing
- [ ] No breaking changes

### DIFF
--- a/flutter/lib/src/view_hierarchy/sentry_tree_walker.dart
+++ b/flutter/lib/src/view_hierarchy/sentry_tree_walker.dart
@@ -54,13 +54,9 @@ class _TreeWalker {
 
     final renderObject = element.renderObject;
     if (renderObject is RenderBox) {
-      final offset = renderObject.localToGlobal(Offset.zero);
-      if (offset.dx > 0) {
-        x = offset.dx;
-      }
-      if (offset.dy > 0) {
-        y = offset.dy;
-      }
+      final offset = renderObject.globalToLocal(Offset.zero);
+      x = offset.dx;
+      y = offset.dy;
       // no z axes in 2d
 
       final size = element.size;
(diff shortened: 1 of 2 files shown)


flutter/lib/src/view_hierarchy/sentry_tree_walker.dart
```diff
@@ -54,13 +54,9 @@ class _TreeWalker {
...
-      final offset = renderObject.localToGlobal(Offset.zero);
-      if (offset.dx > 0) {
-        x = offset.dx;
-      }
-      if (offset.dy > 0) {
-        y = offset.dy;
-      }
+      final offset = renderObject.globalToLocal(Offset.zero);
+      x = offset.dx;
+      y = offset.dy;
```
marandaneto, line 59: @narsaynorath I've noticed negative values for some reason, is this normal?
I can filter out negative and zero values if needed, let me know.

flutter/lib/src/view_hierarchy/sentry_tree_walker.dart
```diff
@@ -54,16 +54,17 @@ class _TreeWalker {
...
     final renderObject = element.renderObject;
     if (renderObject is RenderBox) {
-      final offset = renderObject.localToGlobal(Offset.zero);
-      if (offset.dx > 0) {
-        x = offset.dx;
-      }
-      if (offset.dy > 0) {
-        y = offset.dy;
-      }
+      final offset = renderObject.globalToLocal(Offset.zero);
```
marandaneto, line 57: @ueman ideas why `globalToLocal` return negative values as well?
Found https://stackoverflow.com/questions/36800042/globaltolocal-not-behaving-as-i-expected-pretty-sure-im-just-not-getting-it but the Flutter docs don't state that.
Wondering if I should use something else instead of `Offset.zero`.
We need to get the `x` and `y` relative to the parent.

PR #1306, opened 2023-03-02
Add support for sqflite

### LINKED ISSUE #1201: APM/crumbs support for `sqflite`
### Description

https://pub.dev/packages/sqflite

marandaneto:
## :scroll: Description
Add support for sqflite

Closes https://github.com/getsentry/sentry-dart/issues/1201

## :pencil: Checklist

- [X] All tests passing
- [X] No breaking changes


sqflite/lib/src/sentry_sqflite.dart
```diff
@@ -0,0 +1,33 @@
...
+  int? version,
+  OnDatabaseConfigureFn? onConfigure,
+  OnDatabaseCreateFn? onCreate,
+  OnDatabaseVersionChangeFn? onUpgrade,
+  OnDatabaseVersionChangeFn? onDowngrade,
+  OnDatabaseOpenFn? onOpen,
+  bool readOnly = false,
+  bool singleInstance = true,
+}) {
+  Future<Database> openDatabase() async {
```
marandaneto, line 28: Users would need to use our `openDatabase` in order to track APM for sqflite.
Or they can wrap directly with `SentryDatabase(Database)`

sqflite/lib/src/sentry_database.dart
```diff
@@ -0,0 +1,190 @@
...
+  final Hub _hub;
+
+  // ignore: public_member_api_docs
+  SentryDatabase(
+    this._database, {
+    @internal Hub? hub,
+  }) : _hub = hub ?? HubAdapter();
+
+  @override
+  Batch batch() {
```
marandaneto: We'll likely need a wrapper for `Batch` as well unless this is not really used.

marandaneto:
@ueman @kuhnroyal what do you think about this wrapper approach in case you'd be using `sqflite`?

ueman:
I'm not using sqflite but the approach looks reasonable to me (The app I'm working on sadly uses Hive for our offline storage needs)

dart/lib/src/protocol/sdk_version.dart
```diff
@@ -88,6 +88,12 @@ class SdkVersion {
 
   /// Adds a package
   void addPackage(String name, String version) {
+    for (final item in _packages) {
```
romtsn, line 91: irrelevant to the sqflite, but is there some Set analogy in dart which you could use here maybe?

sqflite/lib/src/sentry_sqflite.dart
marandaneto: Added to code comments/docs

krystofwoldrich:
I looked at `sqflite/src` and it makes sense to me, but I don't have a comparison with the implementations on the other platforms.

sqflite/lib/src/sentry_database.dart
```diff
@@ -0,0 +1,128 @@
...
+
+  @override
+  Future<T> transaction<T>(
+    Future<T> Function(Transaction txn) action, {
+    bool? exclusive,
+  }) {
+    Future<T> future() async {
+      final currentSpan = _hub.getSpan();
+      final span = currentSpan?.startChild(
+        'db',
```
brustolin: m: Cocoa uses db.sql.transaction  for operation and for description a summary of what happened, like: 
`INSERTED 1 'TestEntity', UPDATED 1 'TestEntity', DELETED 1 'TestEntity'`, it would be great if possible to do the same.

sqflite/lib/src/sentry_database.dart
```diff
@@ -0,0 +1,128 @@
...
+    options.sdk.addPackage(packageName, sdkVersion);
+  }
+
+  @override
+  Future<void> close() {
+    Future<void> future() async {
+      final currentSpan = _hub.getSpan();
+      final span = currentSpan?.startChild(
+        'db',
+        description: 'CLOSE',
```
brustolin: l: I think is a good approach to use constants, easier to update later if necessary.

brustolin (commented):
First look, everything seems ok.
Just a couple of comments.

It would be nice if we could clean the query if the user do the wrong thing of building the query without using placeholders. But as long we have this documented that they should use the argument array, is all fine.

sqflite/lib/src/sentry_database.dart
marandaneto: `db.sql.transaction` makes sense, but it's not in the [spec](https://develop.sentry.dev/sdk/performance/span-operations/)
will add it.

marandaneto: We don't have an API to collect the summary of all statements within a transaction, we do that for batch operations, but it'd be a lot of hacky code/internal APIs to get this working since each span will have its own description with its SQL, I guess its fine, but it'd be nice to have, indeed.

marandaneto: https://github.com/getsentry/develop/pull/866

sqflite/lib/src/sentry_database.dart
```diff
@@ -0,0 +1,141 @@
...
+    options.sdk.addPackage(packageName, sdkVersion);
+  }
+
+  @override
+  Future<void> close() {
+    Future<void> future() async {
+      final currentSpan = _hub.getSpan();
+      final span = currentSpan?.startChild(
+        dbOp,
+        description: 'CLOSE',
```
romtsn: `l`: would it make sense to add database name or path here as well (if we can access it)?

sqflite/lib/src/sentry_database.dart
```diff
@@ -0,0 +1,141 @@
...
+  @override
+  Future<T> transaction<T>(
+    Future<T> Function(Transaction txn) action, {
+    bool? exclusive,
+  }) {
+    Future<T> future() async {
+      final currentSpan = _hub.getSpan();
+      final span = currentSpan?.startChild(
+        _dbSqlOp,
+        description: 'TRANSACTION',
```
romtsn: `l`: same here re. database name or path in the description

sqflite/lib/src/sentry_sqflite_database_factory.dart
```diff
@@ -0,0 +1,74 @@
...
+    // ignore: invalid_use_of_internal_member
+    if (!_hub.options.isTracingEnabled()) {
+      return super.openDatabase(path, options: options);
+    }
+
+    Future<Database> openDatabase() async {
+      final currentSpan = _hub.getSpan();
+      final span = currentSpan?.startChild(
+        SentryDatabase.dbOp,
+        description: 'OPEN',
```
romtsn: `l`: same re. including database name or path in the description

sqflite/lib/src/sentry_sqflite.dart
```diff
@@ -0,0 +1,70 @@
...
+      onDowngrade: onDowngrade,
+      onOpen: onOpen,
+      readOnly: readOnly,
+      singleInstance: singleInstance,
+    );
+
+    // ignore: invalid_use_of_internal_member
+    final options = (hub ?? HubAdapter()).options;
+
+    final database =
```
romtsn: `m`: I guess you could also wrap this into a `OPEN` span, right?

sqflite/lib/src/sentry_batch.dart
```diff
@@ -0,0 +1,246 @@
...
+    @internal Hub? hub,
+  }) : _hub = hub ?? HubAdapter();
+
+  @override
+  Future<List<Object?>> apply({bool? noResult, bool? continueOnError}) {
+    Future<List<Object?>> future() async {
+      final currentSpan = _hub.getSpan();
+
+      final span = currentSpan?.startChild(
+        SentryDatabase.dbSqlExecuteOp,
```
romtsn: `m`: can you execute SELECT queries in a batch? if so, I guess we should try to distinguish here between `db.sql.execute` and `db.sql.query` if possible? I think `db.sql.query` have special normalization on backend for N+1 detectors, so probably worth to look into that.

sqflite/lib/src/sentry_batch.dart
```diff
@@ -0,0 +1,246 @@
...
+  Future<List<Object?>> commit({
+    bool? exclusive,
+    bool? noResult,
+    bool? continueOnError,
+  }) {
+    Future<List<Object?>> future() async {
+      final currentSpan = _hub.getSpan();
+
+      final span = currentSpan?.startChild(
+        SentryDatabase.dbSqlExecuteOp,
```
romtsn: `m`: same here wrt `db.sql.query`

sqflite/lib/src/sentry_database_executor.dart
```diff
@@ -0,0 +1,456 @@
...
+
+    return future();
+  }
+
+  @override
+  Future<void> execute(String sql, [List<Object?>? arguments]) {
+    Future<void> future() async {
+      final currentSpan = _parentSpan ?? _hub.getSpan();
+      final span = currentSpan?.startChild(
+        SentryDatabase.dbSqlExecuteOp,
```
romtsn, line 68: `m`: same here wrt to `db.sql.query`

romtsn (approved):
a few minor comments, but LGTM otherwise 🚀 

We are only using `db.sql.query` on Android and not `db.sql.execute` for insert/update/delete, not sure if there's any problem with that, but I think it's fine.

dart/lib/src/protocol/sdk_version.dart
marandaneto: We [do have](https://api.dart.dev/stable/2.19.4/dart-core/Set-class.html) but that would be breaking, I'll consider it in the next refactoring.

sqflite/lib/src/sentry_database.dart
marandaneto: Good idea, we can.

sqflite/lib/src/sentry_batch.dart
marandaneto: There could be any CRUD + Select, that's why its generic (execute), maybe just `db.sql` then?
Even if I track when there's select or not, what if we have both?

sqflite/lib/src/sentry_sqflite.dart
marandaneto: yep, good idea.

sqflite/lib/src/sentry_batch.dart
marandaneto: Changing to `db` for now because of https://github.com/getsentry/sentry/blob/23e48d19c704d620a1345531512a61543ced0537/src/sentry/spans/grouping/strategy/base.py#L107
I'll ask if we can add the new ops to that file.

marandaneto: https://github.com/getsentry/sentry/pull/45630

marandaneto:
@krystofwoldrich @brustolin ready for final review

.github/workflows/file.yml
```diff
@@ -9,6 +9,8 @@ on:
       - 'logging/**'
       - 'flutter/**'
       - 'dio/**'
+      - 'file/**'
```
krystofwoldrich: This is a workflow for the `file` package, I guess here `file/**` should not be ignored?

flutter/example/pubspec.yaml
```diff
@@ -15,13 +15,17 @@ dependencies:
...
   universal_platform: ^1.0.0
   feedback: ^2.0.0
   provider: ^6.0.0
   dio: any # This gets constrained by `sentry_dio`
+  sqflite: any # This gets constrained by `sentry_sqflite`
   logging: any # This gets constrained by `sentry_logging`
   package_info_plus: ^3.0.0
   path_provider: ^2.0.0
+  #sqflite_common_ffi: ^2.0.0
+  #sqflite_common_ffi_web: ^0.3.0
```
krystofwoldrich, line 28: L: When are these needed? Just trying to understand if there should be a comment explaining them or if users of the lib will know.

.github/workflows/file.yml
marandaneto: good catch

flutter/example/pubspec.yaml
marandaneto: This is just the example, so I can easily swap the deps and add sqflite with FFI.

flutter/example/lib/main.dart
```diff
@@ -8,6 +8,11 @@ import 'package:flutter/material.dart';
 import 'package:flutter/services.dart';
 import 'package:logging/logging.dart';
 import 'package:sentry_flutter/sentry_flutter.dart';
+import 'package:sentry_sqflite/sentry_sqflite.dart';
+// import 'package:sentry_sqflite/sentry_sqflite.dart';
```
krystofwoldrich: L: The lines are the same, I guess the comment can be removed?

flutter/example/lib/main.dart
```diff
@@ -394,6 +403,52 @@ class MainScaffold extends StatelessWidget {
...
+    final tr = Sentry.startTransaction(
+      'sqfliteTest',
+      'db',
+      bindToScope: true,
+    );
+
+    // databaseFactory = databaseFactoryFfiWeb; // or databaseFactoryFfi // or SentrySqfliteDatabaseFactory()
+
+    // final sqfDb = await openDatabase(inMemoryDatabasePath);
+    final db = await openDatabaseWithSentry(inMemoryDatabasePath);
```
krystofwoldrich, line 416: L: Where is the `inMemoryDatabasePath` from? Just curious because I don't see it defined above.

krystofwoldrich (approved):
`sentry-flutter / analyze` failed on an unrelated file?

 `sentry-sqflite / analyze / package-analysis` will work after publishing the package, right?
 
 I have some small comments/questions but besides that, all looks good. 🚀

flutter/example/lib/main.dart
marandaneto: part of `package:sqflite/sqflite.dart`

marandaneto:
> `sentry-flutter / analyze` failed on an unrelated file?
> 
> `sentry-sqflite / analyze / package-analysis` will work after publishing the package, right?
> 
> I have some small comments/questions but besides that, all looks good. 🚀

Both correct, fixed.

PR #1312, opened 2023-03-03
Include and set ProGuard uuid in Flutter Options

Wcisang:
## :scroll: Description

I just added a proGuardUuid property to the Options and passed it to Sentry's android native initialization.

Due to the flutter SDK not having support to configure the proguard uuid, it is necessary to perform a large workaround to perform this action.
So this simple modification already makes it possible for Flutter apps to configure the proguard uuid at runtime.

created unit tests for testing defaults and custom values

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/dart/lib/src/sentry_options.dart
+++ b/dart/lib/src/sentry_options.dart
@@ -279,9 +279,6 @@ class SentryOptions {
   /// If enabled, [scopeObservers] will be called when mutating scope.
   bool enableScopeSync = true;
 
-  /// Sets the ProGuard uuid for Android platform.
-  String? proGuardUuid;
-
   final List<ScopeObserver> _scopeObservers = [];
 
   List<ScopeObserver> get scopeObservers => _scopeObservers;
--- a/dart/lib/src/sentry_options.dart
+++ b/dart/lib/src/sentry_options.dart
@@ -279,6 +279,9 @@ class SentryOptions {
   /// If enabled, [scopeObservers] will be called when mutating scope.
   bool enableScopeSync = true;
 
+  /// Sets the ProGuard uuid for Android platform.
+  String? proGuardUuid;
+
   final List<ScopeObserver> _scopeObservers = [];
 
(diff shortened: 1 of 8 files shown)


dart/lib/src/sentry_options.dart
```diff
@@ -279,6 +279,9 @@ class SentryOptions {
   /// If enabled, [scopeObservers] will be called when mutating scope.
   bool enableScopeSync = true;
 
+  /// Sets the ProGuard uuid for Android platform.
+  String? proGuardUuid;
```
marandaneto: Add to `SentryFlutterOptions` instead.

```suggestion
  /// Sets the ProGuard uuid for Android platform.
  String? proguardUuid;
```

Let's align on the naming.

marandaneto:
@Wcisang you could technically set it directly into the [Manifest file on Android](https://docs.sentry.io/product/cli/dif/#proguard-mapping-upload).
Does that help?

Wcisang:
> 

@marandaneto 
Because we use a protection tool that runs after the build in the pipeline and the obfuscation map changes completely every run. And with that, we need to set this configuration at runtime.

dart/lib/src/sentry_options.dart
Wcisang: done! thanks!

Wcisang: done! thanks!

marandaneto (approved):
Thanks @Wcisang 
Before merging, please add an entry to our changelog :)

Wcisang:
> Thanks @Wcisang Before merging, please add an entry to our changelog :)

Done!
When does the next release happen?

marandaneto:
> > Thanks @Wcisang Before merging, please add an entry to our changelog :)
> 
> Done! When does the next release happen?

I can cut a release once CI is happy, just fixed a detekt issue.

Issue #1185, opened 2022-12-13
Set RN metadata to `sentry_client` on Native SDKs


marandaneto:
### Description

Context https://github.com/getsentry/team-mobile/issues/48
iOS should use `PrivateSentrySDKOnly#setSdkName` and Android `Options#setSentryClientName`
For iOS, it requires upgrading to [v8](https://github.com/getsentry/sentry-dart/issues/1161)

marandaneto:
The goal is that the Native SDKs use the SDK name and version from the Hybrid SDK.

krystofwoldrich:
I think this should be reopened as the way how we name the native SDKs changed.
- https://github.com/getsentry/team-mobile/issues/48#issuecomment-1425726754

PR #1320, opened 2023-03-09
Fix SDK names in the native SDKs

marandaneto:
## :scroll: Description
Change SDK names of the Native SDKs when embedded in a Hybrid SDK.
_#skip-changelog_

Fix https://github.com/getsentry/sentry-dart/issues/1185

## :pencil: Checklist

- [X] All tests passing
- [ ] No breaking changes


flutter/lib/src/version.dart
```diff
@@ -2,4 +2,10 @@
...
 
 /// The default SDK name reported to Sentry.io in the submitted events.
-const String sdkName = 'sentry.dart.flutter';
+String sdkName(bool isWeb) => isWeb ? _browserSdkName : _ioSdkName;
+
+/// The default SDK name reported to Sentry.io in the submitted events.
+const String _ioSdkName = 'sentry.dart.flutter';
+
+/// The default SDK name reported to Sentry.io in the submitted events.
+const String _browserSdkName = '$_ioSdkName.browser';
```
marandaneto: I wonder if we want this level of granularity.

flutter/android/src/main/kotlin/io/sentry/flutter/SentryFlutterPlugin.kt
```diff
@@ -165,16 +165,19 @@ class SentryFlutterPlugin : FlutterPlugin, MethodCallHandler, ActivityAware {
...
-      args.getIfNotNull<Map<String, Any>>("sdk") { sdk ->
-        val name = sdk["name"] as? String
-        val version = sdk["version"] as? String
-        if (name != null && version != null) {
-          val sdkVersion = SdkVersion(name, version)
-          options.setSentryClientName("$name/$version")
-          options.setSdkVersion(sdkVersion)
-        }
+      val name = "sentry.java.android.flutter"
+      val version = io.sentry.android.core.BuildConfig.VERSION_NAME
```
marandaneto: Question, should we use the Native SDK's version (Android in this case), or the Flutter SDK version?
I believe the Native SDK version would make more sense/easier here.

marandaneto:
@bruno-garcia can you double-check if the SDK names are matching your spec?

bruno-garcia (commented):
The names make sense to me: basically appending just `.flutter` 👌

flutter/lib/src/version.dart
marandaneto: We can break down the adoption using other tags instead of SDK name.

brustolin (approved):

Issue #1322, opened 2023-03-09
DioError extractor extracts wrong stacktrace


ueman:
I have tested this change excessively over the last few days and the whole update from Sentry 6.x/Dio 4.x to Sentry 7.0/Dio 5.0 is a big step back regarding traceability of the `DioError`. This is in big part due to changes in `dio` which removed the whole source stacktrace. But due to the `DioErrorExtractor` there are now also cases where the inner stacktrace of `DioError.error.stacktrace` is completely missing in the event.

I think the `DioError.stacktrace` needs to be added by default, as it was the case in `6.x` and only the inner cause error/stacktrace should be extracted via the `DioErrorExtractor`.

Something like:
```dart
class DioErrorExtractor extends ExceptionCauseExtractor<DioError> {
  @override
  ExceptionCause? cause(DioError error) {
    if (error.error == null) {
      return null;
    }
    return ExceptionCause(
      error.error,
      error.error is Error ? (error.error as Error).stackTrace : null,
    );
  }
}
```

_Originally posted by @kuhnroyal in https://github.com/getsentry/sentry-dart/issues/1282#issuecomment-1462184777_

ueman:
@kuhnroyal PR incoming? 😄

kuhnroyal:
Yea sure but I think we need to solve the dio problem first, otherwise the `DioError.stackTrace` is almost always `null`.

marandaneto:
@denrase can you take a look?
If the `stackTrace` is null, we can check if `error` is an `Error` and use the https://api.dart.dev/stable/2.19.4/dart-core/Error/stackTrace.html as mentioned.
Sounds like the right thing to do.
What if `stackTrace` isn't null but `error` also has the `error.stackTrace`, which one has precedence? I believe the original `stackTrace`.

kuhnroyal:
> What if `stackTrace` isn't null but `error` also has the `error.stackTrace`, which one has precedence? I believe the original `stackTrace`.

From my understanding, the `DioError. stackTrace` should always be on the original error and has nothing to do with the cause being extracted here.

kuhnroyal:
Here is an example, the issue is named `String DioError inner stacktrace`.
The stacktrace belongs to the original error which was extracted via `SentryDioClientAdapter` - the error simply has no cause.

[...]

PR #1323, opened 2023-03-09
Improve (Dio)Error stackTrace and cause extraction

kuhnroyal:
## :scroll: Description
This is just a very quick idea idea of how to address https://github.com/getsentry/sentry-dart/issues/1322 and also by default support `Error.stackTrace`.

Fixes https://github.com/getsentry/sentry-dart/issues/1322

## :pencil: Checklist

- [ ] All tests passing
- [ ] No breaking changes

### DIFF
--- a/dart/lib/src/recursive_exception_cause_extractor.dart
+++ b/dart/lib/src/recursive_exception_cause_extractor.dart
@@ -17,7 +17,7 @@ class RecursiveExceptionCauseExtractor {
 
     var currentException = exception;
     ExceptionCause? currentExceptionCause =
-        ExceptionCause(exception, stackTrace);
+        ExceptionCause(exception, stackTrace ?? (exception is Error) ? (exception as Error).stackTrace : null);
 
     while (currentException != null &&
         currentExceptionCause != null &&
--- a/dio/lib/src/dio_error_extractor.dart
+++ b/dio/lib/src/dio_error_extractor.dart
@@ -5,12 +5,13 @@ import 'package:sentry/sentry.dart';
 class DioErrorExtractor extends ExceptionCauseExtractor<DioError> {
   @override
   ExceptionCause? cause(DioError error) {
-    if (error.stackTrace == null) {
+    final cause = error.error;
+    if (cause == null) {
       return null;
     }
     return ExceptionCause(
-      error.error ?? 'DioError inner stacktrace',
-      error.stackTrace,
+      cause,
+      (cause is Error) ? cause.stackTrace : null,
     );
   }
 }


dart/lib/src/recursive_exception_cause_extractor.dart
```diff
@@ -17,7 +17,7 @@ class RecursiveExceptionCauseExtractor {
 
     var currentException = exception;
     ExceptionCause? currentExceptionCause =
-        ExceptionCause(exception, stackTrace);
+        ExceptionCause(exception, stackTrace ?? (exception is Error) ? (exception as Error).stackTrace : null);
```
kuhnroyal, line 20: I think this would also be a nice addition in general for all Dart/Flutter `Error` sub-classes.

marandaneto: Yeah but we do that already.
https://github.com/getsentry/sentry-dart/blob/f1314d57fef996236ceca00725cc9c3acce16d85/dart/lib/src/sentry_exception_factory.dart#L27-L29
I guess we just don't do that when its a chained exception.

kuhnroyal: Hmm yea that's true, but then why is there no stacktrace in my screenshots?

marandaneto: There is, your screenshot is a chained exception, the String has a stack trace, but the DioError does not and your fix should solve this, I guess.
`DioError` is an `Exception` and not an `Error` that's why it didn't work automatically.

kuhnroyal: > `DioError` is an `Exception` and not an `Error` that's why it didn't work automatically.

****, yea you are right. Would it be generally useful to cast the error `Object` to `dynamic` and check if it has useable `stackTrace` property? Right there in the `SentryExceptionFactory` I mean, independent of the Dio case here?

marandaneto: I don't think so, if the exception is an `Error` we already do that (it's a known type).
Any other type/chained exception type has to do that beforehand using the extractor, in this case specifically, we just need to fix the extractor IMO.
Unless having a `stackTrace` property in a custom exception is a well-known and used pattern for the most custom exceptions, I don't think it's the case.

kuhnroyal: But then everything is correct, this is what is currently being done. But it is unusable.
Every `DioError` without a cause (`DioError.error`) is named `DioError inner stacktrace` because this is then extract as the cause.

A stackTrace is not a cause, it just belongs to the original error.

kuhnroyal: What if we add a ` StackTrace? stackTrace(T error) => null;` function to `ExceptionCauseExtractor` and use this here for every error: https://github.com/getsentry/sentry-dart/blob/f1314d57fef996236ceca00725cc9c3acce16d85/dart/lib/src/sentry_exception_factory.dart#L27-L29 and fallback to the current behavior if it is still null afterwards.
This can then be overridden for each error type where it is required.

marandaneto: > But then everything is correct, this is what is currently being done. But it is unusable.
> Every DioError without a cause (DioError.error) is named DioError inner stacktrace because this is then extract as the cause.
> 
> A stackTrace is not a cause, it just belongs to the original error.

Nope, that's not how it is, your PR fixes it and makes it like this.

marandaneto: I have to run the code and check how it is and the changes, right now I'm trying to understand based on only looking into the code, I'm not aware of the dio v5 changes as well besides reviewing the migration PR.
I will check this carefully next week and come back to you, sorry about the back and forth.

kuhnroyal: > > But then everything is correct, this is what is currently being done. But it is unusable.
> > Every DioError without a cause (DioError.error) is named DioError inner stacktrace because this is then extract as the cause.
> > A stackTrace is not a cause, it just belongs to the original error.
> 
> Nope, that's not how it is, your PR fixes it and makes it like this.

No worries, with my PR there is no stacktrace at all when there is no `DioError.error`. There is no code that extracts `DioError.stackTrace` in that case.

marandaneto: As mentioned before, this change specifically isn't needed cus https://github.com/getsentry/sentry-dart/blob/f1314d57fef996236ceca00725cc9c3acce16d85/dart/lib/src/sentry_exception_factory.dart#L27-L29 does it automatically.
The problem is that `DioError` isn't an `Error` so the `DioError.stackTrace` is never read in case the given `stackTrace` param is null, ideally we'd fallback to `DioError.stackTrace`.
I believe this should be solved by `ExceptionCause` or similar approach.
Like a way to resolve a better `stackTrace` from the error for known types, not generally from the cause, can be from the root cause as well.

marandaneto: Your suggestion of parsing it to `dynamic` and accessing the `stackTrace` would work in this case but its not scalable because you might have an Exception that has a `stackTrace` field called differently.
ideally we'd have something similar to the `ExceptionCause` but for stack trace.

Such as:

```dart
abstract class ExceptionCauseExtractor<T> {
  ExceptionCause? cause(T error);
  dynamic stackTrace(T error);
  Type get exceptionType => T;
}
```

`DioErrorExtractor` then:

```dart
@override
  dynamic stackTrace(DioError error) => error.stackTrace;
```

And later the `RecursiveExceptionCauseExtractor` does:

```dart
final betterStackTrace = stackTrace ?? extractor?.stackTrace(extractionSourceSource);
```

And use `betterStackTrace` as the `stackTrace` instead of something else.

How does that sound?

marandaneto: Maybe this should not be part of `ExceptionCauseExtractor` cus the `stackTrace` isn't related to the cause at all, but it was easy to hack it up.

marandaneto: This should rather be part of `sentry_exception_factory.dart` as well closer to `if (throwable is Error)`, maybe even before that cus the user defined stack trace extractor would have precedence.

kuhnroyal: Yea I agree, let's call it `ExceptionStackTraceExtractor`?
Each `ExceptionCause` is also individually processed by `SentryExceptionFactory` so this would then also get applied to all extracted causes.

 > This should rather be part of sentry_exception_factory.dart as well closer to if (throwable is Error), maybe even before that cus the user defined stack trace extractor would have precedence.

Not sure, I think the manually provided stackTrace should have precedence.

marandaneto: `ExceptionStackTraceExtractor` sounds good.

> Not sure, I think the manually provided stackTrace should have precedence.

Most likely there will be cases where the provided stackTrace is better or worse than the extracted one.
So that's fine, I'd expect that people don't provide an `ExceptionStackTraceExtractor` for a known type if the provided stack trace is already enough, so it's a fair trade-off.

Would you like to adapt your PR to consider `ExceptionStackTraceExtractor`? a separate PR is also fine.

kuhnroyal: Yea I'll start working on a new PR.

PR #1335, opened 2023-03-15
Add a way to extract stacktraces from custom exception types

### LINKED ISSUE #1322: DioError extractor extracts wrong stacktrace

kuhnroyal:
## :scroll: Description
* add `ExceptionStackTraceExtractor` and use as fallback in `SentryExceptionFactory`
* add `DioStackTraceExtractor` to the dio module which extracts `DioError.stackTrace`

See discussion in https://github.com/getsentry/sentry-dart/pull/1323
Fixes #1322

## :pencil: Checklist

- [ ] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps

CC @ueman @marandaneto

### DIFF
--- a/dart/lib/src/exception_stacktrace_extractor.dart
+++ b/dart/lib/src/exception_stacktrace_extractor.dart
@@ -17,12 +17,12 @@ import 'sentry_options.dart';
 /// class ExceptionWithInner {
 ///   ExceptionWithInner(this.innerException, this.innerStackTrace);
 ///   Object innerException;
-///   StackTrace innerStackTrace;
+///   dynamic innerStackTrace;
 /// }
 ///
 /// class ExceptionWithInnerStackTraceExtractor extends ExceptionStackTraceExtractor<ExceptionWithInner>  {
 ///   @override
-///   StackTrace? cause(ExceptionWithInner error) {
+///   dynamic cause(ExceptionWithInner error) {
 ///     return error.innerStackTrace;
 ///   }
 /// }
@@ -30,6 +30,6 @@ import 'sentry_options.dart';
 /// options.addExceptionStackTraceExtractor(ExceptionWithInnerStackTraceExtractor());
 /// ```
 abstract class ExceptionStackTraceExtractor<T> {
-  StackTrace? stackTrace(T error);
+  dynamic stackTrace(T error);
   Type get exceptionType => T;
 }
--- a/dart/lib/src/exception_stacktrace_extractor.dart
+++ b/dart/lib/src/exception_stacktrace_extractor.dart
@@ -1,0 +1,35 @@
+import 'protocol.dart';
+import 'sentry_options.dart';
+
+/// Sentry handles [Error.stackTrace] by default. For other cases
+/// extend this abstract class and return a custom [StackTrace] of your
+/// exceptions.
+///
+/// Implementing an extractor and providing it through
+/// [SentryOptions.addExceptionStackTraceExtractor] will enable the framework to
+/// extract the inner stacktrace and add it to [SentryException] when no other
+/// stacktrace was provided while capturing the event.
+///
+/// For an example on how to use the API refer to dio/DioStackTraceExtractor or the
+/// code below:
+///
+/// ```dart
+/// class ExceptionWithInner {
+///   ExceptionWithInner(this.innerException, this.innerStackTrace);
+///   Object innerException;
+///   StackTrace innerStackTrace;
+/// }
+///
+/// class ExceptionWithInnerStackTraceExtractor extends ExceptionStackTraceExtractor<ExceptionWithInner>  {
+///   @override
+///   StackTrace? cause(ExceptionWithInner error) {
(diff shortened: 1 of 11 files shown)


kuhnroyal:
Hmm no idea about the failing workflows.

dart/lib/src/exception_stacktrace_extractor.dart
```diff
@@ -0,0 +1,35 @@
...
+///   @override
+///   StackTrace? cause(ExceptionWithInner error) {
+///     return error.innerStackTrace;
+///   }
+/// }
+///
+/// options.addExceptionStackTraceExtractor(ExceptionWithInnerStackTraceExtractor());
+/// ```
+abstract class ExceptionStackTraceExtractor<T> {
+  StackTrace? stackTrace(T error);
```
marandaneto:
```suggestion
  dynamic stackTrace(T error);
```

I don't like that too but a `StackTrace` can be in many forms.
Can be a `Chain`, can be a `Trace`, can be a `String` if coming from Platform exceptions, etc.

dart/lib/src/exception_stacktrace_extractor.dart
```diff
@@ -0,0 +1,35 @@
...
+/// ```dart
+/// class ExceptionWithInner {
+///   ExceptionWithInner(this.innerException, this.innerStackTrace);
+///   Object innerException;
+///   StackTrace innerStackTrace;
+/// }
+///
+/// class ExceptionWithInnerStackTraceExtractor extends ExceptionStackTraceExtractor<ExceptionWithInner>  {
+///   @override
+///   StackTrace? cause(ExceptionWithInner error) {
```
marandaneto:
```suggestion
///   dynamic cause(ExceptionWithInner error) {
```

dio/lib/src/dio_stacktrace_extractor.dart
```diff
@@ -0,0 +1,10 @@
+import 'package:dio/dio.dart';
+import 'package:sentry/sentry.dart';
+
+/// Extracts the inner stacktrace from [DioError]
+class DioStackTraceExtractor extends ExceptionStackTraceExtractor<DioError> {
+  @override
+  StackTrace? stackTrace(DioError error) {
```
marandaneto, line 7:
```suggestion
  dynamic stackTrace(DioError error) {
```

marandaneto:
> Hmm no idea about the failing workflows.

The analyze one is due to https://github.com/dart-lang/pana/issues/1020
All good.
The other one is a timeout, related to https://github.com/dart-lang/test/issues/1622 most likely, so all good as well.

marandaneto (approved):
Thanks @kuhnroyal 
Left a few comments but other than that LGTM

dart/lib/src/exception_stacktrace_extractor.dart
kuhnroyal: Yea but `Chain`/`Trace` implement `StackTrace` and `String` can be converted with `StackTrace.fromString()`.
Native stacks are the only thing where I see the point :) Will change to keep it consistent with the rest of the SDK.

dio/lib/src/dio_stacktrace_extractor.dart
kuhnroyal: This one is fine, for `DioError` the type can be tightened.

dart/lib/src/exception_stacktrace_extractor.dart
kuhnroyal: Done

dart/lib/src/exception_stacktrace_extractor.dart
kuhnroyal: Done

dart/lib/src/exception_stacktrace_extractor.dart
marandaneto: yeah, the boilerplate from `StackTrace.fromString()` is suboptimal if the SDK can do it automatically anyway.

PR #1363, opened 2023-03-27
Fix SentryUserInteractionWidget throwing when Sentry is not enabled

kuhnroyal:
## :scroll: Description

`SentryUserInteractionWidget` always attempts to cast the `hub.options` to `SentryFlutterOptions`- this cast can fail if `Sentry.isEnabled` is false or due do bad configuration. A simple type check was added to avoid this.

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps
Not sure what more to test.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 ### Fixes
 
 - Sync missing properties to the Native SDKs ([#1354](https://github.com/getsentry/sentry-dart/pull/1354))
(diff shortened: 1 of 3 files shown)


marandaneto (approved):
good catch @kuhnroyal 
can you add a changelog entry? thanks

kuhnroyal:
> good catch @kuhnroyal can you add a changelog entry? thanks

Done

PR #1385, opened 2023-04-07
Improve FutureOr usage only when necessary

### LINKED ISSUE #1336: Remove View Hierarchy from Flutter Web
### Description

Added by #1189
View Hierarchy works for Flutter Web but it requires symbolication thru source maps and right now we are focusing on symbolication for Flutter Mobile and macOS (Android, iOS, macOS).
Later on we can enable support for Flutter web again.

marandaneto:
## :scroll: Description
We should only use `FutureOr<T>` when we don't know its type.
We should always use `Future<T>` or `T` when we know the return type.
We should only `await` if its a `Future`, we can check the return type of a `FutureOr<T>` method.
We always `implement` an interface and only `extend` when we need to make use of its default methods.
Tests should only be `async` if its awaiting for a Future.
We should always honor the return type of an interface, for example, `SentryEvent?` instead of `SentryEvent`.

Also fixes https://github.com/getsentry/sentry-dart/issues/1336
Screenshots and View hierarchy should only be added to errors and not transactions.

Tests already exist.

## :pencil: Checklist

- [X] All tests passing
- [X] No breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Changelog
 
+## Unreleased
+
+### Fixes
+
+- Screenshots and View Hierarchy should only be added to errors ([#1385](https://github.com/getsentry/sentry-dart/pull/1385))
+  - View Hierarchy is removed from Web errors since we don't symbolicate minified View Hierarchy yet.
+- More improvements related to not awaiting `FutureOr<T>` if it's not a future ([#1385](https://github.com/getsentry/sentry-dart/pull/1385))
+
 ## 7.4.2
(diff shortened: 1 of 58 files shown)


krystofwoldrich:
@marandaneto Would it be possible to check this by a lint rule?

marandaneto:
> @marandaneto Would it be possible to check this by a lint rule?

only if we make the linting rules ourselves, they don't exist just yet afaik.

krystofwoldrich (approved):
LGTM 🚀

PR #1402, opened 2023-04-21
Add processor count to device info

### LINKED ISSUE #1401: Use `getExtraContext` on iOS
### Description

`sentry-cocoa` has a new method for HybridSDKs which exposes extra context data that originally were available only in `sentry-cocoa` client. 

https://github.com/getsentry/sentry-cocoa/pull/2855

RN implementation: https://github.com/getsentry/sentry-react-native/pull/2984

marandaneto:
https://github.com/getsentry/team-mobile/issues/77
Fixes https://github.com/getsentry/sentry-dart/issues/1401

## :pencil: Checklist

- [X] All tests passing
- [X] No breaking changes

### DIFF
--- a/flutter/ios/Classes/SentryFlutterPluginApple.swift
+++ b/flutter/ios/Classes/SentryFlutterPluginApple.swift
@@ -206,7 +206,7 @@ public class SentryFlutterPluginApple: NSObject, FlutterPlugin {
                 // merge device
                 if let extraDevice = extraContext[deviceStr] as? [String: Any] {
                     if var currentDevice = context[deviceStr] as? [String: Any] {
-                        currentDevice.mergeDictionary(other: extraDevice)
+                        currentDevice.merge(extraDevice) { (current, _) in current }
                         context[deviceStr] = currentDevice
                     } else {
                         context[deviceStr] = extraDevice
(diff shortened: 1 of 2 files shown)


ueman:
https://github.com/getsentry/sentry-dart/blob/fed8146b5b1d465a10ce707039dbe2f9b2ceb9c8/dart/lib/src/event_processor/enricher/io_enricher_event_processor.dart#L104
Does that need to be changed too?

marandaneto:
> https://github.com/getsentry/sentry-dart/blob/fed8146b5b1d465a10ce707039dbe2f9b2ceb9c8/dart/lib/src/event_processor/enricher/io_enricher_event_processor.dart#L104
> 
> 
> Does that need to be changed too?

Nope because `device?.processorCount` is going to be populated, also because `hasNativeIntegration` is true so it won't fallback to `_getDevice`, its still gonna be useful for non native plats.

flutter/ios/Classes/SentryFlutterPluginApple.swift
```diff
@@ -687,4 +718,12 @@ public class SentryFlutterPluginApple: NSObject, FlutterPlugin {
     }
 }
 
+extension Dictionary {
+    mutating func mergeDictionary(other: Dictionary) {
+        for (key, value) in other {
+            self.updateValue(value, forKey: key)
+        }
+    }
+}
```
krystofwoldrich: L: Looks good, could we use [`dict.merging`](https://developer.apple.com/documentation/swift/dictionary/merge(_:uniquingkeyswith:)-m2ub) from Swift 4, or the project uses older version?

krystofwoldrich (approved):
LGTM 🚀

flutter/ios/Classes/SentryFlutterPluginApple.swift
marandaneto: Good point [`a29ac2a` (#1402)](https://github.com/getsentry/sentry-dart/pull/1402/commits/a29ac2aa2583676b4e6daea80fdd1ff1d37451d8)

PR #1427, opened 2023-05-09
Add support to Gradle 8+

davidmartos96:
## :scroll: Description

Hello, currently the plugin doesn't build when Gradle 8+ is used. That is because the property `namespace` is mandatory.

* I updated the Android example project.
* I included the namespace property with backwards compatibility

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 - Fix `event.origin` and `event.environment` on unhandled exceptions ([#1419](https://github.com/getsentry/sentry-dart/pull/1419))
 - Fix authority redaction ([#1424](https://github.com/getsentry/sentry-dart/pull/1424))
+- Adds a namespace for compatibility with AGP 8.0. ([#1427](https://github.com/getsentry/sentry-dart/pull/1427))
 
 ### Dependencies
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,12 +1,17 @@
 # Changelog
 
+## Unreleased
+
+### Fixes
+
+- Adds a namespace for compatibility with AGP 8.0. ([#1427](https://github.com/getsentry/sentry-dart/pull/1427))
+
 ## 7.5.2
 
 ### Fixes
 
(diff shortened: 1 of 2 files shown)


krystofwoldrich:
@marandaneto @davidmartos96 Maybe an integration test to check that the plugin builds with Gradle 7 and 8 would make sense to add?

davidmartos96:
@krystofwoldrich Is there any test of that kind in the repo? I'm not sure how it would work or if it would be worth the effort. 

This change is equivalent to the one the official plugins have recently pushed in latest versions.
Like here: https://pub.dev/packages/shared_preferences_android/changelog
Here is the commit: https://github.com/flutter/packages/commit/a86beafa8912609275225847198c9c0164957d09
The if statement is for Gradle versions not compatible with the namespace attribute. It should build fine for different Gradle versions.

davidmartos96:
@krystofwoldrich Is there any test of that kind in the repo? I'm not sure how it would work or if it would be worth the effort. 

This change is equivalent to the one the official plugins have recently pushed in latest versions.
Like here: https://pub.dev/packages/shared_preferences_android/changelog
Here is the commit: https://github.com/flutter/packages/commit/a86beafa8912609275225847198c9c0164957d09
The if statement is for Gradle versions not compatible with the namespace attribute. It should build fine for different Gradle versions.

marandaneto:
> @marandaneto @davidmartos96 Maybe an integration test to check that the plugin builds with Gradle 7 and 8 would make sense to add?

Since we don't use AGP 8 yet, this is an integration test per see :)
Also every flutter package does this already https://github.com/flutter/packages/commit/a86beafa8912609275225847198c9c0164957d09
Its safe to go, otherwise we'd need to set up another sample app just for that (CI overhead, etc), but I'd do if the change was more fragile.

marandaneto (approved):

marandaneto:
Thanks @davidmartos96

marandaneto:
@davidmartos96 can you add a changelog entry? https://github.com/getsentry/sentry-dart/blob/main/CHANGELOG.md#fixes

davidmartos96:
@marandaneto Done!

PR #1436, opened 2023-05-10
Support package_info_plus 4.x

ThexXTURBOXx:
## :scroll: Description

This PR adds support for `package_info_plus 4.x`

It fixes compatibility problems when using AGP 8.x by supporting a "fixed" version of package_info_plus.
There might be other libraries which do not support that yet, but they can probably be separately updated later.

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Fixes
 
-- Fix battery level convertion for iOS 16.4 ([#1433](https://github.com/getsentry/sentry-dart/pull/1433))
+- Fix battery level conversion for iOS 16.4 ([#1433](https://github.com/getsentry/sentry-dart/pull/1433))
 - Adds a namespace for compatibility with AGP 8.0. ([#1427](https://github.com/getsentry/sentry-dart/pull/1427))
 
 ## 7.5.2
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
 - Fix battery level conversion for iOS 16.4 ([#1433](https://github.com/getsentry/sentry-dart/pull/1433))
 - Adds a namespace for compatibility with AGP 8.0. ([#1427](https://github.com/getsentry/sentry-dart/pull/1427))
 
+### Breaking Changes
+
+- Support `package_info_plus` versions `4.x` which require `min_sdk_version` `>=19` and iOS `11`
+
 ## 7.5.2
 
 ### Fixes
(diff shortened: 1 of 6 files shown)


marandaneto:
@ThexXTURBOXx thanks, can you add a changelog entry? https://github.com/getsentry/sentry-dart/blob/main/CHANGELOG.md

marandaneto:
@ThexXTURBOXx also this:

> /home/runner/work/sentry-dart/sentry-dart/flutter/example/android/app/src/main/AndroidManifest.xml Error:
> 	uses-sdk:minSdkVersion 16 cannot be smaller than version 19 declared in library [:integration_test] /home/runner/work/sentry-dart/sentry-dart/flutter/example/build/integration_test/intermediates/merged_manifest/release/AndroidManifest.xml as the library might be using APIs not available in 16
> 	Suggestion: use a compatible library with a minSdk of at most 16,
> 		or increase this project's minSdk version to at least 19,
> 		or use tools:overrideLibrary="dev.flutter.integration_test" to force usage (may lead to runtime failures)

Let's do the `tools:overrideLibrary` thingie.

ThexXTURBOXx:
I thought about this for a bit and it might be more advisable to add a breaking change instead and increment the [`min_sdk_version`](https://github.com/getsentry/sentry-dart/blob/main/flutter/android/build.gradle#L37) since 16 is not supported anymore anyways (because of the device_info_plus update).

Please let me know what you think.

marandaneto:
> device_info_plus

Based on what I see, the class already checks SDK version for compatibility.
https://github.com/fluttercommunity/plus_plugins/blob/c8f7b6342a7c51eafafae95792775505d2b52ce9/packages/package_info_plus/package_info_plus/android/src/main/kotlin/dev/fluttercommunity/plus/packageinfo/PackageInfoPlugin.kt#L69

This is just to ensure that the SDK compiles, you can still force minAPi 19 in your final app, we compile our sample with the min. supported version as well https://github.com/getsentry/sentry-dart/blob/bf4aed726e9adcfe8b364aef9d0870dd2c4ad5fe/min_version_test/android/app/build.gradle#L51
The AGP plugin checks if there's any class being used that is not supported < 16 and will fail the build, if that happens, then maybe increasing the version is the way to go.

ThexXTURBOXx:
Sorry, I of course meant `package_info_plus`.
I have added a changelog entry and updated the `min_sdk_version`s all over the place and it should build now. Let's see.

ThexXTURBOXx:
Since this is still not working I will close this in favor of #1440

Issue #1457, opened 2023-05-16
Using SentryAssetBundle causes images to flash when navigating between pages


stx:
### Platform

Flutter Web

### Obfuscation

Disabled

### Debug Info

Disabled

### Doctor

[✓] Flutter (Channel stable, 3.10.0, on macOS 13.3.1 22E772610a darwin-arm64, locale en-US)
    • Flutter version 3.10.0 on channel stable at /Users/james/Development/flutter
    • Upstream repository https://github.com/flutter/flutter.git
    • Framework revision 84a1e904f4 (7 days ago), 2023-05-09 07:41:44 -0700
    • Engine revision d44b5a94c9
    • Dart version 3.0.0
    • DevTools version 2.23.1

[✓] Android toolchain - develop for Android devices (Android SDK version 32.1.0-rc1)
    • Android SDK at /Users/james/Library/Android/sdk
    • Platform android-33, build-tools 32.1.0-rc1
    • Java binary at: /Applications/Android Studio.app/Contents/jbr/Contents/Home/bin/java
    • Java version OpenJDK Runtime Environment (build 11.0.15+0-b2043.56-8887301)
    • All Android licenses accepted.

[✓] Xcode - develop for iOS and macOS (Xcode 14.2)
    • Xcode at /Applications/Xcode.app/Contents/Developer
    • Build 14C18
    • CocoaPods version 1.12.0

[✓] Chrome - develop for the web
    • Chrome at /Applications/Google Chrome.app/Contents/MacOS/Google Chrome

[✓] Android Studio (version 2022.1)
    • Android Studio at /Applications/Android Studio.app/Contents
    • Flutter plugin can be installed from:
      🔨 https://plugins.jetbrains.com/plugin/9212-flutter
    • Dart plugin can be installed from:
      🔨 https://plugins.jetbrains.com/plugin/6351-dart
    • Java version OpenJDK Runtime Environment (build 11.0.15+0-b2043.56-8887301)

[✓] VS Code (version 1.78.2)
    • VS Code at /Applications/Visual Studio Code.app/Contents
    • Flutter extension version 3.64.0

[✓] Connected device (3 available)
    • Apollo (mobile) • 00008110-0004752E3AB8801E • ios            • iOS 16.2 20C65
    • macOS (desktop) • macos                     • darwin-arm64   • macOS 13.3.1 22E772610a darwin-arm64
    • Chrome (web)    • chrome                    • web-javascript • Google Chrome 112.0.5615.137
[...]

PR #1462, opened 2023-05-19
`SentryAssetBundle` returns Future by default

marandaneto:
## :scroll: Description
Calls were made `async` and it was `await` to check its length.
Now they return the Future by default and its only `await` when it should be.

Relates to https://github.com/getsentry/sentry-dart/issues/1457

## :pencil: Checklist

- [X] All tests passing
- [X] No breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Enhancements
+
+- `SentryAssetBundle` returns Future by default ([#1462](https://github.com/getsentry/sentry-dart/pull/1462))
+
 ### Dependencies
 
 - Bump Android SDK from v6.18.1 to v6.19.0 ([#1455](https://github.com/getsentry/sentry-dart/pull/1455))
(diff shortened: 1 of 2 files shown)


krystofwoldrich (approved):
LGTM 🚀

stefanosiano:
I'm not sure i get the change done here.
I mean, we return a future instead of executing the call right away (async). But then that future is going to be called anyway, so what does it exactly change?
Also, the reason is to fix the user complain, but it seems like it didn't.
What am i missing? 😅

marandaneto:
> I'm not sure i get the change done here. I mean, we return a future instead of executing the call right away (async). But then that future is going to be called anyway, so what does it exactly change? Also, the reason is to fix the user complain, but it seems like it didn't. What am i missing? 😅

This is more semantically correct since it's an async method that has to `await` on something else.
If it returns a Future, the Future is added to the event loop only and executed whenever it should be executed (the user has the chance to `await` if needed).
Executing right away, you are taking the decision on the user's behalf.

That does not fix the issue that's why it's `Relates to` instead of `Closes to`, the original issues remain open since I can reproduce the issue even without the `SentryAssetBundle`.

PR #1468, opened 2023-05-22
Change SDK name of Sentry Native for Flutter

### LINKED ISSUE #1333: Change SDK name of Sentry Native for Flutter
### Description

Right now it is `sentry.native.android` and this is set at compile time.
Ideally, this would be `sentry.native.android.flutter` or just `sentry.native.flutter`.

denrase:
#skip-changelog

## :scroll: Description

Update SDK names set in native plugins.

Closes  #1333

## :pencil: Checklist

- [ ] All tests passing
- [ ] No breaking changes

### DIFF
--- a/flutter/android/src/main/kotlin/io/sentry/flutter/SentryFlutterPlugin.kt
+++ b/flutter/android/src/main/kotlin/io/sentry/flutter/SentryFlutterPlugin.kt
@@ -41,8 +41,8 @@ class SentryFlutterPlugin : FlutterPlugin, MethodCallHandler, ActivityAware {
   private var autoPerformanceTracingEnabled = false
 
   private val flutterSdk = "sentry.dart.flutter"
-  private val androidSdk = "sentry.native.android.flutter"
-  private val nativeSdk = "sentry.native.android"
+  private val androidSdk = "sentry.java.android.flutter"
+  private val nativeSdk = "sentry.native.android.flutter"
 
   override fun onAttachedToEngine(flutterPluginBinding: FlutterPlugin.FlutterPluginBinding) {
     context = flutterPluginBinding.applicationContext
@@ -182,6 +182,7 @@ class SentryFlutterPlugin : FlutterPlugin, MethodCallHandler, ActivityAware {
 
       options.sdkVersion = sdkVersion
       options.sentryClientName = "$androidSdk/$VERSION_NAME"
+      options.nativeSdkName = nativeSdk
 
       options.setBeforeSend { event, _ ->
         setEventOriginTag(event)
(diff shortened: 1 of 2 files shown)


flutter/android/src/main/kotlin/io/sentry/flutter/SentryFlutterPlugin.kt
```diff
@@ -41,7 +41,7 @@ class SentryFlutterPlugin : FlutterPlugin, MethodCallHandler, ActivityAware {
   private var autoPerformanceTracingEnabled = false
 
   private val flutterSdk = "sentry.dart.flutter"
-  private val androidSdk = "sentry.java.android.flutter"
+  private val androidSdk = "sentry.native.android.flutter"
```
marandaneto: This change isn't needed.
The changes are about the `nativeSdk`

flutter/ios/Classes/SentryFlutterPluginApple.swift
```diff
@@ -12,7 +12,7 @@ import AppKit
 // swiftlint:disable:next type_body_length
 public class SentryFlutterPluginApple: NSObject, FlutterPlugin {
 
-    private static let nativeClientName = "sentry.cocoa.flutter"
+    private static let nativeClientName = "sentry.native.cocoa.flutter"
```
marandaneto: This change isn't needed, the issue is only about the Sentry Native for Android NDK

denrase:
@marandaneto I tried to check the SDK names on sentry.io through the example app, but i'm not seeing those values set, neither for iOS nor Android. Am I looking in the wrong place?

Bildschirmfoto 2023-05-22 um 11 10 07

flutter/ios/Classes/SentryFlutterPluginApple.swift
denrase: So we don't need to align on the coco side that the naming is the same?

marandaneto:
> @marandaneto I tried to check the SDK names on sentry.io through the example app, but i'm not seeing those values set, neither for iOS nor Android. Am I looking in the wrong place?
> 
> Bildschirmfoto 2023-05-22 um 11 10 07

The place is right, but events captured on the Dart layer will be this name.
See the `AndroidExample` class in `main.dart`.

When this method is executed https://github.com/getsentry/sentry-dart/blob/d883d62345e5c21066faf14885eea7afd75b89e1/flutter/example/android/app/src/main/kotlin/io/sentry/samples/flutter/MainActivity.kt#L30
SDK name should be `sentry.java.android.flutter` (it is already).

When those 2 methods are called https://github.com/getsentry/sentry-dart/blob/d883d62345e5c21066faf14885eea7afd75b89e1/flutter/example/android/app/src/main/kotlin/io/sentry/samples/flutter/MainActivity.kt#L33-L38
SDK name it should be called `sentry.native.android.flutter` -> this is the one that needs fixing.

flutter/ios/Classes/SentryFlutterPluginApple.swift
marandaneto: Sentry the confusion, when I meant Sentry Native isn't about Android and iOS, its about https://github.com/getsentry/sentry-native
SentryNative for NDK support.- Android.

marandaneto:
@denrase example https://github.com/getsentry/sentry-react-native/pull/3086/files

marandaneto (approved):

PR #1476, opened 2023-05-24
Wrapped methods return a Future instead of executing right away

marandaneto:
## :scroll: Description
Wrapped methods return a Future instead of executing right away

Follow up https://github.com/getsentry/sentry-dart/pull/1462#issuecomment-1559152056
I was calling the declared `Future` function in the return statement with the impression that I was returning the function, my bad.

## :pencil: Checklist

- [X] All tests passing
- [X] No breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## Unreleased
 
+### Fixes
+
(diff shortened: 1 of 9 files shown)


krystofwoldrich (approved):
Looks good 🚀, now we right away return Future.

PR #1485, opened 2023-05-25
Fix `readTimeoutMillis` wrongly configures `connectionTimeoutMillis` instead of the correct field

fzyzcjy:
Just realized this when reading code. Not tested locally, want to use CI to see whether this is a real bug (i.e. should fix) or not.

## :pencil: Checklist

- [ ] All tests passing
- [ ] No breaking changes


marandaneto:
Good catch @fzyzcjy 
Its time to address https://github.com/getsentry/sentry-dart/issues/1488

marandaneto (approved):

krystofwoldrich (approved):

Issue #1486, opened 2023-05-26
Sentry Flutter Android does not capture any exceptions at all when custom initialization - caused by callback failure, thus would be great to add a log / throw when this happens


fzyzcjy:
### Platform

Flutter Mobile

### Obfuscation

Disabled

### Debug Info

Disabled

### Doctor

latest

### Version

latest

### Steps to Reproduce

Custom initialization:

```kt
class MyFlutterApplication : FlutterApplication() {
    override fun onCreate() {
        super.onCreate()
        initSentry()
    }

    fun initSentry() {
        SentryAndroid.init(this) { options ->
            options.dsn = SENTRY_DSN
            val removed = options.integrations.removeAll { it is TempSensorBreadcrumbsIntegration }
            check(removed)
        }
    }
```

And disable Dart side auto initiailization:

```dart
    options
      ..autoInitializeNativeSdk = !Platform.isAndroid
      ......
```

Then, `Sentry.captureException(...)`.

#### Extra experiment and possible cause analysis

I have compared two cases:

1. The standard default way - let Flutter side auto init Android side
2. The way mentioned above - init Sentry Android manually at FlutterApplication.onCreate, and disable Flutter side from initing Android side

I see both cases call `SentryClient._attachClientReportsAndSend` and `FileSystemTransport.send` happily. Indeed, with the following extra logs, I can see "hi 1" and "hi 2" both.

<details>

![image](https://github.com/getsentry/sentry-dart/assets/5236035/cb8fea5f-9a94-486b-899a-435667583e8c)

</details>

I will update my findings here. If I can fix it, hopefully this issue can be used as a future reference for who wants to manually initialize sentry.

### Expected Result

See exception sent to sentry server

### Actual Result

No exception sent to sentry server

### Are you willing to submit a PR?

None

fzyzcjy:
I see the problem:

```
2023-05-26 08:14:12.449 16445-16445 Sentry                     E  Error in the 'OptionsConfiguration.configure' callback.
```

if enabling the `isDebug` output.

So, now the issue becomes: It would be great to have a more user friendly hint when there are errors. For example, when such initialization fails, we may either:

1. Output a log even if isDebug=false, since this causes a *fatal* error - the whole sentry will silently fail in the whole lifetime
2. Or, let sentry run happily even if this callback fails
[...]

PR #1489, opened 2023-05-26
Check if the Native SDKs are enabled when using autoInitializeNativeSdk false

marandaneto:
## :scroll: Description
Check if the Native SDKs are enabled when using `autoInitializeNativeSdk=false`

Relates to https://github.com/getsentry/sentry-dart/issues/1486

Running autoInitializeNativeSdk mode and checking logs.

## :pencil: Checklist

- [X] All tests passing
- [X] No breaking changes

### DIFF
--- a/flutter/android/src/main/kotlin/io/sentry/flutter/SentryFlutterPlugin.kt
+++ b/flutter/android/src/main/kotlin/io/sentry/flutter/SentryFlutterPlugin.kt
@@ -354,20 +354,25 @@ class SentryFlutterPlugin : FlutterPlugin, MethodCallHandler, ActivityAware {
   }
 
   private fun captureEnvelope(call: MethodCall, result: Result) {
+    if (!Sentry.isEnabled()) {
+      result.error("1", "The Sentry Android SDK is disabled", null)
+      return
+    }
+
     val args = call.arguments() as List<Any>? ?: listOf<Any>()
     if (args.isNotEmpty()) {
       val event = args.first() as ByteArray?
 
       if (event != null && event.isNotEmpty()) {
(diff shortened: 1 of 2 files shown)


flutter/android/src/main/kotlin/io/sentry/flutter/SentryFlutterPlugin.kt
```diff
@@ -354,20 +354,25 @@ class SentryFlutterPlugin : FlutterPlugin, MethodCallHandler, ActivityAware {
   }
 
   private fun captureEnvelope(call: MethodCall, result: Result) {
+    if (!Sentry.isEnabled()) {
```
marandaneto, line 357: @brustolin or @philipphofmann do we need to do this guard for iOS before calling `PrivateSentrySDKOnly.capture`? or does `capture` check `isEnabled` internally already?

marandaneto:
@krystofwoldrich this might be useful for RN as well.

flutter/android/src/main/kotlin/io/sentry/flutter/SentryFlutterPlugin.kt
brustolin: If the SDK in the native side is disabled then we don't capture the envelope.
But I would keep this check in the dart side as well.

marandaneto: `capture` already calls `SentryClient#isDisabled` so that's all we need, to need to double guard in this case, thanks.

stefanosiano (approved):

PR #1503, opened 2023-06-04
Update widgets_binding_observer.dart

abdelrahmanelmarakby:
fix ```widgets_binding_observer.dart:183:17: Error: A non-null value must be returned since the return type 'String' doesn't allow null.``` because switch returns String? not String because there is no default case

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,11 @@
   - [changelog](https://github.com/getsentry/sentry-java/blob/main/CHANGELOG.md#6210)
   - [diff](https://github.com/getsentry/sentry-java/compare/6.20.0...6.21.0)
 
+
+### Fixes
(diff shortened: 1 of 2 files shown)


ueman:
Apparently, there were more states added: https://master-api.flutter.dev/flutter/dart-ui/AppLifecycleState.html

marandaneto:
Makes sense @abdelrahmanelmarakby 
Would you mind adding a [changelog](https://github.com/getsentry/sentry-dart/blob/main/CHANGELOG.md) entry?

abdelrahmanelmarakby:
@marandaneto will add the entry , any thing else to take in consideration?

marandaneto:
@abdelrahmanelmarakby thanks for the PR.
I will close this one in favor of https://github.com/getsentry/sentry-dart/pull/1506 because there are more places doing this and can break in the future, so I decided to fix the other cases as well in a single PR.
Will add the credits to the PR tho :)

Issue #1504, opened 2023-06-04
Newest Flutter master branch breaks Sentry


RossComputerGuy:
### Platform

Flutter Mobile, Flutter Desktop, Flutter Web

### Obfuscation

Enabled

### Debug Info

Enabled

### Doctor

```
[!] Flutter (Channel master, 3.11.0-18.0.pre.76, on Freedesktop.org SDK 22.08 (Flatpak runtime) 5.15.96, locale ja_JP.UTF-8)
    • Flutter version 3.11.0-18.0.pre.76 on channel master at /home/ross/flutter
    ! The flutter binary is not on your path. Consider adding /home/ross/flutter/bin to your path.
    ! The dart binary is not on your path. Consider adding /home/ross/flutter/bin to your path.
    • Upstream repository https://github.com/flutter/flutter
    • Framework revision 0b7415356e (2 days ago), 2023-06-03 07:16:37 -0400
    • Engine revision c838a1b059
    • Dart version 3.1.0 (build 3.1.0-166.0.dev)
    • DevTools version 2.24.0
    • If those were intentional, you can disregard the above warnings; however it is recommended to use "git" directly to perform update checks and upgrades.

[✓] Android toolchain - develop for Android devices (Android SDK version 33.0.2)
    • Android SDK at /home/ross/Android/Sdk
    • Platform android-33, build-tools 33.0.2
    • Java binary at: /app/extra/android-studio/jbr/bin/java
    • Java version OpenJDK Runtime Environment (build 17.0.6+0-17.0.6b802.4-9586694)
    • All Android licenses accepted.

[✗] Chrome - develop for the web (Cannot find Chrome executable at google-chrome)
    ! Cannot find Chrome. Try setting CHROME_EXECUTABLE to a Chrome executable.

[✗] Linux toolchain - develop for Linux desktop
    ✗ clang++ is required for Linux development.
      It is likely available from your distribution (e.g.: apt install clang), or can be downloaded from https://releases.llvm.org/
    • cmake version 3.26.4
    • ninja version 1.11.1
    • pkg-config version 1.9.5

[!] Android Studio (not installed)
    • Android Studio not found; download from https://developer.android.com/studio/index.html
      (or visit https://flutter.dev/docs/get-started/install/linux#android-setup for detailed instructions).

[✓] Connected device (2 available)
    • Pixel 4a 5G (mobile) • 09231JECB00880 • android-arm64 • Android 13 (API 33)
[...]

PR #1506, opened 2023-06-05
Enums use its name instead of non exhaustive switches

marandaneto:
## :scroll: Description
Enums will use its name instead of non-exhaustive switches
Since Dart 2.15 we can use the `name` if they match 1 by 1.

https://github.com/getsentry/sentry-dart/pull/1503 and https://github.com/getsentry/sentry-dart/issues/1504
Follow up by @abdelrahmanelmarakby thanks :)

## :pencil: Checklist

- [X] All tests passing
- [X] No breaking changes

### DIFF
--- a/dart/lib/src/protocol/sentry_device.dart
+++ b/dart/lib/src/protocol/sentry_device.dart
@@ -221,19 +221,6 @@ class SentryDevice {
 
   /// Produces a [Map] that can be serialized to JSON.
   Map<String, dynamic> toJson() {
-    String? orientation;
-
-    switch (this.orientation) {
-      case SentryOrientation.portrait:
-        orientation = 'portait';
-        break;
-      case SentryOrientation.landscape:
-        orientation = 'landscape';
-        break;
-      case null:
-        orientation = null;
-        break;
-    }
     return <String, dynamic>{
       if (name != null) 'name': name,
       if (family != null) 'family': family,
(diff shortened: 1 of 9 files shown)


dart/lib/src/protocol/sentry_device.dart
```diff
@@ -221,27 +221,14 @@ class SentryDevice {
...
-    }
     return <String, dynamic>{
       if (name != null) 'name': name,
       if (family != null) 'family': family,
       if (model != null) 'model': model,
       if (modelId != null) 'model_id': modelId,
       if (arch != null) 'arch': arch,
       if (batteryLevel != null) 'battery_level': batteryLevel,
-      if (orientation != null) 'orientation': orientation,
+      if (orientation != null) 'orientation': orientation!.name,
```
ueman, line 231: Can't this result in a force unwrap exception?

marandaneto: Don't think so because `orientation` is `final`, and we do `orientation != null` before.
I also don't like `!` but if it's `final`, it's safe.

ueman: > and we do orientation != null before.

True, I missed that.

stefanosiano (approved):

krystofwoldrich (approved):

Issue #1509, opened 2023-06-07
Replace `DioError` to `DioException`


marandaneto:
### Description

Right now files are annotated with `deprecated_member_use` due to Dio's breaking change.
`DioError` still exists and is back-compatible, at some point may be removed.

kuhnroyal:
We have removed `DioError` in the `6.0.0` branch - see https://github.com/cfug/dio/pull/1825
So at some point in the future `sentry_dio` needs a breaking change as well.

marandaneto:
Thanks for the heads up.

PR #1510, opened 2023-06-07
Ignore DioError deprecared member use

marandaneto:
## :scroll: Description
_#skip-changelog_

Relates to https://github.com/getsentry/sentry-dart/issues/1509

## :pencil: Checklist

- [ ] All tests passing
- [ ] No breaking changes

### DIFF
--- a/dio/lib/src/sentry_dio_extension.dart
+++ b/dio/lib/src/sentry_dio_extension.dart
@@ -1,3 +1,5 @@
+// ignore_for_file: public_member_api_docs, deprecated_member_use
+
 import 'package:dio/dio.dart';
 import 'package:sentry/sentry.dart';
 import 'dio_error_extractor.dart';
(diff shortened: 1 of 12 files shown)


dio/lib/src/sentry_dio_extension.dart
```diff
@@ -1,3 +1,5 @@
+// ignore_for_file: public_member_api_docs, deprecated_member_use
```
marandaneto:
```suggestion
// ignore_for_file: deprecated_member_use
```

stefanosiano (approved):
This is fine, but should we also provide support for DioException somehow?
Or would it mean to increase the minimum supported version of dio?

marandaneto:
> This is fine, but should we also provide support for DioException somehow? Or would it mean to increase the minimum supported version of dio?

> would it mean to increase the minimum supported version of dio

yes, so better to wait before the adoption of the new version is very relevant.

PR #1543, opened 2023-07-10
Upgrade android emulator and caching

marandaneto:
## :scroll: Description
_#skip-changelog_

1. `Error: Timeout waiting for emulator to boot.`
2. significantly reduce emulator startup time by setting up AVD snapshot caching.
https://github.com/ReactiveCircus/android-emulator-runner

## :pencil: Checklist

- [ ] All tests passing
- [ ] No breaking changes


.github/workflows/flutter_integration_test.yml
```diff
@@ -46,11 +46,36 @@ jobs:
       - name: flutter pub get
         run: flutter pub get
 
+      - name: Gradle cache
+        uses: gradle/gradle-build-action@v2
```
krystofwoldrich, line 50: Should we add the SHA instead?

.github/workflows/flutter_integration_test.yml
```diff
@@ -46,11 +46,39 @@ jobs:
...
+
+      - name: create AVD and generate snapshot for caching
+        if: steps.avd-cache.outputs.cache-hit != 'true'
+        uses: reactivecircus/android-emulator-runner@d94c3fbe4fe6a29e4a5ba47c12fb47677c73656b #pin@v2.28.0
+        with:
+          working-directory: ./flutter/example
+          api-level: 21
+          force-avd-creation: false
+          emulator-options: -no-window -gpu swiftshader_indirect -noaudio -no-boot-anim -camera-back none
+          disable-animations: false
```
krystofwoldrich, line 69: L: Should this also be `true`? I guess we don't need animation anywhere.
```suggestion
          disable-animations: true
```

.github/workflows/flutter_integration_test.yml
marandaneto: This is a verified GH action, no need.

.github/workflows/flutter_integration_test.yml
```diff
@@ -46,11 +46,39 @@ jobs:
...
+          script: echo "Generated AVD snapshot for caching."
+
       - name: launch android emulator & run android integration test
-        uses: reactivecircus/android-emulator-runner@50986b1464923454c95e261820bc626f38490ec0 #pin@v2.27.0
+        uses: reactivecircus/android-emulator-runner@d94c3fbe4fe6a29e4a5ba47c12fb47677c73656b #pin@v2.28.0
         with:
           working-directory: ./flutter/example
           api-level: 21
+          force-avd-creation: false
+          emulator-options: -no-snapshot-save -no-window -gpu swiftshader_indirect -noaudio -no-boot-anim -camera-back none
```
krystofwoldrich, line 80: L: Here it's not that long, so it's good, but in case there would be more params we can use YML multiline string `>` that replaces new lines with spaces.

```yml
          emulator-options: >
            -noaudio
            ...
```

https://github.com/getsentry/sentry-react-native/blob/da1b6a88c430816aafb4ed733863395007d667c8/.github/workflows/e2e.yml#L434-L435

.github/workflows/flutter_integration_test.yml
krystofwoldrich: I've noticed that in the actions' repo it is also with enabled animations, but that must be a typo, or does it have a side effect?
https://github.com/ReactiveCircus/android-emulator-runner

krystofwoldrich (approved):
Looks good 🚀, I left two comments. 

I've noticed the bootup timeout quite a bit in RN too, so I will definitely add this there too.

stefanosiano:
Just a note, I tried to do the same on the Java repo, but the cache actually made tests slower and sometimes they timed out, so I had to revert it.
Let's run it a few times to check it really improves timings

.github/workflows/flutter_integration_test.yml
marandaneto: No idea, I'd keep it as it is tho since it's following the https://github.com/ReactiveCircus/android-emulator-runner example.

PR #1582, opened 2023-08-03
Ensure Breadcrumbs and SentryEvents are always serialisable

### LINKED ISSUE #1581: Logging integration behaviour when message is `Object` is inconsistent with the Logging package
### Platform

Dart

### Obfuscation

Enabled

### Debug Info

Enabled

### Doctor

```
[!] Flutter (Channel unknown, 3.10.6, on macOS 13.4 22F66 darwin-x64, locale en-GB)
    ! Flutter version 3.10.6 on channel unknown at /Users/barabas/src/flutter
      Currently on an unknown channel. Run `flutter channel` to switch to an official channel.
      If that doesn't fix the issue, reinstall Flutter by following instructions at https://flutter.dev/docs/get-started/install.
    ! Unknown upstream repository.
      Reinstall Flutter by following instructions at https://flutter.dev/docs/get-started/install.
    • Framework revision f468f3366c (3 weeks ago), 2023-07-12 15:19:05 -0700
    • Engine revision cdbeda788a
    • Dart version 3.0.6
    • DevTools version 2.23.1
    • If those were intentional, you can disregard the above warnings; however it is recommended to use "git" directly to perform update checks and upgrades.

    • Android SDK at /Users/barabas/Library/Android/sdk
    • Platform android-34, build-tools 34.0.0
    • ANDROID_HOME = /Users/barabas/Library/Android/sdk
    • Java version OpenJDK Runtime Environment (build 17.0.6+0-17.0.6b829.9-10027231)

[✓] Xcode - develop for iOS and macOS (Xcode 14.3.1)
    • Build 14E300c
    • CocoaPods version 1.11.3

[✓] Android Studio (version 2022.3)
    • Java version OpenJDK Runtime Environment (build 17.0.6+0-17.0.6b829.9-10027231)

    • Pixel 5 (mobile) • 0C221FDD4003LH • android-arm64  • Android 13 (API 33)
    • macOS (desktop)  • macos          • darwin-x64     • macOS 13.4 22F66 darwin-x64
    • Chrome (web)     • chrome         • web-javascript • Google Chrome 113.0.5672.126
    ! Error: iPhone XR has recently restarted. Xcode will continue when iPhone XR is unlocked. (code -14)

[✓] Network resources
    • All expected network resources are available.

! Doctor found issues in 1 category.

```

### Version

7.9.0

### Steps to Reproduce

1. Set up an app with Sentry and logging integration
2. Log an object: `Logger('test').info(DemoObject());`

```
class DemoObject {
  final int answer = 42;

  @override
  String toString() {
    return 'DemoObject: $answer';
  }
}
```

### Expected Result

Sentry creates a breadcrumb from the log that matches the output of the app's logger.

### Actual Result

Sentry drops the breadcrumb with an exception.

```
[sentry] [error] Native call `addBreadcrumb` failed
         Invalid argument: Instance of 'DemoObject'
         #0      StandardMessageCodec.writeValue (package:flutter/src/services/message_codecs.dart:465:7)
         #1      StandardMessageCodec.writeValue.<anonymous closure> (package:flutter/src/services/message_codecs.dart:462:9)
         #2      _LinkedHashMapMixin.forEach (dart:collection-patch/compact_hash.dart:625:13)
         #3      StandardMessageCodec.writeValue (package:flutter/src/services/message_codecs.dart:460:13)
         #4      StandardMessageCodec.writeValue.<anonymous closure> (package:flutter/src/services/message_codecs.dart:462:9)
         #5      _LinkedHashMapMixin.forEach (dart:collection-patch/compact_hash.dart:625:13)
         #6      StandardMessageCodec.writeValue (package:flutter/src/services/message_codecs.dart:460:13)
         #7      StandardMessageCodec.writeValue.<anonymous closure> (package:flutter/src/services/message_codecs.dart:462:9)
         #8      _LinkedHashMapMixin.forEach (dart:collection-patch/compact_hash.dart:625:13)
         #9      StandardMessageCodec.writeValue (package:flutter/src/services/message_codecs.dart:460:13)
         #10     StandardMethodCodec.encodeMethodCall (package:flutter/src/services/message_codecs.dart:603:18)
         #11     MethodChannel._invokeMethod (package:flutter/src/services/platform_channel.dart:299:34)
         #12     MethodChannel.invokeMethod (package:flutter/src/services/platform_channel.dart:483:12)
         #13     SentryNativeChannel.addBreadcrumb (package:sentry_flutter/src/sentry_native_channel.dart:59:12)
         #14     SentryNative.addBreadcrumb (package:sentry_flutter/src/sentry_native.dart:72:34)
         #15     NativeScopeObserver.addBreadcrumb (package:sentry_flutter/src/native_scope_observer.dart:29:25)
         #16     Scope.addBreadcrumb.<anonymous closure> (package:sentry/src/scope.dart:195:31)
         #17     Scope._callScopeObservers (package:sentry/src/scope.dart:455:21)
         #18     Scope.addBreadcrumb (package:sentry/src/scope.dart:194:13)
         #19     Hub.addBreadcrumb (package:sentry/src/hub.dart:293:24)
         #20     Sentry.addBreadcrumb (package:sentry/src/sentry.dart:233:12)
         #21     HubAdapter.addBreadcrumb (package:sentry/src/hub_adapter.dart:30:20)
         #22     LoggingIntegration._onLog (package:sentry_logging/src/logging_integration.dart:66:18)
         #23     _RootZone.runUnaryGuarded (dart:async/zone.dart:1594:10)
         #24     _BufferingStreamSubscription._sendData (dart:async/stream_impl.dart:339:11)
         #25     _BufferingStreamSubscription._add (dart:async/stream_impl.dart:271:7)
         #26     _SyncBroadcastStreamController._sendData.<anonymous closure> (dart:async/broadcast_stream_controller.dart:385:20)
         #27     _BroadcastStreamController._forEachListener (dart:async/broadcast_stream_controller.dart:322:15)
         #28     _SyncBroadcastStreamController._sendData (dart:async/broadcast_stream_controller.dart:384:5)
         #29     _BroadcastStreamController.add (dart:async/broadcast_stream_controller.dart:244:5)
         #30     Logger._publish (package:logging/src/logger.dart:287:51)
         #31     Logger.log (package:logging/src/logger.dart:215:18)
         #32     Logger.info (package:logging/src/logger.dart:255:7)
         #33     appMain (package:maestrounite/main.dart:102:8)
         #34     Sentry._init (package:sentry/src/sentry.dart:149:24)
         <asynchronous suspension>
         #35     Sentry.init (package:sentry/src/sentry.dart:70:5)
         <asynchronous suspension>
         #36     SentryFlutter.init (package:sentry_flutter/src/sentry_flutter.dart:83:5)
         <asynchronous suspension>
         #37     main (package:maestrounite/main.dart:69:3)
         <asynchronous suspension>
```

It seems like the caller of `StandardMessageCodec.writeValue` should call `toString()` before passing the Object?

None

maBarabas:
## :scroll: Description

Fixes unserialisable Breadcrumbs and SentryEvents created from log records containing an `object`, `stacktrace` or `error` of a type not supported by the [`StandardMessageCodec`](https://api.flutter.dev/flutter/services/StandardMessageCodec-class.html).

Fixes #1581 

Fixes #1581 

Run the dart tests in the repo.

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps

I'm not sure if there is any utility in uploading a second copy of the object this way. It should always be identical to the existing `message` field of the breadcrumb, so it is just a duplicate. The only information it adds that the log call looked like `log.info(MyObject())`, rather than `log.info('${MyObject}')` which is irrelevant.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+- Ensure Breadcrumbs and SentryEvents are always serialisable ([#1582](https://github.com/getsentry/sentry-dart/pull/1582))
+
 ## 7.9.0
 
 ### Features
--- a/logging/lib/src/extension.dart
+++ b/logging/lib/src/extension.dart
@@ -12,9 +12,9 @@ extension LogRecordX on LogRecord {
       level: level.toSentryLevel(),
       message: message,
       data: <String, Object>{
-        if (object != null) 'LogRecord.object': object!,
-        if (error != null) 'LogRecord.error': error!,
-        if (stackTrace != null) 'LogRecord.stackTrace': stackTrace!,
+        if (object != null) 'LogRecord.object': object!.toString(),
+        if (error != null) 'LogRecord.error': error!.toString(),
+        if (stackTrace != null) 'LogRecord.stackTrace': stackTrace!.toString(),
         'LogRecord.loggerName': loggerName,
         'LogRecord.sequenceNumber': sequenceNumber,
       },
@@ -30,7 +30,7 @@ extension LogRecordX on LogRecord {
       throwable: error,
       // ignore: deprecated_member_use
       extra: <String, Object>{
-        if (object != null) 'LogRecord.object': object!,
+        if (object != null) 'LogRecord.object': object!.toString(),
         'LogRecord.sequenceNumber': sequenceNumber,
       },
     );
--- a/logging/test/extension_test.dart
+++ b/logging/test/extension_test.dart
@@ -2,6 +2,10 @@ import 'package:logging/logging.dart';
 import 'package:sentry_logging/src/extension.dart';
 import 'package:test/test.dart';
 
+class _TestObject {}
+
+class _TestErrorObject {}
+
 void main() {
   test('breadcrumb time is always utc', () {
     final log = LogRecord(Level.CONFIG, 'foo bar', 'test logger');
@@ -13,5 +17,42 @@ void main() {
     final log = LogRecord(Level.CONFIG, 'foo bar', 'test logger');
 
     expect(log.toEvent().timestamp?.isUtc, true);
+  });
+
+  test('user defined objects are converted to string', () {
+    // Every value in the data and extra map must be convertible to by the
+    // StandardMessageCodec. We convert all user supplied types and the
+    // StackTrace to a string to make this property hold.
+
+    final object = _TestObject();
+    final error = _TestErrorObject();
+    final stackTrace = StackTrace.current;
+
+    final log = LogRecord(
+      Level.CONFIG,
+      'foo bar',
+      'test logger',
+      error,
+      stackTrace,
+      null,
+      object,
+    );
+
+    final breadcrumb = log.toBreadcrumb();
+    expect(
+      breadcrumb.data,
+      containsPair('LogRecord.object', object.toString()),
+    );
+    expect(
+      log.toBreadcrumb().data,
+      containsPair('LogRecord.stackTrace', stackTrace.toString()),
+    );
+
+    final event = log.toEvent();
+    expect(
+      // ignore: deprecated_member_use
+      event.extra,
+      containsPair('LogRecord.object', object.toString()),
+    );
   });
 }


ueman:
Shouldn't https://github.com/getsentry/sentry-dart/blob/d78342459141e1798a213da74ad13e89235cbe31/dart/lib/src/utils.dart#L31 be responsible for that? That way you can ensure in a more generic way that everything correctly encodable to json.

maBarabas:
You mean using this callback in combination with [jsonEncode](https://api.dart.dev/stable/3.0.7/dart-convert/jsonEncode.html)? That does seem like a more elegant and would automatically work for any new map entries that are not encodable.

However, if one of the custom user objects has a toJson function, then it will be encoded. This does not seem desirable, as Sentry does not know how to decode it. I guess it would be possible to copy paste from the web UI and decode it using some dart code. This is also more useful than duplicating the `toString` result.

I can also ignore jsonEncode, and manually call the callback on the objects, but then we lose the automatic fallback for new objects.

maBarabas:
Just had a closer look at the logs in my Sentry instance, and actually the events that cause the original issue are there in the web UI even without this fix (using Sentry 6.22.0). It seems like the only change from this PR is that the errors are not logged any more.

Is there another path in the code that submits breadcrumbs? The native call here is definitely failing, so the breadcrumb should not be present.

ueman:
The Flutter to native sync could be a culprit: https://github.com/getsentry/sentry-dart/blob/d78342459141e1798a213da74ad13e89235cbe31/flutter/lib/src/sentry_native_channel.dart#L59 

That method is executed when a breadcrumb is added, which then directly sends it via method channel to the native side. As the object is not serializable, it fails.

But when the actual error is send, the serialization succeeds because the code path is a different one.

This is just a hunch though.

marandaneto:
See https://github.com/getsentry/sentry-dart/issues/1581#issuecomment-1667682938
Making the whole Map serializable would benefit the other methods too, so instead of fixing only crumbs, I'd rather fix for all the cases mentioned above in the comment.

marandaneto:
On Java there's a base interface called `JsonSerializable/JsonDeserializer` that in case any data class implements this interface, we can call `serialize`/`deserialize`, so users that have control under those classes, can implement the interface and write their own custom serializers/deserializers, but it's not really used a lot so I think it's an overkill.
It's enough to fall back to `toString` when normalizing the Map/Json.

The SDK though should not set any field (automatically) that isn't serializable by default, so I think:

```
        if (object != null) 'LogRecord.object': object!,
        if (error != null) 'LogRecord.error': error!,
        if (stackTrace != null) 'LogRecord.stackTrace': stackTrace!,
```

Those should not be part of the crumb anyway, probably an oversight, not just because of serialization.

marandaneto:
This is going to be addressed by https://github.com/getsentry/sentry-dart/pull/1591 which is a more generic approach.

Issue #1106, opened 2022-11-05
Support Profiling


marandaneto:
### Description

Similar to https://docs.sentry.io/platforms/android/profiling/ but for Dart code

Relates issues on the Dart SDK https://github.com/dart-lang/sdk/issues/3686, https://github.com/dart-lang/sdk/issues/37664, https://github.com/dart-lang/sdk/issues/50055, https://github.com/flutter/flutter/issues/37204

---

Note: 
- iOS/macOS profiling is in Alpha
- Android profiling is not supported yet

marandaneto:
Right now that would be possible if you [init](https://docs.sentry.io/platforms/flutter/native-init/) the SDK manually.

Enable Performance and Profiling directly on the Native SDKs, for example, Android.
[docs.sentry.io/platforms/android/performance](https://docs.sentry.io/platforms/android/performance/)
[docs.sentry.io/platforms/android/profiling](https://docs.sentry.io/platforms/android/profiling/)

The same steps for iOS, would work for Android and iOS native code only, not in the Dart bits nor C/C++ code.

bruno-garcia:
Hey @marandaneto , I was talking to @vaind and he might take a look at this to see how hard/what options do we have.

kahest:
Hey @bruno-garcia @vaind are there any updates on this?

vaind:
@kahest no updates yet, just started looking into this recently

vaind:
[This article by a Dart SDK developer](https://medium.com/dartlang/dart-devtools-analyzing-application-performance-with-the-cpu-profiler-3e94a0ec06ae) gives some intro how profiling is implemented in the Dart VM and exposed via DevTools. TLDR: 
* there's native code in the dart VM responsible for sampling
* this code is exposed via a VM service (websocket)
* **this code is not compiled in a release build (a.k.a "product")**,  see for example `#ifndef PRODUCT` in [profiler.cc](https://dart.googlesource.com/sdk/+/15230e41e5febac958dfa9878fd33f4dc72a2e59/runtime/vm/profiler.cc#577)\
* even if it was available in a release build, the VM service opens a port which would be accessible from any other app on the same device, so it would likely not be feasible to use this in production apps anyway. 

Therefore, this looks like a dead end. 

I'll update here if I can find an alternative solution, e.g. isolate stacks sampling from dart directly.

marandaneto:
[...]

PR #1611, opened 2023-08-22
feat: profiling for iOS/macOS

vaind:
## :scroll: Description

This brings initial support for profiling on iOS and macOS, via native Sentry cocoa SDK profiler.

I've tried extracting a couple of preparation PRs but there's still some refactoring going on here, hope that doesn't make reviewing too hard. Namely, `SentryNative` now acts as an adapter to implementations of `SentryNativeBinding`. The original `SentryNativeChannel` is one implementation, `SentryNativeCocoa` is another one. Also, error handling has moved from `SentryNativeChannel` to `SentryNative` so that it can be shared by all binding implementations.

#1106 

## :pencil: Checklist

- [x] I updated the docs if needed - https://github.com/getsentry/sentry-docs/pull/7880
- [x] All tests passing
- [x] No breaking changes

## TODO in this PR

* [x] integration test (because we can't unit test FFI)
* [x] [fix generated FFI bindings](https://github.com/getsentry/sentry-dart/pull/1648) - currently crashes with the latest binding generated by the pinned ffigen version. I've temporarily replaced the binding in this PR to an older version that worked. The issue seems to be with creating an NSObject where an int should be returned from startProfilerForTrace()

## :crystal_ball: Next steps

* [x] add changelog
* [x] #1613 
* [ ] [better symbolication](https://github.com/getsentry/sentry-dart/issues/1106#issuecomment-1717430776)


bruno-garcia:
> ## iOS Performance metrics 🚀
>  	Plain	With Sentry	Diff
> Startup time	1237.53 ms	1255.72 ms	**18.19 ms**
> Size	8.29 MiB	9.39 MiB	**1.10 MiB**

Got a MB larger in this test even after the linker runs?

buenaflor:
@stefanosiano @krystofwoldrich can you take care of this while I'm gone pls

dart/lib/src/sentry_envelope_item.dart
```diff
@@ -97,6 +98,7 @@ class SentryEnvelopeItem {
 
       final newLine = utf8.encode('\n');
       final data = await dataFactory();
+      // TODO the data copy could be avoided - this would be most significant with attachments.
```
bruno-garcia, line 101: Can we have a follow up to address this? prob a ticket tracking is good for now

flutter/example/lib/main.dart
```diff
@@ -327,7 +328,7 @@ class MainScaffold extends StatelessWidget {
                     status: const SpanStatus.internalError());
 
                 await Future.delayed(const Duration(milliseconds: 50));
-
+                findPrimeNumber(1000000); // xxx remove - just for testing
```
bruno-garcia:
```suggestion
                // findPrimeNumber(1000000); // Uncomment to see it with profiling
```

flutter/ios/Classes/SentryFlutterPluginApple.swift
```diff
@@ -550,6 +561,53 @@ public class SentryFlutterPluginApple: NSObject, FlutterPlugin {
         result("")
       }
     }
+
+    private func startProfiler(_ call: FlutterMethodCall, _ result: @escaping FlutterResult) {
```
bruno-garcia: @armcknight could you help review the bridge profiling into cocoa?

bruno-garcia:
@marandaneto feel like one last review? :)

dart/lib/src/profiling.dart
```diff
@@ -0,0 +1,22 @@
+import 'dart:async';
+
+import 'package:meta/meta.dart';
+
+import '../sentry.dart';
+
+@internal
+abstract class ProfilerFactory {
```
marandaneto:
```suggestion
abstract class SentryProfilerFactory {
```

prefix all classes with `Sentry` to avoid naming clashes

dart/lib/src/sentry_envelope_item.dart
vaind: https://github.com/getsentry/sentry-dart/issues/1666

dart/lib/src/profiling.dart
vaind: done

dart/lib/src/sentry_options.dart
```diff
@@ -289,6 +289,17 @@ class SentryOptions {
   /// to be sent to Sentry.
   TracesSamplerCallback? tracesSampler;
 
+  double? _profilesSampleRate;
```
stefanosiano, line 292: we have a [profilesSampler](https://github.com/getsentry/sentry-cocoa/blob/main/Sources/Sentry/Public/SentryOptions.h#L371) option in the other SDKs, which, if set, is prioritized over profilesSampleRate
I'm fine adding it in another PR, if we decide so

flutter/lib/src/profiling.dart
```diff
@@ -0,0 +1,115 @@
...
+  NativeProfiler? startProfiler(SentryTransactionContext context) {
+    if (context.traceId == SentryId.empty()) {
+      return null;
+    }
+
+    final startTime = _native.startProfiler(context.traceId);
+    if (startTime == null) {
+      return null;
+    }
+    return NativeProfiler(_native, startTime, context.traceId, _clock);
```
stefanosiano: just to be sure, is the `profilesSampleRate` being checked by the native profiler?

stefanosiano (approved):
Left just a couple of comments, but looks fine

flutter/lib/src/profiling.dart
vaind: It's used by the the hub when a transaction is started:

```dart
      SentryProfiler? profiler;
      if (_profilerFactory != null &&
          _tracesSampler.sampleProfiling(samplingDecision)) {
        profiler = _profilerFactory?.startProfiler(transactionContext);
      }
```

```dart
  bool sampleProfiling(SentryTracesSamplingDecision tracesSamplingDecision) {
    double? optionsRate = _options.profilesSampleRate;
    if (optionsRate == null || !tracesSamplingDecision.sampled) {
      return false;
    }
    return _sample(optionsRate);
  }
```

bruno-garcia (approved):
lets merge and release a `beta.0` to start?

PR #1629, opened 2023-09-05
[starfish] add attributes to db spans data

### LINKED ISSUE #1594: [Starfish] Add db attributes to database span's span data
To get an overview of this work across all SDKs, please visit:
https://github.com/orgs/getsentry/projects/135

Parent issue: https://github.com/getsentry/team-sdks/issues/19

For the starfish db module it was determined that we need extra database attributes.

`db.name` should be set if possible, but is not required. `db.system` should match the list in [OpenTelemetry's well known conventions](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/database.md#notes-and-well-known-identifiers-for-dbsystem).

| Attribute  | Type | Description  | Examples  | Requirement Level |
|---|---|---|---|---|
| `db.system` | string | An identifier for the database management system (DBMS) product being used. | `sqlite` | Required |
| `db.name` | string | This attribute is used to report the name of the database being accessed. For commands that switch the database, this should be set to the target database (even if the command fails). | `customers`; `main` | Conditionally Required: If applicable. |

These values should be set on `span.data`.

### Motivation

This is being done for the Sentry's [new performance views for databases](https://sentry.sentry.io/starfish/database/). See a WIP example of the view below:

image

buenaflor:
## :scroll: Description
Adds `db.system` and `db.name` to db spans

Closes #1594 

Unit tests

## :pencil: Checklist

- [ ] All tests passing
- [ ] No breaking changes

### DIFF
--- a/sqflite/lib/src/sentry_database.dart
+++ b/sqflite/lib/src/sentry_database.dart
@@ -43,8 +43,7 @@ class SentryDatabase extends SentryDatabaseExecutor implements Database {
   static const dbNameKey = 'db.name';
   @internal
   // ignore: public_member_api_docs
-  static String? dbName;
-  
+  String? dbName;
 
   /// ```dart
   /// import 'package:sqflite/sqflite.dart';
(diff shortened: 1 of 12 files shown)


sqflite/lib/src/sentry_database.dart
```diff
@@ -31,6 +33,19 @@ class SentryDatabase extends SentryDatabaseExecutor implements Database {
...
+  static const dbSystemKey = 'db.system';
+  @internal
+  // ignore: public_member_api_docs
+  static const dbSystem = 'sqlite';
+  @internal
+  // ignore: public_member_api_docs
+  static const dbNameKey = 'db.name';
+  @internal
+  // ignore: public_member_api_docs
+  static String? dbName;
```
brustolin: h: Is this supposed to be static? Is it not possible to run two instances of SentryDatabase in the same project pointing to different sqlite files?

brustolin (commented):
Most things look good. I have just one question below.

sqflite/lib/src/sentry_database.dart
buenaflor: True, it would be better to couple it to the database instance, it's certainly possible to have multiple instances

sqflite/lib/src/sentry_database.dart
```diff
@@ -66,6 +93,7 @@ class SentryDatabase extends SentryDatabaseExecutor implements Database {
       try {
         await _database.close();
 
+        dbName = null;
```
brustolin: m: I dont think you need to clean the name. We could use it to log if the user try to use a closed databased and an exception is thrown.

brustolin (approved):
LGTM, I have just one suggestion.

sqflite/lib/src/sentry_database.dart
buenaflor: sounds good 👍

PR #1622, opened 2023-09-01
refactor: native FFI for Android and Cocoa SDK

vaind:
## :scroll: Description

This is an evaluation/PoC for integration of the iOS/macOS (a.k.a. Cocoa) SDK as well as Android (Java) SDK via direct FFI invocation, avoiding method channels.

Tracking issue: #1444

Currently, this PR only ads the configuration & code generated to make the necessary change & I've tested a full call roundtrip for both SDKs.

## :pencil: Checklist

- [ ] I reviewed submitted code
- [ ] I added tests to verify changes
- [ ] No new PII added or SDK only sends newly added PII if `sendDefaultPii` is enabled
- [ ] I updated the docs if needed
- [ ] All tests passing
- [ ] No breaking changes

## :crystal_ball: Next steps

Decide whether we want to go ahead and implement this fully - replacing the functionality currently implemented in platform-specific method channels.

Issue #1444, opened 2023-05-14
Use `jnigen` and `ffigen` to call Java/Kotlin/Swift/ObjC natively without the `MethodChannels`


marandaneto:
### Description

https://dart.dev/guides/libraries/java-interop and https://dart.dev/guides/libraries/objective-c-interop
Depends on Dart 3, still experimental.
Can probably use `gen` packages (ffigen) for Linux, Windows, etc.
Need to figure out how to call Web without `MethodChannels`

ueman:
For web you don't need any method channels or similar since Dart has JS interopt built-in: https://dart.dev/web/js-interop

vaind:
1. We may need to do this to enable profiling. Starting a profiler via a method channel is an async operation (even though the actual profiler startup call in objective-c/swift is synchronous). However, `Hub.startTransaction()` is synchronous so we cannot wait for the profiler to be actually started before returning from this function. 
2. I don't see why this would be limited to Dart 3, considering it seems to use standard FFI which has been around forever. Do you remember the source for that SDK version limitation @marandaneto?

ueman:
ffigen for Swift was introduced in Dart 2.18 which shipped with Flutter 3.3, see https://medium.com/flutter/announcing-flutter-3-3-at-flutter-vikings-6f213e068793

Then there's dart:ffi (which is part of the Dart SDK), and package:ffi (distributed via pub). ffigen makes use of a specific version of package:ffi. ffigen is also the tool where most of the Swift binding magic and its improvments happens. 

In addition to that, Swift ffi is considered experimental, however cupertino_http (a package using those Swfit/objc bindings) is considered stable.

vaind:
Looking at the wording in the [Dart 2.18 announcement](https://dart.dev/guides/whats-new#august-30-2022-218-release), it looks like it's just a documentation issue (and the availability of a generator), not that it didn't work with an older version of the SDK. [The changelog for 2.18](https://github.com/dart-lang/sdk/blob/main/CHANGELOG.md#2180---2022-08-30) doesn't mention anything (or I'm blind) so that would indicate my assumption may be true. 
I may be wrong but I think we wouldn't even need to bump the minimum Dart version in pubspec. Anyway, I'll likely give this a quick try, considering the need for profiling I've mentioned above. 

(FYI, I am pretty familiar with FFI: https://github.com/objectbox/objectbox-dart/graphs/contributors - the whole thing is based on a C(++) library)

[...]

PR #1631, opened 2023-09-07
feat/refactor: add cocoa FFI binding

vaind:
## :scroll: Description

Based on experiments in #1622 and as a prerequisite for #1611 - this adds Cocoa SDK FFI bindings generated by ffigen.
None of the bindings are actually used at all so this doesn't have any functional changes yet.

However, the following PRs can make use of FFI now. This is necessary for profiling because it needs to be started synchronously, which MethodChannels don't allow us to do.

#skip-changelog - this change should have no effect on library users at this time

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps

- switch the profiling PR to use FFI so that we get correct samples
- consider switching from method channels to FFI (see #1444)


flutter/scripts/generate-cocoa-bindings.sh
```diff
@@ -0,0 +1,29 @@
...
+
+# Remove dependency on script exit (even in case of an error).
+trap "dart pub remove ffigen" EXIT
+
+# Currently we add the dependency only when the code needs to be generated because it depends
+# on Dart SDK 3.2.0 which isn't available on with Flutter stable yet.
+# Leaving the dependency in pubspec would block all contributors.
+# As for why this is coming from a fork - because we need a specific version of ffigen including PR 607 but not PR 601
+# which starts generating code not compatible with Dart SDK 2.17. The problem is they were merged in the wrong order...
+dart pub add 'dev:ffigen:{"git":{"url":"https://github.com/vaind/ffigen","ref":"4d714b0f2d11e9698a1885ec2d5de15f2ed827f1"}}'
```
bruno-garcia: Should we fork this under `getsentry`? Or use straight up the `dart-lang` one?

```suggestion
dart pub add 'dev:ffigen:{"git":{"url":"https://github.com/getsentry/ffigen","ref":"4d714b0f2d11e9698a1885ec2d5de15f2ed827f1"}}'
```

bruno-garcia (changes requested):

flutter/scripts/generate-cocoa-bindings.sh
bruno-garcia: Create a fork: https://github.com/getsentry/ffigen

@vaind will create a new branch with the sha we need, and we'll point to it from here

bruno-garcia (approved):
Please update the issue once the org in the repo is fixed here: https://github.com/getsentry/ffigen/issues/1

flutter/scripts/generate-cocoa-bindings.sh
vaind: done

.github/workflows/update-deps.yml
```diff
@@ -3,11 +3,11 @@ name: Update Dependencies
...
   # Run every day.
   schedule:
-    - cron: '0 3 * * *'
+    - cron: "0 3 * * *"
   # And on on every PR merge so we get the updated dependencies ASAP, and to make sure the changelog doesn't conflict.
   push:
-    branches:
-      - main
+    # branches:
+    #   - main
```
krystofwoldrich: Are the deps updates disabled for main on purpose? I don't see why would we want that?

vaind: no, just realized this morning I also need to install Flutter SDK during the updater workflow so I have it enabled on all pushes now while working on it.

sorry for the confusion

PR #1651, opened 2023-09-19
ci: don't run CI on markdown updates

vaind:
No need to wait for CI to finish when we just add the missing changelog entry to a PR...


krystofwoldrich (approved):
LGTM 🚀

krystofwoldrich:
I've noticed a few style changes, maybe we should add lint for the workflow files to avoid them in the future.

PR #1667, opened 2023-10-03
ci: generate code coverage in flutter integration test

vaind:
This adds code coverage reporting to integration tests, with hopes of getting reports on coverage of cocoa FFI briding code in 	#1611

Not sure if this actually works - can someone who understands codecov UI check if the change looks as expected?

### DIFF
--- a/.github/workflows/flutter.yml
+++ b/.github/workflows/flutter.yml
@@ -109,6 +109,7 @@ jobs:
         with:
           name: sentry_flutter
           file: ./flutter/coverage/lcov.info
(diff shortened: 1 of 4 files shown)


buenaflor:
https://app.codecov.io/gh/getsentry/sentry-dart/pull/1667/tree/flutter

I think the coverage report will only try to match coverage within the same subproject so in your changes it would try to match coverage within flutter/example which I assume we don't want.

There are also no coverage changes for flutter/lib and dart/lib compared to the base branch so I don't think this works in its current state

vaind:
I agree, let's close this

PR #1701, opened 2023-10-25
chore: log warning if both tracesSampleRate and tracesSampler are set

### LINKED ISSUE #981: Notifiy user of invalid trace rate setting
Performance tracing doesn't work if both the tracesSampler and the traceSampler rate is set. There should be some sort of error when the user does it.

This enlightenment come out of a recent discussion on Discord.

buenaflor:
## :scroll: Description
In Dart if both `tracesSampleRate` and `tracesSampler` are set, `tracesSampler` takes precedence and is used.

This is also [documented here](https://docs.sentry.io/platforms/flutter/performance/?original_referrer=https%3A%2F%2Fsentry.io%2F)

Closes #981

Manual testing

## :pencil: Checklist

- [ ] All tests passing
- [ ] No breaking changes


stefanosiano (changes requested):
from the [traces sampler docs](https://docs.sentry.io/platforms/flutter/performance/?original_referrer=https%3A%2F%2Fsentry.io%2F#configure):
`// return a number between 0 and 1 or null (to fallback to configured value)`
I think it's enough to update the log saying that the sampler will then fallback to the sample rate if it returns null

buenaflor:
@stefanosiano do you mean adding an additional log? I don't think those are mutually exclusive, I think it would be useful to have both logged - in case tracesSampler and traceSampleRate are both set and the other case if tracesSampler returns null

dart/lib/src/sentry_traces_sampler.dart
```diff
@@ -14,7 +14,12 @@ class SentryTracesSampler {
   SentryTracesSampler(
     this._options, {
     Random? random,
-  }) : _random = random ?? Random();
+  }) : _random = random ?? Random() {
+    if (_options.tracesSampler != null && _options.tracesSampleRate != null) {
+      _options.logger(SentryLevel.warning,
+          'Both tracesSampler and sampleRate are set - only tracesSampler will be used.');
```
stefanosiano: @buenaflor i mean changing this log that you just added with something like:
`Both tracesSampler and sampleRate are set. TracesSampler will take precedence and fallback to sampleRate when it returns null.`

buenaflor: I see, makes sense

stefanosiano (approved):

PR #1709, opened 2023-11-02
feat: drift apm

### LINKED ISSUE #1566: Support for `drift` APM/crumbs
### Description

https://github.com/simolus3/drift
https://docs.flutter.dev/packages-and-plugins/happy-paths/recommended#background-processing-

buenaflor:
## :scroll: Description

Adds APM integration for [drift](https://pub.dev/packages/drift).

Closes #1566 

Test suite
Manual testing

## :pencil: Checklist

- [ ] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps
Release and docs


buenaflor:
should be good to go for a review now, the ci package analysis is failing but that's normal for now until we make the first release for the integration

dart/lib/src/sentry_trace_origins.dart
```diff
@@ -18,4 +18,7 @@ class SentryTraceOrigins {
       'auto.db.sqflite.database_executor';
   static const autoDbSqfliteDatabaseFactory =
       'auto.db.sqflite.database_factory';
+  static const autoDbDriftQueryExecutor = 'auto.db.drift.database.executor';
```
denrase: Shouldn't the value be named `auto.db.drift.query.executor`? Like it's below for `transaction.executor`?

dart/lib/src/sentry_tracer.dart
```diff
@@ -134,6 +134,7 @@ class SentryTracer extends ISentrySpan {
       // remove from scope
       await _hub.configureScope((scope) {
         if (scope.span == this) {
+          print('removing span from scope');
```
denrase: Guess this is debug code we can remove?

drift/README.md
```diff
@@ -0,0 +1,73 @@
...
+      options.dsn = dsn;
+      options.tracesSampleRate = 1.0;
+      options.debug = true;
+    },
+    appRunner: runApp, // Init your App.
+  );
+}
+
+Future<void> runApp() async {
+  final tr =
```
denrase, line 43: Should we introduce a helper method to start up the main span around the db? Like we do for `sqflite` with `openDatabaseWithSentry`?

drift/README.md
```diff
@@ -0,0 +1,73 @@
...
+      Sentry.startTransaction('drift', 'op', bindToScope: true);
+  final executor = SentryQueryExecutor(
+    () => NativeDatabase.memory(),
+    databaseName: 'my_db_name',
+  );
+  final db = AppDatabase(executor);
+
+  await db.into(db.todoItems).insert(
+        TodoItemsCompanion.insert(
```
denrase, line 52: This example won't compile out of the box, right? As the definition for `TodoItem` is not part of it. I'm not sure if this is needed though here and would just bloat the code where we just need to show where the Sentry hook is.

denrase: Ignore, it's in a sep. file.

drift/example/example.dart
```diff
@@ -0,0 +1,44 @@
...
+      options.dsn = dsn;
+      options.tracesSampleRate = 1.0;
+      options.debug = true;
+    },
+    appRunner: runApp, // Init your App.
+  );
+}
+
+Future<void> runApp() async {
+  final tr = Sentry.startTransaction('drift', 'op', bindToScope: true);
```
denrase, line 24: Continuing from above. If it makes sense to provide the main span with a helper, we can also set the origin on it, does that make sense?

drift/lib/src/sentry_query_executor.dart
```diff
@@ -0,0 +1,170 @@
...
+  // ignore: public_member_api_docs
+  static const dbOp = 'db';
+
+  @internal
+  // ignore: public_member_api_docs
+  static const dbSystemKey = 'db.system';
+
+  @internal
+  // ignore: public_member_api_docs
+  static const dbSystem = 'sqlite';
```
denrase, line 44: Does drift use `sqlite` under the hood? Otherwise, we had this discussion in the other PR where we named this value `flutter_hive`, does it make sense here to go with the same naming?

drift/lib/src/sentry_span_helper.dart
```diff
@@ -0,0 +1,151 @@
...
+    }
+  }
+
+  /// This span is used for the database transaction.
+  @internal
+  ISentrySpan? transactionSpan;
+
+  /// @nodoc
+  @internal
+  T beginTransaction<T>(
```
denrase, line 74: This method shares most of the code with `asyncWrapInSpan`. Maybe we can do some reuse/unification.

drift/lib/src/sentry_transaction_executor.dart
```diff
@@ -0,0 +1,102 @@
...
+      return await _executor.send();
+    });
+  }
+
+  @override
+  SqlDialect get dialect => _executor.dialect;
+
+  @override
+  Future<bool> ensureOpen(QueryExecutorUser user) {
+    return _executor.ensureOpen(user);
```
denrase: No wrap here as well as the remaining methods in this class?

drift/lib/src/sentry_transaction_executor.dart
```diff
@@ -0,0 +1,102 @@
...
+  );
+
+  final String? _dbName;
+
+  /// @nodoc
+  SentryTransactionExecutor(this._executor, Hub hub, {@internal String? dbName})
+      : _hub = hub,
+        _dbName = dbName {
+    _spanHelper.setHub(_hub);
+    beginTransaction();
```
denrase: Calling the constructor is supposed to call `beginTransaction`? Isn't this behaviour we are changing?

denrase (changes requested):
Looking good, mostly questions/suggestions.

- [x] Should we also add sample code to the `flutter/example` app so we & users can easily test the package and check out the created spans on `sentry.io`?

buenaflor:
@denrase 

>  Should we also add sample code to the flutter/example app so we & users can easily test the package and check out the created spans on sentry.io?

Yes, definitely

dart/lib/src/sentry_tracer.dart
buenaflor: maybe I'm remembering wrong but don't we have anything in place that would flag stuff like this?

drift/lib/src/sentry_query_executor.dart
buenaflor: yup, https://drift.simonbinder.eu/faq/#can-i-view-a-drift-database

since drift itself is not a database but uses sqlite under the hood, I think sqlite should be alright, we also use sqlite for sqflite.

drift/lib/src/sentry_span_helper.dart
buenaflor: tbh, since hive also uses the same approach, it might make sense to unify that altogether, what do you think?

drift/lib/src/sentry_transaction_executor.dart
buenaflor: Hm, yeah it's not ideal. good catch

drift/lib/src/sentry_span_helper.dart
buenaflor: maybe not now but at least as soon as we have merged these

dart/lib/src/sentry_trace_origins.dart
buenaflor: true

drift/README.md
buenaflor: afaik sqflite also doesn't work without a transacstion to connect to. `openDatabaseWithSentry` should only be the open span

drift/lib/src/sentry_transaction_executor.dart
buenaflor: hm, yes there should be, thx

kuhnroyal:
@buenaflor Are you aware of the drift `QueryInterceptor` which will be added in the next drift version? https://654abec513b45231d6041540--moor.netlify.app/docs/examples/tracing/

buenaflor:
@kuhnroyal I've been talking with Simon (maintainer of drift) about this Sentry integration and he was the one who recommended the approach for this implementation and it looks like he's decided to add an interceptor for cases like these after.

buenaflor:
image

Created spans from the flutter example

denrase (approved):
LGTM 👍

.craft.yml
```diff
@@ -20,3 +21,5 @@ targets:
       pub:sentry_dio:
       pub:sentry_file:
       pub:sentry_sqflite:
+      # This needs to be added to the Sentry registry before uncommenting
+      # pub:sentry_drift:
```
denrase, line 25: We probably also need this for the hive PR, right?

buenaflor: Yep, we need to do an initial release first and afterwards add it to the registry so the integration can be retrieved in the docs etc.

PR #1723, opened 2023-11-13
Remove renderer on Dart:io platforms

### LINKED ISSUE #1721: Remove renderer info for now
### Description

As described in #1719 `skia` is currently hardcoded.

We should remove it for now until we find a correct way to get the renderer information

ueman:
## :scroll: Description

Fixes #1721

New tests

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Fixes 
 
+- Flutter renderer information was removed on dart:io platforms since it didn't add the correct value
 - Unsupported types with Expando ([#1690](https://github.com/getsentry/sentry-dart/pull/1690))
 
 ### Features
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Fixes 
 
-- Flutter renderer information was removed on dart:io platforms since it didn't add the correct value
+- Flutter renderer information was removed on dart:io platforms since it didn't add the correct value ([#1723](https://github.com/getsentry/sentry-dart/pull/1723))
 - Unsupported types with Expando ([#1690](https://github.com/getsentry/sentry-dart/pull/1690))
 
 ### Features
--- a/flutter/lib/src/sentry_flutter.dart
+++ b/flutter/lib/src/sentry_flutter.dart
@@ -167,8 +167,7 @@ mixin SentryFlutter {
       integrations.add(LoadImageListIntegration(channel));
     }
     final renderer = options.rendererWrapper.getRenderer();
-    if (renderer == FlutterRenderer.skia ||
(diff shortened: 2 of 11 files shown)


flutter/test/event_processor/flutter_enricher_event_processor_test.dart
```diff
@@ -22,7 +22,38 @@ void main() {
       fixture = Fixture();
     });
 
-    testWidgets('flutter context', (WidgetTester tester) async {
+    testWidgets('flutter context on dart:io', (WidgetTester tester) async {
+      if (kIsWeb) {
+        // widget tests don't support onPlatform config
```
ueman, line 27: Do you have any better idea than this workaround?

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### Fixes 
 
+- Flutter renderer information was removed on dart:io platforms since it didn't add the correct value
```
```suggestion
- Flutter renderer information was removed on dart:io platforms since it didn't add the correct value ([#1723](https://github.com/getsentry/sentry-dart/pull/1723))
```

flutter/lib/src/sentry_flutter.dart
```diff
@@ -167,8 +167,7 @@ mixin SentryFlutter {
       integrations.add(LoadImageListIntegration(channel));
     }
     final renderer = options.rendererWrapper.getRenderer();
-    if (renderer == FlutterRenderer.skia ||
-        renderer == FlutterRenderer.canvasKit) {
+    if (!platformChecker.isWeb || renderer == FlutterRenderer.canvasKit) {
```
ueman, line 170: The HTML renderer is the only renderer where screenshots do not work. Therefore, it should be enough to check for `!web` and the CanvasKit renderer

buenaflor: Do you know if we have it documented somewhere that HTML renderer isn't supported for screenshot? 

Might be good to improve the docs for that as well some time since the docs on sentry.io don't specify that, but well, docs in general are still rough on sentry 😓

flutter/test/event_processor/flutter_enricher_event_processor_test.dart
buenaflor: Hm, don't think so but I'm fine with this as long as it's documented

flutter/lib/src/sentry_flutter.dart
ueman: @denrase might know that, since he wrote it. Though, it's pretty well known that the backing API doesn't work on the HTML renderer

buenaflor (approved):

PR #1749, opened 2023-11-22
deps: update `package_info_plus` constraints

buenaflor:
There's a new `5.0.1` version out which is not part of our constraint.

This also fixes the [package-analysis complaint](https://github.com/getsentry/sentry-dart/runs/18923673712) for flutter

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 
 ### Dependencies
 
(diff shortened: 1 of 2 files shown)


krystofwoldrich (approved):
LG 🚀

krystofwoldrich:
We should also add it to changelog. I think.

Issue #1663, opened 2023-09-28
Obfuscated built Flutter app obfuscates Sentry message


b4tchkn:
### Environment

SaaS (https://sentry.io/)

### Steps to Reproduce

1. Obfuscate Flutter apps with using `--obfuscate --split-debug-info=./` options.
2. Upload symbol files to Sentry using `sentry-cli debug-files upload -o <org> -p <project> ./` on  CI.
3. Send class runtimeType or StackTrace.current() to SentryEvent message in Flutter.
4. Check Issues on Sentry.

### Expected Result

To be confirmed messages that have not been obfuscated.

### Actual Result

The message arrives on Sentry obfuscated.

So, If I don't set stackTrace when call `Sentry.captureEvent()`, StackTrace on Sentry will be unknown. before Flutter obfuscation, I could see a normal StackTrace.
If I set stackTrace to StackTrace.current, etc. when call `Sentry.captureEvent()`, I can confirm a normal StackTrace.

Is it a specification that obfuscation occurs when source code objects are included in the message and that StackTrace is obfuscated when StackTrace is not specified?

### Product Area

Issues

stefanosiano:
hey @b4tchkn Could you please provide us a link to the event?
If you don't want to provide a link on this github issue, you can send it via email to stefano.siano@sentry.io
Also, are you experiencing it on a specific platform, like web?

b4tchkn:
@stefanosiano 
This is occurring on iOS and Android.
And sent an email with a link to the Sentry event. Please check it out.

stefanosiano:
hey @b4tchkn 
I had a look at the event, and it seems like it's related to https://github.com/getsentry/sentry-dart/issues/1565 and https://github.com/getsentry/sentry-dart/issues/2805
I'll have a closer look and come later. As pointed in https://github.com/getsentry/sentry-dart/issues/2805, though, it's quite a big scope

stefanosiano:
hey @b4tchkn 
after merging https://github.com/getsentry/sentry-dart/pull/1756, the stacktraces of the events captured with `captureMessage` or `captureEvent` will be deobfuscated. Note that you either have to put your own stacktrace or use the `await` keyword when calling Sentry APIs to get the calling function in the stacktrace, due to technical limitations.
Regarding deobsfuscation of messages, that's a bigger topic that we are discussing internally.

I'm closing this issue for now. Feel free to add a comment in case the fix doesn't work for your case

PR #1756, opened 2023-11-24
Add debug_meta to all events

### LINKED ISSUE #1744: debug_meta not set for non-exception events
### Description

currently, the `debug_meta` field is set only when there is an exception.
 Our `LoadImageListIntegration` sets the `debug_meta` field [only when there is an exception](https://github.com/getsentry/sentry-dart/blob/main/flutter/lib/src/integrations/load_image_list_integration.dart#L28-L35).

stefanosiano:
## :scroll: Description
`load_image_list_integration` now appends `debug_meta info to all non-transaction events with a stacktrace, instead of checking for Exception existence

Events generated through `captureMessage` or `captureEvent` had obfuscated stacktraces attached to them, when using the `attachStacktrace` option, due to the lack of `debug_meta` info.
Fixes https://github.com/getsentry/sentry-dart/issues/1744 and https://github.com/getsentry/sentry-dart/issues/1663

The tests now use event without exception

## :pencil: Checklist

- [ ] All tests passing
- [ ] No breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@
 ### Fixes
 
 - Add debug_meta to all events ([#1756](https://github.com/getsentry/sentry-dart/pull/1756))
+  - Fixes obfuscated stacktraces when `captureMessage` or `captureEvent` is called with `attachStacktrace` option 
+
 ### Features
 
 - Add option to opt out of fatal level for automatically collected errors ([#1738](https://github.com/getsentry/sentry-dart/pull/1738))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Fixes
+
+- Add debug_meta to all events ([#1756](https://github.com/getsentry/sentry-dart/pull/1756))
```
buenaflor, line 7: Can we add another sub-item with: Fixes obfuscated `stacktraces` occurring on `captureMessage` and captureEvent calls when `attachStacktrace`option is used.  

or similar

just to clarify a little bit more what it fixes

buenaflor (approved):
lgtm 👍 

left one comment regarding the changelog

PR #1799, opened 2024-01-01
Add app flavor to `flutter_context`

ueman:
## :scroll: Description
Starting with Flutter 3.16, you have access to the flavor. Therefore, this information can now be added.

Manually

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/flutter/lib/src/event_processor/flutter_enricher_event_processor.dart
+++ b/flutter/lib/src/event_processor/flutter_enricher_event_processor.dart
@@ -153,6 +153,7 @@ class FlutterEnricherEventProcessor implements EventProcessor {
       // See https://github.com/flutter/flutter/issues/83919
       // 'window_is_visible': _window.viewConfiguration.visible,
       if (renderer != null) 'renderer': renderer,
+      if (_appFlavor != null) 'appFlavor': _appFlavor!,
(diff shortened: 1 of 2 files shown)


flutter/lib/src/event_processor/flutter_enricher_event_processor.dart
```diff
@@ -153,6 +153,7 @@ class FlutterEnricherEventProcessor implements EventProcessor {
       // See https://github.com/flutter/flutter/issues/83919
       // 'window_is_visible': _window.viewConfiguration.visible,
       if (renderer != null) 'renderer': renderer,
+      if (_appFlavor != null) 'appFlavor': _appFlavor!,
```
ueman, line 156: I don't think this can be tested without an integration test with a flavor set up. I'm really not sure whether that's worth the effort though.

buenaflor (approved):

flutter/lib/src/event_processor/flutter_enricher_event_processor.dart
buenaflor: That's fine for me, could also be done separately when someone has the chance

PR #1846, opened 2024-01-25
feat: Add sentry widget that includes other sentry widgets

### LINKED ISSUE #1839: Add a central `SentryWidget`
### Description

Let's explore the possibility to add a central `SentryWidget` that allows to enable/disable features of other widgets such as `SentryScreenshotWidget` etc... so we avoid causing a nested Widget structure for the user.

This also helps in eliminating the need for adding new widgets for cases such as https://github.com/getsentry/sentry-dart/issues/1832 since we could potentially just add it to the `SentryWidget`

buenaflor:
## :scroll: Description

`SentryWidget` serves as a container widget that reduces the amount of widget nesting for the user and makes adding functionalities that depend on widgets much easier

Closes #1839 

Manual
Unit Test

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,6 @@
 ### Features
 
 - Add `SentryWidget` ([#1846](https://github.com/getsentry/sentry-dart/pull/1846))
-  - Deprecates `SentryScreenshotWidget` and `SentryUserInteractionWidget`
 - APM for isar ([#1726](https://github.com/getsentry/sentry-dart/pull/1726))
 - Performance monitoring support for isar ([#1726](https://github.com/getsentry/sentry-dart/pull/1726))
 - Tracing without performance for Dio integration ([#1837](https://github.com/getsentry/sentry-dart/pull/1837))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ### Features
 
+- Add `SentryWidget` ([#1846](https://github.com/getsentry/sentry-dart/pull/1846))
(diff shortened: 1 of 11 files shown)


kuhnroyal:
I would suggest to even go further and make the existing widgets check their appropriate options themselves and then always add them to the widget tree. If the options are disabled, they can just `return child`. The widgets don't hurt but the code is clearer and more maintainable and could be tested individually.

Additionally, returning the widgets from functions could lead to unnecessary rebuilds.
Good summary here: https://www.youtube.com/watch?v=IOyq-eTRhvo

buenaflor:
Good suggestions, thx for the hint

CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
 
 ### Features
 
+- Add `SentryWidget` ([#1846](https://github.com/getsentry/sentry-dart/pull/1846))
+- APM for isar ([#1726](https://github.com/getsentry/sentry-dart/pull/1726))
```
krystofwoldrich, line 9: `- APM for isar ([#1726](https://github.com/getsentry/sentry-dart/pull/1726))` was released in 7.15.0. I'm not sure why it wasn't in the changelog

krystofwoldrich (approved):
Looks good! 🚀

I would also add a small code snippet to the changelog, but I leave that up to you.

- [x] Also we should update the docs

CHANGELOG.md
buenaflor: There was a mistake. We didn't add isar to the craft file so it was never published, I removed it from the 7.15.0 changelog

PR #1882, opened 2024-02-19
feat: ttid/ttfd

buenaflor:
## :scroll: Description

TTID
 - approximation via `addPostFrameCallback` in navigator observer (enabled by default)
 - manual via wrapping child with `SentryDisplayWidget(child: MyWidget())` (manual api)

TTFD
- needs to be finished manually with `SentryFlutter.reportFullyDisplayed()`
- auto finishes after 30seconds with deadline exceeded if not manually finished within timeout

Mobile starfish: introduces TTID and TTFD

## :pencil: Checklist

- [ ] All tests passing
- [ ] No breaking changes


dart/lib/src/sentry_measurement.dart
```diff
@@ -39,6 +39,22 @@ class SentryMeasurement {
...
+        unit = DurationSentryMeasurementUnit.milliSecond;
+
+  /// Duration of the time to full display in milliseconds
+  SentryMeasurement.timeToFullDisplay(Duration duration)
+      : assert(!duration.isNegative),
+        name = 'time_to_full_display',
+        value = duration.inMilliseconds,
+        unit = DurationSentryMeasurementUnit.milliSecond;
+
+  // TODO: might wanna move ttid/ttfd to flutter since we don't have it on pure dart
```
philipphofmann: `m`: I agree 👍

dart/lib/src/sentry_trace_origins.dart
```diff
@@ -27,4 +27,7 @@ class SentryTraceOrigins {
   static const autoDbDriftQueryExecutor = 'auto.db.drift.query.executor';
   static const autoDbDriftTransactionExecutor =
       'auto.db.drift.transaction.executor';
+  static const uiLoad = 'ui.load';
+  static const uiTimeToInitialDisplay = 'ui.load.initial_display';
+  static const uiTimeToFullDisplay = 'ui.load.full_display';
```
philipphofmann: `h`: You need to prefix trace origin with either `manual` or `auto` see https://develop.sentry.dev/sdk/performance/trace-origin/. On Cocoa the `uiTimeToFullDisplay` is prefixed with manual because users have to manually report it. Furthermore, we chose `"auto.ui.time_to_display"` and `"manual.ui.time_to_display"`. Please align this.

dart/lib/src/sentry_tracer.dart
```diff
@@ -109,18 +109,23 @@ class SentryTracer extends ISentrySpan {
...
-          if (_rootEndTimestamp.isAfter(oldestChildEndTimestamp)) {
-            _rootEndTimestamp = oldestChildEndTimestamp;
+        DateTime? latestEndTime;
+
+        for (var child in children) {
+          final childEndTimestamp = child.endTimestamp;
+          if (childEndTimestamp != null) {
+            if (latestEndTime == null || childEndTimestamp.isAfter(latestEndTime)) {
+              latestEndTime = child.endTimestamp;
+            }
```
philipphofmann: `l`: You don't use `childEndTimestamp` here ` latestEndTime = child.endTimestamp`. I think you can ditch the extra variable or use it everywhere
```suggestion
          if (child.endTimestamp != null) {
            if (latestEndTime == null || child.endTimestamp.isAfter(latestEndTime)) {
              latestEndTime = child.endTimestamp;
            }
          }
        }
```

flutter/lib/src/integrations/native_app_start_integration.dart
```diff
@@ -21,18 +26,98 @@ class NativeAppStartIntegration extends Integration<SentryFlutterOptions> {
...
+              DateTime.fromMillisecondsSinceEpoch(
+                  nativeAppStart.appStartTime.toInt()),
+              appStartEnd,
+              measurement,
+            );
+
+            _appStartTracker?.setAppStartInfo(appStartInfo);
+          } else {
+            _appStartTracker?.setAppStartInfo(null);
+          }
```
philipphofmann: `l`: You could turn this else branch into an early return above to reduce the nesting. Generally speaking, this code is pretty nested, and you could apply early return also in the code you didn't touch here to reduce the nesting, but I would do that in an extra PR.

flutter/lib/src/integrations/native_app_start_integration.dart
```diff
@@ -21,18 +26,98 @@ class NativeAppStartIntegration extends Integration<SentryFlutterOptions> {
...
+abstract class IAppStartTracker {
+  AppStartInfo? get appStartInfo;
+
+  void setAppStartInfo(AppStartInfo? appStartInfo);
+
+  void onAppStartComplete(Function(AppStartInfo?) callback);
+}
+
+@internal
+class AppStartTracker extends IAppStartTracker {
```
philipphofmann: `l`: Why did you add the interface `IAppStartTracker`? I'm a fan of only creating interfaces if you will have multiple implementations. Some people refer to this as interface overuse. I call it YAGNI (You Aren't Gonna Need It). Right now it looks like you could ditch the interface to simplify things.

flutter/lib/src/integrations/native_app_start_integration.dart
```diff
@@ -21,18 +26,98 @@ class NativeAppStartIntegration extends Integration<SentryFlutterOptions> {
...
 
     options.sdk.addIntegration('nativeAppStartIntegration');
   }
 }
 
 /// Used to provide scheduler binding at call time.
 typedef SchedulerBindingProvider = SchedulerBinding? Function();
+
+@internal
+class AppStartInfo {
```
philipphofmann: `l`: You could change this to an inner class of `AppStartTracker` as this class only makes sense in combination with the `AppStartTracker`. That's a good use case of an inner class.

flutter/lib/src/navigation/display_strategy_evaluator.dart
```diff
@@ -0,0 +1,63 @@
...
+import 'package:meta/meta.dart';
+
+@internal
+class DisplayStrategyEvaluator {
+  static final DisplayStrategyEvaluator _instance =
+      DisplayStrategyEvaluator._internal();
+
+  factory DisplayStrategyEvaluator() {
+    return _instance;
+  }
```
philipphofmann: `l`: Is this the way to define a singleton in Dart? If yes, why do you define this as a singleton? Singletons are global state and global state is a bummer for testability.

flutter/lib/src/navigation/display_strategy_evaluator.dart
```diff
@@ -0,0 +1,63 @@
...
+
+  DisplayStrategyEvaluator._internal();
+
+  final Map<String, bool> _manualReportReceived = {};
+  final Map<String, Timer> _timers = {};
+  final Map<String, Completer<TimeToDisplayStrategy>> _completers = {};
```
philipphofmann: `m`: If I'm not mistaken, the key for each map is the `route`. I think you could merge these three maps into one. You get two major benefits: you don't have to manage the state of three different maps, which then is easy to get wrong. If you want to remove a route, you have to remove it from all three maps. The second benefit is that you use less memory.

flutter/lib/src/navigation/display_strategy_evaluator.dart
```diff
@@ -0,0 +1,63 @@
...
+      _completers[routeName]?.complete(TimeToDisplayStrategy.manual);
+    }
+
+    // Cancel the timer as it's no longer necessary.
+    _timers[routeName]?.cancel();
+    return wasReportedAlready;
+  }
+}
+
+enum TimeToDisplayStrategy {
```
philipphofmann: `l`: I would put the enum above the `DisplayStrategyEvaluator` class, because it's short and you need to see it to understand what the `DisplayStrategyEvaluator` is doing.

philipphofmann: `l`: I would maybe this to `TimeToDisplayMode` or something similar because the suffix strategy reminds me a bit of the strategy pattern, and at first glance, I would assume that `TimeToDisplayStrategy` is a class that performs actual actions, but instead it stands for the way time to display will be reported.

flutter/lib/src/navigation/display_strategy_evaluator.dart
```diff
@@ -0,0 +1,63 @@
+import 'dart:async';
+
+import 'package:meta/meta.dart';
+
+@internal
+class DisplayStrategyEvaluator {
```
philipphofmann: `l`: This class name should reflect that it deals with TTID. `DisplayStrategyEvaluator` doesn't reveal that. Let's have a chat on that in our 1on1.

philipphofmann (commented):
This review is a mostly nitpick session on how to improve the readability of the code, and not checking the actual implementation for Flutter.

dart/lib/src/sentry_tracer.dart
buenaflor: this is a dart type promotion issue:

```dart
```

buenaflor: this is a dart type promotion issue:

```dart
if (child.endTimestamp != null) {
 // child.endTimestamp is still handled as nullable because type promotion for these cases are only supported in higher dart versions
}

// vs

final childEndTimestamp = child.endTimestamp;
if (childEndTimestamp != null) {
  // childEndTimestamp is not nullable
}
  

```
declaring the variable first explcitlity is a small workaround for that issue.

flutter/lib/src/integrations/native_app_start_integration.dart
buenaflor: unfortunately inner classes are not possible in Dart so not sure if we can meaningfully combine those two

philipphofmann (commented):
This is only a quick pass, will give you a more in depth review tomorrow.

flutter/lib/src/event_processor/native_app_start_event_processor.dart
```diff
@@ -10,36 +12,24 @@ class NativeAppStartEventProcessor implements EventProcessor {
...
-      // We've seen app starts with hours, days and even months.
-      if (measurement.value >= _maxAppStartMillis) {
-        return event;
-      }
+    final appStartInfo = await _appStartTracker?.getAppStartInfo();
+    final measurement = appStartInfo?.measurement;
+    if (!didAddAppStartMeasurement &&
+        measurement != null &&
+        measurement.value.toInt() <= _maxAppStartMillis &&
+        event is SentryTransaction) {
```
philipphofmann: `m`: You could check `didAddAppStartMeasurement` and `event is SentryTransaction` before you get the app start info because they are early exits in this method. Now, every event has to wait for retrieving the app start info.

flutter/lib/src/integrations/app_start/app_start_tracker.dart
```diff
@@ -0,0 +1,41 @@
...
+
+import 'package:meta/meta.dart';
+
+import '../../../sentry_flutter.dart';
+
+@internal
+class AppStartInfo {
+  final DateTime start;
+  final DateTime end;
+  final SentryMeasurement measurement;
```
philipphofmann, line 11: `m`: I think the AppStartInfo shouldn't contain the `SentryMeasurement`. It shouldn't know that it exists. Whoever wants to turn the app start info into a measurement should do that themselves. If that is required in multiple locations, we could have a helper function somewhere converting the `AppStartInfo` into a `SentryMeasurement`.

flutter/lib/src/navigation/time_to_initial_display_tracker.dart
```diff
@@ -0,0 +1,126 @@
...
+    final ttidSpan = transaction.startChild(
+      SentrySpanOperations.uiTimeToInitialDisplay,
+      description: '$routeName initial display',
+      startTimestamp: appStartInfo.start,
+    );
+    ttidSpan.origin = SentryTraceOrigins.autoUiTimeToDisplay;
+
+    transaction.setMeasurement(
+        appStartInfo.measurement.name, appStartInfo.measurement.value,
+        unit: appStartInfo.measurement.unit);
```
philipphofmann: `m`: I'm confused about why we must also set the app start measurement here. If I'm not mistaken, the `NativeAppStartEventProcessor.apply` will overwrite it. I think we should have one central place to set the app start measurement and as the TTID measurement.

flutter/lib/src/event_processor/native_app_start_event_processor.dart
```diff
@@ -10,36 +12,24 @@ class NativeAppStartEventProcessor implements EventProcessor {
...
       event.measurements[measurement.name] = measurement;
+      didAddAppStartMeasurement = true;
```
philipphofmann: `m`: Can the `apply` method be called from multiple isolates? If yes, we could experience race conditions for `didAddAppStartMeasurement` and add the app start measurement to two transactions. I'm unfamiliar with Dart's async programming model, so I ask.

flutter/lib/src/integrations/native_app_start_integration.dart
```diff
@@ -21,14 +27,48 @@ class NativeAppStartIntegration extends Integration<SentryFlutterOptions> {
...
+          // This could be due to many different reasons.
+          // If you do the manual init and init the SDK too late and it does not
+          // compute the app start end in the very first Screen.
+          // If the process starts but the App isn't in the foreground.
+          // If the system forked the process earlier to accelerate the app start.
+          // And some unknown reasons that could not be reproduced.
+          // We've seen app starts with hours, days and even months.
+          if (nativeAppStart == null ||
+              measurement == null ||
+              measurement.value >= 60000) {
```
philipphofmann: `m`: We now check if the app start is longer than 60s twice. Here and in `NativeAppStartEventProcessor.apply`. I would like to avoid that. If you want to check it here, it should use the constant from `NativeAppStartEventProcessor`.

Furthermore, we now have an off-by-one error. This checks implementation checks if the `value >= 60000` and  `NativeAppStartEventProcessor.apply` checks `value <= 60000`. So here we could discard 60000 and only send up to 59999 and in `NativeAppStartEventProcessor.apply` we would send up to 60000. Another reason to only have this check once.

flutter/lib/src/event_processor/native_app_start_event_processor.dart
```diff
@@ -10,36 +12,24 @@ class NativeAppStartEventProcessor implements EventProcessor {
...
-    this._native,
-  );
+  final AppStartTracker? _appStartTracker;
 
-  final SentryNative _native;
+  NativeAppStartEventProcessor({
+    AppStartTracker? appStartTracker,
+  }) : _appStartTracker = appStartTracker ?? AppStartTracker();
+
+  bool didAddAppStartMeasurement = false;
```
philipphofmann: `m`: This seems to be a duplicate state of `SentryNative.didFetchAppStart`. It would be better to only store the state once if we already got the app start or added it as a measurement. Why did we add this flag?

flutter/lib/src/navigation/time_to_display_tracker.dart
```diff
@@ -0,0 +1,92 @@
...
+
+    final transaction = await _ttdTransactionHandler
+        .startTransaction(name, arguments, startTimestamp: appStartInfo.start);
+    if (transaction == null) return;
+
+    await _ttidTracker.trackAppStart(transaction, appStartInfo, name);
+
+    if (_enableTimeToFullDisplayTracing) {
+      _ttfdTracker.startTracking(transaction, appStartInfo.start, name);
+    }
```
philipphofmann: `m`: Why is the `trackAppStart` async? I think it shouldn't be. Now, we wait until the ttidTracker finishes app start tracking until we start the ttfdTracker. I think we should start the ttfdTracker immediately.

flutter/lib/src/navigation/time_to_display_tracker.dart
```diff
@@ -0,0 +1,92 @@
...
+        _enableTimeToFullDisplayTracing = enableTimeToFullDisplayTracing,
+        _ttdTransactionHandler = ttdTransactionHandler,
+        _appStartTracker = appStartTracker ?? AppStartTracker(),
+        _ttfdTracker = ttfdTracker ?? TimeToFullDisplayTracker(),
+        _ttidTracker = ttidTracker ?? TimeToInitialDisplayTracker();
+
+  Future<void> startTracking(String? routeName, Object? arguments) async {
+    final startTimestamp = DateTime.now();
+    if (routeName == '/') {
+      routeName = 'root ("/")';
```
philipphofmann, line 37: `m`: Why do you set the routeName to `'root ("/")'`?

flutter/lib/src/navigation/time_to_display_tracker.dart
```diff
@@ -0,0 +1,92 @@
...
+  /// This method listens for the completion of the app's start process via
+  /// [AppStartTracker], then:
+  /// - Starts a transaction with the app start start timestamp
+  /// - Starts TTID and optionally TTFD spans based on the app start start timestamp
+  /// - Finishes the TTID span immediately with the app start end timestamp
+  ///
+  /// We start and immediately finish the TTID span since we cannot mutate the history of spans.
+  Future<void> _trackAppStartTTD(String? routeName, Object? arguments) async {
+    final appStartInfo = await _appStartTracker.getAppStartInfo();
+    final name = routeName ?? SentryNavigatorObserver.currentRouteName;
```
philipphofmann: `m`: We only call this if `isRootScreen == true` above. I wonder why we need this. `routeName` should be non optional for this method.

flutter/lib/src/navigation/time_to_full_display_tracker.dart
```diff
@@ -0,0 +1,99 @@
...
+@internal
+class TTIDEndTimestampProvider implements EndTimestampProvider {
+  @override
+  DateTime? get endTimestamp => TimeToInitialDisplayTracker().endTimestamp;
+}
+
+@internal
+class TimeToFullDisplayTracker {
+  static final TimeToFullDisplayTracker _instance =
+      TimeToFullDisplayTracker._internal();
```
philipphofmann: `h`: This won't work. We store references to the `_transaction`, `_ttfdSpan`, and the `Duration. When the `SentryNavigatorObserver` pushes a new route before the time fires or the user calls `reportFullyDisplayed`, we run into the following problems:
* the timer gets overwritten and never fires
* the ttfd span doesn't get correctly finished

flutter/lib/src/navigation/time_to_full_display_tracker.dart
```diff
@@ -0,0 +1,99 @@
...
+    }
+    return _instance;
+  }
+
+  TimeToFullDisplayTracker._internal();
+
+  DateTime? _startTimestamp;
+  ISentrySpan? _ttfdSpan;
+  Timer? _ttfdTimer;
+  ISentrySpan? _transaction;
```
philipphofmann, line 44: `h`: If I'm not mistaken, we must set the references to `_ttfdSpan` and `transaction` to `null` in `reportFullyDisplayed` and `handleTimeToFullDisplayTimeout` to avoid memory leaks.

philipphofmann (commented):
This is another pass. As this PR is huge, I can't review everything at once.

flutter/lib/src/integrations/app_start/app_start_tracker.dart
buenaflor: hm yeah good point I agree

buenaflor: I guess one reason why I added it was because then I didn't have to take care of checking whether it's a cold or warm start. so we would need to add something like `isColdStart` boolean to app start info, wdyt?

flutter/lib/src/integrations/native_app_start_integration.dart
buenaflor: yup 👍

flutter/lib/src/navigation/time_to_display_tracker.dart
buenaflor: what about reversing it? 

```
    if (_enableTimeToFullDisplayTracing) {
      _ttfdTracker.startTracking(transaction, appStartInfo.start, name);
    }

    await _ttidTracker.trackAppStart(transaction, appStartInfo, name);
```

the benefit of having an await here would be that it's more consistent in tests so we don't have to manually wait x ms for a ttid or ttfd to finish.

flutter/lib/src/navigation/time_to_display_tracker.dart
buenaflor: this is how the navigation observer has dealt with routes `'/'` before so I didn't want to change that

flutter/lib/src/navigation/time_to_display_tracker.dart
buenaflor: good catch, thx

flutter/lib/src/navigation/time_to_full_display_tracker.dart
buenaflor: hm yes thats true

in `didPush` we also finish the transaction before starting the tracking, so we could also maybe clear the state of the ttfd / ttid trackers before we begin the tracking in general but not sure what the expected behaviour is here.

flutter/lib/src/event_processor/native_app_start_event_processor.dart
buenaflor: our isolate support is very limited so it's not even possible to access native app starts other than the main isolate unless you reinitialize the whole sdk again in the isolate, I don't think this is going to be an issue

flutter/lib/src/event_processor/native_app_start_event_processor.dart
buenaflor: That's because `SentryNative.didFetchAppStart` doesn't tell us 'we've already added this measurement to the transaction`

since the event processor is always run for every event checking `if (didFetchAppStart == true)` and then adding the measurement would mean we are always adding measurement because after the app start it's always true.

flutter/lib/src/frame_callback_handler.dart
```diff
@@ -0,0 +1,14 @@
+import 'package:flutter/scheduler.dart';
+import 'package:flutter/widgets.dart';
+
+abstract class FrameCallbackHandler {
+  void addPostFrameCallback(FrameCallback callback, {String debugLabel});
```
stefanosiano: what's the `debugLabel` for? I didn't find any usage

flutter/lib/src/event_processor/native_app_start_event_processor.dart
stefanosiano:
```suggestion
  bool _didAddAppStartMeasurement = false;
```

flutter/lib/src/integrations/app_start/app_start_tracker.dart
```diff
@@ -0,0 +1,41 @@
...
+
+  AppStartInfo(this.start, this.end, this.measurement);
+}
+
+@internal
+class AppStartTracker {
+  static final AppStartTracker _instance = AppStartTracker._internal();
+  factory AppStartTracker() => _instance;
+  AppStartTracker._internal();
```
stefanosiano:
```suggestion
  AppStartTracker._();
```
Let's be consistent with other classes we already have, like [HubAdapter](https://github.com/getsentry/sentry-dart/blob/b0811ccc76dccb91004b36b5f710163daf733d92/dart/lib/src/hub_adapter.dart#L18) and [NoOp classes](https://github.com/getsentry/sentry-dart/blob/b0811ccc76dccb91004b36b5f710163daf733d92/dart/lib/src/noop_client.dart#L7)
This applies for other places in this PR as well

flutter/lib/src/integrations/app_start/app_start_tracker.dart
```diff
@@ -0,0 +1,41 @@
...
+  AppStartInfo? _appStartInfo;
+
+  void setAppStartInfo(AppStartInfo? appStartInfo) {
+    _appStartInfo = appStartInfo;
+    if (!_appStartCompleter.isCompleted) {
+      _appStartCompleter.complete(appStartInfo);
+    } else {
+      _appStartCompleter = Completer<AppStartInfo?>();
+      _appStartCompleter.complete(appStartInfo);
+    }
```
stefanosiano:
```suggestion
   if (_appStartCompleter.isCompleted) {
      _appStartCompleter = Completer<AppStartInfo?>();
    }
    _appStartCompleter.complete(appStartInfo);
```

flutter/lib/src/frame_callback_handler.dart
buenaflor: I misread the signature of `addPostFrameCallback` it's not needed, we can remove it. thx!

flutter/lib/src/event_processor/native_app_start_event_processor.dart
```diff
@@ -2,57 +2,34 @@ import 'dart:async';
...
-      // If you do the manual init and init the SDK too late and it does not
-      // compute the app start end in the very first Screen.
-      // If the process starts but the App isn't in the foreground.
-      // If the system forked the process earlier to accelerate the app start.
-      // And some unknown reasons that could not be reproduced.
-      // We've seen app starts with hours, days and even months.
-      if (measurement.value >= _maxAppStartMillis) {
-        return event;
-      }
+    if (!_didAddAppStartMeasurement || event is! SentryTransaction) {
```
stefanosiano, line 21: shouldn't this be?
```suggestion
    if (_didAddAppStartMeasurement || event is! SentryTransaction) {
```

flutter/lib/src/integrations/native_app_start_integration.dart
```diff
@@ -21,18 +30,64 @@ class NativeAppStartIntegration extends Integration<SentryFlutterOptions> {
...
+        NativeAppStartEventProcessor(appStartTracker: _appStartTracker));
 
     options.sdk.addIntegration('nativeAppStartIntegration');
   }
 }
 
 /// Used to provide scheduler binding at call time.
 typedef SchedulerBindingProvider = SchedulerBinding? Function();
+
+extension NativeAppStartMeasurement on NativeAppStart {
```
stefanosiano, line 83: since we use this only here, we can avoid creating an extension and just use a private function, like
```
SentryMeasurement? _extractAppStartMeasurement(NativeAppStart? appStart, DateTime appStartEnd) {
    if (appStart == null) {
      return null;
    }
    final appStartDateTime =
    DateTime.fromMillisecondsSinceEpoch(appStart.appStartTime.toInt());
    final duration = appStartEnd.difference(appStartDateTime);

    return appStart.isColdStart
        ? SentryMeasurement.coldAppStart(duration)
        : SentryMeasurement.warmAppStart(duration);
  }
```

flutter/lib/src/navigation/sentry_navigator_observer.dart
```diff
@@ -139,7 +173,6 @@ class SentryNavigatorObserver extends RouteObserver<PageRoute<dynamic>> {
     );
 
     _finishTransaction();
-    _startTransaction(previousRoute);
```
stefanosiano, line 142: i guess it's fine to remove the transaction in didPop, but we should highlight it in the changelog as a "breaking change" (or as a behaviour fix)

flutter/lib/src/navigation/sentry_navigator_observer.dart
```diff
@@ -178,77 +211,14 @@ class SentryNavigatorObserver extends RouteObserver<PageRoute<dynamic>> {
...
-  Future<void> _finishTransaction() async {
-    final transaction = _transaction;
-    _transaction = null;
-    if (transaction == null || transaction.finished) {
-      return;
-    }
-    transaction.status ??= SpanStatus.ok();
-    await transaction.finish();
+    await _timeToDisplayTracker?.startTracking(routeName, arguments);
+    completedDisplayTracking?.complete();
```
stefanosiano, line 221:
```suggestion
    _completedDisplayTracking?.complete();
```
What is this doing, exactly? If it's just for tests, let's add a comment

flutter/lib/src/navigation/time_to_display_transaction_handler.dart
```diff
@@ -0,0 +1,80 @@
...
+      return null;
+    }
+
+    if (routeName == null) {
+      return null;
+    }
+
+    final transactionContext = SentryTransactionContext(
+      routeName,
+      SentrySpanOperations.uiLoad,
```
stefanosiano, line 34: We are changing from `navigation` to `ui.load`
We should highlight it in the changelog, too

buenaflor:
Closing - splitting up this pr in smaller parts

PR #1897, opened 2024-02-28
ci: disable package-analysis for now

buenaflor:
disable package-analysis for now

#skip-changelog


stefanosiano (approved):

krystofwoldrich:
Seems like other users of the package are running into too,
- https://github.com/axel-op/dart-package-analyzer/issues/27

Maybe it's something the package maintainers can/will fix.

PR #1905, opened 2024-03-01
refactor: fetch app start in integration instead of event processor

buenaflor:
## :scroll: Description
Refactor the app start to be fetched in integration and not in event processor. The event processor is only used for attaching the measurement in this case.

This is a prerequisite for TTID/TTFD implementation since TTID requires app start as well

This is also a split pr of #1882 

## :pencil: Checklist

- [ ] All tests passing
- [x] No breaking changes

### DIFF
--- a/flutter/lib/src/frame_callback_handler.dart
+++ b/flutter/lib/src/frame_callback_handler.dart
@@ -7,6 +7,9 @@ abstract class FrameCallbackHandler {
 class DefaultFrameCallbackHandler implements FrameCallbackHandler {
   @override
   void addPostFrameCallback(FrameCallback callback) {
-    SchedulerBinding.instance.addPostFrameCallback(callback);
+    try {
+      /// Flutter >= 2.12 throws if SchedulerBinding.instance isn't initialized.
+      SchedulerBinding.instance.addPostFrameCallback(callback);
+    } catch (_) {}
   }
 }
--- a/flutter/lib/src/frame_callback_handler.dart
+++ b/flutter/lib/src/frame_callback_handler.dart
@@ -1,0 +1,12 @@
+import 'package:flutter/scheduler.dart';
+
+abstract class FrameCallbackHandler {
+  void addPostFrameCallback(FrameCallback callback);
+}
+
+class DefaultFrameCallbackHandler implements FrameCallbackHandler {
+  @override
+  void addPostFrameCallback(FrameCallback callback) {
+    SchedulerBinding.instance.addPostFrameCallback(callback);
+  }
+}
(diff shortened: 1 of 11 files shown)


flutter/lib/src/integrations/native_app_start_integration.dart
```diff
@@ -21,18 +53,74 @@ class NativeAppStartIntegration extends Integration<SentryFlutterOptions> {
...
+
           // ignore: invalid_use_of_internal_member
-          _native.appStartEnd = options.clock();
+          // We only assign the current time if it's not already set - this is useful in tests
+          _native.appStartEnd ??= options.clock();
+          final appStartEnd = _native.appStartEnd;
+          final nativeAppStart = await _native.fetchNativeAppStart();
+
+          if (nativeAppStart == null || appStartEnd == null) {
+            setAppStartInfo(null);
```
stefanosiano: maybe this is not needed, as we are going to use the `_didAddAppStartMeasurement` flag in the `apply` method
If for any reason this integration is called twice, we set the appStartInfo and right after that we remove it

stefanosiano (changes requested):
nice
just fix the build errors and add a test for the appStartInfo being set after the integration runs

flutter/lib/src/native/sentry_native.dart
```diff
@@ -27,6 +27,15 @@ class SentryNative {
   /// Flag indicating if app start was already fetched.
   bool get didFetchAppStart => _didFetchAppStart;
 
+  bool _didAddAppStartMeasurement = false;
```
stefanosiano: you can make this public directly, since we have both get and set

flutter/test/mocks.dart
```diff
@@ -171,6 +171,11 @@ class TestMockSentryNative implements SentryNative {
   @override
   bool get didFetchAppStart => _didFetchAppStart;
 
+  bool _didAddAppStartMeasurement = false;
```
stefanosiano: you can make this public, too

flutter/lib/src/frame_callback_handler.dart
```diff
@@ -0,0 +1,12 @@
+import 'package:flutter/scheduler.dart';
+
+abstract class FrameCallbackHandler {
+  void addPostFrameCallback(FrameCallback callback);
+}
+
+class DefaultFrameCallbackHandler implements FrameCallbackHandler {
+  @override
+  void addPostFrameCallback(FrameCallback callback) {
+    SchedulerBinding.instance.addPostFrameCallback(callback);
```
stefanosiano:
```suggestion
    try {
      /// Flutter >= 2.12 throws if SchedulerBinding.instance isn't initialized.    
      SchedulerBinding.instance.addPostFrameCallback(callback);
    } catch (_) {}
```

flutter/lib/src/integrations/native_app_start_integration.dart
```diff
@@ -13,6 +15,36 @@ class NativeAppStartIntegration extends Integration<SentryFlutterOptions> {
   final SentryNative _native;
   final SchedulerBindingProvider _schedulerBindingProvider;
```
stefanosiano: you can replace this with an instance of `FrameCallbackHandler`, as it will call the `SchedulerBinding` only when needed

stefanosiano (approved):
nice! 👍

PR #1910, opened 2024-03-04
feat: ttid

buenaflor:
## :scroll: Description

Implement TTID 

Split pr of #1882 

CI, manual test with spotlight

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes


flutter/lib/src/sentry_flutter_measurement.dart
```diff
@@ -0,0 +1,23 @@
+import '../sentry_flutter.dart';
+
+extension SentryFlutterMeasurement on SentryMeasurement {
+  /// Duration of the time to initial display in milliseconds
+  static SentryMeasurement timeToInitialDisplay(Duration duration) {
```
stefanosiano: you can create new constructors for both ttid and ttfd in `SentryMeasurement` without the need of extensions
if you want to keep the assertion on the duration you can even create a `factory` constructor (still in `SentryMeasurement`)

buenaflor: I thought it'd be nice to separate it since `SentryMeasurement` is in the `dart` package and ttid/ttfd don't exist in only dart

stefanosiano: we already have frames stuff in `SentryMeasurement`, that are not used in pure dart apps, so it's fine to put everything in one place

flutter/lib/src/navigation/time_to_display_transaction_handler.dart
```diff
@@ -0,0 +1,80 @@
...
+  TimeToDisplayTransactionHandler({
+    required Hub? hub,
+    required bool? enableAutoTransactions,
+    required Duration? autoFinishAfter,
+  })  : _hub = hub ?? HubAdapter(),
+        _enableAutoTransactions = enableAutoTransactions,
+        _autoFinishAfter = autoFinishAfter,
+        _native = SentryFlutter.native;
+
+  Future<ISentrySpan?> startTransaction(String? routeName, Object? arguments,
```
stefanosiano: we may move this method inside `TimeToDisplayTracker` and remove this class, as it's only used there

flutter/lib/src/navigation/sentry_navigator_observer.dart
```diff
@@ -179,77 +214,14 @@ class SentryNavigatorObserver extends RouteObserver<PageRoute<dynamic>> {
...
-  Future<void> _finishTransaction() async {
-    final transaction = _transaction;
-    _transaction = null;
-    if (transaction == null || transaction.finished) {
-      return;
-    }
-    transaction.status ??= SpanStatus.ok();
-    await transaction.finish();
+    await _timeToDisplayTracker?.startTracking(routeName, arguments);
+    completedDisplayTracking?.complete();
```
stefanosiano:
```suggestion
    _completedDisplayTracking?.complete();
```

flutter/lib/src/navigation/time_to_display_tracker.dart
```diff
@@ -0,0 +1,92 @@
...
+  final bool _enableTimeToFullDisplayTracing;
+
+  TimeToDisplayTracker({
+    required bool enableTimeToFullDisplayTracing,
+    required TimeToDisplayTransactionHandler ttdTransactionHandler,
+    TimeToInitialDisplayTracker? ttidTracker,
+  })  : _native = SentryFlutter.native,
+        _enableTimeToFullDisplayTracing = enableTimeToFullDisplayTracing,
+        _ttdTransactionHandler = ttdTransactionHandler,
+        _ttidTracker = ttidTracker ?? TimeToInitialDisplayTracker();
```
stefanosiano: we should always set the `isManual` to false when a route is pushed.
If the user use the `SentryDisplayWidget` only in one route, the others will never complete, as we use a singleton that thinks all routes are manual.
If we do it here in `startTracking`, let's make sure the manual API works as expected

buenaflor: I'm setting `isManual = false` by calling the `clear()` method. this happens always before the tracking starts via `_finishTransaction` https://github.com/getsentry/sentry-dart/blob/4dedf37f828d3ea4f5bf3cf7c4a95683379c2492/flutter/lib/src/navigation/sentry_navigator_observer.dart#L142

flutter/lib/src/navigation/time_to_initial_display_tracker.dart
```diff
@@ -0,0 +1,119 @@
...
+  bool _isManual = false;
+  Completer<DateTime>? _trackingCompleter;
+  DateTime? _endTimestamp;
+
+  /// This endTimestamp is needed in the [TimeToFullDisplayTracker] class
+  @internal
+  DateTime? get endTimestamp => _endTimestamp;
+
+  Future<void> trackRegularRoute(ISentrySpan transaction,
+      DateTime startTimestamp, String routeName) async {
```
stefanosiano: you can add a nullable `endTimestamp` and a nullable `origin` parameters, null on regular routes, and have the `trackAppStart` just call this method, as they are nearly identical.
Or you can create a new private method called by both

flutter/lib/src/navigation/time_to_display_tracker.dart
stefanosiano: yeah, just noticed it, forget about it 😄

CHANGELOG.md
```diff
@@ -4,6 +4,16 @@
 
 ### Features
 
+- Add TTID, allows you to measure the time it takes to render the first frame of your screen ([#1910](https://github.com/getsentry/sentry-dart/pull/1910))
```
philipphofmann: `l`:
```suggestion
- Add TTID (time to initial display), which allows you to measure the time it takes to render the first frame of your screen ([#1910](https://github.com/getsentry/sentry-dart/pull/1910))
```

CHANGELOG.md
```diff
@@ -4,6 +4,16 @@
...
+  - Requires using the [routing instrumentation](https://docs.sentry.io/platforms/flutter/integrations/routing-instrumentation/).
+  - Introduces two modes: `automatic` and `manual`.
```
philipphofmann: `l`:
```suggestion
  - Introduces two modes: 
```

dart/lib/src/sentry_measurement.dart
```diff
@@ -39,6 +39,20 @@ class SentryMeasurement {
...
+        name = 'time_to_initial_display',
+        value = duration.inMilliseconds,
+        unit = DurationSentryMeasurementUnit.milliSecond;
+
+  /// Duration of the time to full display in milliseconds
+  SentryMeasurement.timeToFullDisplay(Duration duration)
+      : assert(!duration.isNegative),
+        name = 'time_to_full_display',
+        value = duration.inMilliseconds,
+        unit = DurationSentryMeasurementUnit.milliSecond;
```
philipphofmann: `l`: Isn't this PR only about TTID?
```suggestion
```

dart/lib/src/sentry_tracer.dart
```diff
@@ -109,18 +109,24 @@ class SentryTracer extends ISentrySpan {
...
-            _rootEndTimestamp = oldestChildEndTimestamp;
+        DateTime? latestEndTime;
+
+        for (var child in children) {
+          final childEndTimestamp = child.endTimestamp;
+          if (childEndTimestamp != null) {
+            if (latestEndTime == null ||
+                childEndTimestamp.isAfter(latestEndTime)) {
+              latestEndTime = child.endTimestamp;
+            }
```
philipphofmann: `l`: Out of curiosity, why did you change this?

flutter/lib/src/navigation/sentry_display_widget.dart
```diff
@@ -0,0 +1,60 @@
...
+/// ```
+///
+/// Make sure to configure Sentry's routing instrumentation in your app by following
+/// the guidelines provided in Sentry's documentation for Flutter integrations:
+/// https://docs.sentry.io/platforms/flutter/integrations/routing-instrumentation/
+///
+/// See also:
+/// - [Sentry's documentation on Flutter integrations](https://docs.sentry.io/platforms/flutter/)
+///   for more information on how to integrate Sentry into your Flutter application.
+class SentryDisplayWidget extends StatefulWidget {
```
philipphofmann, line 30: `m`: This small feature could be in an extra PR. Can you split it if it's not too much work?

flutter/lib/src/navigation/sentry_navigator_observer.dart
```diff
@@ -44,57 +52,66 @@ typedef AdditionalInfoExtractor = Map<String, dynamic>? Function(
...
         _setRouteNameAsTransaction = setRouteNameAsTransaction,
         _routeNameExtractor = routeNameExtractor,
-        _additionalInfoProvider = additionalInfoProvider,
-        _native = SentryFlutter.native {
+        _additionalInfoProvider = additionalInfoProvider {
     if (enableAutoTransactions) {
-      // ignore: invalid_use_of_internal_member
       _hub.options.sdk.addIntegration('UINavigationTracing');
     }
+    _timeToDisplayTracker = timeToDisplayTracker ??
```
philipphofmann: `h`: I don't fully understand how the autoFinishAfter now works if you pass in a `timeToDisplayTracker`. How can we ensure that the `TimeToDisplayTracker` passed here uses the correct parameters for `autoFinishAfter` and `enableAutoTransactions`?

flutter/lib/src/navigation/time_to_display_tracker.dart
```diff
@@ -0,0 +1,153 @@
...
+    }
+
+    final transactionContext = SentryTransactionContext(
+      routeName,
+      SentrySpanOperations.uiLoad,
+      transactionNameSource: SentryTransactionNameSource.component,
+      origin: SentryTraceOrigins.autoNavigationRouteObserver,
+    );
+
+    final transaction = _hub?.startTransactionWithContext(
```
philipphofmann: `m`: It seems a bit weird to me that the time to display tracker now starts the transaction. I think we should leave this responsibility to the navigation observer. Why did you choose this approach?

philipphofmann (commented):
First pass.

flutter/lib/src/navigation/time_to_initial_display_tracker.dart
```diff
@@ -0,0 +1,135 @@
...
+            ? SentryTraceOrigins.manualUiTimeToDisplay
+            : SentryTraceOrigins.autoUiTimeToDisplay);
+
+    final duration = Duration(
+        milliseconds: _endTimestamp.difference(startTimestamp).inMilliseconds);
+    final ttidMeasurement = SentryMeasurement.timeToInitialDisplay(duration);
+
+    transaction.setMeasurement(ttidMeasurement.name, ttidMeasurement.value,
+        unit: ttidMeasurement.unit);
+    await ttidSpan.finish(endTimestamp: _endTimestamp);
```
stefanosiano, line 90: why did you remove the `clear()` call?

buenaflor: since we call it in `_finishTransaction` anyway in the navigation observer we are calling it twice

dart/lib/src/sentry_tracer.dart
buenaflor: this is basically a bug fix, will create another pr for this, this is only relevant for ttfd anyway

flutter/lib/src/navigation/time_to_initial_display_tracker.dart
buenaflor: unless you think we should be more defensive here and clear regardless?

stefanosiano: I'd be for clearing here, so the logic behind ttid stays in the ttid tracker and not spreadto other classes

buenaflor: should we maybe clear at the start and at the end? edge case, because it's also possible that users route to another screen before ttid has finished

buenaflor:
~don't review yet, still need to fix the start timestamps~

ready for review

flutter/lib/src/navigation/sentry_navigator_observer.dart
```diff
@@ -242,14 +274,41 @@ class SentryNavigatorObserver extends RouteObserver<PageRoute<dynamic>> {
     await _native?.beginNativeFramesCollection();
   }
 
-  Future<void> _finishTransaction() async {
+  Future<void> _startTimeToDisplayTracking(Route<dynamic>? route) async {
```
philipphofmann, line 268: `l`: You could rename this to `_startTransaction` and merge the current content of `_startTransaction` into this function. Now in `didPush` the code reads like this:
```
_finishTransaction();
_startTimeToDisplayTracking(route);
```

Which is a bit confusing because you finish a transaction and you don't start a new one.

Also you could add an early return in here when `!_enableAutoTransactions` what `_startTransaction` already does.

flutter/lib/src/navigation/time_to_display_tracker.dart
```diff
@@ -0,0 +1,34 @@
...
+
+import 'dart:async';
+
+import 'package:meta/meta.dart';
+
+import '../../sentry_flutter.dart';
+import 'time_to_initial_display_tracker.dart';
+
+@internal
+class TimeToDisplayTracker {
```
philipphofmann, line 11: `l`: This class is a bit obsolete right now, as it is a wrapper around `TimeToInitialDisplayTracker`.

flutter/lib/src/navigation/time_to_initial_display_tracker.dart
```diff
@@ -0,0 +1,132 @@
...
+
+    final tracer = transaction as SentryTracer;
+
+    final ttidSpan = transaction.startChild(
+      SentrySpanOperations.uiTimeToInitialDisplay,
+      description: '${tracer.name} initial display',
+      startTimestamp: startTimestamp,
+    );
+
+    // Set the origin based on provided value or determine based on _isManual
```
philipphofmann: `l`: You can ditch this comment. For me, the code is self explaining.

flutter/lib/src/navigation/time_to_initial_display_tracker.dart
```diff
@@ -0,0 +1,132 @@
...
+  }
+
+  Future<void> _trackTimeToInitialDisplay({
+    required ISentrySpan transaction,
+    required DateTime startTimestamp,
+    DateTime? endTimestamp,
+    String? origin,
+  }) async {
+    // Determine endTimestamp if not provided
+    final _endTimestamp = endTimestamp ?? await determineEndTime();
```
philipphofmann, line 68: `h`: Shouldn't we add some timeout for this? Otherwise, this could last forever.

flutter/lib/src/navigation/sentry_display_widget.dart
```diff
@@ -0,0 +1,60 @@
...
+
+  @override
+  _SentryDisplayWidgetState createState() => _SentryDisplayWidgetState();
+}
+
+class _SentryDisplayWidgetState extends State<SentryDisplayWidget> {
+  @override
+  void initState() {
+    super.initState();
+    TimeToInitialDisplayTracker().markAsManual();
```
philipphofmann, line 49: `h`: So when using this widget once, everything is in manual mode. I think instead it would be good to find a way to be able to use both combined.

flutter/test/navigation/sentry_display_widget_test.dart
```diff
@@ -0,0 +1,115 @@
...
+    final ttidSpan = spans.first;
+    expect(ttidSpan.context.operation,
+        SentrySpanOperations.uiTimeToInitialDisplay);
+    expect(ttidSpan.finished, isTrue);
+    expect(ttidSpan.context.description, 'Current Route initial display');
+    expect(ttidSpan.origin, SentryTraceOrigins.manualUiTimeToDisplay);
+    expect(tracer.measurements, hasLength(1));
+    final measurement = tracer.measurements['time_to_initial_display'];
+    expect(measurement, isNotNull);
+    expect(measurement?.unit, DurationSentryMeasurementUnit.milliSecond);
```
philipphofmann, line 53: `m`: Shouldn't we also add an assert for the value of the measurement?

flutter/test/navigation/sentry_display_widget_test.dart
```diff
@@ -0,0 +1,115 @@
...
+        element.context.operation ==
+        SentrySpanOperations.uiTimeToInitialDisplay);
+
+    expect(spans, hasLength(1));
+    final ttidSpan = spans.first;
+    expect(ttidSpan.context.operation,
+        SentrySpanOperations.uiTimeToInitialDisplay);
+    expect(ttidSpan.finished, isTrue);
+    expect(ttidSpan.context.description, 'Current Route initial display');
+    expect(ttidSpan.origin, SentryTraceOrigins.manualUiTimeToDisplay);
```
philipphofmann, line 46: `m`: I think adding asserts for the duration of the ttidSpan would improve this test.

flutter/test/navigation/time_to_initial_display_tracker_test.dart
```diff
@@ -0,0 +1,173 @@
...
+
+      final children = transaction.children;
+      expect(children, hasLength(1));
+
+      final ttidSpan = children.first;
+      expect(ttidSpan.context.operation,
+          SentrySpanOperations.uiTimeToInitialDisplay);
+      expect(ttidSpan.finished, isTrue);
+      expect(ttidSpan.context.description, 'root ("/") initial display');
+      expect(ttidSpan.origin, SentryTraceOrigins.autoUiTimeToDisplay);
```
philipphofmann, line 44: `m`: Again I think we should assert the duration of the span.

philipphofmann (commented):
As already discussed, my Flutter knowledge is still very limited, and therefore, I'm not confident enough to approve this PR. I added some comments.

flutter/lib/src/navigation/sentry_display_widget.dart
buenaflor: You can use both combined. You just don't wrap the screens you don't want to manually instrument

flutter/test/navigation/sentry_display_widget_test.dart
buenaflor: It's kind of hard to reliably do it in this test

```
      fixture.navigatorObserver.didPush(currentRoute, null);
      await tester.pumpWidget(fixture.getSut());
```

we don't know how long `pumpWidget` is going to take until it finishes, so it's really in the hands of the testing framework

I guess we could add a threshold? something like if it takes longer than 300ms the test fails

flutter/test/navigation/sentry_display_widget_test.dart
buenaflor: Same comment as above

flutter/lib/src/navigation/time_to_display_tracker.dart
buenaflor: TTFD is also going to be added to this class, not sure if I should refactor it now

flutter/lib/src/navigation/sentry_navigator_observer.dart
buenaflor: I think I'd rather rename `_finishTransaction`, `_startTransaction` is already a rather big function, combining all of them would make it a pretty big function

flutter/lib/src/navigation/sentry_display_widget.dart
philipphofmann: Then ignore me. Exellent 💯

flutter/test/navigation/sentry_display_widget_test.dart
philipphofmann: The value should match the value of the span. Can you use that?

flutter/lib/src/navigation/time_to_display_tracker.dart
philipphofmann: Then keep it. We can also ditch it later.

flutter/lib/src/navigation/sentry_navigator_observer.dart
philipphofmann: I don't mind long functions if they clearly communicate what's going on. If the logic is simple and long, I don't mind having it in one function. Up to you.

flutter/test/navigation/sentry_display_widget_test.dart
buenaflor: Added 👍

flutter/lib/src/navigation/sentry_navigator_observer.dart
```diff
@@ -88,13 +97,25 @@ class SentryNavigatorObserver extends RouteObserver<PageRoute<dynamic>> {
   final AdditionalInfoExtractor? _additionalInfoProvider;
   final SentryNative? _native;
 
+  static TimeToDisplayTracker? _timeToDisplayTracker;
+
+  @internal
+  static TimeToDisplayTracker? get timeToDisplayTracker =>
```
stefanosiano: this getter is not used anywhere, so we can remove it

flutter/lib/src/navigation/sentry_navigator_observer.dart
```diff
@@ -88,13 +97,25 @@ class SentryNavigatorObserver extends RouteObserver<PageRoute<dynamic>> {
   final AdditionalInfoExtractor? _additionalInfoProvider;
   final SentryNative? _native;
 
+  static TimeToDisplayTracker? _timeToDisplayTracker;
```
stefanosiano:
```suggestion
  late final TimeToDisplayTracker? _timeToDisplayTracker;
```

buenaflor: refactored it to be initialized in the constructor's initializer list directly so we don't need `late`

flutter/test/navigation/time_to_display_tracker_test.dart
```diff
@@ -0,0 +1,173 @@
...
+
+  group('time to initial display', () {
+    group('in root screen app start route', () {
+      test('matches startTimestamp of transaction', () async {
+        final sut = fixture.getSut();
+
+        await sut.trackRegularRouteTTD(fixture.getTransaction(name: '/'),
+            startTimestamp: fixture.startTimestamp);
+
+        final transaction = fixture.hub.getSpan() as SentryTracer;
```
stefanosiano:
```suggestion
        final transaction = fixture.getTransaction(name: '/') as SentryTracer;
        await sut.trackRegularRouteTTD(transaction, startTimestamp: fixture.startTimestamp);
```
this way it's easier to understand what starts a transaction and what finishes it
apply to other tests as well

flutter/test/navigation/time_to_display_tracker_test.dart
```diff
@@ -0,0 +1,173 @@
...
+                element.context.operation ==
+                SentrySpanOperations.uiTimeToInitialDisplay)
+            .first;
+
+        expect(transaction, isNotNull);
+        expect(transaction.context.operation, SentrySpanOperations.uiLoad);
+        expect(transaction.startTimestamp, ttidSpan.startTimestamp);
+      });
+
+      test('startMeasurement finishes ttid span', () async {
```
stefanosiano:
```suggestion
      test('trackAppStartTTD finishes ttid span', () async {
```
guess that's what the test is for, right?
apply to other tests as well

flutter/test/navigation/time_to_display_tracker_test.dart
```diff
@@ -0,0 +1,173 @@
...
+      });
+
+      group('with manual strategy', () {
+        test('finishes ttid span after reporting with manual api', () async {
+          final sut = fixture.getSut();
+
+          Future.delayed(const Duration(milliseconds: 100), () {
+            fixture.ttidTracker.markAsManual();
+            fixture.ttidTracker.completeTracking();
+          });
```
stefanosiano:
```suggestion
          fixture.ttidTracker.markAsManual();
          Future.delayed(const Duration(milliseconds: 1), () {
            fixture.ttidTracker.completeTracking();
          });
```
all other delays in all tests of this class seems useless, as they pass even without delays

flutter/test/navigation/time_to_display_tracker_test.dart
```diff
@@ -0,0 +1,173 @@
...
+                  element.context.operation ==
+                  SentrySpanOperations.uiTimeToInitialDisplay)
+              .first;
+          expect(ttidSpan, isNotNull);
+
+          await Future.delayed(const Duration(milliseconds: 100));
+
+          expect(ttidSpan.finished, isTrue);
+        });
+      });
```
stefanosiano, line 147: You can add another test for the timeout
Also, we can reduce the default timeout of `FakeFrameCallbackHandler` from 500 to 50 ms, to speed up tests, if it doesn't create issues

flutter/lib/src/navigation/time_to_initial_display_tracker.dart
```diff
@@ -0,0 +1,135 @@
...
+        unit: ttidMeasurement.unit);
+    await ttidSpan.finish(endTimestamp: _endTimestamp);
+  }
+
+  Future<DateTime?>? determineEndTime() {
+    _trackingCompleter = Completer<DateTime?>();
+
+    // If we already know it's manual we can return the future immediately
+    if (_isManual) {
+      return _trackingCompleter?.future;
```
stefanosiano: we can put a timeout here, too

flutter/test/navigation/time_to_display_tracker_test.dart
buenaflor: oh yeah my bad, the delays are not needed anymore since await already waits for everything to finish, wasn't the case before

stefanosiano (approved):
looks good to me! @krystofwoldrich wanna have a look, too?

stefanosiano:
what about merging `SentryDisplayWidget` into `SentryWidget`?
So that if `SentryWidget` is used in a screen (it shouldn't matter where in the hierarchy) we can collect the TTID on "manual" mode.
This is just a question to open a discussion, we would do it in another PR

buenaflor:
`SentryWidget` is meant to be used only once at the root (need to document that properly) so I don't think it's viable currently to merge them.

Gotta check what happens if we use `SentryWidget` multiple times.

But we can explore that option for sure

krystofwoldrich:
Looks good! 🚀 Very exciting, looking forward to TTFD.

PR #1920, opened 2024-03-06
feat: ttfd

buenaflor:
## :scroll: Description

Introduces time to full display

opt-in via `enableTimeToFullDisplayTracing`

Starfish

CI, manual

## :pencil: Checklist

- [ ] All tests passing
- [x] No breaking changes


CHANGELOG.md
```diff
@@ -1,25 +1,33 @@
...
-
-### Features
+## Unreleased
 
+- Add TTFD (time to full display), which allows you to measure the time it takes to render the full screen ([#1920](https://github.com/getsentry/sentry-dart/pull/1920))
+  - Requires using the [routing instrumentation](https://docs.sentry.io/platforms/flutter/integrations/routing-instrumentation/).
+  - Set `enableTimeToFullDisplayTracing = true` in your `SentryFlutterOptions` to enable TTFD
+  - Manually report the end of the full display by calling `SentryFlutter.reportFullyDisplayed()`
+  - If not reported within 30 seconds, the span will be automatically finish with the status `deadline_exceeded`
 - Add TTID (time to initial display), which allows you to measure the time it takes to render the first frame of your screen ([#1910](https://github.com/getsentry/sentry-dart/pull/1910))
```
stefanosiano, line 12: Is TTID not released, yet?

flutter/lib/src/navigation/sentry_navigator_observer.dart
```diff
@@ -82,11 +85,23 @@ class SentryNavigatorObserver extends RouteObserver<PageRoute<dynamic>> {
...
     if (enableAutoTransactions) {
       _hub.options.sdk.addIntegration('UINavigationTracing');
     }
+    _timeToDisplayTracker =
+        timeToDisplayTracker ?? _initializeTimeToDisplayTracker();
+  }
+
+  /// Initializes the TimeToDisplayTracker with the option to enable time to full display tracing.
+  TimeToDisplayTracker _initializeTimeToDisplayTracker(
+      {bool enableTimeToFullDisplayTracing = false}) {
```
stefanosiano: why is it a param? we don't pass it anywhere

flutter/lib/src/navigation/sentry_navigator_observer.dart
```diff
@@ -96,7 +111,11 @@ class SentryNavigatorObserver extends RouteObserver<PageRoute<dynamic>> {
   final RouteNameExtractor? _routeNameExtractor;
   final AdditionalInfoExtractor? _additionalInfoProvider;
   final SentryNative? _native;
-  final TimeToDisplayTracker? _timeToDisplayTracker;
+  late final TimeToDisplayTracker? _timeToDisplayTracker;
+
+  @internal
+  static TimeToDisplayTracker? get timeToDisplayTracker =>
+      SentryNavigatorObserver()._timeToDisplayTracker;
```
stefanosiano:
```suggestion
  static TimeToDisplayTracker? _timeToDisplayTracker;

  @internal
  static TimeToDisplayTracker? get timeToDisplayTracker => _timeToDisplayTracker;
```
This way we avoid creating an instance of the observer just to access the tracker

flutter/lib/src/navigation/sentry_navigator_observer.dart
```diff
@@ -254,55 +272,89 @@ class SentryNavigatorObserver extends RouteObserver<PageRoute<dynamic>> {
   }
 
   Future<void> _finishTimeToDisplayTracking() async {
-    _timeToDisplayTracker?.clear();
+    try {
+      final transaction = _transaction;
+      _transaction = null;
+      if (transaction == null || transaction.finished) {
+        return;
```
stefanosiano, line 288: we are returning too early imho
if a transaction is finished because it was in manual mode, we don't clear the tracker, and it will expect the next transaction to be manual, right?

flutter/lib/src/navigation/sentry_navigator_observer.dart
```diff
@@ -254,55 +272,89 @@ class SentryNavigatorObserver extends RouteObserver<PageRoute<dynamic>> {
...
+      }
 
-    final transaction = _transaction;
-    _transaction = null;
-    if (transaction == null || transaction.finished) {
-      return;
+      // Cancel unfinished child spans, e.g this might happen if the user navigates
+      // away from the current route before TTFD or TTID is finished.
+      for (final child in (transaction as SentryTracer).children) {
```
stefanosiano, line 293: we can't just finish all spans of the transaction, as we would finish the spans started by the user, too.
we should just finish TTID and TTFD (with status deadline_exceeded).
Also, we can remove the transaction from the scope and let it finish in the background. wdyt?

flutter/lib/src/navigation/sentry_navigator_observer.dart
```diff
@@ -254,55 +272,89 @@ class SentryNavigatorObserver extends RouteObserver<PageRoute<dynamic>> {
...
-  Future<void> _startTimeToDisplayTracking(Route<dynamic>? route) async {
-    if (!_enableAutoTransactions) {
-      return;
-    }
+  Future<void> _finishThenStartTimeToDisplayTracking(
+      Route<dynamic>? route) async {
+    try {
+      // We can await inside this function so we can make sure the previous
+      // tracking is finished before starting a new one.
+      await _finishTimeToDisplayTracking();
```
stefanosiano: if we remove the previous transaction from the scope, we don't need to await for finishing it

CHANGELOG.md
buenaflor: craft messed up the merging. I merged the ttid pr during the release of 7.17.0 so craft merged the changelog together in a wrong way when the release finished - but it's not included in the release changelog so all good

buenaflor: we can release it together with ttfd on 7.18.0 this week

flutter/lib/src/navigation/sentry_navigator_observer.dart
buenaflor: it's wrapped in a try / finally with `_clear()` and when we return the finally block is executed as well

flutter/lib/src/navigation/sentry_navigator_observer.dart
buenaflor: > we should just finish TTID and TTFD (with status deadline_exceeded)

👍 

> we can remove the transaction from the scope and let it finish in the background. wdyt?

I think that's fine, probably is better than forcefully finishing it

buenaflor: what if we remove it from the scope and still finish it without awaiting?

stefanosiano: > what if we remove it from the scope and still finish it without awaiting?

yep, that's the idea

flutter/lib/src/navigation/sentry_navigator_observer.dart
buenaflor: sry now I know what you mean, yes I'll fix that

buenaflor: clear now is only called if `clearAfter` is true (in `didPop`) and additionally it is also cleared at the very start of `didPush` so we can make sure we start from a clean state. It's safe to call because didPush happens before the Widget is being built so we wouldn't override any manual flags

flutter/lib/src/navigation/sentry_navigator_observer.dart
```diff
@@ -265,30 +268,35 @@ class SentryNavigatorObserver extends RouteObserver<PageRoute<dynamic>> {
     }
 
     await _hub.configureScope((scope) {
-      scope.span ??= _transaction;
+      scope.span = _transaction;
```
stefanosiano: i guess this was done to preserve a transaction started by the user and bound to the scope
```suggestion
      scope.span ??= _transaction;
```

flutter/lib/src/navigation/sentry_navigator_observer.dart
```diff
@@ -265,30 +268,35 @@ class SentryNavigatorObserver extends RouteObserver<PageRoute<dynamic>> {
...
 
-  Future<void> _finishTimeToDisplayTracking() async {
+  Future<void> _finishTimeToDisplayTracking({bool clearAfter = false}) async {
+    final transaction = _transaction;
+    _transaction = null;
     try {
-      final transaction = _transaction;
-      _transaction = null;
+      _hub.configureScope((scope) {
+        scope.span = null;
```
stefanosiano:
```suggestion
        if (scope.span == transaction) {
            scope.span = null;
        }
```

flutter/lib/src/navigation/time_to_full_display_tracker.dart
```diff
@@ -0,0 +1,117 @@
...
+@internal
+abstract class EndTimestampProvider {
+  DateTime? get endTimestamp;
+}
+
+@internal
+class TTIDEndTimestampProvider implements EndTimestampProvider {
+  @override
+  DateTime? get endTimestamp => TimeToInitialDisplayTracker().endTimestamp;
+}
```
stefanosiano: What about removing unneeded classes?
```suggestion
@internal
typedef EndTimestampProvider = DateTime? Function();

@internal
EndTimestampProvider ttidEndTimestampProvider() => () => TimeToInitialDisplayTracker().endTimestamp;

```

flutter/test/sentry_navigator_observer_test.dart
```diff
@@ -259,46 +263,40 @@ void main() {
       });
     });
 
-    test('do not bind to scope if already set', () {
```
stefanosiano: we can re-add this test, right?

stefanosiano (commented):
I's pretty good
Just few minor details

flutter/test/sentry_navigator_observer_test.dart
buenaflor: yea, now that we only remove the scope span if it's the transaction from navigator observer this test is again useful

flutter/lib/src/navigation/time_to_full_display_tracker.dart
buenaflor: Yeah looks good, didn't think of that 👍

stefanosiano (approved):
LGTM! 👍

PR #1933, opened 2024-03-13
ci: use personal auth token for e2e

buenaflor:
It makes for us easier to differentiate

let's use personal auth token `SENTRY_AUTH_TOKEN_E2E` for e2e test which needs to access the events

everything else:  will use the org token `SENTRY_AUTH_TOKEN` for uploading sourcemaps, debug symbols etc

#skip-changelog


kahest (approved):
LGTM

PR #1996, opened 2024-04-18
Bump versions in v8

buenaflor:
Prepare v8 branch for release

analyze ci will fail here which is expected

#skip-changelog


brustolin (approved):
All looks good.

I wondering,  dont we pump version during release automatically?

buenaflor:
@brustolin we do, it usually happens by ci when we trigger a release but the v8 branch has been up for months so we needed to do it manually (manually run the update script)

PR #1997, opened 2024-04-18
Merge v8 into main

buenaflor:
merges the v8 branch into main from which we then can do a 8.0.0 release

#skip-changelog


CHANGELOG.md
```diff
@@ -117,6 +139,40 @@
   - [changelog](https://github.com/getsentry/sentry-cocoa/blob/main/CHANGELOG.md#8200)
   - [diff](https://github.com/getsentry/sentry-cocoa/compare/8.19.0...8.20.0)
 
+## 8.0.0-beta.2
```
ueman, line 146: This is at the wrong place in the changelog

buenaflor: What do you mean?

ueman: The v8 things should be at the top, right? That's not the case right now.

buenaflor: We discussed this and we'll keep the release order so beta should be fine where it is now

kahest (approved):
🚀

PR #2009, opened 2024-04-23
App start: Add spans to first transaction

buenaflor:
## :scroll: Description

Adds app start spans to the first transaction

Added spans:
- App launch to plugin registration -> measures app launch until the plugin is attached to the flutter engine
- Main isolate setup -> measures time it takes from plugin registration until the earliest point dart is executed (as early as the flutter sdk is initialized)
- First frame render -> measures time between main isolate dart code executing until the first frame

mobile starfish

In combination with sentry navigator observer you'll be able to see app start traces in the app start module

Unit Tests

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes


flutter/lib/src/sentry_flutter.dart
```diff
@@ -43,6 +48,9 @@ mixin SentryFlutter {
   }) async {
     final flutterOptions = SentryFlutterOptions();
 
+    // ignore: invalid_use_of_internal_member
+    mainIsolateStartTime = flutterOptions.clock();
```
stefanosiano: why are you setting it here?
This changes its value, and it shouldn't, as far as i understand
This is why the value in the `AppStartInfo` is different from the one retrieved in `NativeAppStartEventProcessor`

flutter/lib/src/event_processor/native_app_start_event_processor.dart
```diff
@@ -25,6 +33,82 @@ class NativeAppStartEventProcessor implements EventProcessor {
...
+        startTimestamp: appStartInfo.pluginRegistration,
+        endTimestamp: appStartInfo.mainIsolateStart);
+
+    final firstFrameRenderSpan = await _createAndFinishSpan(
+        tracer: transaction,
+        operation: appStartInfo.appStartTypeOperation,
+        description: appStartInfo.firstFrameRenderDescription,
+        parentSpanId: appStartSpan.context.spanId,
+        traceId: transactionTraceId,
+        startTimestamp: SentryFlutter.mainIsolateStartTime,
```
stefanosiano:
```suggestion
        startTimestamp: appStartInfo.mainIsolateStart,
```

flutter/lib/src/event_processor/native_app_start_event_processor.dart
```diff
@@ -25,6 +33,82 @@ class NativeAppStartEventProcessor implements EventProcessor {
...
+      firstFrameRenderSpan
+    ]);
+  }
+
+  Future<SentrySpan> _createAndFinishSpan({
+    required SentryTracer tracer,
+    required String operation,
+    required String description,
+    required SpanId? parentSpanId,
+    required SentryId? traceId,
```
stefanosiano:
```suggestion
    required SpanId parentSpanId,
    required SentryId traceId,
```

flutter/lib/src/sentry_flutter.dart
buenaflor: I think we can remove it.

`static DateTime mainIsolateStartTime = getUtcDateTime();` should be set as soon as SentryFlutter class is loaded, I guess it should be good enough to not have to use `flutterOptions.clock();` within init,

just tested it and there is around a 2-4ms difference

stefanosiano: the main problem is if `SentryFlutter.init` is not called immediately. in that case the difference would be higher.
So yes, the static variable is enough

stefanosiano (approved):
nice!

PR #2017, opened 2024-04-29
App start: Create transaction when no `SentryNavigatorObserver` is present

buenaflor:
## :scroll: Description

Builds upon https://github.com/getsentry/sentry-dart/pull/2009

When `SentryNavigatorObserver` is not present, it will not create a transaction for the app start spans to attach to - so we need to create it manually

Mobile starfish, improve app start instrumentation

Existing tests from https://github.com/getsentry/sentry-dart/pull/2009

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes


buenaflor:
@stefanosiano this is now ready for review pls

stefanosiano (approved):
looks good!

flutter/lib/src/integrations/native_app_start_integration.dart
```diff
@@ -149,6 +149,25 @@ class NativeAppStartIntegration extends Integration<SentryFlutterOptions> {
...
 
       setAppStartInfo(appStartInfo);
+
+      // When we don't have a SentryNavigatorObserver, a TTID transaction
+      // is not created therefore we need to create a transaction ourselves.
+      // We detect this by checking if the currentRouteName is null.
+      // This is a workaround since there is no api that tells us if
+      // the navigator observer exists or not. The currentRouteName is always
+      // set during a didPush triggered by the navigator observer.
+      if (SentryNavigatorObserver.currentRouteName == null) {
```
stefanosiano: we could even have an additional static flag `isAttached` that is set to true when the class is instantiated, and use it together with this check, but not sure if that would actually help some case.
Your call here

buenaflor: hm I think it would be good to have it additionally, I'll add it 👍

buenaflor: I think naming it something like `isCreated` is more accurate since you can also instantiate it earlier and attach it much later

PR #2027, opened 2024-05-02
App start: native spans

buenaflor:
## :scroll: Description

Adds native spans to app start transaction

```[tasklist]
### Blocked by
- [ ] https://github.com/getsentry/sentry-dart/pull/2009
``` 

Improve app start transaction and give insight into the native layer times

Unit, Manual with spotlight

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes


flutter/lib/src/integrations/native_app_start_integration.dart
```diff
@@ -101,12 +106,37 @@ class NativeAppStartIntegration extends Integration<SentryFlutterOptions> {
...
+            // ignore: invalid_use_of_internal_member
+            _hub.options.logger(
+                SentryLevel.warning, 'Failed to parse native span times: $e');
+            continue;
+          }
+        }
+
+        // We want to sort because the native spans are not guaranteed to be in order.
+        // Performance wise this won't affect us since the native span amount is very low.
+        nativeSpanTimes.sort((a, b) => a.start.compareTo(b.start));
```
buenaflor: also happy to hear other recommendations to solve this if this is not sufficient enough

flutter/lib/src/sentry_flutter.dart
```diff
@@ -48,6 +48,9 @@ mixin SentryFlutter {
   }) async {
     final flutterOptions = SentryFlutterOptions();
 
+    // ignore: invalid_use_of_internal_member
+    mainIsolateStartTime = flutterOptions.clock();
```
stefanosiano: this should be removed, same reaason of the other PR (not changing the mainIsolateStartTime after being initialized)

flutter/test/integrations/native_app_start_integration_test.dart
```diff
@@ -207,7 +263,7 @@ void main() {
       final processor = fixture.options.eventProcessors.first;
       tracer = fixture.createTracer();
       final transaction = SentryTransaction(tracer);
-      final enriched =
+      enriched =
```
stefanosiano, line 265: is the removal of `final` an overlook?

stefanosiano (approved):
just a detail, but good for the rest!

flutter/lib/src/integrations/native_app_start_integration.dart
stefanosiano: guess it should be enough

flutter/test/integrations/native_app_start_integration_test.dart
buenaflor: I set `late SentryTransaction enriched;` so I can access it in each test

buenaflor:
@stefanosiano I've updated the span description to `Before Sentry Init Setup `

flutter/lib/src/sentry_flutter.dart
```diff
@@ -48,6 +48,9 @@ mixin SentryFlutter {
   }) async {
     final flutterOptions = SentryFlutterOptions();
 
+    // ignore: invalid_use_of_internal_member
+    sentrySetupStartTime = flutterOptions.clock();
```
stefanosiano:
```suggestion
    sentrySetupStartTime ??= flutterOptions.clock();
```
so if the user initializes the SDK multiple times it's not an issue

flutter/lib/src/sentry_flutter.dart
stefanosiano: just leaving a comment for future readers. Dart static variables are lazy, initialized only when accessed, so we cannot rely on static variables.

stefanosiano:
ok, last thing and can be merged 👍

PR #2028, opened 2024-05-03
fix: event processor blocking transactions from being sent if `autoAppStart` is false

### LINKED ISSUE #2007: Finish a transaction never returns on iOS after being in background
### Platform

Flutter Mobile

### Obfuscation

Disabled

### Debug Info

Disabled

### Doctor

[√] Flutter (Channel stable, 3.19.4, on Microsoft Windows [Version 10.0.19045.4291], locale de-DE)
    • Flutter version 3.19.4 on channel stable at C:\Users\Niklas\fvm\versions\3.19.4             
    • Framework revision 68bfaea224 (5 weeks ago), 2024-03-20 15:36:31 -0700                      
    • Engine revision a5c24f538d                                                                  
    • Dart version 3.3.2                                                                          
    • DevTools version 2.31.1  

### Version

7.18.0

### Steps to Reproduce

How to create a reproducible example:
1. Create new flutter project
2. Add sentry packages
3. Initialize Sentry for instance like here: https://docs.sentry.io/platforms/flutter/#configure
4. Add a `await Sentry.startTransaction("test", "test").finish()` somewhere, for instance in the basic flutter example before counter is incremented
5. Go through the steps below to see the problem

Steps to Reproduce:
1. Run an app with Sentry in release mode on an iOS device
2. Optionally, enable "Fast App Termination" in the developer settings to not have to wait for so long
3. Put the app in the background and wait for some time (~30min with "Fast App Termination")
4. Open the app and somehow start and finish a transaction

If you downgrade to version 7.10.1 (there might be others, but this is the version I was using before) it works as expected. 

### Expected Result

The `transaction.finish(...)` returns after a short amount of time and the transaction is logged in Sentry.

### Actual Result

The `transaction.finish(...)` never returns and the transaction is not logged in Sentry.

No

buenaflor:
## :scroll: Description

if `autoAppStart` is false then the Future wats for `AppStartInfo` forever

Fixes #2007 

Unit tests

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/flutter/lib/src/integrations/native_app_start_integration.dart
+++ b/flutter/lib/src/integrations/native_app_start_integration.dart
@@ -82,12 +82,13 @@ class NativeAppStartIntegration extends Integration<SentryFlutterOptions> {
           nativeAppStart.appStartTime.toInt());
       final pluginRegistrationDateTime = DateTime.fromMillisecondsSinceEpoch(
           nativeAppStart.pluginRegistrationTime);
+      DateTime? appStartEndDateTime;
 
       if (options.autoAppStart) {
         // We only assign the current time if it's not already set - this is useful in tests
         // ignore: invalid_use_of_internal_member
         _native.appStartEnd ??= options.clock();
-        final appStartEndDateTime = _native.appStartEnd;
+        appStartEndDateTime = _native.appStartEnd;
 
         final duration = appStartEndDateTime?.difference(appStartDateTime);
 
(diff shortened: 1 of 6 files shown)


flutter/lib/src/integrations/native_app_start_integration.dart
```diff
@@ -121,21 +134,36 @@ class NativeAppStartIntegration extends Integration<SentryFlutterOptions> {
...
-      {required this.start,
-      required this.end,
-      required this.pluginRegistration,
-      required this.mainIsolateStart});
+  AppStartInfo(
+    this.type, {
+    required this.start,
+    required this.pluginRegistration,
+    required this.mainIsolateStart,
+    DateTime? end,
```
stefanosiano:
```suggestion
    this.end,
```

you can remove the `get` and `set`, and make it public like the other fields, just removing the `final` label

flutter/lib/src/integrations/native_app_start_integration.dart
```diff
@@ -109,10 +112,20 @@ class NativeAppStartIntegration extends Integration<SentryFlutterOptions> {
...
+        // through SentryFlutter.setAppStartEnd and is going to be queried in the event processor
+        final appStartInfo = AppStartInfo(
+            nativeAppStart.isColdStart ? AppStartType.cold : AppStartType.warm,
+            start: appStartDateTime,
+            pluginRegistration: pluginRegistrationDateTime,
+            mainIsolateStart: mainIsolateStartDateTime);
+
+        setAppStartInfo(appStartInfo);
+      }
+    });
```
stefanosiano, line 117: I'd put this whole section after the `if{}` block, removing the `else`.
Just initialize `appStartEndDateTime` to null and set a value in the `if` block.
So we avoid repetition and an `else` branch

buenaflor:
@stefanosiano we might have to release this fix on the v7 branch as well since it affects it too

stefanosiano (approved):

Issue #1193, opened 2022-12-16
Session Replay support for Flutter on Mobile


bruno-garcia:
Add support for Sentry's Session Replay: https://sentry.io/for/session-replay/

Rafik-Belkadi-Reccap:
from how many thumbs up will you start to consider its implementation?

marandaneto:
> from how many thumbs up will you start to consider its implementation?

This is blocked by https://github.com/flutter/flutter/issues/117382

YasserDRIF:
We are using Smartlook currently, Adding this to Sentry would be amazing

bruno-garcia:
We're working on it! Wanna join the early adopter release? Join the waitlist and discussion about the feature: 
* https://github.com/getsentry/sentry/discussions/63138

vaind:
Replay alpha version now available for Android in 8.6.0-alpha.2 - please share any and all feedback.


  To try out replay, you can set following options:

  ```dart
  await SentryFlutter.init(
    (options) {
      ...
      options.experimental.replay.sessionSampleRate = 1.0;
      options.experimental.replay.errorSampleRate = 1.0;
    },
    appRunner: () => runApp(MyApp()),
  );
  ```

  Access is limited to early access orgs on Sentry. If you're interested, [sign up for the waitlist](https://sentry.io/lp/mobile-replay-beta/)

mcosti:
Where should we post the issues we find? In this topic or each in a separate issue?

Here's what I found so far:

```
[sentry] [debug] WidgetFilter obscuring: Text("Editor")
[sentry] [debug] WidgetFilter obscuring: Text("My Festivals")
[sentry] [debug] WidgetFilter obscuring: Text("Messages")
[sentry] [debug] WidgetFilter obscuring: Text("Profile")
[sentry] [error] Replay: failed to capture screenshot.
         'dart:ui/painting.dart': Failed assertion: line 26 pos 10: '<optimized out>': Rect argument contained a NaN value.
         #0      _AssertionError._doThrowNew (dart:core-patch/errors_patch.dart:51:61)
         #1      _AssertionError._throwNew (dart:core-patch/errors_patch.dart:40:5)
         #2      _rectIsValid (dart:ui/painting.dart:26:10)
         #3      _NativeCanvas.drawRect (dart:ui/painting.dart:5936:12)
         #4      ScreenshotRecorder._obscureWidgets (package:sentry_flutter/src/replay/recorder.dart:137:14)
[...]

PR #2032, opened 2024-05-06
Flutter replay for Android 

vaind:
## :scroll: Description

Adds session replay support for Flutter (Android)
* [x] automatic session replay 
* [x] native errors replay
* [x] dart errors replay
* [x] unit test coverage
* [x] changelog entry
* [x] wait for sentry-java release https://github.com/getsentry/sentry-java/pull/3339

- part of #1193

Unit tests and manually

## :pencil: Checklist

- [ ] All tests passing
- [ ] No breaking changes

## :crystal_ball: Next steps

* Feature parity with native Android implementation, e.g. breadcrumbs
* https://github.com/getsentry/sentry/issues/63255
* iOS/macOS support
* integration test - needs either a beforeSend callback or a fake in-process sentry http server.
* docs.sentry.io


flutter/android/src/main/kotlin/io/sentry/flutter/SentryFlutterPlugin.kt
```diff
@@ -52,7 +56,11 @@ class SentryFlutterPlugin : FlutterPlugin, MethodCallHandler, ActivityAware {
       )
   }
 
-  override fun onMethodCall(call: MethodCall, result: Result) {
+  @Suppress("CyclomaticComplexMethod")
+  override fun onMethodCall(
+    call: MethodCall,
+    result: Result,
+  ) {
```
vaind, line 65: I had to use `ktlint --format` because of the boatload of complains by reviewdog 🤷

flutter/lib/src/sentry_flutter_options.dart
```diff
@@ -297,14 +298,25 @@ class SentryFlutterOptions extends SentryOptions {
...
   }
 
   /// The [navigatorKey] is used to add information of the currently used locale to the contexts.
   GlobalKey<NavigatorState>? navigatorKey;
+
+  /// Configuration of experimental features that may change or be removed
+  /// without prior notice. Additionally, these features may not be ready for
+  /// production use yet.
+  @meta.experimental
+  final experimental = _SentryFlutterExperimentalOptions();
```
vaind, line 341: `experimental` is the name introduced in other SDKs so copying the approach here as well.

flutter/lib/src/sentry_replay_options.dart
```diff
@@ -0,0 +1,48 @@
...
+  /// (disabled).
+  double? get sessionSampleRate => _sessionSampleRate;
+
+  /// Indicates the percentage in which the replay for the session will be created. Specifying 0
+  /// means never, 1.0 means always. The value needs to be >= 0.0 and <= 1.0 The default is null
+  /// (disabled).
+  set sessionSampleRate(double? value) {
+    assert(value == null || (value >= 0 && value <= 1));
+    _sessionSampleRate = value;
+  }
```
buenaflor: Do we need the duplicate comments?

flutter/lib/src/native/java/sentry_native_java.dart
```diff
@@ -1,10 +1,104 @@
...
+  late final SentryFlutterOptions _options;
   SentryNativeJava(super.channel);
+
+  @override
+  Future<void> init(SentryFlutterOptions options) async {
+    // We only need these when replay is enabled so let's set it up
+    // conditionally. This allows Dart to trim the code.
+    if (options.experimental.replay.isEnabled) {
+      _options = options;
+      if ((options.experimental.replay.errorSampleRate ?? 0) > 0) {
```
buenaflor, line 24: Just for my understanding: why do we check this condition before adding the event processor if we already know it's enabled?

flutter/lib/src/native/java/sentry_native_java.dart
```diff
@@ -1,10 +1,104 @@
...
+import '../../replay/recorder.dart';
+import '../../replay/recorder_config.dart';
 import '../sentry_native_channel.dart';
 
 // Note: currently this doesn't do anything. Later, it shall be used with
 // generated JNI bindings. See https://github.com/getsentry/sentry-dart/issues/1444
 @internal
 class SentryNativeJava extends SentryNativeChannel {
+  ScreenshotRecorder? _replayRecorder;
+  late final SentryFlutterOptions _options;
```
buenaflor: could this be null instead of late?

flutter/lib/src/native/java/sentry_native_java.dart
```diff
@@ -1,10 +1,104 @@
...
+    //     save as PNG (25401 bytes): 43.110 ms  -- used for the final image
+    //     image size: 25401 bytes
+    //     save to file: 3.677 ms
+    //     onScreenshotRecorded1: 1.237 ms
+    //     released and exiting callback: 0.021 ms
+    ScreenshotRecorderCallback callback = (image) async {
+      var imageData = await image.toByteData(format: ImageByteFormat.png);
+      if (imageData != null) {
+        var timestamp = DateTime.now().millisecondsSinceEpoch;
+        var filePath = path.join(cacheDir, "$timestamp.png");
```
buenaflor:
```suggestion
        final timestamp = DateTime.now().millisecondsSinceEpoch;
        final filePath = path.join(cacheDir, "$timestamp.png");
```

also: should we rather transform the datetime to utc or is that irrelevant?

flutter/lib/src/replay/recorder.dart
```diff
@@ -0,0 +1,134 @@
...
+      final watch = Stopwatch()..start();
+
+      // The desired resolution (coming from the configuration) is usually
+      // rounded to next multitude of 16. Therefore, we scale the image.
+      final srcWidth = renderObject.size.width;
+      final srcHeight = renderObject.size.height;
+      final pixelRatio =
+          min(_config.width / srcWidth, _config.height / srcHeight);
+
+      // First, we synchronously capture the image and enumarete widgets on the main UI loop.
```
buenaflor:
```suggestion
      // First, we synchronously capture the image and enumerate widgets on the main UI loop.
```

buenaflor (commented):
Added some comments, will do another round of review - cool stuff!

flutter/lib/src/native/java/sentry_native_java.dart
vaind: I've added comments: isEnabled means it's enabled for either session capture or error capture. Because the event processor is needed only for error capture, it's added only when error capture is enabled.

flutter/lib/src/native/java/sentry_native_java.dart
vaind: It could but why though? Functionally, it must be set in order to use the native binding. And it's always set in `init()`, which is the function that must be called in order to use the binding.

flutter/lib/src/native/java/sentry_native_java.dart
vaind: > also: should we rather transform the datetime to utc or is that irrelevant?

no need to, millisecondsSinceEpoch is always UTC

flutter/lib/src/sentry_replay_options.dart
vaind: no we don't in dart (copy-paste from another SDK). Cleaned up & rephrased.

flutter/lib/src/native/java/sentry_native_java.dart
buenaflor: kk all good then

buenaflor:
Sorry for the delay, will look over it once more tomorrow

buenaflor (approved):
looks good :rocket:

bruno-garcia:
@vaind release was done. https://github.com/getsentry/sentry-java/releases/tag/7.12.0

PR #2035, opened 2024-05-06
ci: fix expected iOS binary size

vaind:
#skip-changelog

Recent PRs pushed the compiled binary diff over the edge we expect, so let's update it.
image


buenaflor:
thx :pray:

buenaflor (approved):

PR #2044, opened 2024-05-08
ci: cleanup flutter matrix

vaind:

The exclude list was a bit long and we can make this shorter and simpler by flipping the logic to specify targets we actually want.

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes


buenaflor (approved):
👍

Issue #1480, opened 2023-05-25
Support Flutter Web WASM symbolication


marandaneto:
@buenaflor: You can symbolicate Flutter Web Wasm builds with the (hacky) workaround. However this is not considered to be stable and is not an official support. Read [this](<https://github.com/getsentry/sentry-dart/issues/1480#issuecomment-3922413325>) for more information.

This should work as a temporary solution until we properly implement it

### Description

\~JS package -> [https://github.com/getsentry/sentry-javascript/tree/develop/packages/wasm\~](<https://github.com/getsentry/sentry-javascript/tree/develop/packages/wasm~>)<br>~~The SDK needs to fill the ~~`~~debug_meta~~`~~ property with the image list.~~~~<br>~~~~Flutter Web probably exposes the ~~`~~WebAssembly.Module~~`~~ as well, if not, we can leverage that by using the JS SDK directly.~~~~<br>~~~~The [Dart plugin](<https://github.com/getsentry/sentry-dart-plugin>) should upload the DWARF files.~~

~~SDKs cannot depend on ~~`~~html~~`~~ packages nor its transitive dependencies~~<br>\~[https://docs.flutter.dev/platform-integration/web/wasm#requires-preview-js-interop-to-access-browser-and-js-apis\~](<https://docs.flutter.dev/platform-integration/web/wasm#requires-preview-js-interop-to-access-browser-and-js-apis~>)

~~Sentry Flutter SDK isn't compatible as of now.~~~~<br>~~~~Sentry Dart SDK compiles just fine.~~

~~Blocked by [https://github.com/flutter/flutter/issues/127564](<https://github.com/flutter/flutter/issues/127564>) as well~~

marandaneto:
Dart SDK compiles but it does not work either.
We have to guard every call that depends on `dart.library.html` or `dart.library.io` to a no-op call. we do this most of the time but there are probably a few exceptions not considered because we always assume it's either one or the other.

<img width="547" alt="Screenshot 2023-05-25 at 10 41 52" src="https://github.com/getsentry/sentry-dart/assets/5731772/80eda427-097c-449e-8ca9-845a5402997e">

vaind:
runtime parsing can also be improved, currently (dart2wasm used in tests and introduced in #2113) we get this:

```json
{
  event_id: dd5764b21ef04faa84c5c9c4a4e54447,
  timestamp: 2024-06-20T17: 43: 53.646Z,
  platform: other,
  user: {
[...]

PR #2064, opened 2024-05-20
Migrates to package:web and js_interop

josh-burton:
## :scroll: Description
Migrates to package:web and js_interop

Supports compilation via WASM

Tested on a flutter web app project, ran unit tests

## :pencil: Checklist

- [x] All tests passing
- [ ] No breaking changes

## :crystal_ball: Next steps

Version numbers should be bumped. I think technically this is not a breaking change, but perhaps it should be treated as one.

### DIFF
--- a/.github/workflows/min_version_test.yml
+++ b/.github/workflows/min_version_test.yml
@@ -36,7 +36,7 @@ jobs:
 
       - uses: subosito/flutter-action@44ac965b96f18d999802d4b807e3256d5a3f9fa1 # pin@v2.16.0
         with:
-          flutter-version: "3.0.0"
+          flutter-version: "3.13.0"
 
       - name: Build Android
         run: |
@@ -53,7 +53,7 @@ jobs:
 
       - uses: subosito/flutter-action@44ac965b96f18d999802d4b807e3256d5a3f9fa1 # pin@v2.16.0
         with:
-          flutter-version: "3.0.0"
+          flutter-version: "3.13.0"
 
       - name: Build iOS
         run: |
@@ -70,7 +70,7 @@ jobs:
 
       - uses: subosito/flutter-action@44ac965b96f18d999802d4b807e3256d5a3f9fa1 # pin@v2.16.0
         with:
-          flutter-version: "3.0.0"
+          flutter-version: "3.13.0"
 
       - name: Build web
         run: |
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 - Bump Cocoa SDK from v8.25.2 to v8.26.0 ([#2060](https://github.com/getsentry/sentry-dart/pull/2060))
   - [changelog](https://github.com/getsentry/sentry-cocoa/blob/main/CHANGELOG.md#8260)
   - [diff](https://github.com/getsentry/sentry-cocoa/compare/8.25.2...8.26.0)
+- Migrates to package:web and js_interop ([#2064](https://github.com/getsentry/sentry-dart/pull/2064))
 
 ## 8.2.0
 
--- a/dart/lib/src/event_processor/enricher/enricher_event_processor.dart
+++ b/dart/lib/src/event_processor/enricher/enricher_event_processor.dart
@@ -1,7 +1,7 @@
 import '../../event_processor.dart';
 import '../../sentry_options.dart';
 import 'io_enricher_event_processor.dart'
-    if (dart.library.html) 'web_enricher_event_processor.dart';
+    if (dart.library.js_interop) 'web_enricher_event_processor.dart';
 
 abstract class EnricherEventProcessor implements EventProcessor {
   factory EnricherEventProcessor(SentryOptions options) =>
--- a/dart/lib/src/event_processor/enricher/web_enricher_event_processor.dart
+++ b/dart/lib/src/event_processor/enricher/web_enricher_event_processor.dart
@@ -42,12 +42,16 @@ class WebEnricherEventProcessor implements EnricherEventProcessor {
   // https://github.com/getsentry/sentry-javascript/blob/a6f8dc26a4c7ae2146ae64995a2018c8578896a6/packages/browser/src/integrations/useragent.ts
   SentryRequest _getRequest(SentryRequest? request) {
     final requestHeader = request?.headers;
-    final header = requestHeader == null ? <String, String>{} : Map<String, String>.from(requestHeader);
+    final header = requestHeader == null
+        ? <String, String>{}
+        : Map<String, String>.from(requestHeader);
 
     header.putIfAbsent('User-Agent', () => _window.navigator.userAgent);
 
     final url = request?.url ?? _window.location.toString();
-    return (request ?? SentryRequest(url: url)).copyWith(headers: header).sanitized();
+    return (request ?? SentryRequest(url: url))
+        .copyWith(headers: header)
+        .sanitized();
   }
 
   SentryDevice _getDevice(SentryDevice? device) {
@@ -55,9 +59,11 @@ class WebEnricherEventProcessor implements EnricherEventProcessor {
       online: device?.online ?? _window.navigator.onLine,
       memorySize: device?.memorySize ?? _getMemorySize(),
       orientation: device?.orientation ?? _getScreenOrientation(),
-      screenHeightPixels: device?.screenHeightPixels ?? _window.screen.availHeight,
+      screenHeightPixels:
+          device?.screenHeightPixels ?? _window.screen.availHeight,
       screenWidthPixels: device?.screenWidthPixels ?? _window.screen.availWidth,
-      screenDensity: device?.screenDensity ?? _window.devicePixelRatio.toDouble(),
+      screenDensity:
+          device?.screenDensity ?? _window.devicePixelRatio.toDouble(),
     );
   }
 
--- a/dart/lib/src/event_processor/enricher/web_enricher_event_processor.dart
+++ b/dart/lib/src/event_processor/enricher/web_enricher_event_processor.dart
@@ -1,11 +1,11 @@
-import 'dart:html' as html show window, Window;
+import 'package:web/web.dart' as web show window, Window, Navigator;
 
 import '../../../sentry.dart';
 import 'enricher_event_processor.dart';
 
 EnricherEventProcessor enricherEventProcessor(SentryOptions options) {
   return WebEnricherEventProcessor(
-    html.window,
+    web.window,
     options,
   );
 }
@@ -16,7 +16,7 @@ class WebEnricherEventProcessor implements EnricherEventProcessor {
     this._options,
   );
 
-  final html.Window _window;
+  final web.Window _window;
 
   final SentryOptions _options;
 
@@ -42,16 +42,12 @@ class WebEnricherEventProcessor implements EnricherEventProcessor {
   // https://github.com/getsentry/sentry-javascript/blob/a6f8dc26a4c7ae2146ae64995a2018c8578896a6/packages/browser/src/integrations/useragent.ts
   SentryRequest _getRequest(SentryRequest? request) {
     final requestHeader = request?.headers;
-    final header = requestHeader == null
-        ? <String, String>{}
-        : Map<String, String>.from(requestHeader);
+    final header = requestHeader == null ? <String, String>{} : Map<String, String>.from(requestHeader);
 
     header.putIfAbsent('User-Agent', () => _window.navigator.userAgent);
 
     final url = request?.url ?? _window.location.toString();
-    return (request ?? SentryRequest(url: url))
-        .copyWith(headers: header)
-        .sanitized();
+    return (request ?? SentryRequest(url: url)).copyWith(headers: header).sanitized();
   }
 
   SentryDevice _getDevice(SentryDevice? device) {
@@ -59,12 +55,9 @@ class WebEnricherEventProcessor implements EnricherEventProcessor {
       online: device?.online ?? _window.navigator.onLine,
       memorySize: device?.memorySize ?? _getMemorySize(),
       orientation: device?.orientation ?? _getScreenOrientation(),
-      screenHeightPixels: device?.screenHeightPixels ??
-          _window.screen?.available.height.toInt(),
-      screenWidthPixels:
-          device?.screenWidthPixels ?? _window.screen?.available.width.toInt(),
-      screenDensity:
-          device?.screenDensity ?? _window.devicePixelRatio.toDouble(),
+      screenHeightPixels: device?.screenHeightPixels ?? _window.screen.availHeight,
+      screenWidthPixels: device?.screenWidthPixels ?? _window.screen.availWidth,
+      screenDensity: device?.screenDensity ?? _window.devicePixelRatio.toDouble(),
     );
   }
 
@@ -77,14 +70,12 @@ class WebEnricherEventProcessor implements EnricherEventProcessor {
 
   SentryOrientation? _getScreenOrientation() {
     // https://developer.mozilla.org/en-US/docs/Web/API/ScreenOrientation
-    final screenOrientation = _window.screen?.orientation;
-    if (screenOrientation != null) {
-      if (screenOrientation.type?.startsWith('portrait') ?? false) {
-        return SentryOrientation.portrait;
-      }
-      if (screenOrientation.type?.startsWith('landscape') ?? false) {
-        return SentryOrientation.landscape;
-      }
+    final screenOrientation = _window.screen.orientation;
+    if (screenOrientation.type.startsWith('portrait')) {
+      return SentryOrientation.portrait;
+    }
+    if (screenOrientation.type.startsWith('landscape')) {
+      return SentryOrientation.landscape;
     }
     return null;
   }
@@ -100,4 +91,8 @@ class WebEnricherEventProcessor implements EnricherEventProcessor {
       timezone: culture?.timezone ?? DateTime.now().timeZoneName,
     );
   }
+}
+
+extension on web.Navigator {
+  external double? get deviceMemory;
 }
--- a/dart/pubspec.yaml
+++ b/dart/pubspec.yaml
@@ -24,6 +24,7 @@ dependencies:
   meta: ^1.3.0
   stack_trace: ^1.10.0
   uuid: '>=3.0.0 <5.0.0'
+  web: ^0.5.1
 
 dev_dependencies:
   build_runner: ^2.4.2
--- a/flutter/lib/src/renderer/html_renderer.dart
+++ b/flutter/lib/src/renderer/html_renderer.dart
@@ -10,6 +10,9 @@ bool get isCanvasKitRenderer {
   return _windowFlutterCanvasKit != null;
 }
 
+// These values are set by the engine. They are used to determine if the
+// application is using canvaskit or skwasm.
+//
 // See https://github.com/flutter/flutter/blob/414d9238720a3cde85475f49ce0ba313f95046f7/packages/flutter/lib/src/foundation/_capabilities_web.dart#L10
 @JS('window.flutterCanvasKit')
 external JSAny? get _windowFlutterCanvasKit;
--- a/flutter/lib/src/renderer/html_renderer.dart
+++ b/flutter/lib/src/renderer/html_renderer.dart
@@ -1,4 +1,4 @@
-import 'dart:js' as js;
+import 'dart:js_interop';
 
 import 'renderer.dart';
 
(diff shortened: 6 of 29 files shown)


buenaflor:
hey, I see that this is for wasm compilation but I don't think we support symbolication for wasm yet

kevmoo:
This is blocking some customers. Happy to discuss how we can help!

dart/lib/src/event_processor/enricher/web_enricher_event_processor.dart
```diff
@@ -101,3 +92,7 @@ class WebEnricherEventProcessor implements EnricherEventProcessor {
     );
   }
 }
+
+extension on web.Navigator {
+  external double? get deviceMemory;
```
denrase, line 103: Is this used anywhere? If not, we can remove it.

flutter/lib/src/renderer/html_renderer.dart
```diff
@@ -7,6 +7,9 @@ FlutterRenderer? getRenderer() {
 }
 
 bool get isCanvasKitRenderer {
-  final flutterCanvasKit = js.context['flutterCanvasKit'];
-  return flutterCanvasKit != null;
+  return _windowFlutterCanvasKit != null;
 }
+
+// See https://github.com/flutter/flutter/blob/414d9238720a3cde85475f49ce0ba313f95046f7/packages/flutter/lib/src/foundation/_capabilities_web.dart#L10
```
denrase, line 16: We could also add the comment fron the link as decriptions.

```dart
// These values are set by the engine. They are used to determine if the
// application is using canvaskit or skwasm.
// 
// See https://github.com/flutter/flutter/blob/414d9238720a3cde85475f49ce0ba313f95046f7/packages/flutter/lib/src/foundation/_capabilities_web.dart#L10
```

denrase (changes requested):
Thank you for contributing! I agree that we should migrate to this and also maybe release this in a major version bump.

Some small nitpicks & questions. Also, we need to satisfy CI in order to proceed.

- [ ] We are missing a changelog entry
- [ ] integration tests on macOS are failing apparently
- [ ] min-version tests are failing. Do we need to bump dart/flutter versions?
- [ ] analyze is failing. `dart format .` needs to be run & changes committed

buenaflor:
quick fyi: the ci job`sentry-sqflite / analyze / package-analysis` currently fails expectedly so no need to try and fix this here

dart/lib/src/event_processor/enricher/web_enricher_event_processor.dart
josh-burton: It's currently used in the `int? _getMemorySize()` method. This API has been added to the web package so in a future update it can be removed.

flutter/lib/src/renderer/html_renderer.dart
josh-burton: great idea

josh-burton:
@denrase @buenaflor I've addressed the PR comments and hopefully fixed the build failures.

I needed to bump the min version of Flutter has Dart 3.1.0 is required by package:web.

buenaflor:
@josh-burton if it's okay with you we'd like to fork your branch and take over from here, it's easier for us to fix the rest of the ci since we don't need to approve runnning them every time (cc @denrase)

josh-burton:
> @josh-burton if it's okay with you we'd like to fork your branch and take over from here, it's easier for us to fix the rest of the ci since we don't need to approve runnning them every time (cc @denrase)

No problem, go for it.

kevmoo:
Any help we can get here?

buenaflor:
hey we've been looking into it. The main problem we have is the necessity to bump the min versions. as far as I can see we would need to bump it to dart 3.3.0 and flutter 3.19.0 which is pretty much a huge jump for us

Any recommendations to alleviate that?

zs-dima:
@buenaflor 

image

> need to bump it to dart 3.3.0 and flutter 3.19.0 which is pretty much a huge jump for us
> Any recommendations to alleviate that?

buenaflor:
We cannot simply bump to versions that are only a few months old, we want to avoid that even in a new major release. A large part of our customers are not on dart 3.3.0 and flutter 3.19 yet (which are very close to the latest versions) and we don't plan on maintaining two majors for an extended period of time.

@kevmoo do you have guidance here if we have any other option available other than bumping the min versions?

kevmoo:
@buenaflor – this is a tricky spot.

If you want to maintain backward compat, you *could* test for the existence of `dart:js_interop` – see https://dart.dev/guides/libraries/create-packages#conditionally-importing-and-exporting-library-files

I worry that you'll get folks importing this library before it's stable on versions before Dart 3.3 – but it might be work investigating

denrase:
@kevmoo We do this, however this does not preclude us from adding the web package to our dependencies, which automatically raises the dart min SDK version to 3.3. Or am I missing something here?

dart/pubspec.yaml
```diff
@@ -24,6 +24,7 @@ dependencies:
   meta: ^1.3.0
   stack_trace: ^1.10.0
   uuid: '>=3.0.0 <5.0.0'
+  web: ^0.5.1
```
vaind, line 27: Do we actually need to add this dependency? Tests passed locally regardless.
```suggestion
```

flutter/pubspec.yaml
```diff
@@ -27,6 +27,7 @@ dependencies:
   package_info_plus: '>=1.0.0'
   meta: ^1.3.0
   ffi: ^2.0.0
+  web: ^0.5.1
```
vaind, line 30: same question as above
```suggestion
```

.github/workflows/min_version_test.yml
```diff
@@ -36,7 +36,7 @@ jobs:
 
       - uses: subosito/flutter-action@44ac965b96f18d999802d4b807e3256d5a3f9fa1 # pin@v2.16.0
         with:
-          flutter-version: "3.0.0"
+          flutter-version: "3.13.0"
```
vaind, line 39: Let's try keeping this and see how it goes without adding the explicit dependency on `web`:
```suggestion
          flutter-version: "3.0.0"
```

.github/workflows/min_version_test.yml
```diff
@@ -53,7 +53,7 @@ jobs:
...
```
vaind, line 56:
```suggestion
          flutter-version: "3.0.0"
```

.github/workflows/min_version_test.yml
```diff
@@ -70,7 +70,7 @@ jobs:
...
```
vaind, line 73:
```suggestion
          flutter-version: "3.0.0"
```

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@
 - Bump Cocoa SDK from v8.25.2 to v8.26.0 ([#2060](https://github.com/getsentry/sentry-dart/pull/2060))
   - [changelog](https://github.com/getsentry/sentry-cocoa/blob/main/CHANGELOG.md#8260)
   - [diff](https://github.com/getsentry/sentry-cocoa/compare/8.25.2...8.26.0)
+- Migrates to package:web and js_interop ([#2064](https://github.com/getsentry/sentry-dart/pull/2064))
```
vaind, line 15: Maybe this should be listed under features, considering it enables WASM compilation

dart/lib/src/event_processor/enricher/enricher_event_processor.dart
```diff
@@ -1,7 +1,7 @@
 import '../../event_processor.dart';
 import '../../sentry_options.dart';
 import 'io_enricher_event_processor.dart'
-    if (dart.library.html) 'web_enricher_event_processor.dart';
+    if (dart.library.js_interop) 'web_enricher_event_processor.dart';
```
vaind, line 4: is it possible to keep the original code in place for consumers with an older Flutter version? Something like 
```suggestion
    if (dart.library.js_interop) 'web_enricher_event_processor.dart';
    if (dart.library.html) 'web_enricher_event_processor_legacy.dart';
```

vaind (changes requested):
I've given this a brief look and have some ideas that may be able to unblock this while keeping backward compatibility. I may be missing something but it may be worth giving it a try.

kevmoo:
@vaind idea might work. I worry about the fact that `dart:js_interop` was in Dart 3.2 – just not "stable" - https://api.dart.dev/stable/3.2.1/index.html

So we might be in the same spot!

dart/pubspec.yaml
josh-burton: package:web is used in the dart package. Removing this raises a bunch of `depend_on_referenced_packages` analyzer warnings.

kevmoo: Yeah, you have to leave it!

vaind: The question is whether it actually works. We can ignore the warning for some time before cutting the compatibility.

buenaflor:
we have a [poc](https://github.com/getsentry/sentry-dart/pull/2110) for moving the web package implementations to another package called `sentry_web` 

which then inside the dart library we import it conditionally

```dart
import '../../web/noop_window.dart'
  if (dart.library.html) '../../web/http_window.dart'
  if (dart.library.js_interop) '../../web/web_window.dart';
```

inside a user's application if they wanna be able to build for wasm:
```dart
import 'package:sentry_web/sentry_web.dart' as sentry_web;

// set the window in SentryFlutter.init
options.window = sentry_web.SentryWeb.createWindow;
```      

as far as testing goes it looks fine, doesn't break backwards compat on lower dart versions (🤞).

might be a wrong direction though, it was only briefly tested

josh-burton:
> we have a [poc](https://github.com/getsentry/sentry-dart/pull/2110) for moving the web package implementations to another package called `sentry_web`
> 
> which then inside the dart library we import it conditionally
> 
> ```dart
> import '../../web/noop_window.dart'
>   if (dart.library.html) '../../web/http_window.dart'
>   if (dart.library.js_interop) '../../web/web_window.dart';
> ```
> 
> inside a user's application if they wanna be able to build for wasm:
> 
> ```dart
> import 'package:sentry_web/sentry_web.dart' as sentry_web;
> 
> // set the window in SentryFlutter.init
> options.window = sentry_web.SentryWeb.createWindow;
> ```
> 
> as far as testing goes it looks fine, doesn't break backwards compat on lower dart versions (🤞).
> 
> might be a wrong direction though, it was only briefly tested

Great. That sounds like the best approach. Should I close this PR?

vaind:
> we have a [poc](https://github.com/getsentry/sentry-dart/pull/2110) for moving the web package implementations to another package called `sentry_web`
> 
> which then inside the dart library we import it conditionally
> 
> ```dart
> import '../../web/noop_window.dart'
>   if (dart.library.html) '../../web/http_window.dart'
>   if (dart.library.js_interop) '../../web/web_window.dart';
> ```
> 
> inside a user's application if they wanna be able to build for wasm:
> 
> ```dart
> import 'package:sentry_web/sentry_web.dart' as sentry_web;
> 
> // set the window in SentryFlutter.init
> options.window = sentry_web.SentryWeb.createWindow;
> ```
> 
> as far as testing goes it looks fine, doesn't break backwards compat on lower dart versions (🤞).
> 
> might be a wrong direction though, it was only briefly tested

Having custom code for one platform is not ideal because then people need to know they need to change this, i.e. there need to be docs and they have to notice it in the docs. 

Just to double check: has anyone tried if my proposals actually work or did we just assume they won't? Because in that case I would give it a try myself.

dart/pubspec.yaml
buenaflor: I don't quite understand why we can leave it out since we have to use the package

```dart
import 'package:web/web.dart' as web show window, Window, Navigator;

// and in the enricher event processor we use web.window
```

buenaflor:
> Having custom code for one platform is not ideal because then people need to know they need to change this, i.e. there need to be docs and they have to notice it in the docs.

We've already decided that we won't bump the versions not matter what for this use case so this solution is the "worst" case scenario if we can't manage anything more simpler.

> Just to double check: has anyone tried if my proposals actually work or did we just assume they won't? Because in that case I would give it a try myself.

The approach is very similar to the poc.
I'm only confused about the removal of the web package since we have to use a specific property of that package (window) so I'm not sure how it's possible to leave it out.

vaind:
> I'm only confused about the removal of the web package since we have to use a specific property of that package (window) so I'm not sure how it's possible to leave it out.

AFAIU it will work because the consumer app depends on the web package. If it doesn't then our code doesn't use it anyway (it should be tree-shaken/trimmed)

vaind:
let me check this out in a new PR

dart/lib/src/event_processor/enricher/enricher_event_processor.dart
buenaflor: that works and is also done inside the poc

in your scenario example `web_enricher_event_processor_legacy.dart` would use dart:html

and `'web_enricher_event_processor.dart` would use package:web, it can't use dart:html since that would fail the compilation so we're back at the point where we have to add the web package dependency. 

Just looks to me like we can't avoid that unless I see it wrong xd

sigmundch:
@kevmoo, @srujzs, and I were brainstorming a bit here to see what other alternative we could offer from a Dart perspective, and we landed on either branching (having 2 versions of the package under maintenance) or practically what @buenaflor suggested [above](https://github.com/getsentry/sentry-dart/pull/2064#issuecomment-2179312493) (2 versions of the web-specific code, one based on `dart:html`, the other on `dart:js_interop`, but a package version.)

There are a lot of caveats to keep in mind:
* *Import order*: the order of the conditional imports matters, I'd recommend `dart.library.html` to go first. Technically that will keep all JavaScript clients on the old logic for a lot longer (until you are ready to bump this package's SDK constraint and directly use `package:web`). However, I'm inclined to go in that direction because of the risks of inconsistent behavior with using `dart:js_interop` depending on the SDK version. If you have a solution that works across all SDK versions consistently, then by all means put `dart.library.js_interop` first.
* *`dart:js_interop` breaking changes*: `dart:js_interop` has existed since Dart 3.1.0 and has gone through several breaking changes until it graduated as a stable library in Dart 3.3.0. If it's possible to make the logic consistent across all versions of `dart:js_interop`, then great! If not, I believe putting it second in the import order may be good enough. That's because `dart2wasm` only reached stable in Dart 3.4, so we should expect some consistency to follow from the fact that users are likely only using the newer versions of the SDK. 
* *no extension types*: Note that `extension types` were only added in Dart 3.3,  which is what we default to use now for JS-interop APIs. To use `dart:js_interop` without a min-SDK constraint, you will likely need to use older forms of interop. This may be to use `@staticInterop` or rely on the more primitive parts of the API (like `dart:js_interop_unsafe`).
* *dropping `package:web`*: I think this is OK to do, but I wouldn't rely on that dependency existing on user's package. Instead, I would suggest you simply don't use the package altogether. The good news is that all of `package:web` is written using `dart:js_interop`, so it is possible to access directly the browser APIs you need using only the Dart/JS interop layer (e.g. use `JS() external JSObject get window;` to expose window directly, if you need it)

Hope this is helpful.

vaind:
> @kevmoo, @srujzs, and I were brainstorming a bit here to see what other alternative we could offer from a Dart perspective, and we landed on either branching (having 2 versions of the package under maintenance) or practically what @buenaflor suggested [above](https://github.com/getsentry/sentry-dart/pull/2064#issuecomment-2179312493) (2 versions of the web-specific code, one based on `dart:html`, the other on `dart:js_interop`, but a package version.)
> 
> There are a lot of caveats to keep in mind:
> 
> * _Import order_: the order of the conditional imports matters, I'd recommend `dart.library.html` to go first. Technically that will keep all JavaScript clients on the old logic for a lot longer (until you are ready to bump this package's SDK constraint and directly use `package:web`). However, I'm inclined to go in that direction because of the risks of inconsistent behavior with using `dart:js_interop` depending on the SDK version. If you have a solution that works across all SDK versions consistently, then by all means put `dart.library.js_interop` first.
> * _`dart:js_interop` breaking changes_: `dart:js_interop` has existed since Dart 3.1.0 and has gone through several breaking changes until it graduated as a stable library in Dart 3.3.0. If it's possible to make the logic consistent across all versions of `dart:js_interop`, then great! If not, I believe putting it second in the import order may be good enough. That's because `dart2wasm` only reached stable in Dart 3.4, so we should expect some consistency to follow from the fact that users are likely only using the newer versions of the SDK.
> * _no extension types_: Note that `extension types` were only added in Dart 3.3,  which is what we default to use now for JS-interop APIs. To use `dart:js_interop` without a min-SDK constraint, you will likely need to use older forms of interop. This may be to use `@staticInterop` or rely on the more primitive parts of the API (like `dart:js_interop_unsafe`).
> * _dropping `package:web`_: I think this is OK to do, but I wouldn't rely on that dependency existing on user's package. Instead, I would suggest you simply don't use the package altogether. The good news is that all of `package:web` is written using `dart:js_interop`, so it is possible to access directly the browser APIs you need using only the Dart/JS interop layer (e.g. use `JS() external JSObject get window;` to expose window directly, if you need it)
> 
> Hope this is helpful.

Thanks for inputs. I was able to make this work today by dropping the import & using conditional imports (i.e. what I've suggested above). See followup PR https://github.com/getsentry/sentry-dart/pull/2113 - with a little bit of fiddling it seems to work fine in my tests and I don't think it would break anyone currently using the SDK while also enabling people to start using WASM (once we update stack trace parsing & implement offline symbolication for wasm - #1480)

> _Import order_: the order of the conditional imports matters, I'd recommend `dart.library.html` to go first

that's right, that's exactly what I did

> * _`dart:js_interop` breaking changes_: `dart:js_interop` has existed since Dart 3.1.0 and has gone through several breaking changes until it graduated as a stable library in Dart 3.3.0

I'm testing on Dart 3.3+ with the new `package:web` dependency and it seems to work for me.

buenaflor:
Unfortunately the release of the fix is currently blocked, we've got package validations error when trying to publish:

```
  STDERR:dart: Package validation found the following errors:
  dart: * line 3, column 1 of lib/src/platform/_web_platform.dart: This package does not have web in the `dependencies` section of `pubspec.yaml`.
  dart:     ╷
  dart:   3 │ import 'package:web/web.dart' as web;
  dart:     │ ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  dart:     ╵
  dart: * line 3, column 1 of lib/src/event_processor/enricher/web_enricher_event_processor.dart: This package does not have web in the `dependencies` section of `pubspec.yaml`.
  dart:     ╷
  dart:   3 │ import 'package:web/web.dart' as web show window, Window, Navigator;
  dart:     │ ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  dart:     ╵
  dart: * line 3, column 1 of lib/src/origin_web.dart: This package does not have web in the `dependencies` section of `pubspec.yaml`.
  dart:     ╷
  dart:   3 │ import 'package:web/web.dart';
  dart:     │ ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

we're looking into it

sigmundch:
Ah, it makes sense that pub validates that imports correspond to available dependencies. This validates my earlier concern around how to drop the `package:web` dependency. Note that my recommendation still applies. You can use the low-level interop APIs to replace those uses of package:web. Some may be a bit more ugly unfortunately.

For example:

**dart/lib/src/origin_web.dart**:
```dart
import 'package:web/web.dart';

/// request origin, used for browser stacktrace
String get eventOrigin => '${window.location.origin}/';
```

would become:
```dart
import 'dart:js_interop';

/// request origin, used for browser stacktrace
@JS('window.location.origin')
external String get eventOrigin;
```

OR something like:
```dart
import 'dart:js_interop';
import 'dart:js_interop_unsafe';

/// request origin, used for browser stacktrace
String get eventOrigin {
   final location = globalContext['location'] as JSObject;
   final origin = location['origin'] as JSString;
   return '${origin.toDart}/';
}
```

And _web_platform.dart can directly expose the two members it uses:
```dart
import 'dart:js_interop';
import 'dart:js_interop_unsafe';


/// request origin, used for browser stacktrace
String get hostname {
   final location = globalContext['location'] as JSObject;
   final hostname = location['hostname'] as JSString;
   return hostname.toDart;
}
bool matchMediaMatches(String s) {
  final result = globalContext.callMethod('matchMedia', s.toJS) as JSObject;
  return (result['matches'] as JSBoolean).toDart;
}
```

I believe `web_enricher_event_processor.dart` will need a bit more. Bear in mind that you cannot depend on `extenstion type`, since that's only available in Dart 3.3. And while you could use `@staticInterop`, it is not something that we intend to support long term.

vaind:
8.4.0-beta.1 is out with WASM support. Can you give it a try @josh-burton, @kevmoo ?

kevmoo:
Thank you so much!

PR #2076, opened 2024-05-28
fix(app-start): `Cold start` should be `Cold Start`

buenaflor:
Change span description for app start span


narsaynorath (approved):
Perfect! Thanks for being able to make this change so quickly

PR #2097, opened 2024-06-11
refactor: move all native calls to SentryNativeBinding

vaind:
This finishes the transition to SentryNativeBinding. Previously, there was still some code directly using MethodChannels. Merging this PR will open way to integrating sentry-native via FFI (e.g. for desktop platforms), which is what I originally set out to do before finding out these blockers. 

Basically, this PR:

- moves the remaining native calls `captureEnvelope()`, `loadContexts` and `loadDebugImages` to `SentryNativeBinding`
- removes `SentryNative` class that was just a wrapper around `SentryNativeBinding`, with some quirks:
  - error handling moved to `SentryNativeChannel` via the `SentryNativeSafeInvoker` mixin
  - native app start status variables moved to `NativeAppStartIntegration`
- updates classes that depended on `SentryNative` to instead use any implementation of `SentryNativeBinding`
- updates tests to accomodate these chanhes

I know this PR looks pretty large but although it doesn't seem like it, I did try to keep it to the minimum necessary. Hiding whitespace changes helps a little (especially in tests where a foreach loop may have been added/removed):
image

#skip-changelog because these are all internal changes


flutter/test/load_image_list_test.dart
```diff
@@ -1,183 +0,0 @@
-@TestOn('vm')
```
vaind, line 1: this was duplicate of integration/load_image_list_integration.test

flutter/test/sentry_native_channel_test.dart
```diff
@@ -6,239 +6,228 @@ library flutter_test;
...
+import 'sentry_flutter_test.dart';
 
 void main() {
-  group('$SentryNative', () {
-    late Fixture fixture;
+  for (var platform in [
+    MockPlatform.android(),
+    MockPlatform.iOs(),
+    MockPlatform.macOs()
+  ]) {
```
vaind: this is basically the only addition here - running the test for all implementations of SentryNativeBinding.

The diff is a bit messed up here because of whitespace

flutter/lib/src/sentry_flutter.dart
```diff
@@ -161,22 +153,13 @@ mixin SentryFlutter {
...
-
-    // Will enrich events with device context, native packages and integrations
-    if (platformChecker.hasNativeIntegration &&
-        !platformChecker.isWeb &&
-        (platform.isIOS || platform.isMacOS || platform.isAndroid)) {
-      integrations.add(LoadContextsIntegration(channel));
+    final native = _native;
+    if (native != null) {
+      integrations.add(NativeSdkIntegration(native));
+      integrations.add(LoadContextsIntegration(native));
```
buenaflor, line 159: ah looks much nicer now 😄 

if I get it right we only create the native binding for android, ios, macos right so there's no need for the checks we had before?

buenaflor (approved):
nice! 👍

buenaflor:
btw sdk metrics is failing currently, the app size needs to be adjusted again, not related to this pr though

flutter/lib/src/sentry_flutter.dart
vaind: yes, same checks were applied above when actually asigning the variable so it's easier this way

PR #2106, opened 2024-06-17
feat(performance): report total frames, frame delay, slow & frozen frames

buenaflor:
## :scroll: Description

Attaches frame metrics to spans such as total frames, frame delay, slow & frozen frames

Mobile Starfish

Unit tests, manually

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps

```[tasklist]
### Tasks
- [ ] Update docs
- [x] Create issue that Linux, Windows and Web refresh rates are currently locked at 60hz for frame metrics calculation
```


dart/lib/src/protocol/sentry_span.dart
```diff
@@ -16,6 +13,9 @@ class SentrySpan extends ISentrySpan {
   Map<String, List<MetricSummary>>? _metricSummaries;
   late final DateTime _startTimestamp;
   final Hub _hub;
+  bool _isRootSpan = false;
```
buenaflor, line 19: the reason why we add this is because `SentryTracer` is just a wrapper around a `SentrySpan`.

we need this flag so we can call `setMeasurement` and `setData` only on the wrapping tracer for only root spans.

flutter/lib/src/span_frame_metrics_collector.dart
```diff
@@ -0,0 +1,201 @@
...
+
+    _stopwatch.reset();
+    if (_frameCallbackHandler?.hasScheduledFrame == true) {
+      _stopwatch.start();
+    }
+  }
+
+  Future<void> recordSpanFrameMetrics(
+      ISentrySpan span, DateTime endTimestamp) async {
+    final displayRefreshRate = await _native?.displayRefreshRate();
```
buenaflor: I'm not sure whether we always need to fetch the refresh rates here or just do it once when we initialize the frame collector

my thinking was it may update dynamically but is that really the case for ios and android?

flutter/test/span_frame_metrics_collector_test.dart
```diff
@@ -0,0 +1,242 @@
...
+
+    await Future<void>.delayed(Duration(milliseconds: 800));
+
+    await tracer.finish();
+
+    expect(sut.isFrameTrackingPaused, isTrue);
+  });
+}
+
+Matcher _isWithinRange(int expected, [int delta = 5]) => allOf(
```
buenaflor: when testing measuring times the times will not always be accurate so I'm adding this so the test will pass if the measurements are withing a specific range.

 I'm not sure what kind of range is acceptable for the tests

machines running these tests will have different performances so it might be flaky, some slower, some faster

flutter/android/src/main/kotlin/io/sentry/flutter/SentryFlutterPlugin.kt
```diff
@@ -179,6 +181,19 @@ class SentryFlutterPlugin : FlutterPlugin, MethodCallHandler, ActivityAware {
...
 
+  private fun displayRefreshRate(result: Result) {
+    var refreshRate: Int? = null
+
+    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.R) {
+      val display = activity?.get()?.display
+      if (display != null) {
+        refreshRate = display.refreshRate.toInt()
+      }
+    }
```
stefanosiano, line 202:
```suggestion
    } else {
        val display = activity?.get()?.window?.windowManager?.defaultDisplay
        if (display != null) {
          refreshRate = display.refreshRate.toInt()
        }
    }
```

flutter/lib/src/sentry_flutter.dart
```diff
@@ -90,6 +91,9 @@ mixin SentryFlutter {
 
     await _initDefaultValues(flutterOptions, channel);
 
+    flutterOptions
+        .addPerformanceCollector(SpanFrameMetricsCollector(flutterOptions));
```
stefanosiano: should this go inside `_initDefaultValues`? so we are consistent with other stuff

buenaflor: yeh makes sense

flutter/lib/src/span_frame_metrics_collector.dart
buenaflor: or maybe we should fetch it in `onSpanStarted`

flutter/lib/src/span_frame_metrics_collector.dart
```diff
@@ -0,0 +1,212 @@
...
+      return;
+    }
+
+    activeSpans.add(span);
+    startFrameTracking();
+  }
+
+  @override
+  Future<void> onSpanFinished(ISentrySpan span, DateTime endTimestamp) async {
+    if (span is NoOpSentrySpan || !options.enableFramesTracking) {
```
stefanosiano: if `enableFramesTracking` is false the span was never added, so no need to check here, as the user could change the option at runtime (even if it's unsupported)
```suggestion
    if (span is NoOpSentrySpan) {
```

flutter/lib/src/span_frame_metrics_collector.dart
```diff
@@ -0,0 +1,212 @@
...
+
+    if (!_stopwatch.isRunning) {
+      _stopwatch.start();
+    }
+
+    await _frameCallbackHandler?.endOfFrame;
+
+    final frameDuration = _stopwatch.elapsedMilliseconds;
+    // ignore: invalid_use_of_internal_member
+    frameDurations[getUtcDateTime()] = frameDuration;
```
stefanosiano: just for consistency
```suggestion
    frameDurations[options.clock()] = frameDuration;
```

flutter/lib/src/span_frame_metrics_collector.dart
stefanosiano: In theory yes, there are variable refresh rates devices on both Android and iOS
> or maybe we should fetch it in onSpanStarted

yeah, maybe it's better, so we don't await for it here.
Also i would fallback to 60 hz, instead of dropping the frames in case it's null

flutter/lib/src/span_frame_metrics_collector.dart
```diff
@@ -0,0 +1,212 @@
...
+      ISentrySpan span, DateTime endTimestamp) async {
+    final displayRefreshRate = await _native?.displayRefreshRate();
+    if (displayRefreshRate == null) {
+      options.logger(SentryLevel.warning,
+          'Display refresh rate is not available. Dropping the frame metrics');
+      clear();
+      return;
+    }
+
+    activeSpans.removeWhere(
```
stefanosiano: s: maybe we can change `activeSpans` to a set and here just do activeSpans.remove(span)?

flutter/lib/src/span_frame_metrics_collector.dart
```diff
@@ -0,0 +1,212 @@
...
+            value.isBefore(endTimestamp) && value.isAfter(span.startTimestamp))
+        .toList();
+
+    if (timestamps.isEmpty) {
+      options.logger(
+          SentryLevel.info, 'No frame durations available in frame tracker.');
+      return {};
+    }
+
+    final slowFrames = timestamps.where((timestamp) {
```
stefanosiano: I understand the logic here, but we are iterating on the frames 3 times, plus iterating on the filtered lists.
We can do everything in one for loop, just adding the current frame duration to some local variables

stefanosiano (commented):
nice!
just small details, but looks good so far

flutter/lib/src/span_frame_metrics_collector.dart
buenaflor: the flag isn't used anywhere else except here, or maybe should we guard adding the performance collector behind it in SentryFlutter.init?

stefanosiano: i meant that the user could do `options.enableFramesTracking = true` anytime at runtime, even if this shouldn't be done. So checking this flag when the span is added is enough, as if it wasn't added, this function will do nothing.
You can even change it to `span is NoOpSentrySpan || !runningSpans.contains(span)`

buenaflor: oh yeah makes sense, sry misunderstood, I thought the change referred to the `onSpanStarted` function

buenaflor:
@stefanosiano I updated the cocoa refresh rate fetching so it also works on macos. Linux, windows and web don't have specific refresh rate fetching yet so it stays at 60hz for them and I don't think they're a priority currently. Maybe we can create an issue afterwards so we have it tracked at least?

stefanosiano:
> @stefanosiano I updated the cocoa refresh rate fetching so it also works on macos. Linux, windows and web don't have specific refresh rate fetching yet so it stays at 60hz for them and I don't think they're a priority currently. Maybe we can create an issue afterwards so we have it tracked at least?

yep, totally fine!

flutter/lib/src/span_frame_metrics_collector.dart
```diff
@@ -0,0 +1,226 @@
...
+  bool _isTrackingRegistered = false;
+
+  int displayRefreshRate = 60;
+
+  final _stopwatch = Stopwatch();
+
+  SpanFrameMetricsCollector(this.options,
+      {FrameCallbackHandler? frameCallbackHandler,
+      SentryNativeBinding? native,
+      bool isTestMode = false})
```
stefanosiano:
```suggestion
      @internal bool isTestMode = false})
```

flutter/lib/src/span_frame_metrics_collector.dart
```diff
@@ -0,0 +1,226 @@
...
+
+  @override
+  Future<void> onSpanStarted(ISentrySpan span) async {
+    if (span is NoOpSentrySpan || !options.enableFramesTracking) {
+      return;
+    }
+
+    final fetchedDisplayRefreshRate = await _native?.displayRefreshRate();
+    if (fetchedDisplayRefreshRate != null) {
+      options.logger(SentryLevel.info,
```
stefanosiano: I think info would be too spammy for the users
```suggestion
      options.logger(SentryLevel.debug,
```

flutter/lib/src/span_frame_metrics_collector.dart
```diff
@@ -0,0 +1,226 @@
...
+          'Retrieved display refresh rate at $fetchedDisplayRefreshRate');
+      displayRefreshRate = fetchedDisplayRefreshRate;
+    } else {
+      options.logger(SentryLevel.info,
```
stefanosiano: same here
```suggestion
      options.logger(SentryLevel.debug,
```

flutter/lib/src/span_frame_metrics_collector.dart
```diff
@@ -0,0 +1,226 @@
...
+      span.setData(key, value);
+    });
+
+    // This will call the methods on the tracer, not on the span directly
+    if (span is SentrySpan && span.isRootSpan) {
+      frameMetrics.forEach((key, value) {
+        // ignore: invalid_use_of_internal_member
+        span.tracer.setData(key, value);
+      });
+      frameMetrics.forEach((key, value) {
```
stefanosiano: `s`: we can merge both forEach in one

flutter/lib/src/span_frame_metrics_collector.dart
```diff
@@ -0,0 +1,226 @@
...
+  }
+
+  @visibleForTesting
+  Map<String, int> calculateFrameMetrics(
+      ISentrySpan span, DateTime endTimestamp, int displayRefreshRate) {
+    final expectedFrameDuration = ((1 / displayRefreshRate) * 1000).toInt();
+
+    // Filter frame durations within the span's time range
+    final timestamps = frameDurations.keys
+        .takeWhile((value) =>
```
stefanosiano: `l`: sadly `takeWhile` doesn't work here.
If a span is finished and there is a frame in the map before the start of the span, this takeWhile will return an empty collection. 
The problem is `value.isAfter(span.startTimestamp)`
e.g.
span a -> start: 1, end: 10
span b -> start: 3, end: 8
when span  b finishes, there is a frame at t 1 and 2, so the condition of the takeWhile will be false

flutter/lib/src/span_frame_metrics_collector.dart
```diff
@@ -0,0 +1,226 @@
...
+    }
+
+    int slowFramesCount = 0;
+    int frozenFramesCount = 0;
+    int slowFramesDuration = 0;
+    int frozenFramesDuration = 0;
+    int framesDelay = 0;
+
+    for (final timestamp in timestamps) {
+      final frameDuration = frameDurations[timestamp] ?? 0;
```
stefanosiano: `l`: The frames should be removed from the map, so they are not counted in other spans
```suggestion
      final frameDuration = frameDurations.remove(timestamp) ?? 0;
```

buenaflor: In a scenario like this: root span (t: 1-10) and its child span (t: 6-8).

If we remove frame durations after calculating metrics for the child span, 
wouldn't this leave gaps in the data needed for the root span's calculations?

Shouldn't the root span have access to all frame durations within its time range, 
regardless of whether they've been used in child span calculations?

or maybe I misunderstood how we need to calculate it

buenaflor: btw frames are also being removed if active/running spans is not empty
```dart
    activeSpans.remove(span);
    if (activeSpans.isEmpty) {
      clear();
    } else {
      frames.removeWhere(
          (frameTimestamp, _) => frameTimestamp.isBefore(activeSpans.first.startTimestamp));
    }
```

flutter/lib/src/span_frame_metrics_collector.dart
buenaflor: I've updated the implementation that also includes partially contained frames similar to android, hopefully it makes sense

flutter/lib/src/span_frame_metrics_collector.dart
stefanosiano: > Shouldn't the root span have access to all frame durations within its time range,
regardless of whether they've been used in child span calculations?

I recall we said to set them only on the current span, but the i found we have to set to [all concurrent spans](https://develop.sentry.dev/sdk/performance/frames-delay/), so it's good as is

flutter/lib/src/span_frame_metrics_collector.dart
```diff
@@ -0,0 +1,253 @@
...
+        effectiveDelay = max(0, frameDuration - expectedFrameDuration);
+      } else if (framePartiallyContainedInSpan) {
+        final intersectionStart = max(frameStartMs, spanStartMs);
+        final intersectionEnd = min(frameEndMs, spanEndMs);
+        effectiveDuration = intersectionEnd - intersectionStart;
+
+        final fullFrameDelay = max(0, frameDuration - expectedFrameDuration);
+        final intersectionRatio = effectiveDuration / frameDuration;
+        effectiveDelay = (fullFrameDelay * intersectionRatio).round();
+      } else {
```
stefanosiano, line 208: So we don't check all frames
```suggestion
      } else if (frameStartMs > spanEndMs) {
        // Other frames will be newer than this span, as frames are ordered
        break;
      } else {
```

stefanosiano (approved):
Guess we are good. Nice!

buenaflor:
@stefanosiano seems like `CVDisplayLink` which I'm using for the macos refresh rate impl is [deprecated](https://developer.apple.com/documentation/corevideo/cvdisplaylinkcreatewithcgdisplay(_:_:)), I'll try to migrate it to the recommended API

buenaflor:
Updated the implementation, works nicely and detects the refresh rate depending on the screen where the window is currently at

flutter/lib/src/span_frame_metrics_collector.dart
```diff
@@ -0,0 +1,256 @@
...
+      return;
+    }
+
+    if (_isTrackingPaused) return;
+
+    if (!_stopwatch.isRunning) {
+      _stopwatch.start();
+    }
+
+    await _frameCallbackHandler.endOfFrame;
```
FujiKinaga, line 123: @buenaflor 
Sorry for commenting after merge pr🙏
If you use "await" in a FrameCallback function, is there any possibility of delaying all actions?

buenaflor: hi, what do you mean by delaying all actions? do you mean blocking?

FujiKinaga: @buenaflor 
In our product, we experienced an issue with delayed screen transition behaviour, which improved when `enableFramesTracking` was set to `false`. This is probably due to blocked processing in the `FrameCallback`.

At 60 fps, this callback is called up to 60 times per second.
You should not use await in FrameCallback!

buenaflor: I see, do you have a way to reproduce this in a sample app? 

I'll have a look how we can optimize this

buenaflor: > At 60 fps, this callback is called up to 60 times per second. You should not use await in FrameCallback!

this is not the reason. using await here is fine.

See this [issue](https://github.com/getsentry/sentry-dart/issues/2298) for info why this is happening

PR #2145, opened 2024-07-04
ci: temporarily pin Dio in tests to a version that works with WASM 

vaind:
The CI is currently broken because of Dio making a breaking change regarding WASM compatibility without a major version bump (See https://github.com/cfug/dio/issues/2266 for some details.) 

Currently, we cannot make a release because tests started to fail after the latest version was picked up in CI.  We don't want to change the dependency constraint because most users aren't using WASM compilation so we don't want to prevent them from updating Dio. Pinning in tests to a lower version works fine.

CI passes

## :pencil: Checklist

- [ ] All tests passing
- [ ] No breaking changes

### DIFF
--- a/dio/pubspec_overrides.yaml
+++ b/dio/pubspec_overrides.yaml
@@ -1,3 +1,5 @@
 dependency_overrides:
   sentry:
     path: ../dart
+  # Temporary, see https://github.com/cfug/dio/issues/2266
+  dio: 5.4.3+1


AlexV525:
We've published Dio v5.5.0+1 to solve the issue.

buenaflor (approved):

buenaflor:
ci looks fine, guess we don't need this pr anymore

PR #2152, opened 2024-07-08
Fix truncated stacktraces in unhandled errors

buenaflor:
## :scroll: Description

When the stacktrace we receive from an unhandled error is either empty or null we internally call `StackTrace.current` to grab it but it is truncated because the async nature of the error handling in the integration might be creating a new execution context, losing the original stack.

What we currently have (doesn't help us debug the issue at all since we only see the frame up to a certain point): 
Screenshot 2024-07-09 at 12 46 05

What this fix improves on (now we can see the full context on where it happened):
Screenshot 2024-07-09 at 12 50 12

This will most definitely change grouping for these (although I don't expect them to be the usual type of event)

Manual

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ### Fixes
 
+- Now captures meaningful stack traces when unhandled errors have empty or missing stack traces ([#2152](https://github.com/getsentry/sentry-dart/pull/2152))
+  - This will affect grouping for unhandled errors that have empty or missing stack traces.
 - App starts hanging for 30s ([#2140](https://github.com/getsentry/sentry-dart/pull/2140))
   - Time out for app start info retrieval has been reduced to 10s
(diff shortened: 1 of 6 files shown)


stefanosiano (approved):
Looks good
Can you also check the result with captureMessage (both sync and async) to understand if there are issues with the current grouping (if they return different stacktraces)?

buenaflor:
@stefanosiano the sentry view says that it crashed in non-app stacktrace utils ...etc  which is not true and is misleading, it didn't crash there, is there a way to prevent that from showing? do we need to annotate those frames somehow?

or should we just trim away the sentry frames in those cases?

stefanosiano:
> @stefanosiano the sentry view says that it crashed in non-app stacktrace utils ...etc which is not true and is misleading, it didn't crash there, is there a way to prevent that from showing? do we need to annotate those frames somehow?
> 
> or should we just trim away the sentry frames in those cases?

I don't think we should trim sentry frames. 
The best way would be to correctly mark in-app and system frames.
Anyway, I'm ok to merge it 🚀

PR #2154, opened 2024-07-09
Record dropped spans in client reports

### LINKED ISSUE #2137: Report dropped spans in client reports
### Description

Context: https://github.com/getsentry/projects/issues/70

Cases we need to implement this in Dart/Flutter:
All platforms
- sampling
- beforeSend / beforeTransaction
- event processors

Web, Linux, Windows:
- rate limiter
- network error

For iOS, macOS and Android the rate limiter and network error is implemented by the native sdks

```[tasklist]
### Fixed by
- [x] Cocoa
- [ ] https://github.com/getsentry/sentry-dart/pull/2154
- [ ] https://github.com/getsentry/sentry-dart/pull/2144
```

buenaflor:
## :scroll: Description

Reports the dropped spans as well in client reports, not only the transaction

For all platforms this change reports dropped spans through:
- sampling
- beforeSend/beforeSendTransaction -> either dropping the whole transaction or keeping the transaction but only dropping a few spans
- event processors with same report behaviour as beforeSend

For Web, Linux, Windows (through HttpTransport):
- rate limiter
- network errors

rate limit or network errors on ios, macos or android are handled by the native sdks

Related #2137 

Unit tests

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/dart/lib/src/sentry_client.dart
+++ b/dart/lib/src/sentry_client.dart
@@ -404,7 +404,7 @@ class SentryClient {
     SentryEvent event,
     Hint hint,
   ) async {
-    SentryEvent? eventOrTransaction = event;
+    SentryEvent? processedEvent = event;
     final spanCountBeforeCallback =
         event is SentryTransaction ? event.spans.length : 0;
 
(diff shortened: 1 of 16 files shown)


dart/lib/src/sentry_envelope_item.dart
```diff
@@ -12,7 +12,10 @@ import 'sentry_user_feedback.dart';
 
 /// Item holding header information and JSON encoded data.
 class SentryEnvelopeItem {
-  SentryEnvelopeItem(this.header, this.dataFactory);
+  /// The original, non-encoded object, used when direct access to the source data is needed.
+  Object? originalObject;
```
buenaflor, line 16: used so we don't have to deserialize the envelope when we want to access the number of spans in a transaction for example

buenaflor:
@stefanosiano hey could you review this one for me since you did the implementation on java/android as well pls :pray:

dart/lib/src/sentry_client.dart
```diff
@@ -475,7 +495,8 @@ class SentryClient {
         }
       }
       if (processedEvent == null) {
-        _recordLostEvent(event, DiscardReason.eventProcessor);
+        _options.recorder
+            .recordLostEvent(DiscardReason.eventProcessor, _getCategory(event));
         _options.logger(SentryLevel.debug, 'Event was dropped by a processor');
         break;
       }
```
stefanosiano, line 518: Can you add the same logic of the beforeSend callback here?
report discarded spans from a processor even if the transaction is not dropped 
(note that computation should continue in that case, so no need to `break`)

stefanosiano (approved):
Very good!
Just add the case of dropped span in event processor and you're good to go!

PR #2163, opened 2024-07-11
feat: replay breadcrumbs (android)

vaind:
## :scroll: Description

Adds breadcrumbs to replays

#skip-changelog

## :pencil: Checklist

- [ ] All tests passing
- [ ] No breaking changes

## :crystal_ball: Next steps
* [ ] Improve touch events in general - currently, they're only added for widgets that have keys
* [ ] improve network breadcrumbs - we can capture request and response size

### DIFF
--- a/dart/test/protocol/breadcrumb_test.dart
+++ b/dart/test/protocol/breadcrumb_test.dart
@@ -86,7 +86,7 @@ void main() {
           level: SentryLevel.fatal,
           reason: 'OK',
           statusCode: 200,
-          requestDuration: Duration.zero,
+          requestDuration: Duration(milliseconds: 55),
           timestamp: DateTime.now(),
           requestBodySize: 2,
           responseBodySize: 3,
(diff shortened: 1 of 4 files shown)


flutter/android/src/main/kotlin/io/sentry/flutter/SentryFlutterPlugin.kt
```diff
@@ -425,7 +425,7 @@ class SentryFlutterPlugin : FlutterPlugin, MethodCallHandler, ActivityAware {
     if (args.isNotEmpty()) {
       val event = args.first() as ByteArray?
       val containsUnhandledException = args[1] as Boolean
-      if (event != null && event.isNotEmpty() && containsUnhandledException != null) {
```
vaind, line 428: never null because of the cast above

dart/test/protocol/breadcrumb_test.dart
```diff
@@ -103,17 +103,43 @@ void main() {
           'method': 'GET',
           'status_code': 200,
           'reason': 'OK',
-          'duration': '0:00:00.000000',
+          'duration': '0:00:00.055000',
```
bruno-garcia, line 106: isnt' duration an int?

@romtsn @brustolin ?

romtsn: it's not rrweb crumb but sentry crumb here, so the data bag is arbitrary

vaind: yes, this is not getting to rrweb

buenaflor (approved):

PR #2170, opened 2024-07-15
Implement error type identifier to mitigate obfuscated Flutter issue titles

buenaflor:
## :scroll: Description

Experimented a bit with improving the flutter issues title

This won't fix it for all error/exception types but at least we can by default support the standard ones and add more for specific integrations.

If there are still types that we don't support the user can extend it.

### Before

image

### After

image

This will be enabled by default and users have the option to opt-out

Improve upon Flutter issues title

## :pencil: Checklist

- [ ] All tests passing
- [ ] No breaking changes


dart/lib/src/dart_error_type_identifier.dart
```diff
@@ -0,0 +1,19 @@
...
+    if (error is TypeError) return 'TypeError';
+    if (error is ArgumentError) return 'ArgumentError';
+    if (error is StateError) return 'StateError';
+    if (error is UnsupportedError) return 'UnsupportedError';
+    if (error is UnimplementedError) return 'UnimplementedError';
+    if (error is ConcurrentModificationError)
+      return 'ConcurrentModificationError';
+    if (error is OutOfMemoryError) return 'OutOfMemoryError';
+    if (error is StackOverflowError) return 'StackOverflowError';
+    return null;
```
vaind: There are a few more common types: https://api.dart.dev/stable/3.4.4/dart-core/dart-core-library.html#exceptions

ueman:
A small note: This approach disables tree shaking for every referenced exception type. Since these are all coming from within the framework, the impact is likely rather small, I guess.

stefanosiano (approved):
nice!

stefanosiano:
> A small note: This approach disables tree shaking for every referenced exception type. Since these are all coming from within the framework, the impact is likely rather small, I guess.

For reference: we are now importing only the exceptions from the libraries, through `import ... show ...`, to avoid any potential issue

CHANGELOG.md
```diff
@@ -4,6 +4,29 @@
 
 ### Improvements
 
+- Add error type identifier to improve obfuscated Flutter issue titles ([#2170](https://github.com/getsentry/sentry-dart/pull/2170))
+  - Example: transforms issue titles from `GA` to `FlutterError` or `minified:nE` to `FlutterError`
+  - This is enabled automatically and will change grouping if you already have issues with obfuscated titles
+  - You can add your custom exception identifier if there are exceptions that we do not identify out of the box
+  - If you want to disable this feature, set `enableExceptionTypeIdentification` to `false` in your Sentry options
```
kahest, line 10: it's great that we have this option 👍 I would put this note one bullet above, so that it's connected to the callout about being opt out and that it changes (fixes!) grouping

buenaflor:
> we are now importing only the exceptions from the libraries, through import ... show ..., to avoid any potential issue

afaik it shouldn't directly affect tree shaking but it's good in this case nonetheless so we can be a bit more explicit about what we import in that class

PR #2200, opened 2024-07-26
Bugfix: add-sample-rate-native

JulianBissekkou:
## :scroll: Description
During development I discovered that the sample rate is not correctly forwarded to the native sdks.

This PR makes sure that initialization of the native sdks also uses the configured sampleRate and traceSampleRate from the FlutterSentryOptions.

I tested this by setting a breakpoint and ensuring that my configuration was applied to the native options. If there is a better way, let me know :) 

## :pencil: Checklist

- [ ] All tests passing
- [x] No breaking changes


JulianBissekkou:
I checked "No breaking changes" since I think this is not breaking. However, the sdk might behave differently but I would say its no the correct expected behavior.

buenaflor:
hey, do you use the native sdk alongside the flutter sdk?

buenaflor:
I'm trying to find the context as to why this was not added before. It's hard to imagine that this was just an oversight but it could be.

buenaflor:
Example: our react native SDK also doesn't set the sample rates during init: https://github.com/getsentry/sentry-react-native/blob/979a258538e6b140d912f7d74b89f684c504b67f/android/src/main/java/io/sentry/react/RNSentryModuleImpl.java#L182

JulianBissekkou:
@buenaflor 
I am only using the flutter sdk and my application is running on android and iOS and I was curious how the implementation of the initialization is done and then I also found that some of the properties where not parsed.

I was also suprised, but my assumption was that nobody found this since most react native and flutter developers are interested in the errors that are thrown in the TS or Dart code and the native exceptions are very rare. 
The missing change in the sampleRate is therefore nothing that most people would directly notice since it doesn't affect many events.

I wasn't sure if there is some kind of magic initialization that I was missing so take your time with reviewing this. 
Looking forward hearing from you :)

buenaflor:
After some discussion I don't think this should be added since there's the possibility of double instrumentation since the native sdks themselves have instrumentation for certain integrations. For example the cocoa sdk has all these [automatic instrumentation](https://docs.sentry.io/platforms/apple/guides/ios/tracing/instrumentation/automatic-instrumentation/)  which might result in duplicate data being sent. 

In terms of feature set and design we also want to keep our hybrid sdks as consistent as possible to one another.

So adding this change brings a lot of implications and is actually a rather big decision.

JulianBissekkou:
@buenaflor 
Can you explain why we get duplicate instrumentation? We are only updating configuration parameters of the iOS sdks. This can also be done by developers that are using the native iOS integration. The flutter/dart sdk will only provide data to sentry for the dart specific part.

buenaflor:
@JulianBissekkou if we pass `tracesSampleRate` as well then we would get duplicate appstart performance transactions, one implemented in native ios/android and one implemented in flutter. Just one of many examples.

here is what we should do: I agree that `sampleRate` should be passed, this controls errors and crashes so it makes sense to pass it. but `tracesSampleRate` should stay as it is because of the mentioned reason.

flutter/android/src/main/kotlin/io/sentry/flutter/SentryFlutter.kt
```diff
@@ -119,6 +119,12 @@ class SentryFlutter(
     data.getIfNotNull<Int>("readTimeoutMillis") {
       options.readTimeoutMillis = it
     }
+    data.getIfNotNull<Double>("sampleRate") {
+      options.sampleRate = it
+    }
+    data.getIfNotNull<Double>("tracesSampleRate") {
+      options.tracesSampleRate = it
+    }
```
buenaflor:
```suggestion
```

flutter/lib/src/native/sentry_native_channel.dart
```diff
@@ -66,6 +67,8 @@ class SentryNativeChannel
       'readTimeoutMillis': options.readTimeout.inMilliseconds,
       'appHangTimeoutIntervalMillis':
           options.appHangTimeoutInterval.inMilliseconds,
+      'tracesSampleRate': options.tracesSampleRate,
```
buenaflor:
```suggestion
```

flutter/ios/Classes/SentryFlutter.swift
```diff
@@ -70,6 +70,12 @@ public final class SentryFlutter {
         if let appHangTimeoutIntervalMillis = data["appHangTimeoutIntervalMillis"] as? NSNumber {
             options.appHangTimeoutInterval = appHangTimeoutIntervalMillis.doubleValue / 1000
         }
+        if let sampleRate = data["sampleRate"] as? NSNumber? {
+            options.sampleRate = sampleRate
+        }
+        if let tracesSampleRate = data["tracesSampleRate"] as? NSNumber? {
+            options.tracesSampleRate = tracesSampleRate
+        }
```
buenaflor:
```suggestion
```

JulianBissekkou:
@buenaflor Thanks for explaining.
I applied your suggestions 👍🏽

buenaflor:
@JulianBissekkou some tests are failing but it should be a very simple fix (you can ignore the danger ci)

also could you add this changelog entry under `# Fixes` in the unreleased section

```
- Pass `sampleRate` to native SDKs ([#2200](https://github.com/getsentry/sentry-dart/pull/2200))
```

denrase (changes requested):
Thank you for contributing! 

- [ ] Please also add tests for the new values our native tests `SentryFlutterTests.swift` and `SentryFlutterTest.kt`.
- [ ] Update the test expectations in `init_native_sdk_test.dart` to include the new value

JulianBissekkou:
I will make this PR ready later this week :) Thanks for your feedback

denrase (approved):
LGTM, only issue left is a linting issue in the Kotlin file:

```
SentryFlutterTest.kt:162:28: error: Missing trailing comma before ")"
```

JulianBissekkou:
Closing as other PRs have already implemented that feature.

PR #2207, opened 2024-08-01
Feat/add ignore transactions and ignore errors #1391

### LINKED ISSUE #1391: Add ignoreTransactions and ignoreErrors
### Background:
To further simplfiy ignoring certain events from being sent to sentry, it is also suggested to provide `ignoreTransactions` and `ignoreErrors`

More info here develop docs updated: https://develop.sentry.dev/sdk/features/#event-sampling

- [ ] ignoreErrors
- [ ] ignoreTransactions
- [ ] SDK docs

Info: don't forget about client reports

martinhaintz:
## :scroll: Description
Closes #1391

In the Simulator and wrote test cases.

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## 8.6.0
+## Unreleased
 
 ### Features 
 - Add `ignoreTransactions` and `ignoreErrors` to options ([#2207](https://github.com/getsentry/sentry-dart/pull/2207))
@@ -15,6 +15,8 @@
     appRunner: () => runApp(MyApp()),
   );
   ```
+
+## 8.6.0
 
 ### Improvements
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Unreleased 
+- Add ignoreTransactions and ignoreErrors to options ([#1391](https://github.com/getsentry/sentry-dart/issues/1391)) ([#2207](https://github.com/getsentry/sentry-dart/pull/2207))
 
 ### Improvements
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,17 @@
 
 ### Features 
 - Add `ignoreTransactions` and `ignoreErrors` to options ([#2207](https://github.com/getsentry/sentry-dart/pull/2207))
+  ```dart
+  await SentryFlutter.init(
+    (options) {
+      options.dsn = 'https://examplePublicKey@o0.ingest.sentry.io/0';
+      options.ignoreErrors = ["my-error", "error-.*"];
+      options.ignoreTransactions = ["my-transaction", "transaction-.*"];
+      ...
+    },
+    appRunner: () => runApp(MyApp()),
+  );
+  ```
 
 ### Improvements
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,8 +10,8 @@
   await SentryFlutter.init(
     (options) {
       options.dsn = 'https://examplePublicKey@o0.ingest.sentry.io/0';
-      options.ignoreErrors = ["my-error", "^error\-.*\$"];
-      options.ignoreTransactions = ["my-transaction", "^transaction\-.*\$"];
+      options.ignoreErrors = ["my-error", "^error-.*\$"];
+      options.ignoreTransactions = ["my-transaction", "^transaction-.*\$"];
       ...
     },
     appRunner: () => runApp(MyApp()),
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,9 @@
 # Changelog
 
 ## Unreleased 
-- Add ignoreTransactions and ignoreErrors to options ([#1391](https://github.com/getsentry/sentry-dart/issues/1391)) ([#2207](https://github.com/getsentry/sentry-dart/pull/2207))
+
+### Features 
+- Add `ignoreTransactions` and `ignoreErrors` to options ([#2207](https://github.com/getsentry/sentry-dart/pull/2207))
 
 ### Improvements
 
--- a/dart/lib/src/sentry_client.dart
+++ b/dart/lib/src/sentry_client.dart
@@ -90,7 +90,7 @@ class SentryClient {
         'Error was ignored as specified in the ignoredErrors options.',
       );
       _options.recorder
-          .recordLostEvent(DiscardReason.beforeSend, _getCategory(event));
+          .recordLostEvent(DiscardReason.ignored, _getCategory(event));
       return _emptySentryId;
     }
 
--- a/dart/lib/src/sentry_client.dart
+++ b/dart/lib/src/sentry_client.dart
@@ -191,7 +191,7 @@ class SentryClient {
   }
 
   bool _isIgnoredError(SentryEvent event) {
-    if (event.message == null) {
+    if (event.message == null || _options.ignoreErrors.isEmpty) {
       return false;
     }
     final combinedRegexPattern = _options.ignoreErrors.join('|');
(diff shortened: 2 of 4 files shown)


dart/lib/src/sentry_options.dart
```diff
@@ -184,6 +184,43 @@ class SentryOptions {
   /// sent. Events are picked randomly. Default is null (disabled)
   double? sampleRate;
 
+  /// The ignoreErrors tells the SDK which errors should be not sent to the sentry server.
+  /// If an null or an empty list is used, the SDK will send all transactions.
+  Iterable<String>? ignoreErrors;
+
+  bool isIgnoredError(SentryEvent event) {
```
buenaflor: we can move both `isIgnoredError` and `isIgnoredTransaction` to `sentry_client.dart` and mark them as private like so `_isIgnoredError(..)` `_isIgnoredTransaction(..)`

dart/lib/src/sentry_options.dart
```diff
@@ -184,6 +184,43 @@ class SentryOptions {
   /// sent. Events are picked randomly. Default is null (disabled)
   double? sampleRate;
 
+  /// The ignoreErrors tells the SDK which errors should be not sent to the sentry server.
+  /// If an null or an empty list is used, the SDK will send all transactions.
+  Iterable<String>? ignoreErrors;
```
buenaflor: Let's use `List`, I don't think it really matters much but we can keep it consistent with the rest of the code

dart/lib/src/sentry_options.dart
```diff
@@ -184,6 +184,43 @@ class SentryOptions {
...
+    bool ignore = regExp.firstMatch(event.message!.formatted) != null;
+
+    return ignore;
+  }
+
+  /// The ignoreTransactions tells the SDK which transactions should be not sent to the sentry server.
+  /// If null or an empty list is used, the SDK will send all transactions.
+  Iterable<String>? ignoreTransactions;
+
+  bool isIgnoredTransaction(SentryTransaction transaction) {
```
buenaflor: same comments as above

dart/lib/src/sentry_client.dart
```diff
@@ -84,6 +84,16 @@ class SentryClient {
     dynamic stackTrace,
     Hint? hint,
   }) async {
+    if (_options.isIgnoredError(event)) {
+      _options.logger(
+        SentryLevel.debug,
+        'Error was ignored as specified in the ignoredErrors options.',
+      );
+      _options.recorder
+          .recordLostEvent(DiscardReason.beforeSend, _getCategory(event));
```
buenaflor:
```suggestion
          .recordLostEvent(DiscardReason.ignored, _getCategory(event));
```

dart/lib/src/sentry_client.dart
```diff
@@ -351,6 +361,14 @@ class SentryClient {
       return _emptySentryId;
     }
 
+    if (_options.isIgnoredTransaction(preparedTransaction)) {
+      _options.logger(
```
buenaflor, line 374: client report recorder is missing

dart/test/sentry_client_test.dart
```diff
@@ -1033,6 +1033,42 @@ void main() {
...
   });
 
+  group('SentryClient ignored errors', () {
+    late Fixture fixture;
+
+    setUp(() {
+      fixture = Fixture();
+      fixture.options.ignoreErrors = ["my-error", "error-.*"];
+    });
+
```
buenaflor, line 1043: let's also add a case where it's partially matching 

e.g the message is "partially matching my-error" and it should ignore it because it matched my-error

dart/lib/src/sentry_options.dart
martinhaintz: good point, will change this.

dart/lib/src/sentry_client.dart
martinhaintz: i will add it.

dart/test/sentry_client_test.dart
```diff
@@ -1033,6 +1033,97 @@ void main() {
...
+    test('ignore Error "my-error"', () async {
```
buenaflor: just a nitpick: generally we tend to name the tests more generic such as 

`drop event if error message fully matches ignoreError value`

or something like that, so we don't tie the test name directly to the tested value

dart/lib/src/sentry_options.dart
```diff
@@ -184,6 +184,14 @@ class SentryOptions {
   /// sent. Events are picked randomly. Default is null (disabled)
   double? sampleRate;
 
+  /// The ignoreErrors tells the SDK which errors should be not sent to the sentry server.
+  /// If an null or an empty list is used, the SDK will send all transactions.
+  List<String>? ignoreErrors;
```
buenaflor:
```suggestion
  List<String> ignoreErrors = [];
```

we can keep this by default as empty so we save ourselves from null checks and !

same for `ignoreTransactions`

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Unreleased 
+- Add ignoreTransactions and ignoreErrors to options ([#1391](https://github.com/getsentry/sentry-dart/issues/1391)) ([#2207](https://github.com/getsentry/sentry-dart/pull/2207))
```
```suggestion

### Features 
- Add `ignoreTransactions` and `ignoreErrors` to options ([#2207](https://github.com/getsentry/sentry-dart/pull/2207))
```

linking the PR should be good enough. 

also I like to add a little code example below to show users how they could utilize these options.

example: https://github.com/getsentry/sentry-dart/blob/main/CHANGELOG.md#features-7

dart/lib/src/sentry_client.dart
```diff
@@ -379,6 +415,21 @@ class SentryClient {
...
+    if (_options.ignoreTransactions == null ||
+        _options.ignoreTransactions!.isEmpty) {
+      return false;
+    }
+
+    String combinedRegexPattern = _options.ignoreTransactions!.join('|');
+
+    RegExp regExp = RegExp(combinedRegexPattern);
+
+    bool ignore = regExp.firstMatch(transaction.tracer.name) != null;
```
buenaflor:
```suggestion
    final combinedRegexPattern = _options.ignoreTransactions!.join('|');

    final regExp = RegExp(combinedRegexPattern);

    final ignore = regExp.firstMatch(transaction.tracer.name) != null;
```

let's keep immutability as default, same for `_isIgnoredError`

buenaflor (commented):
nice 👍  left a couple comments but looks good!

buenaflor:
@martinhaintz there are still tests failing from the change

dart/test/sentry_client_test.dart
```diff
@@ -1033,6 +1033,128 @@ void main() {
...
+      await client.captureEvent(event);
+
+      expect((fixture.transport).called(1), true);
+    });
+
+    test(
+        'send event if error message does only partially match ignoreErrors value',
+        () async {
+      final event =
+          SentryEvent(message: SentryMessage("partially matching my-error"));
```
buenaflor: maybe it's just my misunderstanding, shouldn't this case be dropped?

From the js sdk docs:
> When using strings, partial matches will be filtered out

I might be wrong though

dart/test/sentry_client_test.dart
```diff
@@ -1033,6 +1033,128 @@ void main() {
...
+
+    test(
+        'send transaction if name does only partially match ignoreTransaction value',
+        () async {
+      final client = fixture.getSut();
+      final fakeTransaction = fixture.fakeTransaction();
+      fakeTransaction.tracer.name = "partially matching my-transaction";
+      await client.captureTransaction(fakeTransaction);
+
+      expect((fixture.transport).called(1), true);
```
buenaflor: same question as above

dart/test/sentry_client_test.dart
buenaflor: @martinhaintz

martinhaintz: you are right. I fixed the unit-tests and changed them more clearly. I already changed the docs too.

buenaflor (approved):

PR #2209, opened 2024-08-01
feat: iOS replay support

vaind:
## :scroll: Description

adds iOS to the supported platforms for replay.

part of #1193 

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes


flutter/ios/Classes/SentryFlutterReplayScreenshotProvider.m
```diff
@@ -0,0 +1,46 @@
...
+}
+
+- (instancetype _Nonnull)initWithChannel:
+    (FlutterMethodChannel *_Nonnull)channel {
+  if (self = [super init]) {
+    self->channel = channel;
+  }
+  return self;
+}
+
```
buenaflor, line 18: curious why we need to implement screenshot provider and breadcrumb converter in objc, is it for performance?

flutter/lib/src/native/cocoa/sentry_native_cocoa.dart
```diff
@@ -1,17 +1,76 @@
...
+              _replayId = replayId;
+              hub.configureScope((s) {
+                // ignore: invalid_use_of_internal_member
+                s.replayId = replayId;
+              });
+            }
+
+            Uint8List? imageBytes;
+            await _replayRecorder?.capture((image) async {
+              var imageData =
```
buenaflor:
```suggestion
              final imageData =
```

flutter/ios/Classes/SentryFlutterPluginApple.swift
```diff
@@ -174,6 +179,14 @@ public class SentryFlutterPluginApple: NSObject, FlutterPlugin {
         case "resumeAppHangTracking":
             resumeAppHangTracking(result)
 
+        case "sendReplayForEvent":
+#if canImport(UIKit) && !SENTRY_NO_UIKIT && (os(iOS) || os(tvOS))
```
buenaflor, line 183: `|| os(tvOS)` 

we also support replay for tvos?

buenaflor (approved):
looks good, just a couple minor comments/questions

flutter/ios/Classes/SentryFlutterPluginApple.swift
vaind: no (it's not even supported by Flutter officially AFAIK), the condition is exactly the same as in cocoa core code

flutter/ios/Classes/SentryFlutterReplayScreenshotProvider.m
vaind: the types we need from cocoa are not visible outside their module so we need to do it in objc which ignores swift visibility rules.

PR #2214, opened 2024-08-05
feat: add span level measurements #1855

### LINKED ISSUE #1855: Support span level measurements
### Description

https://github.com/getsentry/team-mobile/issues/172

Example of Java SDK implementation: https://github.com/getsentry/sentry-java/pull/3219

martinhaintz:
## :scroll: Description
Add support for span level measurements #1855 

Closes #1855  

Unittests

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Improvements
 
-- Add support for span level measurements.
+- Add support for span level measurements. ([#1855](https://github.com/getsentry/sentry-dart/issues/1855))
 - Add error type identifier to improve obfuscated Flutter issue titles ([#2170](https://github.com/getsentry/sentry-dart/pull/2170))
   - Example: transforms issue titles from `GA` to `FlutterError` or `minified:nE` to `FlutterError`
   - This is enabled automatically and will change grouping if you already have issues with obfuscated titles
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Improvements
 
+- Add support for span level measurements.
 - Add error type identifier to improve obfuscated Flutter issue titles ([#2170](https://github.com/getsentry/sentry-dart/pull/2170))
   - Example: transforms issue titles from `GA` to `FlutterError` or `minified:nE` to `FlutterError`
   - This is enabled automatically and will change grouping if you already have issues with obfuscated titles
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,13 @@
 # Changelog
 
+## Unreleased
+
+- Add support for span level measurements. ([#1855](https://github.com/getsentry/sentry-dart/issues/1855))
+
 ## 8.6.0
 
 ### Improvements
 
-- Add support for span level measurements. ([#1855](https://github.com/getsentry/sentry-dart/issues/1855))
 - Add error type identifier to improve obfuscated Flutter issue titles ([#2170](https://github.com/getsentry/sentry-dart/pull/2170))
   - Example: transforms issue titles from `GA` to `FlutterError` or `minified:nE` to `FlutterError`
   - This is enabled automatically and will change grouping if you already have issues with obfuscated titles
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### Improvements
 
+- Add support for span level measurements. ([#1855](https://github.com/getsentry/sentry-dart/issues/1855))
```
buenaflor: You always need to add new changelog entries under the `## Unreleased` section, the `Danger` CI also warns you with a failure

if the unreleased section doesn't exist (because there has been a release recently), you have to create one e.g

```markdown
## Unreleased

### Features

- Add support for span level measurements. ([#1855](https://github.com/getsentry/sentry-dart/issues/1855))
```

dart/lib/src/noop_sentry_span.dart
```diff
@@ -99,4 +99,8 @@ class NoOpSentrySpan extends ISentrySpan {
 
   @override
   LocalMetricsAggregator? get localMetricsAggregator => null;
+
+  @override
+  // TODO: implement measurements
+  Map<String, SentryMeasurement> get measurements => throw UnimplementedError();
```
buenaflor: This should return an empty map

buenaflor (changes requested):

dart/lib/src/sentry_tracer.dart
```diff
@@ -340,11 +338,15 @@ class SentryTracer extends ISentrySpan {
...
-      return;
+    _rootSpan.setMeasurement(name, value, unit: unit);
+  }
+
+  void setMeasurementFromChild(String name, num value,
+      {SentryMeasurementUnit? unit}) {
+    // We don't want to overwrite the root span measurement, if it comes from a child.
+    if (!_rootSpan.measurements.containsKey(name)) {
+      setMeasurement(name, value, unit: unit);
     }
```
buenaflor: what if we handle all the measurements directly in the sentrytracer instead of keeping them in the root span? I think that would be easier so we don't need to handle recursive calls and just do `_tracer.setMeasurement(...)` from the spans

CHANGELOG.md
martinhaintz: sorry, thanks for the hint.

dart/lib/src/sentry_tracer.dart
martinhaintz: makes sense. I will change this.

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+- Add support for span level measurements. ([#2214](https://github.com/getsentry/sentry-dart/pull/2214))
```
buenaflor, line 7:
```suggestion
### Features

- Add support for span level measurements. ([#2214](https://github.com/getsentry/sentry-dart/pull/2214))
```

dart/lib/src/sentry_tracer.dart
```diff
@@ -341,10 +339,19 @@ class SentryTracer extends ISentrySpan {
...
       return;
     }
-    final measurement = SentryMeasurement(name, value, unit: unit);
-    _measurements[name] = measurement;
+    _measurements[name] = SentryMeasurement(name, value, unit: unit);
+  }
+
+  void setMeasurementFromChild(String name, num value,
+      {SentryMeasurementUnit? unit}) {
+    // We don't want to overwrite span measurement, if it comes from a child.
```
buenaflor, line 351: is not overwriting done in the java sdk as well?

martinhaintz: Here is the implementation from the java platform: [SentryTracer.java](https://github.com/getsentry/sentry-java/blob/main/sentry/src/main/java/io/sentry/SentryTracer.java#L761)
`public void setMeasurementFromChild(final @NotNull String name, final @NotNull Number value) {
    // We don't want to overwrite the root span measurement, if it comes from a child.
    if (!root.getMeasurements().containsKey(name)) {
      setMeasurement(name, value);
    }
  }`

buenaflor (approved):

PR #2227, opened 2024-08-12
Support allowUrls, denyUrls

### LINKED ISSUE #626: Support `allowUrls` , `denyUrls`
denyUrls
A list of strings or regex patterns that match error URLs that should not be sent to Sentry. By default, all errors will be sent. This is a "contains" match to the entire file URL. As a result, if you add foo.com to it, it will also match on https://bar.com/myfile/foo.com. By default, all errors will be sent.

allowUrls
A list of strings or regex patterns that match error URLs which should exclusively be sent to Sentry. By default, all errors will be sent. This is a "contains" match to the entire file URL. As a result, if you add foo.com to it, it will also match on https://bar.com/myfile/foo.com. By default, all errors will be sent.

Similar to JS https://docs.sentry.io/platforms/javascript/configuration/options/
https://docs.sentry.io/product/accounts/quotas/manage-event-stream-guide/#inboundfilters

martinhaintz:
## :scroll: Description
Add allowUrls and denyUrls parameter to the options on the web platform. the user can now specify on which urls, events should be ignored.

Close #626 

unittest

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes


dart/lib/src/sentry.dart
```diff
@@ -8,6 +8,7 @@ import 'run_zoned_guarded_integration.dart';
 import 'event_processor/enricher/enricher_event_processor.dart';
 import 'environment/environment_variables.dart';
 import 'event_processor/deduplication_event_processor.dart';
+import 'event_processor/web_url_filter_event_processor.dart';
```
ueman: You need to use a conditional import here, otherwise you'll break Sentry on non-web platforms. You can take a look at other event processors, there's a bunch which use that pattern already.

martinhaintz: Thanks for the hint. I was still working on this issue, but your comment saved me some time.

dart/lib/src/utils/regex_utils.dart
```diff
@@ -0,0 +1,6 @@
+bool isMatchingRegexPattern(String value, List<String> regexPattern,
```
buenaflor, line 4:
```suggestion
@internal
bool isMatchingRegexPattern(String value, List<String> regexPattern,
```

dart/lib/src/sentry_options.dart
```diff
@@ -184,12 +184,29 @@ class SentryOptions {
   /// sent. Events are picked randomly. Default is null (disabled)
   double? sampleRate;
 
+  /// (Web only) Errors only occurring on these Urls will be handled and sent to sentry.
+  /// If an null or an empty list is used, the SDK will send all errors.
```
buenaflor: the list can't be null as it's not nullable

```suggestion
  /// If an empty list is used, the SDK will send all errors.
```

dart/lib/src/sentry_options.dart
```diff
@@ -184,12 +184,29 @@ class SentryOptions {
...
+  /// To use regex add the `^` and the `$` to the string.
+  ///
+  /// If used on a platform other than Web, this setting will be ignored.
+  List<String> allowUrls = [];
+
+  /// (Web only) Errors occurring on these Urls will be ignored and are not sent to sentry.
+  /// If an null or an empty list is used, the SDK will send all errors.
```
buenaflor:
```suggestion
  /// If an empty list is used, the SDK will send all errors.
```

dart/lib/src/event_processor/url_filter/web_url_filter_event_processor.dart
```diff
@@ -0,0 +1,36 @@
...
+    if (isMatchingRegexPattern(url, _options.allowUrls)) {
+      if (_options.denyUrls.isNotEmpty &&
+          isMatchingRegexPattern(url, _options.denyUrls)) {
+        return null;
+      } else {
+        return event;
+      }
+    } else {
+      return null;
+    }
```
buenaflor: can we split this condition up?

something like this

```dart
if (_options.allowUrls.isNotEmpty &&
    !isMatchingRegexPattern(url, _options.allowUrls)) {
  return null;
}

if (_options.denyUrls.isNotEmpty &&
    isMatchingRegexPattern(url, _options.denyUrls)) {
  return null;
}
```

dart/lib/src/sentry_options.dart
```diff
@@ -184,12 +184,29 @@ class SentryOptions {
...
+  /// In combination with `allowUrls` you can block subdomains of the domains listed in allowUrls.
+  /// To use regex add the `^` and the `$` to the string.
+  ///
+  /// If used on a platform other than Web, this setting will be ignored.
+  List<String> denyUrls = [];
```
buenaflor: I think since these are flutter specific options both options should be moved to `sentry_flutter_options.dart`

buenaflor (changes requested):

dart/lib/src/event_processor/url_filter/web_url_filter_event_processor.dart
martinhaintz: I was also unhappy with this code, but your suggestion looks better. I will implement this.

martinhaintz:
Docs updated here: [11067](https://github.com/getsentry/sentry-docs/pull/11067)

flutter/lib/src/event_processor/url_filter/web_url_filter_event_processor.dart
```diff
@@ -0,0 +1,40 @@
+// We would lose compatibility with old dart versions by adding web to pubspec.
+// ignore: depend_on_referenced_packages
+import 'package:web/web.dart' as web show window, Window;
```
buenaflor, line 3: since we are accessing `window`, do we need to create two separate event processors here, a html event processor and web event processor to keep compat with wasm? @vaind

flutter/test/event_processor/url_filter_event_processor_test.dart
```diff
@@ -0,0 +1,173 @@
+import 'package:flutter/foundation.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:sentry_flutter/sentry_flutter.dart';
+import 'package:sentry_flutter/src/event_processor/url_filter/url_filter_event_processor.dart';
+
+// can be tested on command line with
+// `flutter test --platform=chrome test/event_processor/url_filter_event_processor_test.dart`
+void main() {
+  if (kIsWeb) {
```
vaind: How about we use 
```
@TestOn('browser')
```

flutter/lib/src/event_processor/url_filter/web_url_filter_event_processor.dart
vaind: I'd have expected CI to fail as it is. Actually, it seems like we don't run wasm tests on flutter test. Let me fix that: https://github.com/getsentry/sentry-dart/pull/2231

dart/lib/src/sentry_options.dart
```diff
@@ -186,10 +186,12 @@ class SentryOptions {
 
   /// The ignoreErrors tells the SDK which errors should be not sent to the sentry server.
   /// If an null or an empty list is used, the SDK will send all transactions.
+  /// To use regex add the `^` and the `$` to the string.
   List<String> ignoreErrors = [];
 
   /// The ignoreTransactions tells the SDK which transactions should be not sent to the sentry server.
   /// If null or an empty list is used, the SDK will send all transactions.
+  /// To use regex add the `^` and the `$` to the string.
```
vaind, line 194: this implies that if you don't add these characters, regex won't be used. That is not the case from what I can tell - to me it looks like we always use regex

martinhaintz: yes, you are right. I will change this.

scripts/publish_validation/bin/publish_validation.dart
```diff
@@ -34,7 +34,8 @@ void main(List<String> arguments) async {
     'lib/src/integrations/connectivity/web_connectivity_provider.dart: This package does not have web in the `dependencies` section of `pubspec.yaml`',
     'lib/src/event_processor/enricher/web_enricher_event_processor.dart: This package does not have web in the `dependencies` section of `pubspec.yaml`',
     'lib/src/origin_web.dart: This package does not have web in the `dependencies` section of `pubspec.yaml`',
-    'lib/src/platform/_web_platform.dart: This package does not have web in the `dependencies` section of `pubspec.yaml`'
+    'lib/src/platform/_web_platform.dart: This package does not have web in the `dependencies` section of `pubspec.yaml`',
+    'lib/src/event_processor/url_filter/web_url_filter_event_processor.dart: This package does not have web in the dependencies section of pubspec.yaml',
```
buenaflor:
```suggestion
    'lib/src/event_processor/url_filter/web_url_filter_event_processor.dart: This package does not have web in the `dependencies` section of `pubspec.yaml`',
```

flutter/lib/src/event_processor/url_filter/html_url_filter_event_processor.dart
```diff
@@ -0,0 +1,38 @@
+import 'dart:html' as html show window, Window;
+
+import '../../../sentry_flutter.dart';
+import 'url_filter_event_processor.dart';
+// ignore: implementation_imports
+import 'package:sentry/src//utils/regex_utils.dart';
```
buenaflor: ci is complaining 

```suggestion
import 'package:sentry/src/utils/regex_utils.dart';
```

flutter/lib/src/event_processor/url_filter/web_url_filter_event_processor.dart
```diff
@@ -0,0 +1,40 @@
...
+
+import '../../../sentry_flutter.dart';
+import 'url_filter_event_processor.dart';
+// ignore: implementation_imports
+import 'package:sentry/src//utils/regex_utils.dart';
```
buenaflor:
```suggestion
import 'package:sentry/src/utils/regex_utils.dart';
```

flutter/lib/src/event_processor/url_filter/html_url_filter_event_processor.dart
```diff
@@ -0,0 +1,38 @@
...
+    this._options,
+  );
+
+  final html.Window _window = html.window;
+
+  final SentryFlutterOptions _options;
+
+  @override
+  SentryEvent? apply(SentryEvent event, Hint hint) {
+    final url = event.request?.url ?? _window.location.toString();
```
buenaflor: ~I don't think `event.request?.url` is correct here since it is the http request url and not the current window url~ We have to check if it plays nicely with failed http requests for example since the request field is also used in those scenarios

we may have to use the stackframes to receive the url if we don't have other options

flutter/lib/src/event_processor/url_filter/web_url_filter_event_processor.dart
```diff
@@ -0,0 +1,40 @@
...
+    this._options,
+  );
+
+  final web.Window _window = web.window;
+
+  final SentryFlutterOptions _options;
+
+  @override
+  SentryEvent? apply(SentryEvent event, Hint hint) {
+    final url = event.request?.url ?? _window.location.toString();
```
buenaflor: same comment as above

buenaflor (changes requested):

flutter/lib/src/event_processor/url_filter/html_url_filter_event_processor.dart
martinhaintz: yes this is true. I thought this is part of this feature, but we should keep this in mind if some problems or discussions are occurring.

buenaflor: did you check if this is this not an issue?

buenaflor: cc @martinhaintz

martinhaintz: the unit tests are all passing, so from this perspective it works. 
in the issue #626 it is not exactly defined, if only the current web application with a specific url or also the request which are sent to a specific url should be allowed/denied.
it looks like the javascript implementation distinguish between the application url which can be set in [`global options`](https://docs.sentry.io/platforms/javascript/configuration/options/#deny-urls) and scripts which are loaded from other sources in the [`inboundFilters`](https://docs.sentry.io/platforms/javascript/configuration/integrations/inboundfilters/#allowurls). 
It is not applying the rules based on request to other resources. But this can also be a handy feature.

Any suggestions how we should continue?

buenaflor: > It is not applying the rules based on request to other resources

this was my understanding here as well, let's focus on current web application urls first and foremost. so that's why I think `event.request?.url` won't directly work since it also includes http request urls

buenaflor: as far as I know the most accurate way to get the information is going through the stackframes because there we will know where an error happened

martinhaintz: checking `_window.location.toString()`will not be sufficient?

buenaflor: depends, let's take this example: denyUrl: example.com/deny

user triggers an error in example.com (which should be allowed) and then quickly navigates to example.com/deny but the event processing is not done yet and fetches `window.location` at a later time which would be example.com/deny and now we think we have to reject this error

just some thoughts, it might not be true and might be worth investigating, using `window.location` is definitely easier

but in any case we definitely know the correct path would be in the stackframes regardless of when the event would be processed

martinhaintz: good example. I didn't thought about that edge case. I will change this.

buenaflor: also I think we'll have to check the absolute path of the stackframes, afaik that includes the url in flutter web

flutter/lib/src/event_processor/url_filter/html_url_filter_event_processor.dart
```diff
@@ -0,0 +1,54 @@
...
+  final SentryFlutterOptions _options;
+
+  @override
+  SentryEvent? apply(SentryEvent event, Hint hint) {
+    final frames = _getStacktraceFrames(event);
+    final lastPath = frames?.first?.absPath;
+
+    if (lastPath == null) {
+      return event;
+    }
```
buenaflor, line 27: have you tried if this works with a release build? 

you'll also need the sentry_dart_plugin to upload source maps if you haven't

buenaflor (approved):

vaind:
analyzer seems to have started failing with this PR

martinhaintz:
@vaind thanks for the hint. I will fix it in the next hour.

PR #2256, opened 2024-08-28
Symbolicate Dart stacktrace on Flutter Android and iOS without debug images from native sdks

buenaflor:
## :scroll: Description

Dart StackTraces can be symbolicated with `Sentry.init` on Flutter iOS, macOS and Android

Uses the stacktrace header, specifically `isolateDsoBase / vmDsoBase` & `build_id` to construct a synthetic debug image which allows symbolication to work also on dart stacktraces without debug images that we get from the cocoa and android sdks.

Screenshot 2024-08-28 at 18 09 28

There's a need for using only sentry-dart on Flutter due to app size limitations, this should allow for some symbolication.

Unit test, local testing

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps
Get the code_file path somehow, is probably not possible with a pure dart setup


dart/lib/src/debug_image_extractor.dart
```diff
@@ -0,0 +1,164 @@
...
+    final reversedByteArray = Uint8List.fromList([
+      ...byteArray.sublist(0, 4).reversed,
+      ...byteArray.sublist(4, 6).reversed,
+      ...byteArray.sublist(6, 8).reversed,
+      ...byteArray.sublist(8, 10),
+      ...byteArray.sublist(10),
+    ]);
+
+    return Uuid.unparse(reversedByteArray);
+  }
```
buenaflor: not sure how performant this is

flutter/lib/src/integrations/load_native_debug_images_integration.dart
```diff
@@ -4,11 +4,16 @@ import 'package:sentry/sentry.dart';
 import '../native/sentry_native_binding.dart';
 import '../sentry_flutter_options.dart';
 
+// ignore: implementation_imports
+import 'package:sentry/src/load_dart_debug_images_integration.dart'
+    show NeedsSymbolication;
+
 /// Loads the native debug image list for stack trace symbolication.
-class LoadImageListIntegration extends Integration<SentryFlutterOptions> {
+class LoadNativeDebugImagesIntegration
```
stefanosiano: Is there a valid reason to change the name of the integration?
I don't want to break users that e.g. remove the integration based on the class name

stefanosiano (commented):
looks good overall!
just a question on an integration name

dart/lib/src/debug_image_extractor.dart
stefanosiano: hopefully we call it only once, so it shouldn't be a problem

flutter/lib/src/integrations/load_native_debug_images_integration.dart
buenaflor: now that we're introducing another debug image loading intrgration "load images list" gets kinda ambiguous 

but true probably not a good idea to change it now

I'd personally still prefer to mark it as renaming todo and change it in the next major for example

dart/lib/src/debug_image_extractor.dart
buenaflor: yup we do

stefanosiano (approved):

CHANGELOG.md
```diff
@@ -4,6 +4,11 @@
 
 ### Features
 
+ - Add `enableDartSymbolication` option to Sentry.init() for **Flutter iOS, macOS and Android** ([#2256](https://github.com/getsentry/sentry-dart/pull/2256))
+    - This flag enables symbolication of Dart stack traces when native debug images are not available.
+    - Useful when using Sentry.init() instead of SentryFlutter.init() in Flutter projects for example due to size limitations.
+    - `true` by default but automatically set to `false` when using SentryFlutter.init() because the SentryFlutter fetches debug images from the native SDK integrations.
+- Support allowUrls and denyUrls for Flutter Web ([#2227](https://github.com/getsentry/sentry-dart/pull/2227))
```
vaind, line 11: This line for another PR got duplicated here

vaind: FYI: I'll fix it in one of the other PRs while resolving merge conflicts

buenaflor: thx

PR #2285, opened 2024-09-12
refactor: use native spotlight integrations on Flutter Android, iOS, macOS

### LINKED ISSUE #2282: refactor Spotlight to use native android and cocoa Spotlight integrations when using Flutter
### Description

Currently the Spotlight transport is implemented on the Dart layer and is a decorator for the existing transport but it does not show the final envelope that is actually being sent because on Android for example the envelope is still being processed after `captureEnvelope` is called

So this is the goal: 
- Android uses the Android spotlight integration
- Cocoa uses the Cocoa spotlight integration
- Web keeps using SpotlightHttpTransport because we don't have a JS SDK integration yet but this may change soon
- Linux, Windows will also keep using SpotlightHttpTransport

buenaflor:
## :scroll: Description

use native spotlight integrations on Flutter Android, iOS, macOS

continue using the spotlight http transport on the Flutter layer for Web (as long as we don't have a JS SDK integration), Linux and Windows

The idea behind this PR is to use spotlight to send envelopes as late as possible so only after we are sure this envelope is not going to be modified or serialized/deserialized again. 
Currently we use a http transport decorator which happens on the Flutter layer but it might happen that the envelope is still processed afterwards which is the case for Android.

Closes https://github.com/getsentry/sentry-dart/issues/2282

Unit tests, local tests

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes


dart/test/sentry_client_test.dart
```diff
@@ -1767,12 +1767,71 @@ void main() {
...
+      fixture.options.spotlight = Spotlight(enabled: true);
+      fixture.getSut();
+
+      expect(fixture.options.transport is SpotlightHttpTransport, isTrue);
+    });
+
+    test(
+        'Spotlight enabled should set transport to SpotlightHttpTransport on Linux',
+        () async {
+      fixture.options.platformChecker = MockPlatformChecker(isWebValue: true);
```
martinhaintz: shouldn't this line look like this:
```dart
fixture.options.platformChecker = MockPlatformChecker(
        platform: MockPlatform.linux(),
      );
```

buenaflor: ah yes, thanks for catching!

martinhaintz (approved):

Issue #896, opened 2022-06-24
Support native crash reporting for Windows


bruno-garcia:
Sentry SDK for Flutter supports Windows since it's preview versions: https://github.com/getsentry/sentry-dart/pull/407
Since then, Flutter has made Windows support GA: https://medium.com/flutter/announcing-flutter-for-windows-6979d0d01fed

But it does not capture a native crash. The goal is to use [sentry-native](https://github.com/getsentry/sentry-native) with the crashpad backend to capture minidumps on Windows.

This was recently done on the Unity SDK:

* https://github.com/getsentry/sentry-unity/issues/527

This tracking issue contains the items we would consider when adding support to Flutter. Including:

- [ ] Offline caching on by default
- [ ] Crash free rate to rely on the native layer (ask native: did I crash last run?)
- [ ] Scope sync so data set to the SDK in Dart, shows up in native crashes
- [x] Make sure we can bundle `crashpad_handler.exe` with the final app
- [x] Upload symbols automatically during build. (might require changes to: https://github.com/getsentry/sentry-dart-plugin)

Related:
* https://github.com/getsentry/sentry-dart/issues/433
* https://github.com/getsentry/sentry-dart/issues/487
* https://github.com/flutter/flutter/issues/108974
* https://github.com/getsentry/sentry-native/issues/818

marandaneto:
Consider this [issue](https://github.com/dart-lang/sdk/issues/51726) when tackling this.

Issue #433, opened 2021-04-26
Support `split-debug-info` for Flutter Desktop (Linux and Windows)


ueman:
We need to test 
- [x] `split-debug-info` and `obfuscate` on macOS
- [x] #2431 `split-debug-info` and `obfuscate` on Linux
- [x] #2286 `split-debug-info` and `obfuscate` on Windows

Those options have not been tested. They might require changes in the SDK and/or on the server.
It could be that they already work, and in this case we don't need to do anything.

We need to update the docs after this has been tested.

marandaneto:
let's add the windows/macos/linux support to this script

https://github.com/getsentry/sentry-dart/blob/main/flutter/example/run.sh

the way of uploading debug files is the same as iOS/Android -> https://github.com/getsentry/sentry-dart/blob/main/flutter/example/run.sh#L58-L60

we need to see if frames are symbolicated, file/function names are available and line number is also correct.

I believe macOS won't work if Flutter has the same format as iOS debug files, but let's try.

@bruno-garcia you may need to help us on Windows, would you be willing to do so? :)

bruno-garcia:
Didn't seem to work: https://sentry.io/organizations/sentry-sdks/issues/2367778946/?project=5428562&query=is%3Aunresolved

We would need to add the debug_meta with the DLL's build_id of sorts. @jan-auer might have pointers.

![image](https://user-images.githubusercontent.com/1633368/116334065-99189e00-a7a2-11eb-8cb3-283d235caa7d.png)

bruno-garcia:
![image](https://user-images.githubusercontent.com/1633368/116334232-db41df80-a7a2-11eb-8e5c-4487ad565fc9.png)

`.so` on Windows? They really went all in with ELF it seems

marandaneto:
indeed, we'd need to install the `LoadAndroidImageListIntegration` integration for Flutter Desktop, but not only, on Android, `sentry-native` is also responsible for loading that.
That means, either we ship sentry-native for Linux/Windows or we find a way to do it via Native code or `ffi`

macOS requires that too, so Sentry Cocoa needs to expose a way of getting the debug image list.
https://github.com/getsentry/sentry-cocoa/issues/1089

ueman:
The [Flutter docs](https://flutter.dev/docs/deployment/obfuscate) state that `obfuscate` is not supported on Windows and Linux.

bruno-garcia:
[...]

PR #2286, opened 2024-09-12
feat: windows native error & obfuscation support

vaind:
## :scroll: Description

This adds support for:
- catching native crashes on Windows
- server-side symbolication of obfuscated dart errors on Windows

* #896 
* #433

Added tests

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes


dart/lib/src/sentry_stack_trace_factory.dart
```diff
@@ -9,22 +9,28 @@ import 'sentry_options.dart';
 class SentryStackTraceFactory {
   final SentryOptions _options;
 
-  final _absRegex = RegExp(r'^\s*#[0-9]+ +abs +([A-Fa-f0-9]+)');
-  final _frameRegex = RegExp(r'^\s*#', multiLine: true);
-
+  static final _absRegex = RegExp(r'^\s*#[0-9]+ +abs +([A-Fa-f0-9]+)');
+  static final _frameRegex = RegExp(r'^\s*#', multiLine: true);
+  static final _buildIdRegex = RegExp(r"build_id: '([A-Fa-f0-9]+)'");
+  static final _baseAddrRegex = RegExp(r'vm_dso_base: ([A-Fa-f0-9]+)');
```
buenaflor: we have some overlapping implementations with https://github.com/getsentry/sentry-dart/blob/main/dart/lib/src/debug_image_extractor.dart

e.g regex, `_StackInfo` and `_DebugInfo` etc...

maybe we could merge / reuse them in whatever way you think is better to avoid more duplications

vaind: sure, I'll update this PR deduplicate

dart/lib/src/protocol/sentry_stack_frame.dart
```diff
@@ -91,6 +91,7 @@ class SentryStackFrame {
   /// The "package" the frame was contained in.
   final String? package;
 
+  // TODO what is this? doesn't seem to be part of the spec https://develop.sentry.dev/sdk/event-payloads/stacktrace/
   final bool? native;
```
buenaflor, line 95: hm not sure if this is even used in the sentry product

buenaflor (commented):
overall looks good 👍 

since this is a big PR I'll do another pass later

flutter/lib/src/native/c/sentry_native.dart
```diff
@@ -0,0 +1,389 @@
...
+    final cValue = toNativeUtf8();
+    final result = SentryNative.native.value_new_string(cValue.cast());
+    malloc.free(cValue);
+    return result;
+  }
+}
+
+extension on int {
+  binding.sentry_value_u toNativeValue() {
+    if (this > 0x7FFFFFFF) {
```
buenaflor:
```suggestion
    if (this < -0x80000000 && this > 0x7FFFFFFF) {
```

buenaflor: I guess we can also check this to be safe

flutter/lib/src/native/c/sentry_native.dart
```diff
@@ -0,0 +1,389 @@
...
+    logger(SentryLevel.warning,
+        'SentryNative: unsupported data for for conversion: ${value.runtimeType} ($value)');
+    return null;
+  }
+}
+
+extension on String {
+  binding.sentry_value_u toNativeValue() {
+    final cValue = toNativeUtf8();
+    final result = SentryNative.native.value_new_string(cValue.cast());
```
buenaflor, line 341: is there a chance when `cValue.cast()` throws an exception that `cValue` is never freed?

flutter/lib/src/native/c/sentry_native.dart
vaind: you're right, i've fixed this

flutter/lib/src/native/c/sentry_native.dart
vaind: I don't believe cast() can throw

vaind: but I'll check

vaind: According to my tests this doesn't throw even with invalid agruments. Also, the generated code has the same pattern, e.g. https://github.com/getsentry/sentry-dart/blob/c328ffcf053d131ae2e276be1d448cf4abd22e88/flutter/lib/src/native/cocoa/binding.dart#L23-L29
```dart

  ffi.Pointer<ObjCSel> _registerName1(String name) {
    final cstr = name.toNativeUtf8();
    final sel = _sel_registerName(cstr.cast());
    pkg_ffi.calloc.free(cstr);
    return sel;
  }
```

buenaflor (approved):
nice 👍

dart/lib/src/protocol/sentry_stack_frame.dart
marandaneto: probably legacy or part of the `other` https://github.com/getsentry/relay/blob/8d83992a99478d95dc1acc82c0f437a95165fc30/relay-event-schema/src/protocol/stacktrace.rs#L192-L194
added a long time ago https://github.com/getsentry/sentry-dart/pull/155/files#diff-aee2d40f67826457e0abbaf8d299d5355bf4458244dde1e7ecb46f00b2c3b539
fine to just remove it if not used in the SDK itself

PR #2290, opened 2024-09-14
test: rethrow exceptions

vaind:
## :scroll: Description

This adds rehtrows when running in test mode as that seems to have not been finished in previous PRs

Should have no functional changes outside tests and the test changes are only to fix/workaround some issues that were present but hidden before `rethrow` was enabled.

Related to https://github.com/getsentry/sentry-dart/issues/1122#issuecomment-1318379866
and #1315

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes


dart/lib/src/sentry_baggage.dart
```diff
@@ -55,6 +55,7 @@ class SentryBaggage {
           exception: exception,
           stackTrace: stackTrace,
         );
+        // TODO rehtow in automatedTestMode
```
buenaflor: is this comment on purpose or is it an oversight

dart/lib/src/sentry_envelope_item.dart
```diff
@@ -132,6 +132,7 @@ class SentryEnvelopeItem {
       // TODO the data copy could be avoided - this would be most significant with attachments.
       return [...itemHeader, ...newLine, ...data];
     } catch (e) {
+      // TODO rethrow in automatedTestMode
```
buenaflor: same comment as the other todo

dio/test/mocks.dart
```diff
@@ -3,6 +3,11 @@ import 'package:sentry/src/transport/rate_limiter.dart';
 
 final fakeDsn = 'https://abc@def.ingest.sentry.io/1234567';
 
+SentryOptions defaultTestOptions() {
+  // ignore: invalid_use_of_internal_member
+  return SentryOptions(dsn: fakeDsn)..automatedTestMode = true;
+}
```
buenaflor, line 9: why do we need another `defaultTestOptions` for dio if we have on in `test_utils`?

buenaflor: and same question for the other packages, `drift` test and so on

vaind: it's a different package so yeah

dart/lib/src/sentry_envelope_item.dart
vaind: Similar to above, except this one actually has another TODO so I believe the refactoring would also solve the try-catch-rethrow sittuation

dart/lib/src/sentry_baggage.dart
vaind: I've updated the TODO: we currently don't have options available here and trickling it down would require quite a few changes so left it as a TODO.

dio/test/mocks.dart
buenaflor: got it, what about importing with `import '../../dart/test/test_utils.dart';` ? 

would that work?

vaind: it does but I'm not a fan of the idea of relative filesystem path imports from other packages, even though it's just for tests. Sounds hacky and not sure it's worth for this (there are worse deduplication candidates)

buenaflor: sounds good, then let's keep it this way

buenaflor (approved):

PR #2291, opened 2024-09-16
block app start if appLaunchedInForeground is not set.

### LINKED ISSUE #2215: Use `appLaunchedInForeground` to determinate invalid app start data on Android
### Description

Android SDK recently added check for `appLaunchedInForeground` which help removing invalid AppStart data, which were recorded in background. And are far from the actual application start.

- https://github.com/getsentry/sentry-java/pull/3550

martinhaintz:
## :scroll: Description
Blocking app starts if "appLaunchedInForeground" is false.

close #2215 

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes


buenaflor (commented):
looks good

@stefanosiano can you double check if this is the correct behaviour we want?

flutter/android/src/main/kotlin/io/sentry/flutter/SentryFlutterPlugin.kt
```diff
@@ -171,6 +171,11 @@ class SentryFlutterPlugin : FlutterPlugin, MethodCallHandler, ActivityAware {
 
     val appStartMetrics = AppStartMetrics.getInstance()
 
+    if (!appStartMetrics.isAppLaunchedInForeground) {
+      Log.w("Sentry", "App not launched in Foreground")
```
buenaflor:
```suggestion
      Log.w("Sentry", "Invalid app start data: app not launched in foreground")
```

let's be a bit more specific here

flutter/android/src/main/kotlin/io/sentry/flutter/SentryFlutterPlugin.kt
```diff
@@ -171,6 +171,11 @@ class SentryFlutterPlugin : FlutterPlugin, MethodCallHandler, ActivityAware {
 
     val appStartMetrics = AppStartMetrics.getInstance()
 
+    if (!appStartMetrics.isAppLaunchedInForeground) {
```
stefanosiano: this is good
we also have [another guard](https://github.com/getsentry/sentry-java/blob/61c8d80abb339b6ed87466e712d5a8d06aa97fba/sentry-android-core/src/main/java/io/sentry/android/core/performance/AppStartMetrics.java#L278-L286) in Android (which should be useful for iOS, too): if the appStartSpan takes more than 1 minute, we drop it
Can you add it?

martinhaintz: @stefanosiano thanks for the hint. 
I changed it to 
```kotlin
if (!appStartMetrics.isAppLaunchedInForeground ||
  appStartMetrics.appStartTimeSpan.durationMs > 1.toDuration(DurationUnit.MINUTES).inWholeMilliseconds
) {
  Log.w(
    "Sentry",
    "Invalid app start data: app not launched in foreground or app start took too long (>60s)",
  )
  result.success(null)
}
```
Did you mean it that way?

stefanosiano (approved):
Good!

flutter/android/src/main/kotlin/io/sentry/flutter/SentryFlutterPlugin.kt
stefanosiano: yep

PR #2293, opened 2024-09-16
Make response body accessible via `hint` in `beforeSendTransaction` callback for SentryHttpClient

### LINKED ISSUE #2220: Capture HTTP Response Body for `SentryHttpClient`
### Problem Statement

I would like to automatically capture the following information for all outgoing requests made by `SentryHttpClient`: 
- request URL
- request method
- request headers
- request body
- response status code
- response headers
- response body

This should be opt-in, similar to `networkDetailAllowUrls` in the JavaScript SDK: https://docs.sentry.io/platforms/javascript/session-replay/configuration/#network-details

### Solution Brainstorm

https://github.com/getsentry/rfcs/blob/main/text/0022-response-context.md

Yes

martinhaintz:
## :scroll: Description
The request body is now also available via the `hint` in `beforeSendTransaction` so the user can do further steps. 
  ```dart
  options.beforeSendTransaction = (transaction, hint) {
    final firstHint = transaction.spans[0].hint;
    final firstResponseBody = firstHint?.get(TypeCheckHint.httpResponse);
    final secondHint = transaction.spans[1].hint;
    final secondResponseBody = secondHint?.get(TypeCheckHint.httpResponse);
    // user can now use it
    return transaction;
  };
  ```

close #2220 

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes


dart/lib/src/sentry_options.dart
```diff
@@ -278,6 +278,8 @@ class SentryOptions {
   ///   because the connection was interrupted.
   /// Use with [SentryHttpClient] or `sentry_dio` integration for this to work,
   /// or iOS native where it sets the value to `enableCaptureFailedRequests`.
+  @Deprecated(
+      "All Request and Responses are now logged, if `sendDefaultPii` is `true` and `maxRequestBodySize` and `maxResponseBodySize` conditions are met.")
   bool captureFailedRequests = true;
```
buenaflor: let's rethink this, I don't think deprecating this flag is the right way to go

buenaflor:
@kahest JS Replay uses `captureNetworkBodies` for both request and response

https://github.com/getsentry/sentry-javascript/pull/7589

wdyt if we also adopt this flag?

dart/lib/src/sentry_options.dart
kahest: we still want to provide this flag to disable capturing of failed network requests completely, so yes we shouldn't deprecate this

kahest:
> @kahest JS Replay uses `captureNetworkBodies` for both request and response
> 
> [getsentry/sentry-javascript#7589](https://github.com/getsentry/sentry-javascript/pull/7589)
> 
> wdyt if we also adopt this flag?

I think this has since been [dropped](https://github.com/getsentry/sentry-javascript/pull/7953) in favor of `networkDetailAllowUrls` - see [docs](https://docs.sentry.io/platforms/javascript/session-replay/configuration/#network-details). I also don't see references to the option in code anymore.

I like the new option - it makes it more granular and explicit, though of course it's more work. It's still only available for JS SR I think, so not widely adopted. If there's no competing option for other HTTP integrations on other SDKs, I'm fine with adopting this

buenaflor:
Ah I missed that, thx. I'll have a look at it

buenaflor:
@kahest I haven't found anything specific in other SDKs that allow sending http response bodies other than replay.

However it's possible a user can do this:

```dart
beforeSend: (event, hint) async {
    final response = hint.get(TypeCheckHint.httpResponse);
    if (response is StreamedResponse) {
        final body = getResponseBody(response)
        // user can now use it
    }
}
```
this also aligns with what the js http integration would like to add: https://github.com/getsentry/sentry-javascript/issues/12544

kahest:
@buenaflor IIUC this would work for http+dio? I like the idea, it's very flexible and allows users to decide based on endpoint etc. if they want to add, but of course it's more complex to use than a switch

buenaflor:
> this would work for http+dio

yes but only errors so this won't work for transactions (like the user in the referenced issue wants) because we don't have a system in place to pair hints with transactions

I'll take a look for alternatives

buenaflor:
before I look at the PR, does this also work for transactions? because afaik beforeSend is not triggered for transactions @martinhaintz

buenaflor:
In order to make this fully compatible with tracing we should add the hint to the `beforeSendTransaction` API

It is available in our sdks so imo it's safe to implement it:
- [.net](https://docs.sentry.io/platforms/dotnet/configuration/filtering/#using-before-send-transaction)
- [javascript](https://github.com/getsentry/sentry-javascript/blob/0aa7f1d8b44f452892cd6c6d18d41f3b2568b050/packages/types/src/options.ts#L312C1-L316C1)

martinhaintz:
> before I look at the PR, does this also work for transactions? because afaik beforeSend is not triggered for transactions @martinhaintz

it wasn't done previously, but I implemented it.

martinhaintz:
> In order to make this fully compatible with tracing we should add the hint to the `beforeSendTransaction` API
> 
> It is available in our sdks so imo it's safe to implement it:
> 
> * [.net](https://docs.sentry.io/platforms/dotnet/configuration/filtering/#using-before-send-transaction)
> * [javascript](https://github.com/getsentry/sentry-javascript/blob/0aa7f1d8b44f452892cd6c6d18d41f3b2568b050/packages/types/src/options.ts#L312C1-L316C1)

this is by far, the better solution. I will change it tomorrow.

dart/lib/src/utils/http_deep_copy_streamed_response.dart
```diff
@@ -0,0 +1,28 @@
...
+
+/// Helper to deep copy the StreamedResponse of a web request
+@internal
+Future<List<StreamedResponse>> deepCopyStreamedResponse(
+    StreamedResponse originalResponse, int copies) async {
+  final List<int> bufferedData = [];
+
+  await for (final List<int> chunk in originalResponse.stream) {
+    bufferedData.addAll(chunk);
+  }
```
vaind: This will cause out-of-memory on large responses that are meant to be streamed and not copied to memory.

vaind (changes requested):
This can break existing applications and cause OOM errors in some edge cases. Are there other implementation options that wouldn't have this issue?

For example, only do the capture of the response body with limits on response size (read the response header first)

martinhaintz:
> This can break existing applications and cause OOM errors in some edge cases. Are there other implementation options that wouldn't have this issue?
> 
> For example, only do the capture of the response body with limits on response size (read the response header first)

@vaind You are right. Thanks for the hint. I will take a different approach.

buenaflor:
> This can break existing applications

true, but I'm not sure if adding a `hint` to `beforeSendTransaction` would warrant a new major, although I'm not opposed to marking this as a v9 feature

> only do the capture of the response body with limits on response size

we generally don't want to support this with response bodies (as far as I can tell after some searching other sdks don't allow users to capture response bodies automatically with a max size flag), so we'd like to make this dependent on the user as much as possible

imo I'd prefer using hints for this so the user can decide themselves whether to add it to their events

martinhaintz:
@vaind I implemented a cache for the `streamedResponse` see `streamed_response_copier.dart`

~@buenaflor @vaind A copied stream is now available via the hint with `TypeCheckHint.httpResponse`. I am not sure if it is the best way to hold the `hint` in the `SentryTracer`, but as the response is from/inside a `children` this was the only way I could return it as a `hint`. 
Let me know about your feedback.~

_UPDATE:
This is not a good idea. For every `transaction` we have only one `hint` element which is a `key/value pair` with `TypeCheckHint.httpResponse` for the response. However, a transaction can contain multiple `spans`, with every span containing their own response element. Therefore I think the better solution would be to access the `hint` via `transaction.spans[0].hint` in the `beforeSendTransaction`._

I also updated the analyzer in some plugins to ignore the `mocks.mocks.dart` files, because it threw some warnings after rebuilding.

martinhaintz:
This is the current approach for accessing the hints for two spans inside a transaction:
```dart
options.beforeSendTransaction = (transaction, hint) {
        final firstHint = transaction!.spans[0].hint
        final secondHint = transaction!.spans[1].hint
        return transaction;
      };
```

@vaind @buenaflor should I revert the changes regarding the parameter `options.beforeSendTransaction = (transaction, hint) {}` and remove the `hint`, or leave it there to have the `hint` already there for the future?

buenaflor:
We can add the hint for beforeSendTransaction there but probably will mark this as a v9 feature unless we think this is okay to do in v8

martinhaintz:
@vaind In your previous review you requested some changes, which I (hopefully) addressed. Can you please review again? Thanks!

buenaflor:
Once we are done here, I'll change the base branch to [v9](https://github.com/getsentry/sentry-dart/tree/v9)

CHANGELOG.md
```diff
@@ -4,6 +4,17 @@
...
+- Make response body accessible via hint in `beforSend` callback for failed web requests or if tracing is enabled in `SentryHttpClient` ([#2293](https://github.com/getsentry/sentry-dart/pull/2293))
+  ```dart
+  options.beforeSendTransaction = (transaction, hint) {
+    final firstHint = transaction!.spans[0].hint;
+    final firstResponseBody = firstHint.get(TypeCheckHint.httpResponse);
+    final secondHint = transaction!.spans[1].hint;
+    final secondResponseBody = secondHint.get(TypeCheckHint.httpResponse);
+    // user can now use it
+    return transaction;    
+  };
```
vaind: Can you please check this code in the sample app and update/fix based on the linter warnings? e.g. transaction is non-nullable so the assertion is unnecessary.

vaind (commented):
This is pretty large so I won't have the time to properly review this any sooner than Wednesday. 

As for what I've noticed from a quick look:
1. it's a breaking change so should be merged to v9 right?
2. I don't see how you've addressed my remark about OOM on large responses. Could you elaborate please?

martinhaintz:
> 2. I don't see how you've addressed my remark about OOM on large responses. Could you elaborate, please?

As the responseBody is only accessible once, we must somehow cache it to make sure, that the return value of the `send` method contains the responseBody and also make the responseBody available via the `hint` in the `beforeSendTransaction`. So I don't know how we can address the OOM on large responses. If you have something in mind, feel free to share.

vaind:
> > 2. I don't see how you've addressed my remark about OOM on large responses. Could you elaborate, please?
> 
> As the responseBody is only accessible once, we must somehow cache it to make sure, that the return value of the `send` method contains the responseBody and also make the responseBody available via the `hint` in the `beforeSendTransaction`. So I don't know how we can address the OOM on large responses. If you have something in mind, feel free to share.

1. If the `Content-Length` header is present, we can read that first. This way we can check and compare the value to `MaxResponseBodySize` to decide whether we're even going to try and read the body.
2. If the body is reasonable small (some fixed amount, TBD, maybe 1MiB, we can just read it as you did)
3. if it's larger or we don't know (because the header is not present), we can create a pass-through response stream. This way when the user actually accesses the body, we'll collect it as well. Once it exceeds `MaxResponseBodySize`, we'll drop whatever we have already collected to clear the memory and just stream transparently to the user.
  Once the body is read, we can update the context. We should be able to achieve this with a `Completer`.

> The request body is now also available via the hint in beforeSendTransaction so the user can do further steps.

Where does this decision come from? The linked RFC describes putting this information into the response context 
- https://github.com/getsentry/rfcs/blob/main/text/0022-response-context.md#proposal-namely-option-2
- https://develop.sentry.dev/sdk/data-model/event-payloads/contexts/#response-context (docs don't seem to be updated yet)

buenaflor:
> Where does this decision come from? The linked RFC describes putting this information into the response context

the use case is to read the response payloads which is as far as I can see not part of the RFC, only the body size. and since it's highly pii sensitive imo it's more preferrable to let users have the decision of what to do with the payload, [see javascript](https://github.com/getsentry/sentry-javascript/issues/12544#issuecomment-2178590992) or [.net](https://github.com/getsentry/sentry-dotnet/issues/3594)

martinhaintz:
> 1. If the `Content-Length` header is present, we can read that first. This way we can check and compare the value to `MaxResponseBodySize` to decide whether we're even going to try and read the body.

As I remember correctly, we decided to ignore the "MaxResponseBodySize" for this feature.

> if it's larger or we don't know (because the header is not present), we can create a pass-through response stream. This way when the user actually accesses the body, we'll collect it as well. Once it exceeds MaxResponseBodySize, we'll drop whatever we have already collected to clear the memory and just stream transparently to the user.
Once the body is read, we can update the context. We should be able to achieve this with a Completer.

I don't think we should capture the response body based on if the response body from the the original request is accessed. At least, this is not what was discussed in the [issue](https://github.com/getsentry/sentry-dart/issues/2220). It 
 also would complicate the behavior for the user who looks through sentry and wonders, why some events have response bodies and some not.
 
@buenaflor what do you think?

buenaflor:
imo I'd just make it simpler:
- no headers present, we don't read the body because we don't know the size of it
- otherwise if a header exists, we introduce a hard limit, maybe 0.15mb or higher, this is what [javascript does for replay](https://github.com/getsentry/sentry-javascript/blob/6f903d5b7b6b6a935a204d526ea93c22791d865b/packages/replay-internal/src/constants.ts#L32C1-L34C1) and read the body if it is under the threshold
- if it exceeds the hard limit we don't read the body

wdyt?

vaind:
> imo I'd just make it simpler:
> 
> * no headers present, we don't read the body because we don't know the size of it
> * otherwise if a header exists, we introduce a hard limit, maybe 0.15mb or higher, this is what [javascript does for replay](https://github.com/getsentry/sentry-javascript/blob/6f903d5b7b6b6a935a204d526ea93c22791d865b/packages/replay-internal/src/constants.ts#L32C1-L34C1) and read the body if it is under the threshold
> * if it exceeds the hard limit we don't read the body
> 
> wdyt?

sounds good, although why not use the existing option (MaxResponseBodySize) for the limit?

buenaflor:
> sounds good, although why not use the existing option (MaxResponseBodySize) for the limit?

we could do that, we'd have to adjust the enum value `MaxResponseBodySize.always` or rename it to keep in mind the hard limit

we were considering removing it since no other SDK supports this option

so ultimately this would be the change
- either remove or possibly rename `maxResponseBodySize.always` since we can't always add it due to size limits
- if possible we read the body and make them available in hints, similar to .net and javascript
- automatically capturing the body data will be removed from our dio integration

or instead of not reading the body we just truncate it up to the limit

vaind:
> we were considering removing it since no other SDK supports this option

That makes sense, no reason to use it than.

buenaflor:
I'll close this PR for now, we will continue this for v9

PR #2299, opened 2024-09-19
Validation for fetchedDisplayRefreshRate

Volsavr:
## :scroll: Description
Additional validation for displayRefreshRate received from the native level. 

We noticed that following code: 

`
final fetchedDisplayRefreshRate = await _native?.displayRefreshRate();
`

sometimes returns 0. It happens only on macOS Monterey and below (At least in our cases).  

Assigning 0 to fetchedDisplayRefreshRate leads to exception:

`
 Unsupported operation: Infinity or NaN toInt 
#0      SpanFrameMetricsCollector.calculateFrameMetrics (package:sentry_flutter/src/span_frame_metrics_collector.dart)
#1      SpanFrameMetricsCollector.onSpanFinished (package:sentry_flutter/src/span_frame_metrics_collector.dart:80)
#2      SentrySpan.finish (package:sentry/src/protocol/sentry_span.dart:81)
#3      SentryTracer.finish (package:sentry/src/sentry_tracer.dart:139)
#4      SentryTracer._finishedCallback (package:sentry/src/sentry_tracer.dart:282)
#5      SentrySpan.finish (package:sentry/src/protocol/sentry_span.dart:94)
`

Also it may lead to ui issues in the app (we faced with issues  related to displaying svg resources).  

Using default displayRefreshRate (60) prevents exception appearing and ui issues. 

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/flutter/lib/src/span_frame_metrics_collector.dart
+++ b/flutter/lib/src/span_frame_metrics_collector.dart
@@ -12,6 +12,8 @@ import 'native/sentry_native_binding.dart';
 @internal
 class SpanFrameMetricsCollector implements PerformanceContinuousCollector {
   static const _frozenFrameThresholdMs = 700;
+  static const _defaultRefreshRate = 60;
+
   static const totalFramesKey = 'frames.total';
   static const framesDelayKey = 'frames.delay';
   static const slowFramesKey = 'frames.slow';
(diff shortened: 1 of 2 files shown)


stefanosiano:
Hi @Volsavr thanks for opening the PR!
Would you mind adding a test in `span_frame_metrics_collector_test.dart`?
We need that to avoid regressions in the future.
You can copy the test `captures metrics with display refresh rate of 60 if native refresh rate is null` replacing `null` with `0`, possibly with a comment on macOS Monterey returning 0.
Then i'll be happy to approve it!

buenaflor:
I don't think we should set it to a default refresh rate. a couple weeks ago we decided that we want to keep this accurate or not do it so if we can't return a proper refresh rate it should just disable the instrumentation instead of applying a default rate

I can prepare a PR for fixing that

buenaflor:
We have added validations here https://github.com/getsentry/sentry-dart/pull/2307

PR #2307, opened 2024-09-25
fix: only start frame tracking if we receive valid refresh data

buenaflor:
## :scroll: Description

Refresh data should not be null and bigger than 0

Fixes frame tracking being started with invalid display refresh data 

Unit test

## :pencil: Checklist

- [ ] All tests passing
- [x] No breaking changes

### DIFF
--- a/flutter/lib/src/span_frame_metrics_collector.dart
+++ b/flutter/lib/src/span_frame_metrics_collector.dart
@@ -39,7 +39,7 @@ class SpanFrameMetricsCollector implements PerformanceContinuousCollector {
   bool get isTrackingRegistered => _isTrackingRegistered;
   bool _isTrackingRegistered = false;
 
-  int displayRefreshRate = 60;
+  int? displayRefreshRate;
 
   final _stopwatch = Stopwatch();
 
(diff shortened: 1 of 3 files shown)


kahest:
does this supercede https://github.com/getsentry/sentry-dart/pull/2299?

flutter/lib/src/span_frame_metrics_collector.dart
```diff
@@ -59,25 +62,28 @@ class SpanFrameMetricsCollector implements PerformanceContinuousCollector {
...
-    activeSpans.add(span);
-    startFrameTracking();
   }
 
   @override
   Future<void> onSpanFinished(ISentrySpan span, DateTime endTimestamp) async {
-    if (span is NoOpSentrySpan || !activeSpans.contains(span)) return;
+    if (span is NoOpSentrySpan ||
+        !activeSpans.contains(span) ||
+        displayRefreshRate == null) return;
```
kahest: `m`: should this also check for `displayRefreshRate == 0`?

kahest (commented):
left 2 questions to clarify, otherwise LGTM

buenaflor:
> does this supercede https://github.com/getsentry/sentry-dart/pull/2299?

yep, the other PR is not applicable anymore since we don't want to use a default refresh rate

flutter/lib/src/span_frame_metrics_collector.dart
buenaflor: It doesn't need to since onSpanStarted validates it, this is only so we can safely unwrap the variable

but we can be more defensive here and check the == 0 condition regardless

kahest: gotcha - both is fine for me then 👍

kahest (approved):

PR #2311, opened 2024-09-26
fix(frame-tracking): app lag when long running span finishes

### LINKED ISSUE #2298: App lags when navigating to different screen
### Platform

Flutter Mobile

### Obfuscation

Enabled

### Debug Info

Enabled

### Doctor

```
➜  ~ flutter doctor -v
[!] Flutter (Channel main, 3.7.13-0.0.pre.2, on macOS 14.4 23E214 darwin-arm64, locale en-MY)
    • Flutter version 3.7.13-0.0.pre.2 on channel main at /Users/zhenhao.ng/OlaProjects/flutter
    ! Upstream repository git@github.com:olachat/flutter.git is not a standard remote.
      Set environment variable "FLUTTER_GIT_URL" to git@github.com:olachat/flutter.git to dismiss this error.
    • Framework revision 1a4884310c (12 months ago), 2023-10-04 12:28:33 +0800
    • Engine revision 1a65d409c7
    • Dart version 2.19.6
    • DevTools version 2.20.1
    • If those were intentional, you can disregard the above warnings; however it is recommended to use "git" directly to perform update checks and upgrades.

[!] Android toolchain - develop for Android devices (Android SDK version 34.0.0)
    • Android SDK at /Users/zhenhao.ng/Library/Android/sdk
    • Platform android-35, build-tools 34.0.0
    • Java binary at: /Library/Java/JavaVirtualMachines/zulu-8.jdk/Contents/Home/bin/java
    • Java version OpenJDK Runtime Environment (Zulu 8.74.0.17-CA-macos-aarch64) (build 1.8.0_392-b08)
    ✗ Android license status unknown.
      Run `flutter doctor --android-licenses` to accept the SDK licenses.
      See https://flutter.dev/docs/get-started/install/macos#android-setup for more details.

[✓] Xcode - develop for iOS and macOS (Xcode 15.3)
    • Xcode at /Applications/Xcode 15.3.app/Contents/Developer
    • Build 15E204a
    • CocoaPods version 1.15.2

[!] Android Studio (version 2024.1)
    ✗ Unable to find bundled Java version.
    • Try updating or re-installing Android Studio.

[✓] VS Code (version 1.92.2)
    • Flutter extension version 3.96.0

    • sdk gphone64 arm64 (mobile) • emulator-5554 • android-arm64  • Android 14 (API 34) (emulator)
    • macOS (desktop)             • macos         • darwin-arm64   • macOS 14.4 23E214 darwin-arm64
    • Chrome (web)                • chrome        • web-javascript • Google Chrome 129.0.6668.58
    ! Error: Browsing on the local area network for Flamy's iPhone. Ensure the device is unlocked and attached with a cable or associated with the same local area network as this Mac.
      The device must be opted into Developer Mode to connect wirelessly. (code -27)

[✓] HTTP Host Availability
    • All required HTTP hosts are available

! Doctor found issues in 3 categories.
```

### Version

8.6.0

### Steps to Reproduce

- Use the app for more than approximately 20 minutes.
- App will start to lag when navigating to other screen.

This issue only starts happening when we enabled Sentry in our new version of app.

When checking with CPU profiler, it seems Sentry is using much of the time. I'm wondering if `SentryNavigatorObserver` is causing our app to lag?

Image

Looks related to this PR comment with `span_frame_metrics_collector.dart`:
https://github.com/getsentry/sentry-dart/pull/2106#discussion_r1762199270

Thanks in advance.

### Expected Result

The app should not lag when navigating to other screen.

### Actual Result

After using the app for more than approximately 20 minutes, the app will start lagging when navigating to other screen,

None

buenaflor:
## :scroll: Description

Basically when a long lasting span runs in the background (e.g 20 minutes) without ever finishing, the frame tracking keeps accumulating frames and thus takes a long time to process it.

Fixes #2298 

Unit test

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/flutter/lib/src/sentry_flutter_options.dart
+++ b/flutter/lib/src/sentry_flutter_options.dart
@@ -276,6 +276,13 @@ class SentryFlutterOptions extends SentryOptions {
   /// Defaults to `true`
   bool enableFramesTracking = true;
 
+  /// The period (in minutes) for which to retain frame information.
+  /// Used to calculate the maximum number of frames to store for active spans.
+  /// If the number of frames exceeds the maximum, the frame tracking is cancelled
+  /// and frame metrics are cleared for the current active spans.
+  /// Default value is 3 minutes.
+  int framesTrackingRetentionPeriod = 3;
+
   /// By using this, you are disabling native [Breadcrumb] tracking and instead
   /// you are just tracking [Breadcrumb]s which result from events available
   /// in the current Flutter environment.
--- a/flutter/lib/src/sentry_flutter_options.dart
+++ b/flutter/lib/src/sentry_flutter_options.dart
@@ -276,12 +276,9 @@ class SentryFlutterOptions extends SentryOptions {
   /// Defaults to `true`
   bool enableFramesTracking = true;
 
-  /// The period (in minutes) for which to retain frame information.
-  /// Used to calculate the maximum number of frames to store for active spans.
-  /// If the number of frames exceeds the maximum, the frame tracking is cancelled
-  /// and frame metrics are cleared for the current active spans.
-  /// Default value is 3 minutes.
-  int framesTrackingRetentionPeriod = 3;
+  /// Maximum number of frames to track before clearing.
+  /// Default is 10800 frames, this equals to 3 minutes of frames at 60fps.
+  int maxFramesToTrack = 10800;
 
   /// By using this, you are disabling native [Breadcrumb] tracking and instead
   /// you are just tracking [Breadcrumb]s which result from events available
--- a/flutter/lib/src/sentry_flutter_options.dart
+++ b/flutter/lib/src/sentry_flutter_options.dart
@@ -276,7 +276,8 @@ class SentryFlutterOptions extends SentryOptions {
   /// Defaults to `true`
   bool enableFramesTracking = true;
 
-  /// Maximum number of frames to track before clearing.
+  /// Maximum number of frames to track at a time before clearing.
+  /// Clearing cancels frame tracking for all active spans.
   /// Default is 10800 frames, this equals to 3 minutes of frames at 60fps.
   int maxFramesToTrack = 10800;
 
--- a/flutter/lib/src/sentry_flutter_options.dart
+++ b/flutter/lib/src/sentry_flutter_options.dart
@@ -276,11 +276,6 @@ class SentryFlutterOptions extends SentryOptions {
   /// Defaults to `true`
   bool enableFramesTracking = true;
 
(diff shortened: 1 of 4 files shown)


flutter/lib/src/sentry_flutter_options.dart
```diff
@@ -276,6 +276,13 @@ class SentryFlutterOptions extends SentryOptions {
   /// Defaults to `true`
   bool enableFramesTracking = true;
 
+  /// The period (in minutes) for which to retain frame information.
+  /// Used to calculate the maximum number of frames to store for active spans.
+  /// If the number of frames exceeds the maximum, the frame tracking is cancelled
+  /// and frame metrics are cleared for the current active spans.
+  /// Default value is 3 minutes.
+  int framesTrackingRetentionPeriod = 3;
```
ueman: Why don't you use `Duration`? That way you will not have any confusion regarding the unit:

```suggestion
  Duration framesTrackingRetentionPeriod = Duration(minutes: 3);
```

Also, your calculations further down below will also be simpler :D

buenaflor: I think I'll just use a fixed number like 10800 as default, it's a pretty confusing option tbh

flutter/lib/src/sentry_flutter_options.dart
```diff
@@ -276,6 +276,11 @@ class SentryFlutterOptions extends SentryOptions {
   /// Defaults to `true`
   bool enableFramesTracking = true;
 
+  /// Maximum number of frames to track at a time before clearing.
+  /// Clearing cancels frame tracking for all active spans.
+  /// Default is 10800 frames, this equals to 3 minutes of frames at 60fps.
+  int maxFramesToTrack = 10800;
```
brustolin: l: I think a duration would be more easy to understand instead of number of frames

brustolin (approved):
One suggestion, but not blocking

flutter/lib/src/sentry_flutter_options.dart
buenaflor: the confusing part is that the callback we use from flutter doesn't actually trigger every x fps, it triggers only when new frames are drawn.

If the user doesn't do anything no scrolling etc... after 3 minutes the frames tracked might be 100 or so which doesnt align with a Duration option

stefanosiano (approved):

buenaflor:
we will keep the limit internal for now until we properly fix it. 

we might not need to keep the references to the normal frames in the future which would not require a frame limit

Issue #2302, opened 2024-09-19
metrics_aggregator_test is flaky


vaind:
### Description

I see this fail from time to time

```
❌ test/metrics/metrics_aggregator_test.dart: emit in same time bucket same metric with different tags emit different metrics (failed)
  Expected: <2>
    Actual: <1>
  
  package:matcher                                  expect
  test/metrics/metrics_aggregator_test.dart 189:7  main.<fn>.<fn>
  ```

**Comment by vaind** [comment]
Also some more in https://github.com/getsentry/sentry-dart/actions/runs/10942003756/job/30377979535

```
❌ test/metrics/metrics_aggregator_test.dart: span local aggregation emit calls add (failed after test completion)
❌ test/metrics/metrics_aggregator_test.dart: span local aggregation emit calls add (failed after test completion)
❌ test/metrics/metrics_aggregator_test.dart: span local aggregation emit set (failed after test completion)
❌ test/metrics/metrics_aggregator_test.dart: span local aggregation emit set (failed after test completion)
❌ test/metrics/metrics_aggregator_test.dart: span local aggregation emit counter (failed after test completion)
❌ test/metrics/metrics_aggregator_test.dart: span local aggregation emit counter (failed after test completion)
❌ test/metrics/metrics_aggregator_test.dart: span local aggregation emit gauge (failed after test completion)
❌ test/metrics/metrics_aggregator_test.dart: span local aggregation emit gauge (failed after test completion)
❌ test/metrics/metrics_aggregator_test.dart: span local aggregation emit distribution (failed after test completion)
❌ test/metrics/metrics_aggregator_test.dart: span local aggregation emit distribution (failed after test completion)
```

a sample stack trace:

```
ClientException: XMLHttpRequest error., uri=https://sentry.example.com/api/1/envelope/
  org-dartlang-sdk:///lib/_internal/js_runtime/lib/core_patch.dart 835:36   Object.StackTrace.current
  package:http/src/browser_client.dart 98:22                                BrowserClient.send.<fn>
  package:stack_trace                                                       StackZoneSpecification._run
  org-dartlang-sdk:///lib/async/zone.dart 1407:46                           StaticClosure._rootRunUnary
  org-dartlang-sdk:///lib/async/zone.dart 1307:34                           _CustomZone.runUnary
  org-dartlang-sdk:///lib/async/future_impl.dart 127:29                     _Future._propagateToListeners.handleValueCallback
[...]

PR #2312, opened 2024-09-26
Deprecate: metrics api

### LINKED ISSUE #2277: Deprecate metrics api
### Description

Now with the recent changes in the vision of metrics, it is being deprecated.

Continue reading [here](https://sentry.zendesk.com/hc/en-us/articles/26369339769883-Upcoming-API-Changes-to-Metrics) for more information

buenaflor:
In this PR: remove metrics tests, deprecate public metrics api

## :scroll: Description

Why we deprecate it: https://sentry.zendesk.com/hc/en-us/articles/26369339769883-Upcoming-API-Changes-to-Metrics

Closes #2277


buenaflor:
@stefanosiano is this the only place that needs the deprecated tag?

philipphofmann (commented):
What drove you to remove the tests?

buenaflor:
@philipphofmann some of the metrics tests are very flaky ([as described here](https://github.com/getsentry/sentry-dart/issues/2302)) and metrics in its current state will be gone as a product so these tests won't be useful anymore and only a liability going forward, if we need them for a future implementation we can easily reference them again in the git history

dart/lib/src/sentry.dart
```diff
@@ -316,6 +316,8 @@ class Sentry {
   static ISentrySpan? getSpan() => _hub.getSpan();
 
   /// Gets access to the metrics API for the current hub.
+  @Deprecated(
+      'Metrics will be deprecated and removed in the next major release. Sentry will reject all metrics sent after October 7, 2024. Learn more: https://sentry.zendesk.com/hc/en-us/articles/26369339769883-Upcoming-API-Changes-to-Metrics')
```
philipphofmann, line 320: `m`: Aren't metrics experimental? If yes, we can also remove them in a nonmajor update. I don't think it makes sense to keep the code if, anyways, we are going to reject all metrics after October 7th.

buenaflor: Hmm, I don't have any strong opinion on that. 

@kahest what do you think?

kahest: I'd deprecate now and leave some time to remove - not because it needs to be a major (can be a minor in a few weeks), but to give app maintainers at least a bit of time to learn about the deprecation and remove the code before we break their builds. even if it's likely a very small change they need to do, we don't need to force them to do it right now

buenaflor: Sounds good, I guess we can approve this PR 

I'd like to include that in the release

stefanosiano:
> @stefanosiano is this the only place that needs the deprecated tag?

I'd put the Hub.metricsApi() one, too, in case a user uses the hub directly (it should warn the user about it being internal, but let's ensure they know its deprecation)

stefanosiano (approved):

kahest (approved):

PR #2315, opened 2024-09-27
chore: Add action to warn about potentially risky PR changes

kahest:
## :scroll: Description

Implement a mechanism that automatically adds an "alert" comment to PRs if they include changes to files we think have higher potential to break things than others.

same as https://github.com/getsentry/sentry-java/pull/3726

✅ by adding the file-filter.yml introduced in this PR to the list of potentially risky changes and ensuring that an alert is shown
Screenshot 2024-09-27 at 16 08 33

## :pencil: Checklist

- [ ] ~I added tests to verify changes~
- [ ] ~No new PII added or SDK only sends newly added PII if `sendDefaultPii` is enabled~
- [ ] ~I updated the docs if needed~
- [ ] ~All tests passing~
- [ ] ~No breaking changes~

## :crystal_ball: Next steps

#skip-changelog


.github/file-filters.yml
```diff
@@ -0,0 +1,4 @@
+# This is used by the action https://github.com/dorny/paths-filter
+
+high_risk_code: &high_risk_code
+  - ".github/file-filters.yml"
```
kahest: this is for testing and will be removed before merging
need to align on set of potentially higher-risk files before merging as well

buenaflor: cool! 

is this how we reference file paths?

```yml
high_risk_code: &high_risk_code
  - "dart/lib/src/risky_file.dart"  
  - "flutter/lib/src/risky_file.dart"
```

buenaflor (commented):
nice addition 👍

.github/file-filters.yml
kahest: yep, just full repo paths, you can check https://github.com/getsentry/sentry-cocoa/blob/main/.github/file-filters.yml for inspiration

buenaflor: Okay, I'll think about the code paths that would benefit from this

buenaflor: What do we define as risky? Is this strictly defined as code paths that are prone to crashing?

kahest: the way to think about this is - what can break the users in significant ways which is not currently automatically ensured/validated/tested/etc. in a different way - also try to strike a good balance between missing things and over-reporting (I wouldn't expect this warning to need to be shown in every other PR, for example)

kahest:
@buenaflor do you have more additions in mind? I wanna wrap this up soon-ish :)

buenaflor:
No, I was about to write that this is good now :D 

If I see something over the next few days that's alarming I'll add it

kahest:
> No, I was about to write that this is good now :D
> 
> If I see something over the next few days that's alarming I'll add it

cool 👍 this is meant to be a living list anyway, whenever something makes sense to add, we do it. just need a ✅ now 😅

buenaflor (approved):

PR #2329, opened 2024-10-07
Feat/only send debug images referenced in the stacktrace for events

### LINKED ISSUE #1755: Only send debug images referenced in the stacktrace for events
### Description

Ref: https://github.com/getsentry/team-mobile/issues/158

```[tasklist]
### Blocked by
- [x] Sentry Cocoa `8.39.0`
```

martinhaintz:
## :scroll: Description
Instead of sending all debug images in case of errors, we only want to send the referenced debug images of the stacktrace to be sent.

close #1755 

build local with `flutter build macos --release --obfuscate --split-debug-info=./` and looked at sentry for the imageadresses. Then run `flutter packages pub run sentry_dart_plugin` and looked for further info in sentry.

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes


flutter/ios/Classes/SentryFlutterPluginApple.swift
```diff
@@ -277,8 +277,14 @@ public class SentryFlutterPluginApple: NSObject, FlutterPlugin {
         }
     }
 
-    private func loadImageList(result: @escaping FlutterResult) {
-      let debugImages = PrivateSentrySDKOnly.getDebugImages() as [DebugMeta]
+    private func loadImageList(_ call: FlutterMethodCall, result: @escaping FlutterResult) {
+      guard let imageAddresses = call.arguments as? Set<String> else {
+          print("Arguments is not a list of string")
+          result(FlutterError(code: "X", message: "Cannot get image for address", details: nil))
```
martinhaintz: @buenaflor is there a list for the error codes? Which code should I use?

flutter/test/integrations/load_image_list_test.dart
```diff
@@ -26,8 +26,10 @@ void main() {
 
     setUp(() async {
       fixture = IntegrationTestFixture(LoadImageListIntegration.new);
-      when(fixture.binding.loadDebugImages())
+      when(fixture.binding.loadDebugImages(<String>{"0x6f80b000"}))
           .thenAnswer((_) async => imageList);
+      // when(fixture.binding.loadDebugImages(any))
+      //     .thenAnswer((_) async => imageList);
```
martinhaintz: @buenaflor how can I capture all `Set<String>` for `loadDebugImages`? I tried it with `any`, `any<Set<String>>()`, `captureAny` and also `argThat(isA<Set<String>>())` but nothing worked.

martinhaintz:
@buenaflor any ideas, what would be the best way to test this? `print` and `log` are disabled in Flutter for release builds.

flutter/ios/Classes/SentryFlutterPluginApple.swift
buenaflor: I don't think we should send back an error, if it fails for whatever reason let's just fall back to the previous way of getting the debug image addresses

flutter/lib/src/integrations/load_image_list_integration.dart
```diff
@@ -33,12 +34,44 @@ class _LoadImageListIntegrationEventProcessor implements EventProcessor {
...
+      final images = await _native.loadDebugImages(imageAddresses);
       if (images != null) {
         return event.copyWith(debugMeta: DebugMeta(images: images));
       }
     }
 
     return event;
   }
+
+  Set<String> _collectImageAddressesFromStackTrace(SentryStackTrace trace) {
```
vaind: maybe pass the set as an argument to avoid creating intermediate sets on each invocation.

flutter/lib/src/integrations/load_image_list_integration.dart
```diff
@@ -33,12 +34,44 @@ class _LoadImageListIntegrationEventProcessor implements EventProcessor {
...
+    Set<String> imageAddresses = {};
+    for (var frame in trace.frames) {
+      if (frame.imageAddr != null) {
+        imageAddresses.add(frame.imageAddr!);
```
vaind: Actually, this isn't going to work because for obfuscated code we may only have instruction addresses, not image addresses.

flutter/test/integrations/load_image_list_test.dart
martinhaintz: The parameter of the mocked method has to be nullable, even if the parameter of the original Method is not nullable. After this change, `any` is working.

flutter/lib/src/integrations/load_image_list_integration.dart
martinhaintz: Thanks for the hint, I wasn`t aware of that.

flutter/lib/src/integrations/load_image_list_integration.dart
martinhaintz: @vaind can you please explain it a little bit more, what you mean? Thanks.

buenaflor: we need to use the instruction addresses to grab the images like so

https://github.com/getsentry/sentry-react-native/blob/59a53cbd0983e4bc111169feedd7c8d7d19df9b6/ios/RNSentry.mm#L220C1-L225C1

flutter/lib/src/integrations/load_image_list_integration.dart
buenaflor: we need to use the instruction addresses to grab the images like so

https://github.com/getsentry/sentry-react-native/blob/59a53cbd0983e4bc111169feedd7c8d7d19df9b6/ios/RNSentry.mm#L220C1-L225C1

flutter/lib/src/integrations/load_image_list_integration.dart
buenaflor:
```dart
void _collectInstructionAddresses(
  SentryEvent event,
  Set<String> instructionAddresses,
)
```

he probably means like this

vaind: exactly, thanks @buenaflor

flutter/ios/Classes/SentryFlutterPluginApple.swift
```diff
@@ -277,8 +277,26 @@ public class SentryFlutterPluginApple: NSObject, FlutterPlugin {
...
+                  if let image = image {
+                      let imageAddress = sentry_formatHexAddressUInt64(image.address)!
+                    imagesAddresses.insert(imageAddress)
+                  }
+              }
+          }
+          debugImages = SentryDependencyContainer.sharedInstance().debugImageProvider.getDebugImages(forAddresses: imagesAddresses, isCrash: false) as [DebugMeta]
+      } else {
+          debugImages = PrivateSentrySDKOnly.getDebugImages() as [DebugMeta]
+      }
```
buenaflor, line 298: this doesn't fall back to `PrivateSentrySDKOnly.getDebugImages()` when either the insruction address parsing throws or the resulting imagesAddresses is still empty

we need to make sure that in all cases this will always return the full debug image list as fallback

flutter/ios/Classes/SentryFlutterPluginApple.swift
```diff
@@ -277,8 +277,26 @@ public class SentryFlutterPluginApple: NSObject, FlutterPlugin {
         }
     }
 
-    private func loadImageList(result: @escaping FlutterResult) {
-      let debugImages = PrivateSentrySDKOnly.getDebugImages() as [DebugMeta]
+    private func loadImageList(_ call: FlutterMethodCall, result: @escaping FlutterResult) {
+      var debugImages: [DebugMeta] = []
+      
+      if let arguments = call.arguments as? Set<String> , !arguments.isEmpty{
```
buenaflor:
```suggestion
      if let arguments = call.arguments as? Array<String> , !arguments.isEmpty{
```

flutter/ios/Classes/SentryFlutterPluginApple.swift
```diff
@@ -277,8 +277,26 @@ public class SentryFlutterPluginApple: NSObject, FlutterPlugin {
...
+          var imagesAddresses: Set<String> = []
+
+          for argument in arguments {
+              if let instructionAddress = UInt64(argument) {
```
buenaflor:
```suggestion
              let hexDigits = argument.replacingOccurrences(of: "0x", with: "")
              if let instructionAddress = UInt64(hexDigits, radix: 16) {
```

flutter/ios/Classes/SentryFlutterPluginApple.swift
buenaflor:
```suggestion
      }
      if (debugImages.isEmpty) {
          debugImages = PrivateSentrySDKOnly.getDebugImages() as [DebugMeta]
      }
```

flutter/lib/src/native/sentry_native_channel.dart
```diff
@@ -178,10 +178,11 @@ class SentryNativeChannel
...
 
   @override
-  Future<List<DebugImage>?> loadDebugImages() =>
+  Future<List<DebugImage>?> loadDebugImages(Set<String> instructionAddresses) =>
       tryCatchAsync('loadDebugImages', () async {
-        final images = await channel
-            .invokeListMethod<Map<dynamic, dynamic>>('loadImageList');
+        final images =
+            await channel.invokeListMethod<Map<dynamic, Set<String>>>(
+                'loadImageList', instructionAddresses);
```
buenaflor:
```suggestion
            await channel.invokeListMethod<Map<dynamic, dynamic>>(
                'loadImageList', instructionAddresses);
```

flutter/lib/src/native/sentry_native_binding.dart
```diff
@@ -52,7 +52,7 @@ abstract class SentryNativeBinding {
   Future<Map<String, dynamic>?> collectProfile(
       SentryId traceId, int startTimeNs, int endTimeNs);
 
-  Future<List<DebugImage>?> loadDebugImages();
+  Future<List<DebugImage>?> loadDebugImages(Set<String> instructionAddresses);
```
buenaflor:
```suggestion
  Future<List<DebugImage>?> loadDebugImages(List<String> instructionAddresses);
```

flutter/lib/src/integrations/load_image_list_integration.dart
```diff
@@ -33,12 +34,41 @@ class _LoadImageListIntegrationEventProcessor implements EventProcessor {
...
+      if (event.threads != null && event.threads!.isNotEmpty) {
+        for (var thread in event.threads!) {
+          if (thread.stacktrace != null) {
+            _collectImageAddressesFromStackTrace(
+                thread.stacktrace!, instructionAddresses);
+          }
+        }
+      }
+
+      final images = await _native.loadDebugImages(instructionAddresses);
```
buenaflor:
```suggestion
      final images = await _native.loadDebugImages(instructionAddresses.toList());
```

buenaflor (commented):
I took a quick look and fixed it locally. here's several things that went wrong:

- platform channels do not support `Set`, read [here](https://docs.flutter.dev/platform-integration/platform-channels#codec) about the supported types, which means we need to use `List` to pass the instruction addresses from flutter to native. We can continue using sets when parsing the addresses but when communicating between platform layers these need to be converted to `List`. generally this may be something that we can improve for type safety when interacting through platform channels. I was surprised that this didn't trigger any error or warning
- `UInt64(instructionAddr)` is not correct because the [base it uses to convert the String is base 10](https://developer.apple.com/documentation/swift/int/init(_:radix:)) but the instruction addresses we are dealing with are hex so we need to adjust the radix/base to 16 => `UInt64(instructionAddr, radix: 16)` but this won't still work because the instruction addresses we pass in contain the `0x` prefix which we also need to remove before we convert it to `UInt64`
- In the native channel `loadDebugImages` function you changed `Map<dynamic, dynamic>` to `Map<dynamic, Set<String>>` which doesn't work because DebugMeta fields won't match this
- if the debug images that we parsed from the stacktrace is empty, we should always return all debug images with `PrivateSentrySDKOnly.getDebugImages()`. This did not work because it was only triggered by the condition if the passed instruction addresses are null or empty and not if the parsed images afterwards are empty

buenaflor:
as per [this comment](https://github.com/getsentry/sentry-dart/issues/1755#issuecomment-2411102685) we will not use `getDebugImagesForAddresses` due to performance reasons. 

Instead we should use `SentryDebugImageProvider.getDebugImagesFromCacheForFrames` or `SentryDebugImageProvider. getDebugImagesFromCacheForThreads`, however I think the former is the one more suited for us which will be part of Sentry Cocoa `8.39.0`.

This PR is blocked until cocoa `8.39.0` is released

martinhaintz:
Changed the `SentryDebugImageProvider.getDebugImages` to `SentryDebugImageProvider.getDebugImagesForImageAddressesFromCache` and tested it via sentry.

flutter/lib/src/native/sentry_native_channel.dart
```diff
@@ -186,8 +186,15 @@ class SentryNativeChannel
   @override
   Future<List<DebugImage>?> loadDebugImages(SentryStackTrace stackTrace) =>
       tryCatchAsync('loadDebugImages', () async {
-        final images = await channel
-            .invokeListMethod<Map<dynamic, dynamic>>('loadImageList');
+        Set<String> instructionAddresses = {};
+        for (var frame in stackTrace.frames) {
```
buenaflor:
```suggestion
        for (final frame in stackTrace.frames) {
```

buenaflor:
lgtm now 👍 only one small change left

buenaflor (approved):

PR #2337, opened 2024-10-08
enhancement(frame-tracking): only store slow and frozen frames for calculation

buenaflor:
## :scroll: Description

Reduce the number of frames being stored at once, we don't need to keep track of frames that are within expected duration, only the slow and frozen ones

Improve performance of calculation and memory footprint

Existing tests assert correctness of calculations

## :pencil: Checklist

- [x] ~I added tests to verify changes~
- [x] All tests passing
- [x] No breaking changes


stefanosiano (approved):
nice!

PR #2347, opened 2024-10-11
fix(ttfd): unfinished ttfd should match ttid duration

buenaflor:

Set unfinished ttid, ttfd to cancelled

This might happen if the user navigates to another screen while ttfd is not finished

Unit tests

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -64,6 +64,8 @@ Navigator.push(
 - iOS replay integration when only `onErrorSampleRate` is specified ([#2306](https://github.com/getsentry/sentry-dart/pull/2306))
 - Fix TTID timing issue ([#2326](https://github.com/getsentry/sentry-dart/pull/2326))
 - Start missing TTFD for root screen transaction ([#2332](https://github.com/getsentry/sentry-dart/pull/2332))
+- Unfinished TTID or TTFD spans should be set to status `cancelled` ([#2347](https://github.com/getsentry/sentry-dart/pull/2347))
+  - This might happen if a user navigates to another screen while either span is unfinished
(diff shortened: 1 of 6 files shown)


stefanosiano:
not sure about this
we already specify the deadline_exceeded [in the docs](https://docs.sentry.io/platforms/flutter/integrations/routing-instrumentation/#time-to-full-display)
Also, the same thing is done in other SDKs. The idea was that the user can immediately understand something is wrong with their instrumentation, which is more difficult to realize with the canceled status
Is there a reason/demand for this change?

buenaflor:
@stefanosiano 

> Also, the same thing is done in other SDKs. The idea was that the user can immediately understand something is wrong with their instrumentation, which is more difficult to realize with the canceled status

it still finishes with `deadline exceeded` if it actually times out after 30 seconds but it's also possible that a user navigates away after 2 seconds. we can keep it as `deadline_exceeded` to keep in line with other sdks if they do the same in that scenario, I don't have a strong opinion about it

the rest of the changes still make sense imo: setting ttfd span duration to ttid span duration if ttid duration is available and additional test

stefanosiano:
> it still finishes with deadline exceeded if it actually times out after 30 seconds but it's also possible that a user navigates away after 2 seconds. we can keep it as deadline_exceeded to keep in line with other sdks if they do the same in that scenario, I don't have a strong opinion about it

Yeah, the others keep it as `deadline_exceeded` even when the transaction (navigation) is finished before a frame is drawn

> the rest of the changes still make sense imo: setting ttfd span duration to ttid span duration if ttid duration is available and additional test

👍

stefanosiano (approved):

PR #2361, opened 2024-10-15
redact screenshots via view hierarchy

### LINKED ISSUE #1956: Redact Screenshots via View Hierarchy
We have Screenshots, and we have View Hierarchy. For Mobile SR, we use VH to redact SS-based Replays - we should "backport" this functionality to the SS feature.

Rationale: SS can contain PII and other sensitive information, therefore the feature is not viable for some use cases and orgs. Redacting SS via VH enables wider adoption and improves experience for those who already use it.

Let's evaluate:
1) Implement screeshot redacting and use it for session replay later
2) Wait for session replay and reuse its implementation for SS and VH

martinhaintz:
## :scroll: Description
Switching to View Hierarchy for screenshot generation.

close #1956 

Example App and Unittests

## :pencil: Checklist

- [x] All tests passing
- [ ] No breaking changes

## :crystal_ball: Next steps
- [ ] Update the Docs


flutter/lib/src/sentry_replay_options.dart
```diff
@@ -32,110 +33,17 @@ class SentryReplayOptions {
     _onErrorSampleRate = value;
   }
 
-  /// Mask all text content. Draws a rectangle of text bounds with text color
-  /// on top. Currently, only [Text] and [EditableText] Widgets are redacted.
-  /// Default is enabled.
-  var maskAllText = true;
```
buenaflor, line 38: Ideally we'll align with the other mobile sdks how to provide these options (replay + screenshot redacting) since they'll also implement screenshot redacting.

unless I missed something and this has been decided already :D

martinhaintz: @buenaflor I am currently not aware of any implementation for screenshot redaction. 
But the options for `Session Replay` are already aligned (at least React Native and Flutter using the same options), and therefore I use the same options for `Screenshot` as `Session Replay`builds on top of `Screenshot`.

martinhaintz:
@vaind @buenaflor I moved the redacting features from the SR(Screen Replay) config to the SS(ScreenShot) config and derived from it. I also copied the options from SR for the SS functionality. 

- Should we keep the settings separate, so the user can enable redacting for SR but disable it for SS? 

- Why is the SR feature experimental? Should the redacting settings for SS also stay experimental?

vaind:
> Why is the SR feature experimental? Should the redacting settings for SS also stay experimental?

There still may be changes to replay as well as the redaction logic, therefore I'd say it should be marked experimental for screenshot redaction too. The main purpose is that it lets users know they should use this feature with care as it may have some issues. Also it may change without a major release.

flutter/lib/src/event_processor/screenshot_event_processor.dart
```diff
@@ -75,83 +75,34 @@ class ScreenshotEventProcessor implements EventProcessor {
...
-        var imageResult = _getImage(renderObject, pixelRatio);
-        Image image;
-        if (imageResult is Future<Image>) {
-          image = await imageResult;
-        } else {
-          image = imageResult;
-        }
-        // At the time of writing there's no other image format available which
-        // Sentry understands.
+    Uint8List? _screenshotCache;
```
vaind: "cache" is a little misleading here. We don't cache it for future use

flutter/lib/src/event_processor/screenshot_event_processor.dart
```diff
@@ -75,83 +75,34 @@ class ScreenshotEventProcessor implements EventProcessor {
...
-      return repaintBoundary.toImage(pixelRatio: pixelRatio);
+  Future<Uint8List?> _convertImageToUint8List(Image image) async {
+    final byteData = await image.toByteData(format: ImageByteFormat.png);
+
+    final bytes = byteData?.buffer.asUint8List();
+    if (bytes?.isNotEmpty == true) {
+      return bytes;
+    } else {
+      _options.logger(
+          SentryLevel.debug, 'Screenshot is 0 bytes, not attaching the image.');
```
vaind, line 119: Ever happened to you while testing? Looks like a bug.

flutter/lib/src/screenshot/recorder.dart
```diff
@@ -21,7 +23,13 @@ class ScreenshotRecorder {
   bool warningLogged = false;
 
   ScreenshotRecorder(this.config, this.options) {
-    final maskingConfig = options.experimental.replay.buildMaskingConfig();
+    SentryMaskingConfig maskingConfig;
+    if (config is ScheduledScreenshotRecorderConfig) {
+      maskingConfig = options.experimental.replay.buildMaskingConfig();
+    } else {
+      maskingConfig = options.experimental.screenshot.buildMaskingConfig();
+    }
```
vaind: Do we need/want to have two distinct configurations? Sounds like it's going to be a pain to configure and is asking for trouble because stuff will be masked in one capture and not in the other.

flutter/lib/src/screenshot/sentry_screenshot_quality.dart
```diff
@@ -17,4 +17,30 @@ enum SentryScreenshotQuality {
         return 854;
     }
   }
+
+  int calculateHeight(int width, int height) {
```
vaind: doesn't need to be exposed - this is an implementation detail (a.k.a `@internal`)

flutter/lib/src/screenshot/recorder.dart
```diff
@@ -82,12 +90,21 @@ class ScreenshotRecorder {
...
+        Image finalImage;
+        if (config is ScheduledScreenshotRecorderConfig) {
+          finalImage = await picture.toImage((srcWidth * pixelRatio).round(),
+              (srcHeight * pixelRatio).round());
+        } else {
+          final targetHeight = config.quality
+              .calculateHeight(srcWidth.toInt(), srcHeight.toInt());
+          final targetWidth = config.quality
+              .calculateWidth(srcWidth.toInt(), srcHeight.toInt());
+          finalImage = await picture.toImage(targetWidth, targetHeight);
```
vaind: the original replay is equivalent to screenshot "full" quality AFAICT. I think we can integrate the configs and keep the replay default to "full" for now until we implement quality settings throughout the whole stack for replay.

flutter/lib/src/sentry_replay_options.dart
vaind: @bruno-garcia @romtsn @brustolin any thoughts on the common configuration for masking/redacting screenshots as well as replay?

I'd say we could have the masking stuff in:

`options.experimental.screenshot.mask*`

while anything really replay-specific would stay under `.replay`. Of course the experimental bit would be dropped later from the path.

vaind (commented):
I realize this is still a draft but wanted to have a first look early. Looks like you're on a good path to integrating these two features, I've just had some comments what could possibly be improved.

flutter/lib/src/sentry_replay_options.dart
romtsn: yes, that makes a lot of sense to me! I like having sub-options/namespace for specific features

flutter/lib/src/event_processor/screenshot_event_processor.dart
martinhaintz: yeah, you are right. thanks for the hint.

flutter/lib/src/event_processor/screenshot_event_processor.dart
martinhaintz: No, not that I am aware of. This [code](https://github.com/getsentry/sentry-dart/blob/8f95e333de1ac9c67a8a928fe77cce146c505818/flutter/lib/src/event_processor/screenshot_event_processor.dart#L127-L131) was already here before.
Maybe @denrase knows more?

flutter/lib/src/screenshot/recorder.dart
martinhaintz: @vaind Good point, I also discussed it with @buenaflor. 
What about, having global settings that apply for both, and have override options for `SR` and `SS` for power users? Or should we start with the one for both and wait until the community asks for a different behavior?

vaind: > Or should we start with the one for both and wait until the community asks for a different behavior?

IMO, yes, keep it simple

flutter/lib/src/sentry_redaction_options.dart
```diff
@@ -0,0 +1,111 @@
+import 'package:flutter/services.dart';
+import 'package:flutter/widgets.dart';
+import 'package:meta/meta.dart';
+
+import '../sentry_flutter.dart';
+import 'screenshot/masking_config.dart';
+import 'screenshot/widget_filter.dart';
+
+/// Configuration of the experimental redaction feature.
+class SentryRedactingOptions {
```
vaind: WDYAT about naming this `SentryPrivacyOptions` and the name in options.experimental.privacy?

martinhaintz: yeah, I like this idea.

flutter/lib/src/event_processor/screenshot_event_processor.dart
martinhaintz: @vaind I looked up the `toByteData` method, and it will return `null` if an error occurs and also throws an exception. I would rewrite this, with a try/catch block, but we can also remove this code, if you think, that we don`t need it. WDYT?

flutter/lib/src/sentry_flutter.dart
```diff
@@ -243,6 +245,19 @@ mixin SentryFlutter {
...
+      return;
+    } else if (options.screenshot.attachScreenshot == true &&
+        !options.experimental.replay.isEnabled) {
+      options.experimental.privacy = SentryPrivacyOptions()
+        ..maskAllText = false
+        ..maskAllImages = false;
+    } else {
+      options.experimental.privacy = SentryPrivacyOptions();
+    }
+  }
```
buenaflor: maybe also makes sense to add a short comment why we are doing it this way to make it a bit more clear

vaind (commented):
I've started looking into this and it all looks pretty good, but I'll have a closer look later today/early tomorrow, especially around options.experimental.privacy and how a user would set it (in the sample app). I think it should be possible for the user to not have to create the instance itself (i.e. it can have a non-nullable public getter)

flutter/lib/src/sentry_flutter.dart
```diff
@@ -243,6 +245,27 @@ mixin SentryFlutter {
...
 
+  /// Screen redaction was previously introduced with the SessionReplay feature.
+  /// Screen redaction is enabled by default for SessionReplay.
+  /// As we also to use this feature for Screenshot, which previously was not
+  /// capable of redacting the screenshot, we need to disable redaction for Screenshot by default
+  /// so we don`t break the existing behavior.
+  /// As we have only one central place to configure the redaction,
+  /// we need to set the redaction options to full fill the above default settings.
+  /// The plan is to unify this behaviour with the next major release.
+  static void _setRedactionOptions(SentryFlutterOptions options) {
```
```suggestion
  static void _setMaskingOptions(SentryFlutterOptions options) {
```

minor: since we changed the wording from redact -> mask, it might make sense to change it

flutter/lib/src/sentry_flutter.dart
```diff
@@ -243,6 +245,27 @@ mixin SentryFlutter {
...
   }
 
+  /// Screen redaction was previously introduced with the SessionReplay feature.
+  /// Screen redaction is enabled by default for SessionReplay.
+  /// As we also to use this feature for Screenshot, which previously was not
+  /// capable of redacting the screenshot, we need to disable redaction for Screenshot by default
+  /// so we don`t break the existing behavior.
+  /// As we have only one central place to configure the redaction,
+  /// we need to set the redaction options to full fill the above default settings.
+  /// The plan is to unify this behaviour with the next major release.
```
```suggestion
  /// Masking behaviour
  /// - If only Screenshot is enabled: masking is disabled by default.
  /// - If both Screenshot and Session Replay are enabled: masking is enabled for both by default.
  /// - If the user explicitly sets masking to false: masking is disabled for both features.
  /// We don't want to break the existing screenshot integration which is not masked by default.
  /// The plan is to unify screenshot and replay masking with the next major release.
```

suggestion to keep it a bit shorter

flutter/lib/src/sentry_privacy_options.dart
```diff
@@ -0,0 +1,111 @@
...
+import 'package:meta/meta.dart';
+
+import '../sentry_flutter.dart';
+import 'screenshot/masking_config.dart';
+import 'screenshot/widget_filter.dart';
+
+/// Configuration of the experimental privacy feature.
+class SentryPrivacyOptions {
+  /// Mask all text content. Draws a rectangle of text bounds with text color
+  /// on top. Currently, only [Text] and [EditableText] Widgets are redacted.
```
buenaflor:
```suggestion
  /// on top. Currently, only [Text] and [EditableText] Widgets are masked.
```

let's keep the naming consistent

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ### Enhancements
 
 - Cache parsed DSN ([#2365](https://github.com/getsentry/sentry-dart/pull/2365))
+- Switching from traditional screenshot to view hierarchy for screenshots which allows redacting ([#2361](https://github.com/getsentry/sentry-dart/pull/2361))
```
buenaflor: the wording is a bit confusing from a user perspective we can improve it by saying`Allow Screenshots to be masked for privacy reasons` or something like that

Also show an exampl how a user can enable it and please add the screenshot & replay masking behaviour (the one that I commented) as sub bullet points for more context so it's clear how it behaves

buenaflor: The changelog entry should be under `## Features`, if there is none please create it

buenaflor (changes requested):
left some small feedback regarding changing redact -> mask in comments etc... and some changelog changes

generally looks good to me but I'll have ivan give the final ✅

flutter/lib/src/sentry_screenshot_options.dart
```diff
@@ -0,0 +1,32 @@
...
+import '../sentry_flutter.dart';
+
+/// Configuration of the screenshot feature.
+class SentryScreenshotOptions {
+  /// Automatically attaches a screenshot when capturing an error or exception.
+  ///
+  /// Requires adding the [SentryScreenshotWidget] to the widget tree.
+  /// Example:
+  /// runApp(SentryScreenshotWidget(child: App()));
+  /// The [SentryScreenshotWidget] has to be the root widget of the app.
```
vaind: We should point users to use the single root widget:
```suggestion
  /// The [SentryWidget] has to be the root widget of the app.
```

flutter/lib/src/sentry_flutter_options.dart
```diff
@@ -380,11 +393,7 @@ class SentryFlutterOptions extends SentryOptions {
...
-}
 
-/// Callback being executed in [ScreenshotEventProcessor], deciding if a
-/// screenshot should be recorded and attached.
-typedef BeforeScreenshotCallback = FutureOr<bool> Function(
-  SentryEvent event, {
-  Hint? hint,
-});
+  /// Privacy configuration for masking sensitive data in the Screenshot and Session Replay.
+  SentryPrivacyOptions? privacy;
```
vaind: I've made suggestion how we can make this more user-friendly not having to initialize the field) in another PR on top of this one: #2382

vaind (changes requested):
generally looks good, please have a look at my PR mentioned below. 

However, there's some bug with the sizing because the image is captured too small (with transparent offset on the right and bottom that is actually part of the image):
https://sentry-sdks.sentry.io/issues/4722238716/events/f071c8f05dac430ba64411635fcb49da/
image

When previously it would show up correctly (size-wise, ignore the content of this one):
https://sentry-sdks.sentry.io/issues/4722238716/events/7479041de7b14cfcaf870d55ac1023ad/
image

martinhaintz:
@vaind could you please have a look at the tests `clears replay ID from context` and `captures images`? thanks.

flutter/lib/src/sentry_flutter_options.dart
```diff
@@ -385,11 +398,33 @@ class SentryFlutterOptions extends SentryOptions {
...
+  /// Note: this will change in a future SDK major release to enable screenshot
+  /// redaction by default for all captures.
+  SentryPrivacyOptions get privacy {
+    // If the user explicitly sets the privacy setting, we use that.
+    // Otherwise, we use the default settings, which is no redaction for screenshots
+    // and full redaction for session replay.
+    // This property must only by accessed by user code otherwise it defeats the purpose.
+    _privacy ??= SentryPrivacyOptions();
+    return _privacy!;
+  }
```
martinhaintz: @vaind what will/should happen, if the user disables all `privacy` settings? Will this also disable masking for `Session Replay`? Or will this only affect the `Screenshot` part?

vaind: > Will this also disable masking for Session Replay

yes

CHANGELOG.md
```diff
@@ -13,6 +13,21 @@
...
+  await SentryFlutter.init(
+    (options) {
+      ...
+      options.experimental.privacy.maskAllText = true;
+      options.experimental.privacy.maskAllImages = true;
+      options.experimental.privacy.maskAssetImages = true;
+    },
+    appRunner: () => runApp(MyApp()),
+  );
+  ```
```
vaind: This is a bit of a mouthful. Also misleading about being disabled for screenshot, because the attached code example would actually enable masking for screenshots too. 

Maybe:

````suggestion
- Support for screenshot PII content redaction (masking) ([#2361](https://github.com/getsentry/sentry-dart/pull/2361))
  By default, masking is enabled for SessionReplay. To also enable it for screenshots captured with events, you just can specify `options.experimental.privacy`:
  ```dart
  await SentryFlutter.init(
    (options) {
      ...
      // the defaults are:
      options.experimental.privacy.maskAllText = true;
      options.experimental.privacy.maskAllImages = true;
      options.experimental.privacy.maskAssetImages = false;
      // you cal also set up custom masking, for example:
      options.experimental.privacy.mask<WebView>();
    },
  );
  ```
  Actually, just accessing this field will cause it to be initialized with the default settings to mask all text and images:
  ```dart
  await SentryFlutter.init(
    (options) {
      ...
      // this has a side-effect of creating the default privacy configuration, thus enabling Screenshot redaction:
      options.experimental.privacy;
    },
  );
  ```  
````

File: flutter/lib/src/sentry_privacy_options.dart [hunk 17]
```diff
@@ -0,0 +1,111 @@
+import 'package:flutter/services.dart';
+import 'package:flutter/widgets.dart';
+import 'package:meta/meta.dart';
+
+import '../sentry_flutter.dart';
+import 'screenshot/masking_config.dart';
+import 'screenshot/widget_filter.dart';
+
+/// Configuration of the experimental privacy feature.
+class SentryPrivacyOptions {
```
vaind, line 10: Is this just moved from SentryReplayOptions or have there been any changes? Just so that I know how detailed the review should be.

flutter/lib/src/sentry_flutter_options.dart
```diff
@@ -385,11 +398,33 @@ class SentryFlutterOptions extends SentryOptions {
...
   final replay = SentryReplayOptions();
-}
 
-/// Callback being executed in [ScreenshotEventProcessor], deciding if a
-/// screenshot should be recorded and attached.
-typedef BeforeScreenshotCallback = FutureOr<bool> Function(
-  SentryEvent event, {
-  Hint? hint,
-});
+  /// Privacy configuration for masking sensitive data in the Screenshot and Session Replay.
```
vaind: I know I've originally written this in the other PR but that doesn't make it correct :D
```suggestion
  /// Privacy configuration for masking sensitive data in screenshots and Session Replay.
```

flutter/lib/src/sentry_screenshot_options.dart
```diff
@@ -0,0 +1,32 @@
...
+  /// relevant if `attachScreenshot` is set to true. When false is returned
+  /// from the function, no screenshot will be attached.
+  BeforeScreenshotCallback? beforeScreenshot;
+
+  /// Only attach a screenshot when the app is resumed.
+  bool attachScreenshotOnlyWhenResumed = false;
+
+  /// The quality of the attached screenshot
+  SentryScreenshotQuality screenshotQuality = SentryScreenshotQuality.high;
+}
```
vaind: This is "namespaced" under `options.screenshot` so we don't need to repeat "screenshot" in field names anymore. Maybe:

```suggestion
class SentryScreenshotOptions {
  /// Automatically attaches a screenshot when capturing an error or exception.
  ///
  /// Requires adding the [SentryWidget] to the widget tree.
  /// Example:
  /// runApp(SentryWidget(child: App()));
  /// The [SentryWidget] has to be the root widget of the app.
  bool attach = false;

  /// Sets a callback which is executed before capturing screenshots. Only
  /// relevant if `attachScreenshot` is set to true. When false is returned
  /// from the function, no screenshot will be attached.
  BeforeScreenshotCallback? beforeCapture;

  /// Only attach a screenshot when the app is resumed.
  bool attachOnlyWhenResumed = false;

  /// The quality of the attached screenshot
  SentryScreenshotQuality quality = SentryScreenshotQuality.high;
}
```

also, `  /// Only attach a screenshot when the app is resumed.` makes no sense to me. Do you know what it does? If so can you maybe update the comment, even though it's. unrelated to this PR?

flutter/lib/src/native/java/sentry_native_java.dart
```diff
@@ -40,8 +40,8 @@ class SentryNativeJava extends SentryNativeChannel {
             _startRecorder(
               call.arguments['directory'] as String,
               ScheduledScreenshotRecorderConfig(
-                width: call.arguments['width'] as int,
-                height: call.arguments['height'] as int,
+                targetWidth: call.arguments['targetWidth'] as int,
+                targetHeight: call.arguments['targetHeight'] as int,
```
vaind: can we avoid the rename? doesn't seem necessary. Also, this being a "configuration", not a status it's implicit that it's a "target" width/height, just like it's a target frame rate at which we want to capture.

flutter/lib/src/screenshot/recorder.dart
```diff
@@ -47,9 +61,17 @@ class ScreenshotRecorder {
...
+      final srcHeight = renderObject.size.height.toInt();
+
+      // In Session Replay the target size is already set and should not be changed.
+      // For Screenshots, we need to calculate the target size based on the quality setting.
+      config.targetHeight ??=
+          config.quality.calculateHeight(srcWidth, srcHeight);
+      config.targetWidth ??= config.quality.calculateWidth(srcWidth, srcHeight);
+
+      var pixelRatio =
+          config.getPixelRatio(srcWidth.toDouble(), srcHeight.toDouble());
```
vaind: I understand you've made these changes to fix the issue with screenshots not filling the whole space. I'm pretty confident these changes aren't necessary. The pixel ratio is actually what needs to be coming based on the config. And it must be a single value because we don't want to skew the image.

flutter/lib/src/screenshot/recorder_config.dart
```diff
@@ -0,0 +1,26 @@
...
+
+import 'package:meta/meta.dart';
+
+import '../../sentry_flutter.dart';
+
+@internal
+class ScreenshotRecorderConfig {
+  int? targetWidth;
+  int? targetHeight;
+  final SentryScreenshotQuality quality;
```
vaind: Quality is unnecessary here. It's just width/height setting and should be passed as such.

flutter/lib/src/screenshot/recorder_config.dart
```diff
@@ -0,0 +1,26 @@
+import 'dart:math';
+
+import 'package:meta/meta.dart';
+
+import '../../sentry_flutter.dart';
+
+@internal
+class ScreenshotRecorderConfig {
+  int? targetWidth;
+  int? targetHeight;
```
vaind: I'd prefer if we could avoid `target` prefix here, see my other comment below, but in short, this is a configuration, not status so it always is "target", i.e. "what we want" (where "source" would be the input/what we have)

flutter/lib/src/event_processor/screenshot_event_processor.dart
```diff
@@ -1,20 +1,27 @@
...
 
 class ScreenshotEventProcessor implements EventProcessor {
   final SentryFlutterOptions _options;
 
-  ScreenshotEventProcessor(this._options);
+  late final ScreenshotRecorder _recorder;
+
+  ScreenshotEventProcessor(this._options) {
+    _recorder = ScreenshotRecorder(
+      ScreenshotRecorderConfig(quality: _options.screenshot.screenshotQuality),
```
vaind: Instead of passing quality here, we should pass the width & height that corresponds to the configured quality.

Something like this:

```suggestion
      final size = _options.screenshot.quality.targetResolution();
      _recorder = ScreenshotRecorder(
        ScreenshotRecorderConfig(width: size, height: size),
```

flutter/lib/src/screenshot/sentry_screenshot_quality.dart
```diff
@@ -17,4 +21,34 @@ enum SentryScreenshotQuality {
         return 854;
     }
   }
+
+  @internal
+  int calculateHeight(int deviceWidth, int deviceHeight) {
```
vaind: this should be unnecessary, taking int account my other comments/change suggestions.

flutter/lib/src/screenshot/sentry_screenshot_quality.dart
```diff
@@ -17,4 +21,34 @@ enum SentryScreenshotQuality {
...
+        return targetResolution()!;
+      } else {
+        var ratio = targetResolution()! / deviceWidth;
+        return (deviceHeight * ratio).round();
+      }
+    }
+  }
+
+  @internal
+  int calculateWidth(int deviceWidth, int deviceHeight) {
```
vaind: this should be unnecessary, taking int account my other comments/change suggestions.

vaind (changes requested):
This is on a very good track IMO. Just some changes sizing and it should be good to go (I haven't done any manual testing yet, would do that after the changes with a final review)

flutter/lib/src/sentry_privacy_options.dart
martinhaintz: yes, you are right, I just moved them out of the `SentryReplayOptions`.

flutter/lib/src/native/java/sentry_native_java.dart
martinhaintz: I renamed it, because it was hard to distinguish between multiple variables which were all called `width` and `height` with some are having the `sourceWidth`, `logicalWidth`, `deviceWidth` and `targetWidth`. So for me it wasn't obvious and a little confusing.
But sure, I can undo the renaming. Can I add some comments, stating that this is the `targetWidth`?

flutter/lib/src/screenshot/recorder.dart
martinhaintz: I refactored it, based on your input. I think it is now way cleaner as before.

martinhaintz:
@vaind how, is the change of the device resolution handled in Session Replay? 
Lets say we have a mac desktop application and it is started in portrait mode (w:600, h: 1200), then the recorder config at startup is set to this target resolution. If the user resizes the windows to landscape (w:1200, h: 600), then recorder config for the target size is still in portrait mode, although everything in the window is rendered for landscape.

flutter/lib/src/sentry_screenshot_options.dart
martinhaintz: @vaind I found this in the docs: [Screenshot Integration Background Crash](https://docs.sentry.io/platforms/flutter/troubleshooting/#screenshot-integration-background-crash)
So I think we should leave this setting there. WDYT?

CHANGELOG.md
```diff
@@ -13,6 +13,33 @@
   
 ### Features
 
+- Support for screenshot PII content redaction (masking) ([#2361](https://github.com/getsentry/sentry-dart/pull/2361))
+  By default, masking is enabled for SessionReplay. To also enable it for screenshots captured with events, you just can specify `options.experimental.privacy`:
```
vaind:
```suggestion
  By default, masking is enabled for SessionReplay. To also enable it for screenshots captured with events, you can specify `options.experimental.privacy`:
```

CHANGELOG.md
```diff
@@ -114,6 +141,7 @@
 
 - Avoid sending too many empty client reports when Http Transport is used ([#2380](https://github.com/getsentry/sentry-dart/pull/2380))
 - Cache parsed DSN ([#2365](https://github.com/getsentry/sentry-dart/pull/2365))
+- Switching from traditional screenshot to view hierarchy for screenshots which allows redacting ([#2361](https://github.com/getsentry/sentry-dart/pull/2361))
```
vaind: Duplicate entry. 
```suggestion
```

flutter/lib/src/event_processor/screenshot_event_processor.dart
```diff
@@ -1,20 +1,30 @@
...
-  ScreenshotEventProcessor(this._options);
+  late final ScreenshotRecorder _recorder;
+
+  ScreenshotEventProcessor(this._options) {
+    final targetResolution = _options.screenshot.quality.targetResolution();
+
+    _recorder = ScreenshotRecorder(
+      ScreenshotRecorderConfig(
+        width: targetResolution?.toInt(),
+        height: targetResolution?.toInt(),
```
vaind:
```suggestion
        width: targetResolution,
        height: targetResolution,
```

flutter/test/replay/scheduled_recorder_test.dart
```diff
@@ -38,7 +38,7 @@ class _Fixture {
     sut = ScheduledScreenshotRecorder(
       ScheduledScreenshotRecorderConfig(
         width: 1000,
-        height: 1000,
+        height: 750,
```
vaind: The original test should work even with the changes:

```suggestion
        height: 1000,
```

flutter/test/event_processor/screenshot_event_processor_test.dart
```diff
@@ -69,7 +69,7 @@ void main() {
   testWidgets('does not add screenshot attachment with html renderer',
       (tester) async {
     await _addScreenshotAttachment(tester, FlutterRenderer.html,
-        added: false, isWeb: true);
+        added: true, isWeb: true);
```
vaind: Are you sure screenshots work with the html renderer? Is it tested properly? I haven't noticed any new tests specifically targeting web

vaind:
```suggestion
        added: false, isWeb: true);
```

flutter/lib/src/screenshot/recorder.dart
```diff
@@ -47,9 +58,31 @@ class ScreenshotRecorder {
...
+        targetWidth = binding.window.physicalSize.width.toInt();
+
+        pixelRatio = max(targetWidth, targetHeight) / max(srcWidth, srcHeight);
+      } else {
+        pixelRatio =
+            max(config.width!, config.height!) / max(srcWidth, srcHeight);
+
+        targetWidth = (srcWidth * pixelRatio).toInt();
+        targetHeight = (srcHeight * pixelRatio).toInt();
+      }
```
vaind: As I've said in the previous run, we don't need these changes and I really really wouldn't want to touch this because it affects masking. While the changes you've made don't seem to break it at the moment, they do add complexity to the code so it would be easier to break it in the future. I'm talking from experience here because there was a hard-to-spot issue in the logic here before, during the initial development, that caused replays to not be masked fully around edges in some specific situations.

I've created a PR that reverts these changes to the original code and everything seems to work just fine https://github.com/getsentry/sentry-dart/pull/2409 with a small test adjustment. Also tested manually on Android.

The only difference is you're using physical window size here but there's no point to do that, we already have the resolution defined by `renderObject`. Window size actually doesn't matter - it's entirely possible that we're not even capturing the whole application screen, only a select part (whatever is wrapped in `SentryWidget`)

vaind (changes requested):

flutter/lib/src/sentry_screenshot_options.dart
vaind: well, another thing to keep track of in future Flutter version bumps then.

flutter/lib/src/sentry_screenshot_options.dart
```diff
@@ -0,0 +1,32 @@
...
+  /// The [SentryWidget] has to be the root widget of the app.
+  bool attach = false;
+
+  /// Sets a callback which is executed before capturing screenshots. Only
+  /// relevant if `attach` is set to true. When false is returned
+  /// from the function, no screenshot will be attached.
+  BeforeScreenshotCallback? beforeCapture;
+
+  /// Only attach a screenshot when the app is resumed.
+  bool attachOnlyWhenResumed = false;
```
vaind:
```suggestion
  /// Only attach a screenshot when the app is resumed.
  /// See https://docs.sentry.io/platforms/flutter/troubleshooting/#screenshot-integration-background-crash
  bool attachOnlyWhenResumed = false;
```

flutter/lib/src/event_processor/screenshot_event_processor.dart
vaind: since it's an old code, and it's not really broken, I'm fine with leaving it as is. In an unlikely event there'd be an error, it would be caught by `ScreenshotRecorder` anyway.

vaind:
> @vaind how, is the change of the device resolution handled in Session Replay? 

Currently, it's an open issue. I've checked how this behaves and what needs to be done and I'll make a PR to address this

> Lets say we have a mac desktop application

Just FYI that doesn't support session replay at the moment, but the platform doesn't matter for the argument you're making. Just wanted to clarify for anyone else reading this.


> and it is started in portrait mode (w:600, h: 1200), then the recorder config at startup is set to this target resolution. If the user resizes the windows to landscape (w:1200, h: 600), then recorder config for the target size is still in portrait mode, although everything in the window is rendered for landscape.

For screenshots themselves (i.e. this PR) that shouldn't matter because we're setting a "constraint", not really dimensions (`SentryScreenshotQuality.targetResolution()` returns a single value and we use that as the target either for width or for height).

For replays, this needs to change and we'll need to respect changes in resolution in the PR-to-be I've mentioned above. Basically, we'll need to stop, reconfigure & restart replay when a resolution changes.

flutter/test/replay/scheduled_recorder_test.dart
martinhaintz: But, you previously wrote, that the `height` and `width` in the `RecorderConfig` is the `target height` and `target width`. Therefore the resulting image must be `1000x1000` regarding the config. But the test is looking for an image of the size of `1000x750`.

flutter/test/event_processor/screenshot_event_processor_test.dart
martinhaintz: yes it's working, I just added some tests and previously tested it locally.

flutter/test/replay/scheduled_recorder_test.dart
vaind: Hmm, I guess I should have been clearer. These dimensions are width & height the replay recorder wants. That doesn't mean we should skew the image. Therefore, the only other option if we were to respect the dimensions precisely while keeping the actual screenshot with the right aspect ratio would be painting the image with the original aspect ratio over a blank image but that's not really necessary because just encoding the image of different dimensions to a video achieves the same. And because it's not necessary (and it would be slower), then there's no point in doing that. BTW that's also how ScreenshotQuality handled screenshot capture even before this PR, although it had only one "targetResolution()" for both width & height.

As for the test case. I believe it's a good test that the desired width/height are respected but the aspect ratio is kept, which we verify later in the test case. It's true that the test case with `height: 750` would also pass, it just wouldn't test that we respect the aspect ratio.

vaind: > It's true that the test case with height: 750 would also pass, it just wouldn't test that we respect the aspect ratio.

Maybe worth adding an explicit test case if you think that would be clearer.

flutter/test/event_processor/screenshot_event_processor_test.dart
vaind: In that case it should have another changelog item IMO - something about adding support for screenshot capture with html renderer (flutter web)

flutter/lib/src/screenshot/recorder.dart
martinhaintz: In #2409 [here](https://github.com/getsentry/sentry-dart/pull/2409#discussion_r1842042755) I noticed that you removed the line `options.screenshotQuality = SentryScreenshotQuality.low;` in the `main.dart`. Was this intentional or by accident?

For the the Quality `low`, `medium`, and `high` everything works with your suggested changes in #2409, but for `full` the output image is about the size as the `low` setting, as it uses the `logical resolution` and not the `device resolution` therefore it would break the previous behaviour.

To capture screenshots with quality `full` and keep the previous functionality, we need to get either the `device resolution` or the `devicePixelRatio`. As both values can change, for `Web` and `Desktop` applications, due to resizing of the window, setting the `ScreenshotRecorderConfig` to a fixed value at app start does not work. As it will stay the same even after resizing the window. Therefore the calculation needs to be done, before every screenshot capture. But this approach is only working if the `widget` is covering the complete device display, which is no longer a must.

For me it looks like, the best/only/easiest solution would be to deprecate the `full` setting in the `SentryScreenshotQuality` due to the change of the approach of generating the screenshot. This would also make sense, regarding the possibility, of only capturing a screenshot of a specific widget, which is neither related to the `device resolution` nor to the `device orientation`.
By ignoring the `full` setting, the user will still get screenshots, but with a lower resolution, as with the previous setting, and therefore the change is not that big.

@vaind @buenaflor WDYT? Should we deprecate the `full` or do you have another suggestion?

Should we exchange the `full`, with `extraHigh`? But this would be a breaking change, because it will no longer return `null` and the `enum name` changes and the resolution will also change.

martinhaintz:
> For replays, this needs to change and we'll need to respect changes in resolution in the PR-to-be I've mentioned above. Basically, we'll need to stop, reconfigure & restart replay when a resolution changes.

I understand this behavior, but as a user it would be nice, to have a video and see, that the resizing of the video probably resulted in a crash.

flutter/lib/src/screenshot/recorder.dart
vaind: > In #2409 [here](https://github.com/getsentry/sentry-dart/pull/2409#discussion_r1842042755) I noticed that you removed the line `options.screenshotQuality = SentryScreenshotQuality.low;` in the `main.dart`. Was this intentional or by accident?

intentional, I was thinking we should be testing the default value for quality when testing manually, so that it's the same as what users see by default.
 
> For the the Quality `low`, `medium`, and `high` everything works with your suggested changes in #2409, but for `full` the output image is about the size as the `low` setting, as it uses the `logical resolution` and not the `device resolution` therefore it would break the previous behaviour.

Interesting, it worked for me as expected on Windows. Which platforms have you tried this on so far?


> To capture screenshots with quality `full` and keep the previous functionality, we need to get either the `device resolution` or the `devicePixelRatio`. 

That is a very good point. We should grab devicePixelRatio and take it into account when calculating the pixelRatio based on constraints: https://github.com/getsentry/sentry-dart/pull/2409/files#diff-b603ed629ae7bb928af4f892324700593bfeb22ecff70a571e58345034b881a4R61
That should be a small change, I can have a look if you want. 

> As both values can change, for `Web` and `Desktop` applications, due to resizing of the window, setting the `ScreenshotRecorderConfig` to a fixed value at app start does not work. As it will stay the same even after resizing the window. Therefore the calculation needs to be done, before every screenshot capture. But this approach is only working if the `widget` is covering the complete device display, which is no longer a must.

For screenshots, we don't read or set window size at the start, we read it when capturing the screnshot: https://github.com/getsentry/sentry-dart/pull/2409/files#diff-b603ed629ae7bb928af4f892324700593bfeb22ecff70a571e58345034b881a4R59-R60. For replay, this shall change too when we support orientation change.

The configuration for `full` has both `width` and `height` as null, therefore, the calculated pixelRatio (as in the original code or in the revert PR, not in the current version here) would be 1.0 so that's not a problem. 

> For me it looks like, the best/only/easiest solution would be to deprecate the `full` setting in the `SentryScreenshotQuality` due to the change of the approach of generating the screenshot. This would also make sense, regarding the possibility, of only capturing a screenshot of a specific widget, which is neither related to the `device resolution` nor to the `device orientation`. By ignoring the `full` setting, the user will still get screenshots, but with a lower resolution, as with the previous setting, and therefore the change is not that big.
> 
> @vaind @buenaflor WDYT? Should we deprecate the `full` or do you have another suggestion?
> 
> Should we exchange the `full`, with `extraHigh`? But this would be a breaking change, because it will no longer return `null` and the `enum name` changes and the resolution will also change.

I don't think of the screenshot quality as the function of window size, but of the size of the captured screenshot, which is always coming from the ScreenshotWidget size. I'd keep the setting, with my above suggestion to change the pixelRatio calculation, it should work the same as before.

vaind:
> I understand this behavior, but as a user it would be nice, to have a video and see, that the resizing of the video probably resulted in a crash.

I believe it's still a single replay, just a different segment

flutter/lib/src/screenshot/recorder.dart
martinhaintz: > > In #2409 [here](https://github.com/getsentry/sentry-dart/pull/2409#discussion_r1842042755) I noticed that you removed the line `options.screenshotQuality = SentryScreenshotQuality.low;` in the `main.dart`. Was this intentional or by accident?
> 
> intentional, I was thinking we should be testing the default value for quality when testing manually, so that it's the same as what users see by default.

But we want to keep it in the example app right? You removed it only for testing purposes right?

> > For the the Quality `low`, `medium`, and `high` everything works with your suggested changes in #2409, but for `full` the output image is about the size as the `low` setting, as it uses the `logical resolution` and not the `device resolution` therefore it would break the previous behaviour.
> 
> Interesting, it worked for me as expected on Windows. Which platforms have you tried this on so far?

My last test was with the `Pixel 7 Pro API 35` emulator on my mac.

> > To capture screenshots with quality `full` and keep the previous functionality, we need to get either the `device resolution` or the `devicePixelRatio`.
> 
> That is a very good point. We should grab devicePixelRatio and take it into account when calculating the pixelRatio based on constraints: https://github.com/getsentry/sentry-dart/pull/2409/files#diff-b603ed629ae7bb928af4f892324700593bfeb22ecff70a571e58345034b881a4R61 That should be a small change, I can have a look if you want.

But then we don't need the `getPixelRatio` anymore, because we will get the `devicePixelRatio` from [`options.bindingUtils.instance.window.devicePixelRatio`](https://api.flutter.dev/flutter/dart-ui/FlutterView/devicePixelRatio.html) and this is fixed for every device and not related to the orientation and the size of the `window/widget`. So we can set this, on app start.
In addition the whole `ScreenshotRecorderConfig` will not be needed anymore, because the `width` and `height` properties were only used, for calculating the `ratio` but this approach is not needed anymore.

> > As both values can change, for `Web` and `Desktop` applications, due to resizing of the window, setting the `ScreenshotRecorderConfig` to a fixed value at app start does not work. As it will stay the same even after resizing the window. Therefore the calculation needs to be done, before every screenshot capture. But this approach is only working if the `widget` is covering the complete device display, which is no longer a must.
> 
> For screenshots, we don't read or set window size at the start, we read it when capturing the screnshot: https://github.com/getsentry/sentry-dart/pull/2409/files#diff-b603ed629ae7bb928af4f892324700593bfeb22ecff70a571e58345034b881a4R59-R60. For replay, this shall change too when we support orientation change.
> 
> The configuration for `full` has both `width` and `height` as null, therefore, the calculated pixelRatio (as in the original code or in the revert PR, not in the current version here) would be 1.0 so that's not a problem.

Depends on the definition of a `problem`.
I just tested it with the current `main` branch, and choose the `full` setting and it resulted in an image with the size of `1440 × 3036 pixels` which matches the screen resolution of the `pixel 7 pro` emulator(`1440x3120 pixels`).

As already written, with the `ratio`of 1.0 the resulting output image has the size of the `logical resolution`, which is about the size as the `low` setting. Therefore it will break the current behaviour, which the user gets in the current release. 
> 
> > For me it looks like, the best/only/easiest solution would be to deprecate the `full` setting in the `SentryScreenshotQuality` due to the change of the approach of generating the screenshot. This would also make sense, regarding the possibility, of only capturing a screenshot of a specific widget, which is neither related to the `device resolution` nor to the `device orientation`. By ignoring the `full` setting, the user will still get screenshots, but with a lower resolution, as with the previous setting, and therefore the change is not that big.
> > @vaind @buenaflor WDYT? Should we deprecate the `full` or do you have another suggestion?
> > Should we exchange the `full`, with `extraHigh`? But this would be a breaking change, because it will no longer return `null` and the `enum name` changes and the resolution will also change.
> 
> I don't think of the screenshot quality as the function of window size, but of the size of the captured screenshot, which is always coming from the ScreenshotWidget size. I'd keep the setting, with my above suggestion to change the pixelRatio calculation, it should work the same as before.

At least the behaviour of the current stable, results in a screenshot with the device resolution. But I agree, that this behaviour is not 100% compatible with our new approach.

vaind: > But then we don't need the `getPixelRatio` anymore, because we will get the `devicePixelRatio` from [`options.bindingUtils.instance.window.devicePixelRatio`](https://api.flutter.dev/flutter/dart-ui/FlutterView/devicePixelRatio.html) and this is fixed for every device and not related to the orientation and the size of the `window/widget`. So we can set this, on app start.
> In addition the whole `ScreenshotRecorderConfig` will not be needed anymore, because the `width` and `height` properties were only used, for calculating the `ratio` but this approach is not needed anymore.

We need to take the configuration into account alongside the devicePixelRatio, otherwise anything but full-resolution images would be captured incorrectly.

vaind: > But we want to keep it in the example app right? You removed it only for testing purposes right?

Not really, I was thinking of removing that setting in the sample app. Besides being an example, it's also an app used for manual testing. My thinking is that we should be testing the default quality.

vaind: > Depends on the definition of a `problem`. I just tested it with the current `main` branch, and choose the `full` setting and it resulted in an image with the size of `1440 × 3036 pixels` which matches the screen resolution of the `pixel 7 pro` emulator(`1440x3120 pixels`).
> 
> As already written, with the `ratio`of 1.0 the resulting output image has the size of the `logical resolution`, which is about the size as the `low` setting. Therefore it will break the current behaviour, which the user gets in the current release.

You're right. I've written this part of the response before going back to the devicePixelRatio - that needs to be considered.

flutter/lib/src/screenshot/recorder.dart
```diff
@@ -49,7 +58,10 @@ class ScreenshotRecorder {
       // On iOS, the screenshot resolution is not adjusted.
       final srcWidth = renderObject.size.width;
       final srcHeight = renderObject.size.height;
-      final pixelRatio = config.getPixelRatio(srcWidth, srcHeight);
+
+      final pixelRatio = config.getPixelRatio(srcWidth, srcHeight) ??
+          // ignore: deprecated_member_use
+          options.bindingUtils.instance!.window.devicePixelRatio;
```
vaind: Maybe with this we could avoid deprecated code & non-null assertion:

```suggestion
          MediaQuery.of(context).devicePixelRatio;
```

flutter/test/screenshot/recorder_config_test.dart
```diff
@@ -0,0 +1,22 @@
+import 'package:flutter_test/flutter_test.dart';
+import 'package:sentry_flutter/src/screenshot/recorder_config.dart';
+
+void main() async {
+  group('$ScreenshotRecorderConfig', () {
+    test('defaults', () {
+      var sut = ScreenshotRecorderConfig();
+      expect(sut.height, isNull);
+      expect(sut.width, isNull);
```
vaind, line 9:
```suggestion
      expect(sut.width, isNull);
      expect(sut.getPixelRatio(100, 100), isNull);
```

CHANGELOG.md
```diff
@@ -13,6 +13,34 @@
   
 ### Features
 
+- Screenshots are now supported for Flutter Web via HTML Renderer ([#2361](https://github.com/getsentry/sentry-dart/pull/2361))
```
vaind: I don't see screenshots when testing with the HTML renderer (`flutter run -d chrome --web-renderer html`):
https://sentry-sdks.sentry.io/issues/6051629474/events/b89bcf0150444b7586205ce8c0afdb42/

vaind (commented):
The html renderer screenshots don't seem to work for me + some minor suggestions. Once the code is finished (i.e. someone else has reviewed it and haven't requested any changes), I'll check all platforms manually, especially with the masking stuff in replays.


Also, the HTML renderer is likely going away in the future, replaced by canvaskit and skwasm so maybe not worth trying to get screenshots working in it now.

flutter/test/screenshot/recorder_config_test.dart
martinhaintz: I have the line `expect(ScreenshotRecorderConfig().getPixelRatio(100, 100), null);` already in line 14. I don't think having it twice would make sense.

CHANGELOG.md
martinhaintz: very strange. you are right. It is not working. Looks like I made a mistake during testing. I probably forgot to change to the `latest` issue in the sentry app. I will revert these changes, later or tomorrow.

flutter/test/screenshot/recorder_config_test.dart
vaind: missed that

martinhaintz:
@buenaflor @vaind In `screenshot_event_processor.dart` the `beforeScreenshot` callback is executed first and afterwards it will check if the screenshot functionality is available for this renderer. Is this a bug or is this the expected behaviour?

IMHO I would assume, that the `beforeScreenshot`is not called, if the screenshot feature itself is not available for the platform. Therefore I would move the platform check before the callback handling.

```dart
    final beforeScreenshot = _options.screenshot.beforeCapture;
    if (beforeScreenshot != null) {
      try {
        final result = beforeScreenshot(event, hint: hint);
        bool takeScreenshot;
        if (result is Future<bool>) {
          takeScreenshot = await result;
        } else {
          takeScreenshot = result;
        }
        if (!takeScreenshot) {
          return event;
        }
        _options.logger(
          SentryLevel.error,
          'The beforeScreenshot callback threw an exception',
          exception: exception,
          stackTrace: stackTrace,
        );
        if (_options.automatedTestMode) {
          rethrow;
        }
      }
    }

    final renderer = _options.rendererWrapper.getRenderer();

    if (_options.platformChecker.isWeb &&
        renderer != FlutterRenderer.canvasKit) {
      _options.logger(
        SentryLevel.debug,
        'Cannot take screenshot with ${renderer?.name} renderer.',
      );
      return event;
    }
```

martinhaintz:
@vaind I reverted the changes regarding the `html` renderer, went through all files changed in this PR and tested it on the android emulator on `Pixel 7 Pro API 35` for quality `low`, `high` and `full`, with and without `redaction` and the screenshot size and the redaction looks good.

Let me know, if I can support you in testing.

vaind (approved):

vaind (approved):
LGTM, thanks for all the changes.

flutter/lib/src/sentry_flutter_options.dart
```diff
@@ -385,11 +396,33 @@ class SentryFlutterOptions extends SentryOptions {
...
-}
 
-/// Callback being executed in [ScreenshotEventProcessor], deciding if a
-/// screenshot should be recorded and attached.
-typedef BeforeScreenshotCallback = FutureOr<bool> Function(
-  SentryEvent event, {
-  Hint? hint,
-});
+  /// Privacy configuration for masking sensitive data in screenshots and Session Replay.
+  /// Screen content masking redaction is:
```
buenaflor: nit: shouldn't we use masking exclusively? seems like we're inconsistent with terminology across the codebase

buenaflor (approved):
just one last comment

flutter/lib/src/sentry_flutter_options.dart
martinhaintz: I think `masking` is the preferred terminology or what do you mean @vaind?

buenaflor: but this is not really blocking this pr, just wanna make sure we use the right term in the future

vaind: yeah, I believe "masking" is more descriptive of what's happening (we render a mask over the content). Redacting is a more generic term and includes masking as one way to "redact" so I wouldn't be too concerned with existing uses.

buenaflor: @martinhaintz feel free to merge when you are ready

flutter/lib/src/sentry_flutter_options.dart
```diff
@@ -183,22 +181,35 @@ class SentryFlutterOptions extends SentryOptions {
...
+  bool get attachScreenshotOnlyWhenResumed => screenshot.attachOnlyWhenResumed;
+  set attachScreenshotOnlyWhenResumed(bool value) =>
+      screenshot.attachOnlyWhenResumed = value;
 
   /// Sets a callback which is executed before capturing screenshots. Only
   /// relevant if `attachScreenshot` is set to true. When false is returned
   /// from the function, no screenshot will be attached.
-  BeforeScreenshotCallback? beforeScreenshot;
+  @Deprecated('Use `screenshot.beforeCapture` instead')
+  BeforeScreenshotCallback? get beforeScreenshot => screenshot.beforeCapture;
```
buenaflor: I'm not sure whether we should keep this consistent with other sdks by keeping`before` callbacks in the root options.

We're also planning on adding a `beforeCaptureViewHierarchy` so having the before screenshot nested would be inconsistent then, our other sdks also don't nest this

martinhaintz: I understand. It makes sense to keep it consistent with the other `SDKs`. I will move the `SentryScreenshotOptions` into the `SentryOptions`.

flutter/lib/src/sentry_flutter_options.dart
martinhaintz: I will review the changes, and update the wording from `redacting` to `masking` or similar.

martinhaintz:
I will now start to update the docs, based on the changes of this PR.

vaind:
> I will now start to update the docs, based on the changes of this PR.

We should merge prior to docs changes as these should normally come live only when a release is being made

PR #2362, opened 2024-10-15
test(flutter): UI e2e with Maestro (experimental)

buenaflor:
some experiments with Maestro in CI

#skip-changelog


vaind:
Are there benefits in using Maestro instead of flutter built-in integration testing? We already have these tests in https://github.com/getsentry/sentry-dart/blob/main/flutter/example/integration_test

From my limited experience with Maestro, it's pretty limited in what it can do and cumbersome if you want to get around its limitations. Also, has basically no assertions and very limited scripting support.

buenaflor:
I've tried adding some tests for app start measurements in the current integration test which ended up being extremely flaky, also I think the e2e are disabled because of flakiness? (might be wrong)

and @krystofwoldrich had a good experience with maestro so I wanted to try it out

PR #2365, opened 2024-10-17
enh: cache parsed dsn

### LINKED ISSUE #2364: Cache parsed DSN
### Description

We parse the dsn multiple times when we need it. We can parse it once and cache its result

buenaflor:
## :scroll: Description

Parses and caches the dsn when accessing it.

Previously we parse it every time when we access it instead of caching it

Closes #2364 

Unit test

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/dart/lib/src/sentry_options.dart
+++ b/dart/lib/src/sentry_options.dart
@@ -25,19 +25,24 @@ class SentryOptions {
 
   /// The DSN tells the SDK where to send the events to. If an empty string is
   /// used, the SDK will not send any events.
-  String? get dsn => _dsn;
+  String? dsn;
 
-  set dsn(String? value) {
-    _dsn = value;
-
-    assert(_dsn != null);
-    assert(_dsn!.isNotEmpty);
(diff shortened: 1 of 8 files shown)


dart/lib/src/sentry_options.dart
```diff
@@ -27,6 +27,17 @@ class SentryOptions {
   /// used, the SDK will not send any events.
   String? dsn;
 
+  /// Evaluates and parses the DSN. May throw an exception if the DSN is invalid.
+  @internal
+  late final Dsn parsedDsn = _parseDsn();
```
stefanosiano: the dsn could change in theory, even if not supported.
I'm thinking about the case when a user reinits the SDK with the same options changing the dsn. We would still have the old parsedDsn.

Let's reevaluate it when the dsn is set?
wdyt, since it's an edge case?

buenaflor: 👍 updated it

stefanosiano (approved):

PR #2375, opened 2024-10-24
feat: warning in screenshot/replay creation when a potentially sensitive widget is not masked

vaind:
## :scroll: Description

When a potentionally sensitive-content-containing widget is encountered and it is not masked (or unmasked explicitly), print a warning (in debug mode). 

While testing this I noticed we don't have a TextField in tests which doesn't inherit from Text or FormField so added it although it's unrelated to this PR

This should help users realize that they need to make adjustments to the privacy configuration. The code is only included in debug builds so it won't affect release builds in any way.

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes


flutter/lib/src/sentry_privacy_options.dart
```diff
@@ -50,12 +51,39 @@ class SentryPrivacyOptions {
...
     if (maskAllText) {
       rules.add(
           const SentryMaskingConstantRule<Text>(SentryMaskingDecision.mask));
       rules.add(const SentryMaskingConstantRule<EditableText>(
           SentryMaskingDecision.mask));
     }
+
+    // In Debug mode, check if users explicitly mask (or unmask) widgets that
+    // look like they should be masked, e.g. Videos, WebViews, etc.
+    if (kDebugMode) {
```
buenaflor: wdyt of using `options.platformChecker.isDebugMode()`? 

but doesn't really make a difference I think

only matters if we add an additional test case where we assert that the warning isn't added for release builds, then we can mock it

buenaflor (approved):
looks good, just one question but not really blockign

flutter/lib/src/sentry_privacy_options.dart
vaind: Thanks, I've made the changes and added test cases for non-debug builds

PR #2384, opened 2024-11-01
Added frame rate and relative frame delay to span metrics

YamenSarah:
## :scroll: Description

- Added new measurement `frames.relative_delay` to spans. This metric is the frame delay divided by the span duration. It provides a relative measure of frame delay to better assess the impact of frame delay.
- Added new measurement `frames.rate` to spans. This metric estimates the frame rate based on the span duration and the number of frames.

The goal of this MR is to add some new metrics to improve the user's ability to prioritize poor performing spans.

In reviewing the documentation of [frames delay](https://develop.sentry.dev/sdk/telemetry/traces/frames-delay/#:~:text=Now%2C%20when%20prioritizing%20the%20user%2Dperceived%20degraded%20responsiveness%20by%20frames%20delay%2C%20we%20get%201%2C%203%2C%202%20instead%20of%203%2C%202%2C%201%20when%20looking%20at%20slow%20and%20frozen%20frames.), I noticed the example prioritization scenarios don't account for how long the span was.

For example if Scenario 1 had a frame delay of 2400ms but hypothetically the span duration was 5x the span duration of Scenario 3, wouldn't that mean that over the entire span, Scenario 3 actually had less frame delay?

Hence the metric of relative_delay.

I added frames.rate as the variables needed to calculate it were present.

Wrote one test for relative frame delay
Otherwise, tested locally to evaluate the reported estimated frame rate.

## :pencil: Checklist

- [x] I added tests to verify changes*
- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,8 @@
     ),
   );
   ```
+- Added new measurement `frames.relative_delay` to spans. This metric is the frame delay divided by the span duration. It provides a relative measure of frame delay to better assess the impact of frame delay.
+- Added new measurement `frames.rate` to spans. This metric estimates the frame rate based on the span duration and the number of frames.
 
 ### Enhancements
 
--- a/flutter/lib/src/span_frame_metrics_collector.dart
+++ b/flutter/lib/src/span_frame_metrics_collector.dart
@@ -16,6 +16,8 @@ class SpanFrameMetricsCollector implements PerformanceContinuousCollector {
   static const framesDelayKey = 'frames.delay';
   static const slowFramesKey = 'frames.slow';
   static const frozenFramesKey = 'frames.frozen';
+  static const estimatedFrameRateKey = 'frames.rate';
+  static const relativeFrameDelayKey = 'frames.relative_delay';
 
   final SentryFlutterOptions options;
   final FrameCallbackHandler _frameCallbackHandler;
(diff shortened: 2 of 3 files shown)


buenaflor:
we don't need to estimate the frame rate, we already know the frame rate since we fetch it from the native layer

maybe you have some thoughts on the relative delay proposal @philipphofmann

philipphofmann:
I'm happy that you're looking into frame delay, @YamenSarah. We don't calculate such metrics on the client side; instead, we plan to calculate them on the backend because we have all we need on the backend to do so, and adding it on the client increases the payload size. Our new mobile module already shows that data. We still need to display it properly in all performance-relevant places. This will happen in the upcoming months.
CleanShot 2024-11-07 at 19 43 41@2x

YamenSarah:
@buenaflor Where is the frame rate reported currently? 
@philipphofmann That sounds reasonable. Just to confirm, I'm not talking about average frame delay. I'm talking about frame delay relative to the span duration.
I've been unable to do this in the discover page myself as your Equation feature does not let you use aggregate fields in an Equation, and I want to see (AVG(frame_delay) / AVG(span_duration)) etc..

buenaflor:
> Where is the frame rate reported currently?

We do not report the frame rate right now, it's only known internally. 

If you need the frame rate we can expose it or you can do the calculation as you've done before

> I've been unable to do this in the discover page myself as your Equation feature does not let you use aggregate fields in an Equation, and I want to see (AVG(frame_delay) / AVG(span_duration)) etc..

If you want to send custom measurements you can also do so in the beforeSend hook:

```dart
options.beforeSendTransaction = (transaction) {
  // calculate your relative delay
  transaction.measurements['frames_relative_delay'] =
      SentryMeasurement('frames_relative_delay', relativeDelay);
  return transaction;
};
```

You can then query it on the discover page

philipphofmann:
We don't have frame delay yet available in Discover. We still need to add it, but I can't give you an ETA at the moment.

PR #2405, opened 2024-11-12
feat: use `addTimingsCallback` instead of `addPostFrameCallback` to determine app start end

buenaflor:
## :scroll: Description

Improve app start measurements

The `addTimingsCallback` contains the accurate data for the first frame unlike `addPostFramecallback` which is not correct as it depends on when `SentryFlutter.init` is called

`addTimingsCallback` is always reported immediately for the first frame so it's very suitable to use here, see https://api.flutter.dev/flutter/scheduler/SchedulerBinding/addTimingsCallback.html

Unit tests, manual tests

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes


flutter/lib/src/integrations/native_app_start_integration.dart
```diff
@@ -62,8 +72,12 @@ class NativeAppStartIntegration extends Integration<SentryFlutterOptions> {
         if (options.automatedTestMode) {
           rethrow;
         }
+      } finally {
+        _frameCallbackHandler.removeTimingsCallback(timingsCallback);
```
buenaflor, line 76: we only care about the first frame timing so after we processed it we can just remove the callback

flutter/test/fake_frame_callback_handler.dart
```diff
@@ -1,33 +1,33 @@
...
   }
 
   @override
-  bool hasScheduledFrame = true;
+  void addTimingsCallback(TimingsCallback callback) {
+    timingsCallback = callback;
+  }
 
   @override
-  Future<void> get endOfFrame => Future<void>.value();
```
buenaflor, line 32: removed dead code here

buenaflor:
cc @stefanosiano @denrase please take a look

flutter/lib/src/integrations/native_app_start_integration.dart
```diff
@@ -28,15 +29,24 @@ class NativeAppStartIntegration extends Integration<SentryFlutterOptions> {
...
+      // Set immediately to prevent multiple executions
+      // we only care about the first frame
+      _allowProcessing = false;
+
       try {
         DateTime? appStartEnd;
         if (options.autoAppStart) {
           // ignore: invalid_use_of_internal_member
-          appStartEnd = options.clock();
+          appStartEnd = DateTime.fromMicrosecondsSinceEpoch(timings.first
```
denrase, line 48: Can we always assume `timings` is not empty? If not, we'd crash here accessing `timings.first`. Maybe we should check if it's not empty, just to be safe.

flutter/lib/src/integrations/native_app_start_integration.dart
denrase: Can we be sure that `timingsCallback` here is the same reference as the outside `timingsCallback`? I did try and it does work as expected.

Bildschirmfoto 2024-11-18 um 11 10 30

flutter/lib/src/integrations/native_app_start_integration.dart
denrase: Ok, saw that the tests handle this case. I assume we catch the exception and this is the intended way. 👍

denrase (approved):
Looking good. Only had 2 comment, but those cases work fine when i tested in the example app. 👍

flutter/lib/src/integrations/native_app_start_integration.dart
buenaflor: yes but let me add an additional widget test for this

flutter/lib/src/integrations/native_app_start_integration.dart
buenaflor: yup, we catch it and the app start won't trigger

PR #2406, opened 2024-11-12
feat(js-sdk): add script loader to set up Sentry Javascript SDK

buenaflor:
## :scroll: Description

This will be merged into `feat/js-sdk-integration` branch, not main so it's easier to split up PRs and at the end just merge `feat/js-sdk-integration` into main

PR Nr 1 of JS SDK integration

This adds the JS SDK CDN to the document's header so we can access the JS SDK through Dart

Unit test, integration test, manual check of html document

Screenshot 2024-11-12 at 17 02 12

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps

#skip-changelog


.github/workflows/flutter.yml
```diff
@@ -4,6 +4,7 @@ on:
     branches:
       - main
       - release/**
+      - feat/**
```
buenaflor: will remove before merging, this is so ci runs since we merge into another feat branch

flutter/lib/src/web/sentry_js_bundle.dart
```diff
@@ -0,0 +1,21 @@
...
+    'url': 'https://browser.sentry-cdn.com/$jsSdkVersion/bundle.tracing.min.js',
+    'integrity':
+        'sha384-GmCdCRO2s8VuYopAldiuHl/uns+EWDcLodj8AW810pK14r3vPQxoDNsMxnitCt18'
+  }
+];
+
+@internal
+const debugScripts = [
+  {
+    'url': 'https://browser.sentry-cdn.com/$jsSdkVersion/bundle.tracing.js',
```
philipphofmann, line 20: `m`: Why don't we add integrity for the debug script as well?

flutter/test/web/sentry_script_loader_test.dart
```diff
@@ -0,0 +1,89 @@
...
+
+    test('loads debug scripts when debug is enabled', () async {
+      final sut = fixture.getSut(debug: true);
+
+      await sut.load();
+
+      final scripts = querySelectorAll('script[src*="sentry-cdn"]');
+      for (final script in scripts) {
+        final element = script;
+        expect(element.src, isNot(contains('.min.js')));
```
philipphofmann: `l`: We could check if it endsWith `/bundle.tracing.js`. The URL could be garbage, such as `adsfasdf`, and the test would be green if we only validate for `isNot(contains('.min.js'))`.

flutter/test/web/sentry_script_loader_test.dart
```diff
@@ -0,0 +1,89 @@
...
+
+    test('loads production scripts by default', () async {
+      final sut = fixture.getSut();
+
+      await sut.load();
+
+      final scripts = querySelectorAll('script[src*="sentry-cdn"]');
+      for (final script in scripts) {
+        final element = script;
+        expect(element.src, contains('.min.js'));
```
philipphofmann: `l`: We could be a bit more specific. If the scr is `gino.is.great.min.js` the test would be green.

philipphofmann (commented):
Quick feedback before our pair programming session.

flutter/lib/src/web/script_loader/html_script_dom_api.dart
```diff
@@ -0,0 +1,18 @@
...
+    ..src = src
+    ..crossOrigin = 'anonymous'
+    ..onLoad.listen((_) => completer.complete())
+    ..onError.listen((event) => completer.completeError('Failed to load $src'));
+
+  if (integrity != null) {
+    script.integrity = integrity;
+  }
+
+  document.head?.append(script);
```
philipphofmann: `h`: We should ensure that the Sentry JS SDK gets added before all other scripts because otherwise, our JS SDK can't report problems with the scripts loaded before it; see https://docs.sentry.io/platforms/javascript/#install.

flutter/lib/src/web/script_loader/web_script_dom_api.dart
```diff
@@ -0,0 +1,27 @@
...
+  }
+  // JS SDK needs to be loaded before everything else
+  final head = document.head;
+  if (head != null) {
+    if (head.hasChildNodes()) {
+      head.insertBefore(script, head.firstChild);
+    } else {
+      head.append(script);
+    }
+  }
```
buenaflor, line 53: @lforst do you know if there is a risk in 'injecting' the script into the head like this, e.g being flagged by websites or anything like that? we do that that so we can use the JS SDK in Flutter Web

lforst: I am actually not sure. The only thing I can think of (except for maybe lighthouse punishing you or whatever) is CSP potentially being a pain.

buenaflor: ah got it, so ig it's preferrable that user's just manually add the script in their html file

buenaflor: @lforst 

https://pub.dev/packages/google_identity_services_web

okay so I found one package from google with similar problems where they use one of their javascript sdks and they offer two options for installation:
- option 1: add it yourself in html file
- option 2: they inject the script into the document's head like we do here (we do it on sdk init) and they don't mention any potential issues

so based on that I assume it's fine?

lforst: I guess so!

flutter/test/web/sentry_script_loader_test.dart
buenaflor: changed it to be more specific `$jsSdkVersion/bundle.tracing.js`

flutter/lib/src/web/sentry_js_bundle.dart
buenaflor: added it

flutter/example/integration_test/web_sdk_test.dart
```diff
@@ -0,0 +1,54 @@
...
+      final Map<String, dynamic> options = {
+        'dsn': app.exampleDsn,
+        'beforeSend': beforeSendFn,
+        'defaultIntegrations': [],
+      };
+
+      final sentry = context['Sentry'] as JsObject;
+      sentry.callMethod('init', [JsObject.jsify(options)]);
+
+      sentry.callMethod('captureMessage', [expectedMessage.toJS]);
```
philipphofmann: 💯. @buenaflor did you try removing the script and check if the test fails?

buenaflor: actually no, only the expect, I'll adjust it

buenaflor: test is now more robust

flutter/example/integration_test/utils.dart
```diff
@@ -0,0 +1,21 @@
...
+/// testWidgets override Flutter.onError by default
+/// If a fail happens during integration tests this would complain that
+/// the FlutterError.onError was overwritten and wasn't reset to its
+/// state before asserting.
+///
+/// This function needs to be executed before assertions.
+Future<void> restoreFlutterOnErrorAfter(Future<void> Function() fn) async {
+  final originalOnError = FlutterError.onError!;
+  await fn();
+  final overriddenOnError = FlutterError.onError!;
```
denrase: Would the test crash/abort here if `FlutterError.onError` is null, since we force unwrap with `!`?

flutter/lib/src/web/script_loader/web_script_dom_api.dart
```diff
@@ -0,0 +1,64 @@
...
+  return completer.future;
+}
+
+void injectMetaTag(Map<String, String> attributes) {
+  final HTMLMetaElement meta =
+      document.createElement('meta') as HTMLMetaElement;
+  for (final MapEntry<String, String> attribute in attributes.entries) {
+    meta.setAttribute(attribute.key, attribute.value);
+  }
+  document.head!.appendChild(meta);
```
denrase: Does it make sense to check if head is not null?

flutter/test/integrations/web_sdk_integration_test.dart
```diff
@@ -0,0 +1,62 @@
...
+  FakeSentryScriptLoader(super.options);
+
+  int loadScriptsCalls = 0;
+
+  @override
+  Future<void> loadWebSdk(List<Map<String, String>> scripts,
+      {String trustedTypePolicyName = defaultTrustedPolicyName}) {
+    loadScriptsCalls += 1;
+
+    return super
```
denrase, line 59: Do we need this to be a subclass instead of interface implementation for this test? If all we check is if the method was called, so we don't need potential side-effects by calling the super implementation.

flutter/test/web/sentry_script_loader_test.dart
```diff
@@ -0,0 +1,120 @@
...
+      expect(scriptElements.first.src,
+          endsWith('$jsSdkVersion/bundle.tracing.min.js'));
+    });
+  });
+}
+
+class Fixture {
+  final options = defaultTestOptions();
+
+  SentryScriptLoader getSut({bool debug = false}) {
```
denrase: `getSut()` was always called without the `debug`. We can remove the param if we don't need it in a test case.

denrase (commented):
Looking good, just some minor comments/questions from my side.

flutter/lib/src/web/script_loader/web_script_dom_api.dart
buenaflor: good catch, this should actually not be here, only in tests

buenaflor: I'll remove the fn

flutter/test/integrations/web_sdk_integration_test.dart
buenaflor: we don't necessarily need it, just thought it's easier and didn't want to create an interface just to test this

flutter/example/integration_test/utils.dart
buenaflor: hm I'm not sure, but I guess it's safer to make it work with optionals without force unwrap

flutter/test/integrations/web_sdk_integration_test.dart
buenaflor: wdyt? you prefer creating an interface here?

denrase (approved):

PR #2414, opened 2024-11-15
ci: comment high risk only once

### LINKED ISSUE #2413: ci: fix duplicate commenting of high risk change job
### Description

The high-risk ci job notifies the PR with a comment that the PR contains changes in file that we deem as 'risky'.

However this is triggered for every pushed commit so let's fix that it only does it once

buenaflor:
closes #2413 all the duplicate comments are only noise 

Tested this by checking if only one comment exists on multiple commits

#skip-changelog


brustolin (approved):
LGTM

PR #2415, opened 2024-11-16
fix: add style to feedback AppBar title Text

ewann:
## :scroll: Description
Add `Theme.of` to `AppBar` in the same way it is already applied to other text elements in this widget

In the absence of this change, the title text will not adopt theming configured in the app. 

As a consequence, it can be impossible for end users to read this text.

This is not a change per-se, it seems more like an omission in the existing implementation

Change made locally and then added as a dependency to a flutter project exhibiting the symptom mentioned above

## :pencil: Checklist

- [ ] All tests passing
- [ ] No breaking changes

## :crystal_ball: Next steps

Based on https://github.com/getsentry/sentry-dart/blob/18e6fd72f17e921a090358b11e34f88663a759db/CONTRIBUTING.md I'm unclear if there is more expected from my side.

### DIFF
--- a/flutter/lib/src/feedback/sentry_feedback_widget.dart
+++ b/flutter/lib/src/feedback/sentry_feedback_widget.dart
@@ -63,7 +63,9 @@ class _SentryFeedbackWidgetState extends State<SentryFeedbackWidget> {
   Widget build(BuildContext context) {
     return Scaffold(
       appBar: AppBar(
-        title: Text(widget.title),
+        title: Text(widget.title,
+          style: Theme.of(context).textTheme.titleLarge,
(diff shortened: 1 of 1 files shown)


ueman:
The current code is correct.
In order to change the text style in the AppBar, you have to adapt [`AppBarTheme.titleTextStyle`](https://api.flutter.dev/flutter/material/AppBarTheme/titleTextStyle.html).

See also the source code of the AppBar here: https://github.com/flutter/flutter/blob/41c273f8c66c4a5a373cea8b13605c6d7c813829/packages/flutter/lib/src/material/app_bar.dart#L925

PR #2421, opened 2024-11-19
chore: Update PR template

philipphofmann:
Remove I reviewed the code and add updating the wizard and the docs because we often forget to update these.

#skip-changelog


buenaflor:
we don't support sentry-wizard on flutter (yet), [that's tbd](https://github.com/getsentry/sentry-dart/issues/2424)

PR #2429, opened 2024-11-21
test: skip flaky sentry_navigator_observer test case

vaind:
Seen this test fail multiple times recently so created an issue to fix this.


buenaflor (approved):
thx, I also wanted to do this

PR #2431, opened 2024-11-21
feat: support obfuscation for Linux apps

vaind:
## :scroll: Description

It was a small step to add Linux support after #2286, so here it is.

part of #433

## :pencil: Checklist

- [x] All tests passing
- [ ] No breaking changes


buenaflor (approved):
looks good, I'm not really experienced with it but interested in how compatibility looks like across different linux distros with this impl? 

will prolly also be important for docs to clarify

vaind:
> looks good, I'm not really experienced with it but interested in how compatibility looks like across different linux distros with this impl?
> 
> will prolly also be important for docs to clarify

Because we don't ship binary distributions but instead let the native part be built together with the final Flutter app, any issues should be limited to missing build dependencies (most likely just libcurl4-openssl-dev which is already a very common dependency anyway)

PR #2439, opened 2024-11-21
fix: context defualts on native platforms other than iOS/Android

vaind:
## :scroll: Description

I've noticed windows doesn't have device context filled in. It's because originally we expected native integrations to sync from native iOS/Android to dart via `LoadContexts()`. This is not available with sentry-native which doesn't have such APIs.

To fix this, I've changed the logic to always set contexts in `IoEnricherEventProcessor` and made it so that it runs after the native `_LoadContextsIntegrationEventProcessor`. This way, the former will fill in the blanks all the time.

## :pencil: Checklist

- [ ] All tests passing
- [ ] No breaking changes


buenaflor (approved):
👍

PR #2444, opened 2024-11-25
Change EnvelopeItem dataFactory to FutureOr

vaind:
Review suggestion for #2417 change envelope-item data factory to FutureOr

Looking at the uses of the data factory, it's almost exclusively synchronous so I think we can change it to `FutureOr`. Even though this is changing a public API, it shouldn't require any changes by the callers because the type is compatible with Future.

#skip-changelog

### DIFF
--- a/dart/lib/src/sentry_envelope.dart
+++ b/dart/lib/src/sentry_envelope.dart
@@ -123,7 +123,8 @@ class SentryEnvelope {
     final newLineData = utf8.encode('\n');
     for (final item in items) {
       try {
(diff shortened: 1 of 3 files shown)


denrase (approved):
Looking good. Honestly, I'm not sure if this API is even used, maybe @buenaflor Has some insights?

buenaflor:
> Looking good. Honestly, I'm not sure if this API is even used, maybe @buenaflor Has some insights?

no idea, I doubt that someone uses this api or the number is negligible slim

the change is fine for me

PR #2476, opened 2024-12-10
test(integration): enable e2e

buenaflor:
#skip-changelog

### DIFF
--- a/.github/workflows/flutter_test.yml
+++ b/.github/workflows/flutter_test.yml
@@ -82,7 +82,7 @@ jobs:
           force-avd-creation: false
           avd-name: avd-x86_64-31
           emulator-options: -no-window -gpu swiftshader_indirect -noaudio -no-boot-anim -camera-back none
-          disable-animations: true
+          disable-animations: false
           script: echo 'Generated AVD snapshot for caching.'
 
       - name: build apk
@@ -99,7 +99,7 @@ jobs:
           force-avd-creation: false
           avd-name: avd-x86_64-31
           emulator-options: -no-snapshot-save -no-window -accel on -gpu swiftshader_indirect -noaudio -no-boot-anim -camera-back none
-          disable-animations: false
+          disable-animations: true
           script: ./gradlew testDebugUnitTest
 
       - name: launch android emulator & run android integration test
@@ -112,7 +112,7 @@ jobs:
           force-avd-creation: false
           avd-name: avd-x86_64-31
           emulator-options: -no-snapshot-save -no-window -accel on -gpu swiftshader_indirect -noaudio -no-boot-anim -camera-back none
-          disable-animations: false
+          disable-animations: true
           script: flutter test integration_test/all.dart --dart-define SENTRY_AUTH_TOKEN_E2E=$SENTRY_AUTH_TOKEN_E2E --verbose
 
   cocoa:
--- a/.github/workflows/flutter_test.yml
+++ b/.github/workflows/flutter_test.yml
@@ -81,7 +81,7 @@ jobs:
           arch: x86_64
           force-avd-creation: false
           avd-name: avd-x86_64-31
-          emulator-options: -no-snapshot-save -no-window -gpu swiftshader_indirect -noaudio -no-boot-anim -camera-back none
+          emulator-options: -no-window -gpu swiftshader_indirect -noaudio -no-boot-anim -camera-back none
           disable-animations: true
           script: echo 'Generated AVD snapshot for caching.'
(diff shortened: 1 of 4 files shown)


flutter/example/integration_test/integration_test.dart
```diff
@@ -162,66 +163,75 @@ void main() {
...
+    var output = find.byKey(const Key('output'));
+    late Fixture fixture;
+
+    setUp(() {
+      fixture = Fixture();
+    });
+
+    testWidgets('captureException', (tester) async {
+      late Uri uri;
+      await restoreFlutterOnErrorAfter(() async {
```
vaind, line 179: Looks like something for setup/teardown functions?

buenaflor: I think it needs to be restored before expect calls are made, wouldn't work in setup or teardown

I'll double check

vaind: ah, if that's the case, could you please add a comment to the function so anyone coming back to this won't have the same idea as I've just had?

.github/workflows/flutter_test.yml
```diff
@@ -13,7 +13,7 @@ on:
       - "flutter/**"
 
 env:
-  SENTRY_AUTH_TOKEN: ${{ secrets.SENTRY_AUTH_TOKEN }}
+  SENTRY_AUTH_TOKEN_E2E: ${{ secrets.SENTRY_AUTH_TOKEN_E2E }}
 
 jobs:
```
vaind, line 18: FYI it's now preferrable (faster) to run android emulator on linux: https://github.com/ReactiveCircus/android-emulator-runner#running-hardware-accelerated-emulators-on-linux-runners

buenaflor: oh yeah, will change it

buenaflor: oh yeah, will change it

.github/workflows/flutter_test.yml
```diff
@@ -35,6 +35,12 @@ jobs:
...
         sdk: ["stable", "beta"]
     steps:
+      - name: Enable KVM group perms
+        run: |
+          echo 'KERNEL=="kvm", GROUP="kvm", MODE="0666", OPTIONS+="static_node=kvm"' | sudo tee /etc/udev/rules.d/99-kvm4all.rules
+          sudo udevadm control --reload-rules
+          sudo udevadm trigger --name-match=kvm
+          
       - name: checkout
         uses: actions/checkout@v4
```
vaind: to fix CI

```suggestion
      - name: checkout
        uses: actions/checkout@v4

      - name: Enable KVM group perms
        run: |
          echo 'KERNEL=="kvm", GROUP="kvm", MODE="0666", OPTIONS+="static_node=kvm"' | sudo tee /etc/udev/rules.d/99-kvm4all.rules
          sudo udevadm control --reload-rules
          sudo udevadm trigger --name-match=kvm
          
```

vaind (approved):
LGTM, hopefully it's going to be stable 🤞

vaind:
CI doesn't work so I've changed this back to draft so it's not tempting anyone to merge (because of the approval)

buenaflor:
yeah not sure why, emulator won't start for some reason after changing the runner to linux

vaind:
> yeah not sure why, emulator won't start for some reason after changing the runner to linux

Works here, maybe you can find a difference
- https://github.com/getsentry/sentry-java/blob/main/.github/workflows/integration-tests-ui-critical.yml
- https://github.com/getsentry/sentry-dotnet/blob/main/.github/workflows/device-tests-android.yml

flutter/lib/src/screenshot/widget_filter.dart
```diff
@@ -233,9 +233,11 @@ extension on Element {
 @internal
 extension Opaqueness on Color {
   @pragma('vm:prefer-inline')
+  // ignore: deprecated_member_use
   bool get isOpaque => alpha == 0xff;
 
   @pragma('vm:prefer-inline')
+  // ignore: deprecated_member_use
   Color asOpaque() => isOpaque ? this : Color.fromARGB(0xff, red, green, blue);
```
buenaflor, line 255: @vaind analyzer complains, I'm ignoring this if you're fine with it

I guess it was deprecated in the new version

vaind: It's ok, I'll have a look in a separate PR

vaind (approved):

PR #2477, opened 2024-12-10
deprecate: manual ttid widget

### LINKED ISSUE #2449: Deprecate manual TTID / `SentryDisplayWidget`
## Current state

We currently allow two options:
- automatic ttid which calls `addPostFrameCallback` within `didPush` of the navigator observer
- manual ttid which calls `addPostFrameCallback` within a widget's `initState` that you need to wrap around the widget you want to measure from

## Why deprecate the manual version?
After some tests I've come to the conclusion that the automatic version is more accurate than we think (see timeline), there is usually a difference of around 2-5ms (give or take) to the actual frame duration which is negligible. 

By deprecating it / removing it in v9 we reduce complexity of the implementation and usage but still report accurate data.
Image

buenaflor:
Closes https://github.com/getsentry/sentry-dart/issues/2449

### DIFF
--- a/flutter/test/navigation/time_to_display_tracker_test.dart
+++ b/flutter/test/navigation/time_to_display_tracker_test.dart
@@ -101,7 +101,6 @@ void main() {
         expect(ttidSpan?.origin, SentryTraceOrigins.manualUiTimeToDisplay);
       });
 
-      // skipping test, flaky
       test('completes with timeout when not completing the tracking', () async {
         final sut = fixture.getSut();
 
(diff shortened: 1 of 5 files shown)


flutter/test/navigation/time_to_display_tracker_test.dart
```diff
@@ -109,7 +110,7 @@ void main() {
 
         final transaction = fixture.getTransaction() as SentryTracer;
         await sut.track(transaction, startTimestamp: fixture.startTimestamp);
-      });
+      }, skip: true);
```
krystofwoldrich: I think this should be a separate PR.

krystofwoldrich (approved):
One nit, otherwise looks good.

flutter/test/navigation/time_to_display_tracker_test.dart
buenaflor: the test is a manual TTID test, related but let's do it consistently, I'll open another PR

PR #2481, opened 2024-12-11
refactor: use breadth-first search in widget-filter

vaind:
## :scroll: Description

Stack traces in the widget filters can get pretty deep (I've seen 800+ frames on an app that wasn't crazy complicated) so while we don't seem to be hitting any stack-depth limits, better not tempt that and let's flatten the recursion instead.

#skip-changelog

## :pencil: Checklist

- [ ] All tests passing
- [ ] No breaking changes

### DIFF
--- a/flutter/lib/src/screenshot/widget_filter.dart
+++ b/flutter/lib/src/screenshot/widget_filter.dart
@@ -14,6 +14,7 @@ class WidgetFilter {
   late WidgetFilterColorScheme _scheme;
   late double _pixelRatio;
   late Rect _bounds;
+  late List<Element> _visitList;
   final _warnedWidgets = <int>{};
 
   /// Used to test _obscureElementOrParent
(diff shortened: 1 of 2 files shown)


buenaflor (approved):

buenaflor:
using a dequeue I guess it boils down to whether this makes a difference right?

```dart
final currentList = _visitList;
_visitList = [];

// vs 

final element = queue.removeFirst;
```

vaind:
> using a dequeue I guess it boils down to whether this makes a difference right?
> 
> ```dart
> final currentList = _visitList;
> _visitList = [];
> 
> // vs 
> 
> final element = queue.removeFirst;
> ```

Yep. This function is a hotspot so it would make sense to measure. We don't currently have any (micro)benchmarks in the repo but I was wanting to add some for a long time now

Issue #2309, opened 2024-09-26
mask overlay in flutter session replay


thisames:
### Platform

Flutter Mobile

### Obfuscation

Enabled

### Debug Info

Enabled

### Doctor

Dart version 3.4.3

### Version

8.9.0

### Steps to Reproduce

I'm testing the replay session in Flutter and noticed that widget masks overlap when navigating between screens.

It seems that the filter.items always contains the total number of widgets in the context, not just the widgets for each screen.

If I'm on screen 2, it picks up all the widgets from both screen 1 and 2, instead of just those on screen 2.

https://github.com/user-attachments/assets/a804bdfa-6eb4-43a0-b6fc-1a413a10a47d

I basically tested with these configurations in my project, but I also tested just the example from the library itself and got the same result.

```dart
const String exampleDsn =
    'https://dnsexample';

/// This is an exampleUrl that will be used to demonstrate how http requests are captured.
const String exampleUrl = 'https://jsonplaceholder.typicode.com/todos/';

const _channel = MethodChannel('example.flutter.sentry.io');
var _isIntegrationTest = false;

Future<void> main() async {
  await setupSentry(
    () => runApp(
      SentryWidget(
        child: DefaultAssetBundle(
          bundle: SentryAssetBundle(),
          child: MaterialApp(
            navigatorObservers: [SentryNavigatorObserver()],
            title: 'Navigation Basics',
            home: FirstRoute(),
          ),
        ),
      ),
    
```),
    exampleDsn,
  );
}

Future<void> setupSentry(
  AppRunner appRunner,
  String dsn, {
  bool isIntegrationTest = false,
  BeforeSendCallback? beforeSendCallback,
}) async {
  await SentryFlutter.init(
    (options) {
      options.dsn = exampleDsn;
      options.tracesSampleRate = 1.0;
      options.profilesSampleRate = 1.0;
      options.reportPackages = false;
      options.addInAppInclude('sentry_flutter_example');
      options.considerInAppFramesByDefault = false;
      options.attachThreads = true;
      options.enableWindowMetricBreadcrumbs = true;
      options.sendDefaultPii = true;
      options.reportSilentFlutterErrors = true;
      options.attachScreenshot = true;
      options.screenshotQuality = SentryScreenshotQuality.low;
      options.attachViewHierarchy = true;
      options.debug = true;
      options.spotlight = Spotlight(enabled: true);
      options.enableTimeToFullDisplayTracing = true;
      options.enableMetrics = true;

[...]

PR #2486, opened 2024-12-12
fix: widget masks overlap when navigating between screens.

thisames:
## :scroll: Description

At first I thought it would be necessary to implement a way to scrape widgets by context, as I did in posthog, but I realized that your code was already correct, it just needs one adjustment.

applying `debugVisitOnstageChildren` to your children does exactly what we were looking for. a way to "separate the elements that are hidden and not interactive from the active elements in the current context of the screen"

widget masks overlap when navigating between screens.

#2309 

I just did several tests with different navigations and applications, checking and debugging the widgets

https://github.com/user-attachments/assets/24b0e94e-0d79-42a4-ba71-2d7651423d1a

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes


flutter/lib/src/screenshot/widget_filter.dart
```diff
@@ -80,7 +80,7 @@ class WidgetFilter {
         break;
       case SentryMaskingDecision.continueProcessing:
         // If this element should not be obscured, visit and check its children.
-        element.visitChildElements(_visitList.add);
+        element.debugVisitOnstageChildren(_visitList.add);
```
vaind: This does not seem to be 100% reliable. So far in my testing, I've seen one occasion when masks were off. Never seen that before. I'll check the implementations of debugVisitOnstageChildren().

In the following screenshot (and the correct one for reference), it seems way off. Note: this has been taken while scrolling rapidly.
image
image

vaind (changes requested):

flutter/lib/src/screenshot/widget_filter.dart
thisames: What app are you testing with? I would like to reproduce the same time, but besides that, in my tests if I scroll quickly with `visitChildElements` the same thing will happen (masks were off).

try comparing the test you just did using `visitChildElements` and `debugVisitOnstageChildren`.

vaind: > What app are you testing with? 

https://github.com/vaind/spotube/tree/sentry

>  in my tests if I scroll quickly with `visitChildElements` the same thing will happen (masks were off).
> try comparing the test you just did using `visitChildElements` and `debugVisitOnstageChildren`.

Interesting, I've never noticed that before. If you have a good repro you could shere for this, I'd like to take a look.

vaind: This was an unrelated issue, with a workaround/fix implemented in #2553

vaind (approved):

vaind:
I've created an in-repo PR so that CI runs completely: #2576 
but then messed up a bit and force-pushed here instead of to the other PR so this shows as empty - sorry about that. In any case, your original commit is in the internal PR and will be merged together with the changelog. Thank you for your contribution @thisames

thisames:
> I've created an in-repo PR so that CI runs completely: #2576 but then messed up a bit and force-pushed here instead of to the other PR so this shows as empty - sorry about that. In any case, your original commit is in the internal PR and will be merged together with the changelog. Thank you for your contribution @thisames

no problem, good job

PR #2493, opened 2024-12-13
refactor: make screenshot recorder capture() synchronous so it's harder to break accidentally with future changes

vaind:
## :scroll: Description

Capturing screenshots has two critical parts: 
- capture the image
- determine where to put mask rectangles over that image

These two must take place without any asynchronous gap. 

Therefore, I'm removing `async` from `ScreenshotRecorder.capture()` so that someone doesn't accidentally shift around some code in the future and introduce an `await` between these two steps. 

Also, I'm changing the WidgetFilter usage so that it's not reused between runs as that could cause the `.items` list to get overwritten if another screenshot starts capturing before a previous one has finished processing.

Additionally, this PR moves the actual handling of the replay screenshot and converting it to the image to `SchedulerBinding.scheduleTask()` so that it runs between frames. 

For event screenshot attachments, I'm keeping leaving it as is, otherwise captureEvent() would wait indefinitely if the app never becomes idle; this may happen in tests where you have to explicitly call `TestWidgetsFlutterBinding.idle()` so it could actually break sentry-sdk user tests if they awaited for `captureEvent()` with real http backend. For us, it happens in integration tests.

Also, it adds [Timeline](https://api.flutter.dev/flutter/dart-developer/Timeline-class.html) events that help when profiling (won't do anything in release builds)

#skip-changelog

## :pencil: Checklist

- [ ] All tests passing
- [ ] No breaking changes


flutter/lib/src/screenshot/recorder.dart
```diff
@@ -110,34 +191,28 @@ class ScreenshotRecorder {
...
+        final finalImage = await picture.toImage(width, height);
+        Timeline.finishSync(); // Sentry::screenshotToImage
         try {
-          await callback(finalImage);
+          Timeline.startSync('Sentry::screenshotCallback', flow: flow);
+          _completer.complete(await callback(finalImage));
+          Timeline.finishSync(); // Sentry::screenshotCallback
         } finally {
-          finalImage.dispose(); // image needs to be disposed manually
+          finalImage.dispose(); // image needs to be disposed-of manually
```
buenaflor, line 212: do we need a  `_completer.complete(null)` in finally if `await callback(finalImage)` throws?

buenaflor (commented):
looks good, only got one question

flutter/lib/src/screenshot/recorder.dart
vaind: Exception is handled in in executeTask().onError callback. I've also added the completer return in this case, thanks for the suggestion. See [`804834f` (#2493)](https://github.com/getsentry/sentry-dart/pull/2493/commits/804834fb97d6c6b17f43b8b1ea14597106572395)

buenaflor (approved):
lgtm 👍

Issue #2491, opened 2024-12-12
SentryFlutter.Init is pausing the UI thread for some reason


noti0nS:
## Workaround for this issue

**Edit by Sentry employee: Meanwhile as a workaround until the fix is out, instead of using `WidgetsFlutterBinding.ensureInitialized()` you can replace it with `SentryWidgetsFlutterBinding.ensureInitialized()`** 

### Platform

Flutter Mobile Android

### Obfuscation

Enabled

### Debug Info

Enabled

### Doctor

[√] Flutter (Channel stable, 3.27.0, on Microsoft Windows [Version 10.0.22631.4602], locale pt-BR)
    • Flutter version 3.27.0 on channel stable at C:\dev\flutter\stable
    • Upstream repository https://github.com/flutter/flutter.git
    • Framework revision 8495dee1fd (2 days ago), 2024-12-10 14:23:39 -0800
    • Engine revision 83bacfc525
    • Dart version 3.6.0
    • DevTools version 2.40.2

[√] Windows Version (Installed version of Windows is version 10 or higher)

[√] Android toolchain - develop for Android devices (Android SDK version 34.0.0)
    • Android SDK at C:\Users\Josel\AppData\Local\Android\sdk
    • Platform android-34, build-tools 34.0.0
    • Java binary at: C:\Program Files\Android\Android Studio\jbr\bin\java
    • Java version OpenJDK Runtime Environment (build 17.0.6+0-b2043.56-10027231)
    • All Android licenses accepted.

[√] Chrome - develop for the web
    • Chrome at C:\Program Files\Google\Chrome\Application\chrome.exe

[√] Visual Studio - develop Windows apps (Visual Studio Community 2022 17.12.3)
    • Visual Studio at C:\Program Files\Microsoft Visual Studio\2022\Community
    • Visual Studio Community 2022 version 17.12.35527.113
    • Windows 10 SDK version 10.0.22621.0

[√] Android Studio (version 2022.3)
    • Android Studio at C:\Program Files\Android\Android Studio
    • Flutter plugin can be installed from:
       https://plugins.jetbrains.com/plugin/9212-flutter
    • Dart plugin can be installed from:
       https://plugins.jetbrains.com/plugin/6351-dart
    • Java version OpenJDK Runtime Environment (build 17.0.6+0-b2043.56-10027231)

[√] VS Code (version 1.96.0)
    • VS Code at C:\Users\Josel\AppData\Local\Programs\Microsoft VS Code
[...]

PR #2494, opened 2024-12-13
fix: SentryWidgetsFlutterBinding initializing even if a binding already exists

buenaflor:
## :scroll: Description

When user already initialized their WidgetsBinding with

`WidgetsFlutterBinding.ensureInitialized()` 

and then runs our SDK, the SDK tries to also initialize `SentryFlutterWidgetsBinding` but does it even if one already exists. This leads to unwanted side effects such as first frame being broken.

This wasn't caught because it only fails in profile/release mode.

Fixes https://github.com/getsentry/sentry-dart/issues/2491

integration test in profile mode, manual test with release mode

## :pencil: Checklist

- [ ] All tests passing
- [x] No breaking changes


.github/workflows/flutter.yml
```diff
@@ -84,13 +84,13 @@ jobs:
         if: matrix.target == 'web'
         run: |
           flutter test --platform chrome --test-randomize-ordering-seed=random --exclude-tags canvasKit
-          flutter test --platform chrome --test-randomize-ordering-seed=random --tags canvasKit --web-renderer canvaskit
```
buenaflor, line 87: ci fails in beta, we don't need this since canvasKit is the default anyway

.github/workflows/flutter.yml
```diff
@@ -84,13 +84,13 @@ jobs:
...
+          flutter test --platform chrome --test-randomize-ordering-seed=random --tags canvasKit
```
buenaflor, line 87: ci fails in flutter beta (with error `can't find --web-renderer option`, most likely removed in latest flutter beta), we don't need this since canvasKit is the default anyway

stefanosiano (approved):
nice!

PR #2517, opened 2024-12-18
make Sentry Native usage opt-in

buenaflor:
Only enable sentry-native usage when the `enableSentryNative` flag is true (feel free to suggest another name)

Usually we shouldn't change behaviour of features released in a minor but since this causes multiple issues imo it make sense to do it.

Note: this will be opt-out in v9

Not sure what to categorize this in the changelog


vaind (changes requested):
There's no need for any changes in the runtime part, we need to make compilation opt-in. I'll prep a PR if you don't feel comfortable with cmake.

buenaflor:
sure, I'll leave you to it then 👍

PR #2520, opened 2024-12-19
fix: mask sizing on scaling transform widget children

vaind:
## :scroll: Description

While testing, I've found a situation when there's a `Transform` applied on a widget tree. This would cause the size to be reported smaller than it actually renders on screen(shot). This PR changes how we calculate the bounding box to fix the issue.

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes


buenaflor (approved):
👍

PR #2530, opened 2024-12-20
iOS copy native replay screenshot in-memory to native

vaind:
## :scroll: Description

Sending binary data (uint8list) over method channels is rather slow so instead, we allocate the memory directly in dart and only send the address and length & restore the object in objective-c.

Following screenshots show profiling flamechart (combined data) over 20-ish seconds before and after this PR. I've highlighted the relevant part
Screenshot 2024-12-20 at 20 33 16
Screenshot 2024-12-20 at 20 35 24

## :pencil: Checklist

- [ ] All tests passing
- [ ] No breaking changes


flutter/lib/src/native/native_memory.dart
```diff
@@ -0,0 +1,45 @@
...
+
+  factory NativeMemory.fromJson(Map<dynamic, dynamic> json) {
+    final length = json['length'] as int;
+    final ptr = Pointer<Uint8>.fromAddress(json['address'] as int);
+    return NativeMemory._(ptr, length);
+  }
+
+  /// Frees the underlying native memory.
+  /// You must not use this object after freeing.
+  void free() => pkg_ffi.malloc.free(pointer);
```
buenaflor, line 36: do we need to free this in the Flutter layer? or is this only for tests?

vaind: Currently, it's only used in tests, because there's no native counterpoint to actually free the memory.

buenaflor: gotcha, shall we add a small comment for that? or add a `@visibleForTests`?

vaind: added a comment & annotation

buenaflor (approved):

PR #2534, opened 2024-12-23
ci: deduplicate and cleanup actions

vaind:
Ubuntu-latest will soon change to ubuntu-24.04 so I was checking out what needs to be changes and noticed we have some duplicate code on flutter packages. This merges the code, similar to what we do for dart packages.

changes:
- add a shared flutter build action and use it
- synchronize matrix configs among flutter packages
- enable flutter-beta channel on windows
- removes `flutter upgrade` - there's really no reason to do this in CI after we've just installed latest version
- splits `wasm` && `js` builds to separate matrix items
- removes android/ios targets in packages that actually don't test it (they just ran `flutter test` for the current platform)
- remove flutter_integration_test.yml - duplicate of flutter_test.yml
- update conditional path triggers for workflows to include local reusable actions & workflows
- add missing dart 3.5 build


buenaflor (approved):
nice thx 👍

PR #2546, opened 2025-01-02
feat(js-sdk): support release health

buenaflor:
## :scroll: Description

Built on top of PR https://github.com/getsentry/sentry-dart/pull/2489

Adds support for sessions / release health.

## :pencil: Checklist

- [ ] All tests passing
- [ ] No breaking changes


dart/lib/src/native/c/sentry_native.dart
bruno-garcia, line 1: this empty file needed?

dart/lib/src/sentry_options.dart
```diff
@@ -276,7 +276,7 @@ class SentryOptions {
   /// breadcrumbs.
   /// In a Flutter environment, this setting also toggles recording of `debugPrint` calls.
   /// `debugPrint` calls are only recorded in release builds, though.
-  bool enablePrintBreadcrumbs = true;
+  bool enablePrintBreadcrumbs = false;
```
bruno-garcia, line 279: Is this change needed to be part of the web support? it could have side effects of itself worth tracking as a separate change probably done before the big web support.

bruno-garcia (commented):
the definition of session in JS is about to change on v9
might be a good idea to line it up, so only ship the Dart on JS session using the v9 sdk. you could piggy back on their alpha to ship your on alphas etc

dart/lib/src/sentry_options.dart
buenaflor: this is temporary for testing, on web with the js sdk integration this seems to run into a recursion so I've disabled it for now I will revert it back after I fixed the issue

buenaflor:
note: currently the PR that this is based on is not merged yet so bunch of unrelated changes will show up for now

buenaflor:
we'll wait for JS v9 before implementing sessions

PR #2553, opened 2025-01-07
fix/screenshot masking during changes

vaind:
## :scroll: Description

We're facing an issue: the tree walked with `visitChildElements()` is out of sync to what is currently rendered by `RenderRepaintBoundary.toImage()`, even though there's no async gap between these two. This causes masks to be off during repaints, e.g. when scrolling a view or when text is rendered in different places between two screens. This is most easily reproducible when there's no animation between the two screens. 

For example, Spotube's Search vs Library (2nd and 3rd bottom bar buttons). The following composition of three subsequent frames shows the transition. Note the special mode of rendering which prints masks (in the color that the text was in the original button) and then writes the text that is being masked in bright green. This way you can see that the masks move over to their new position (about 30 px higher then before) when the screen changes, but the captured image still renders the previous version of the screen. I've added a red guideline to make this change easier to see.

image

I've created [an issue upstream](https://github.com/flutter/flutter/issues/161228#issuecomment-2577230350) to try and get this resolved "properly". As an interim solution, we're taking two subsequent screenshots (two frames) and only actually capture a screenshot for replay if the two are exactly the same. 

Mostly manually as I could not reproduce the masking issue programatically. There are some new and updated tests to cover what I could.

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,11 +6,15 @@
 
 - Add `beforeCapture` for View Hierarchy ([#2523](https://github.com/getsentry/sentry-dart/pull/2523))
   - View hierarchy calls are now debounced for 2 seconds.
-  
+
 ### Enhancements
 
 - Replay: improve iOS native interop performance ([#2530](https://github.com/getsentry/sentry-dart/pull/2530))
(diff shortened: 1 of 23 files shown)


buenaflor (approved):
looks good 👍

PR #2564, opened 2025-01-09
ci(linux): fix `Failed to load dynamic library 'libsqlite3.so'`

buenaflor:
drift and sqflite tests are failing because libsqlite3 is missing on Linux

Example: https://github.com/getsentry/sentry-dart/actions/runs/12671780639/job/35314287776

#skip-changelog

### DIFF
--- a/.github/workflows/drift.yml
+++ b/.github/workflows/drift.yml
@@ -37,7 +37,9 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
-      - run: sudo apt-get -y install libsqlite3-dev
+      - name: Install libsqlite3
+        if: matrix.target == 'linux'
+        run: sudo apt-get -y install libsqlite3-dev
 
       - uses: ./.github/actions/flutter-test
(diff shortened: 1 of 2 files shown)


philprime (approved):

.github/workflows/drift.yml
```diff
@@ -37,6 +37,10 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
+      - name: Install libsqlite3
+        if: matrix.target == 'linux'
+        run: sudo apt-get -y install libsqlite3-dev
+
```
vaind, line 43: I think we should do this in the shared action even though it may not be necessary on all flutter builds: https://github.com/getsentry/sentry-dart/blob/0c0805414333c4e379a41ab179dd5749b49dec27/.github/actions/flutter-test/action.yml#L29

buenaflor: yea we can do that, I'll move it in another PR

PR #2571, opened 2025-01-13
Remove metrics API

### LINKED ISSUE #2560: Remove metrics API
### Description

It's not used anymore and can be removed

buenaflor:
Removing public API is usually done in majors but metrics was in beta before and the product is completely shut down so it's just dead code at this point

Closes #2560


buenaflor:
cc @stefanosiano

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Removals
```
buenaflor: Not sure how else to call this header

vaind: Sometimes it's called Changes. In this case you could also say "Breaking changes"

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
...
+
+- Remove Metrics API ([#2571](https://github.com/getsentry/sentry-dart/pull/2571))
```
vaind, line 7: maybe worth saying this never reached maturity from beta since technically it's breaking

flutter/lib/src/hub.dart
```diff
@@ -0,0 +1 @@
+
```
vaind: What's this? Do we need to keep this empty file?

vaind (approved):

flutter/lib/src/hub.dart
buenaflor: I'll remove it, this happened a couple times now, no idea how that sneaks in 😓

stefanosiano (approved):
looks good!

PR #2573, opened 2025-01-13
enha: replay: use rawRGBA for replay native integration on ios

vaind:
## :scroll: Description

Dart PNG conversion is rather slow (about 350 ms for a single screenshot) and may be unnecessary. We can leave it up to the native code to handle images in bitmap. 

## :pencil: Checklist

- [ ] All tests passing
- [ ] No breaking changes

## :crystal_ball: Next steps

Android is a bit more complicated to implement and would benefit from bumping to a higher Dart SDK version (3.5)


buenaflor (approved):
👍

PR #2644, opened 2025-01-29
flutter-example: migrate legacy buildscript

buenaflor:
#skip-changelog

Flutter beta CI jobs errored urging us to upgrade the example from the legacy buildscript to using plugin blocks

I also removed the nativeBundle dependency as I wasn't able to build with it (it also only supports AGP up to 8.0.2) and replace it with updating the cmakelists

See https://docs.flutter.dev/release/breaking-changes/flutter-gradle-plugin-apply


denrase (approved):
Looks good 👍

PR #2659, opened 2025-02-05
wip: v9

vaind:
This will be merged (without squashing) back to main when ready and we want to do a release.

#skip-changelog


buenaflor (approved):
@kahest can we re-enable merging without squashing pls

Issue #2345, opened 2024-10-10
Add Flutter E2E & Integration Test


buenaflor:
### Description

```[tasklist]
### Tasks
- [x] re-enable e2e tests
- [x] valid envelope test
- [ ] invalid/corrupt envelope test
- [ ] check that a replay was captured (see RN e2e tests for API endpoints to check)
- [ ] symbolication & source context
```

buenaflor:
For symbolication & source context e2e test we could use maestro. I dont think the flutter's integration test will be able to test that<br><br>We could at the very least test it for android, not sure about macos and ios though, I dont think you can build an obfuscated build for the simulator

PR #2670, opened 2025-02-07
refactor: use JNI for Android integration

vaind:
## :scroll: Description

This is a first step towards migrating our Android plugin glue code to JNI (part of #1444). 

### Done in this PR:
- added JNI and JNIGEN, currently pinning to a specific version because `package:jni` is in a nonstable release cycle (0.x) so there may be breaking changes even between minor releases. 
- use JNI for replay screenshot capture to improve performance
- offload the native call & bitmap creation to another isolate

### :crystal_ball: Next steps

- Integration tests for replay capture (#2345)
- incrementally migrate more of the android plugin to JNI

Manually & udpated tests

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes


vaind:
@HosseinYousefi do you mind taking a look?

flutter/lib/src/native/java/android_replay_recorder.dart
```diff
@@ -48,3 +73,141 @@ class AndroidReplayRecorder extends ScheduledScreenshotRecorder {
...
+
+    return _AndroidNativeReplayWorker._(receivePort, sendPort);
+  }
+
+  _AndroidNativeReplayWorker._(this._responses, this._commands) {
+    _responses.listen(_handleResponsesFromIsolate);
+  }
+
+  Future<Object?> nativeAddScreenshot(_WorkItem item) async {
+    if (_closed) throw StateError('Closed');
```
HosseinYousefi, line 114: Note that this code path will not release the `jBuffer` eagerly, which is fine if we're throwing the state error to crash.

flutter/lib/src/native/java/android_replay_recorder.dart
```diff
@@ -48,3 +73,141 @@ class AndroidReplayRecorder extends ScheduledScreenshotRecorder {
...
+
+  /// This is the actual Android native implementation, the rest is just plumbing.
+  static void _handleCommandsToIsolate(
+    ReceivePort receivePort,
+    SendPort sendPort,
+  ) {
+    // Android Bitmap creation is a bit costly so we reuse it between captures.
+    native.Bitmap? bitmap;
+
+    final _nativeReplay = native.SentryFlutterPlugin$Companion(null)
```
HosseinYousefi, line 143: Please open an issue for JNIgen to remove the synthetic "default constructor" marker added by the Kotlin compiler from the generated code and just pass `null`.

flutter/lib/src/native/java/android_replay_recorder.dart
```diff
@@ -48,3 +73,141 @@ class AndroidReplayRecorder extends ScheduledScreenshotRecorder {
...
+              bitmap!.getHeight() != item.height) {
+            bitmap!.release();
+            bitmap = null;
+          }
+        }
+
+        // https://developer.android.com/reference/android/graphics/Bitmap#createBitmap(int,%20int,%20android.graphics.Bitmap.Config)
+        // Note: in the currently generated API this may return null so we null-check below.
+        bitmap ??= native.Bitmap.createBitmap$3(
+            item.width, item.height, native.Bitmap$Config.ARGB_8888);
```
HosseinYousefi, line 164: `.ARGB_8888` creates a references which is not eagerly released.

flutter/lib/src/native/java/android_replay_recorder.dart
```diff
@@ -1,40 +1,65 @@
...
 
 import '../../../sentry_flutter.dart';
 import '../../replay/scheduled_recorder.dart';
+import '../../replay/scheduled_recorder_config.dart';
 import '../../screenshot/screenshot.dart';
-import '../sentry_safe_method_channel.dart';
+import 'binding.dart' as native;
 
+// Note, this is currently not unit-tested because mocking the JNI calls is
+// cumbersome, see https://github.com/dart-lang/native/issues/1794
```
HosseinYousefi, line 15: Please let me know what your ideal mocking solution would look like in the issue.

flutter/microbenchmarks/lib/src/jni_bench.dart
```diff
@@ -0,0 +1,95 @@
+import 'dart:ffi';
+import 'dart:math';
+
+import 'package:jni/jni.dart';
+import 'package:benchmarking/benchmarking.dart';
+import 'package:flutter/foundation.dart';
+
+Future<void> execute() async {
```
HosseinYousefi, line 8: Out of curiousity, have you measured the perfomance improvement between the previous way of using method channels vs. using jni?

HosseinYousefi (commented):
I checked the JNI related parts and overall LGTM with one small comment.

flutter/lib/src/native/java/android_replay_recorder.dart
vaind: Thanks for pointing it out, I've actually had a suspicion that may be the case but it doesn't matter in this case as it's only executed once for each orientation change so it's fine if it's collected by GC

flutter/microbenchmarks/lib/src/jni_bench.dart
vaind: Not per se. I have profiled it previously and with the new code and the biggest benefit we get is that we can now use `Bitmap` directly from raw RGBA so we don't have to convert to PNG (which is in hundreds of milliseconds, even though it's async) and to write an intermediary file (two IO ops - write & read). So there really is no one-to-one comparison between the approaches.
Another benefit was being able to move the work to another isolate altogether. Actually, I need to check whether I can pass an Uint8List efficiently between isolates. Last time I checked it didn't work.

flutter/lib/src/native/java/android_replay_recorder.dart
vaind: thx. I've moved jbuffer creation to the isolate so this is resolved now

CHANGELOG.md
```diff
@@ -10,6 +12,10 @@
 - Remove user segment ([#2687](https://github.com/getsentry/sentry-dart/pull/2687))
 - Remove `options.autoAppStart` and `setAppStartEnd` ([#2680](https://github.com/getsentry/sentry-dart/pull/2680))
 
+### Enhancements
+
+- Replay: improve iOS native interop performance by using JNI ([#2670](https://github.com/getsentry/sentry-dart/pull/2670))
```
```suggestion
- Replay: improve Android native interop performance by using JNI ([#2670](https://github.com/getsentry/sentry-dart/pull/2670))
```

buenaflor (approved):
can't comment much on the jni parts but lgtm so far

CHANGELOG.md
vaind: copypasta 🍝

PR #2679, opened 2025-02-10
[v9]: bump drift min version and use `QueryInterceptor` instead of `QueryExecutor`

### LINKED ISSUE #2165: [v9]: remove `SentryQueryExecutor` in favour of using `QueryInterceptor` in Drift
### Description

`QueryInterceptor` is specifically made for hooking into database calls to monitor them whereas `QueryExecutor` is an internal api that may break at some point

```[tasklist]
### Tasks
- [x] https://github.com/getsentry/sentry-dart/pull/2715
- [ ] https://github.com/getsentry/sentry-dart/pull/2679
```

buenaflor:

Use better api for instrumenting drift operations

Part of #2165 

## :pencil: Checklist

- [ ] All tests passing
- [ ] No breaking changes

## :crystal_ball: Next steps

Deprecate it in the main branch


dart/lib/src/constants.dart
```diff
@@ -0,0 +1,29 @@
...
+  static const String dbOpen = 'db.open';
+  static const String dbClose = 'db.close';
+}
+
+@internal
+class SentrySpanData {
+  static const String dbSystemKey = 'db.system';
+  static const String dbNameKey = 'db.name';
+
+  static const String dbSystemSqlite = 'db.sqlite';
```
buenaflor, line 20: tbd in other PR, add other db systems here

dart/lib/src/constants.dart
```diff
@@ -0,0 +1,30 @@
...
+  static const String uiTimeToFullDisplay = 'ui.load.full_display';
+  static const String dbSqlQuery = 'db.sql.query';
+  static const String dbSqlTransaction = 'db.sql.transaction';
+  static const String dbSqlBatch = 'db.sql.batch';
+  static const String dbOpen = 'db.open';
+  static const String dbClose = 'db.close';
+}
+
+@internal
+class SentrySpanData {
```
buenaflor, line 16: tbd in other PR, remove `SpanDataConvention` and put the constants here

drift/example/example.dart
```diff
@@ -22,18 +22,36 @@ Future<void> main() async {
...
+    await db.into(db.todoItems).insert(
+          TodoItemsCompanion.insert(
+            title: 'This is a test thing',
+            content: 'test',
+          ),
+        );
+
+    await db.batch((batch) {
+      batch.insertAll(db.todoItems, [
+        TodoItemsCompanion.insert(title: 'First entry', content: 'My content'),
```
denrase, line 50: We could add a `,` after `'My content'`, so its formatted the same as the item below.

drift/lib/src/constants.dart
```diff
@@ -0,0 +1 @@
+const String integrationName = 'SentryDriftTracing';
```
denrase: Because i'm seeing this just now, dont we start integration names lowercase? So `sentryDriftTracing`

drift/lib/src/sentry_query_interceptor.dart
```diff
@@ -0,0 +1,164 @@
...
+  late final SentrySpanHelper _spanHelper;
+  bool _isDbOpen = false;
+
+  SentryQueryInterceptor({required String databaseName, @internal Hub? hub})
+      : _dbName = databaseName {
+    _spanHelper = SentrySpanHelper(
+      SentryTraceOrigins.autoDbDriftQueryInterceptor,
+      hub: hub,
+    );
+    final options = hub?.options;
```
denrase: Don't we need to fallback to `HubAdapter`? We do this in SentrySpanHelper, but nut here. Wenn we call the class like in the example, options will be null here, unless i'm missing something.

```dart
SentryQueryInterceptor(databaseName: 'your_db_name'),
```

drift/lib/src/sentry_query_interceptor.dart
```diff
@@ -0,0 +1,164 @@
...
+        description,
+        () async => execute(),
+        dbName: _dbName,
+        operation: operation,
+      );
+
+  @override
+  Future<bool> ensureOpen(QueryExecutor executor, QueryExecutorUser user) {
+    if (!_isDbOpen) {
+      _isDbOpen = true;
```
denrase: Technically we set `_isDbOpen = true` before the db is open, as this is only the case when `super.ensureOpen(executor, user)` completes. Does this bother us?

denrase: Before we would await fore setting the flag.

Bildschirmfoto 2025-02-14 um 10 18 35

drift/lib/src/sentry_span_helper.dart
```diff
@@ -65,91 +49,71 @@ class SentrySpanHelper {
...
-    String description,
     T Function() execute, {
     String? dbName,
   }) {
-    final currentSpan = _hub.getSpan();
-    final span = currentSpan?.startChild(
-      SentryQueryExecutor.dbOp,
-      description: description,
+    final scopeSpan = _hub.getSpan();
+    _parentSpan = scopeSpan?.startChild(
```
denrase: Does it make a difference if we assign `_parentSpan` sooner? Before we did it in the finally block. Also, why the rename to `_parentSpan` from `transactionSpan`, was quite fitting, no?

Also one question, are nested transactions possible and is this a case we should handle with some kind of span stack?

denrase (changes requested):

drift/example/example.dart
buenaflor: dart formatting does this automatically, not sure why it's inconsistent here, it removes the `,` when I add it

drift/lib/src/constants.dart
buenaflor: usually we do but this was the way it was previously, I was thinking about not changing it because it might break some data querying for our dashboards but I guess it's fine since it's for a major

drift/lib/src/sentry_query_interceptor.dart
buenaflor: yeah I think the previous method is more consistent

drift/lib/src/sentry_query_interceptor.dart
buenaflor: thx for noticing, the tests didn't catch that because we pass the hub directly

will correct it

drift/lib/src/sentry_span_helper.dart
buenaflor: > Does it make a difference if we assign _parentSpan sooner?

no, I only added it to the top for visibility

> why the rename to _parentSpan from transactionSpan

transactionSpan is kinda confusing since we're mixing sentry transaction and drift transaction terminology

also I updated it to keep track of the parent span in general, not only that one span, so the possible values right now are: scope span or drift transaction span

buenaflor: > nested transactions possible and is this a case we should handle with some kind of span stack?

nested transactions are supported by drift, but I didn't think of this

we should support for it imo, I'll have a look

buenaflor:
updated impl to support nested transactions

image

drift/lib/src/sentry_span_helper.dart
```diff
@@ -9,16 +9,17 @@ class SentrySpanHelper {
...
 
   Future<T> asyncWrapInSpan<T>(
     String description,
     Future<T> Function() execute, {
     String? dbName,
     String? operation,
   }) async {
-    final currentSpan = _parentSpan ?? _hub.getSpan();
-    final span = currentSpan?.startChild(
+    final span = _parentSpan?.startChild(
```
denrase: Wouldn't we have the issue that no span were created if there is no parent span? In the case that upon initialization, the hub does not have a span.

denrase: Also won't attach to the new one if the hub span changes when calling this method.

denrase (changes requested):

drift/lib/src/sentry_span_helper.dart
buenaflor: ~~that's the behaviour we have for all other integrations as well, they require an active transaction~~ (sry my fault, I understand the issue)

buenaflor: maybe we should check for the hub span when we try to instrument

buenaflor: also I think you commented an older commit, I was confused at first because I changed the impl to use a stack

drift/lib/src/sentry_span_helper.dart
```diff
@@ -94,38 +107,46 @@ class SentrySpanHelper {
...
     } catch (exception) {
-      parentSpan?.throwable = exception;
-      parentSpan?.status = SpanStatus.internalError();
+      parentSpan.throwable = exception;
+      parentSpan.status = SpanStatus.internalError();
 
       rethrow;
     } finally {
-      await parentSpan?.finish();
+      await parentSpan.finish();
```
denrase, line 141: Shouldn't we also remove a span here if it came from the stack, or is this done somewhere else?

denrase (approved):

drift/lib/src/sentry_span_helper.dart
buenaflor: it's removed above:   ` _transactionStack.removeLast()` but yea putting it in finally is better

PR #2707, opened 2025-02-12
chore: ignore dart:html warnings

buenaflor:
#skip-changelog

we're gonna move away from dart:html in v9 so it's fine to ignore it in v8


krystofwoldrich (approved):
LGTM 🚀

PR #2715, opened 2025-02-13
deprecate: drift `SentryQueryExecutor`

buenaflor:
This will be replaced in v9 by `SentryQueryInterceptor`


denrase (approved):
Analyzer issues remaining, as we use the now deprecated class in tests.

PR #2726, opened 2025-02-17
fix: auto ip guarded by `sendDefaultPii`

buenaflor:

Previously wasn't guarded by `sendDefaultPii` because of https://github.com/getsentry/sentry-dart/issues/1562

https://github.com/getsentry/team-sdks/issues/118 states it should only be added if `sendDefaultPii = true`
Unit test

## :pencil: Checklist

- [ ] All tests passing
- [ ] No breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 ### Behavioral changes
 
 - ⚠️ Auto IP assignment for `SentryUser` is now guarded by `sendDefaultPii` ([#2726](https://github.com/getsentry/sentry-dart/pull/2726))
-  - If you rely on Sentry automatically processing the IP address of the user, set `options.sendDefaultPii = true` or manually set the ip address of the `SentryUser` to `{{auto}}`
+  - If you rely on Sentry automatically processing the IP address of the user, set `options.sendDefaultPii = true` or manually set the IP address of the `SentryUser` to `{{auto}}`
 
 ### Features
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+### Behavioral changes
+
 - ⚠️ Auto IP assignment for `SentryUser` is now guarded by `sendDefaultPii` ([#2726](https://github.com/getsentry/sentry-dart/pull/2726))
   - If you rely on Sentry automatically processing the IP address of the user, set `options.sendDefaultPii = true` or manually set the ip address of the `SentryUser` to `{{auto}}`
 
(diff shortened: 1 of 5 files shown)


buenaflor:
cc @kahest, would this be considered breaking if we fix this in v8?

kahest:
> cc @kahest, would this be considered breaking if we fix this in v8?

as discussed - yes we should ship it in a v8 minor with very clear callout on the change log, similar to other platforms (e.g. .NET, Java/Android)

kahest (commented):
apart from missing changelog, LGTM

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+- ⚠️ Auto IP assignment for `SentryUser` is now guarded by `sendDefaultPii` ([#2726](https://github.com/getsentry/sentry-dart/pull/2726))
```
buenaflor, line 7: @kahest I added it to the top without a category for most visibility, wdyt?

kahest: placement makes sense 👍 please also add a header called "Behavioral changes"

dart/test/sentry_client_test.dart
```diff
@@ -1040,7 +1053,8 @@ void main() {
       expect(capturedEvent.user?.email, fakeEvent.user!.email);
     });
 
-    test('event has a user without IP address', () async {
+    test('event has a user without IP address and sendDefaultPii = true',
+        () async {
       final client = fixture.getSut(sendDefaultPii: true);
 
       final event = fakeEvent.copyWith(user: fakeUser);
```
denrase, line 1063: In the test `event has a user with IP address` we assume the user has an ip-address. Here we assume the user has none, but we do not set it to null anywhere in the test. We need to set it to null before calling captureEvent.

Bildschirmfoto 2025-02-19 um 14 32 58

denrase (changes requested):

dart/test/sentry_client_test.dart
buenaflor: we are overriding the user on the event with `copyWith` because `fakeEvent` by default has a user already.  the overridden user has a null ip address

 we use the `event` within `captureEvent` and not `fakeEvent`

buenaflor: you need to check `expect(event.user!.ipAddress, isNull);`

denrase (approved):

dart/test/sentry_client_test.dart
buenaflor: actually let me add that assertion to be safe

PR #2733, opened 2025-02-19
ci: fix analyzer for microbenchmarking

buenaflor:
#skip-changelog


buenaflor:
closingt this as this is ignored in release jobs and will be fixed in v9 by bumping the min dart version to 3.5

PR #2742, opened 2025-02-20
Add Flutter runtime information

### LINKED ISSUE #416: Report version of Flutter
Came up on: https://github.com/flutter/flutter/issues/61814

The goal is to include the 'runtime' interface with the Dart and/or Flutter version.
If there's still no API for this, perhaps we can get this at compile time

ueman:
## :scroll: Description

- Fixes https://github.com/getsentry/sentry-dart/issues/416
- Relevant Flutter PR https://github.com/flutter/flutter/pull/140783
- Edit: It got reverted again https://github.com/flutter/flutter/pull/163753
- Edit Edit: Let's try again: https://github.com/flutter/flutter/pull/163761

Fun fact: This fixes the oldest currently open issue in this repo

https://github.com/getsentry/sentry-dart/issues/416

I'm not entirely sure how to test this, since Dart defines aren't mockable.

You can test this manually by switching to the master channel of Flutter and building an app.

## :pencil: Checklist

- [ ] All tests passing
- [ ] No breaking changes

## :crystal_ball: Next steps

Wait until the Flutter PR is merged, then merge this.


CHANGELOG.md
```diff
@@ -16,6 +16,8 @@
 - Use `loadDebugImagesForAddresses` API for Android ([#2706](https://github.com/getsentry/sentry-dart/pull/2706))
   - This reduces the envelope size and data transferred across method channels
   - If debug images received by `loadDebugImagesForAddresses` are empty, the SDK loads all debug images as fallback
+- Add Flutter runtime information ([#2742](https://github.com/getsentry/sentry-dart/pull/2742))
+  - This will start working as soon as [this code](https://github.com/flutter/flutter/pull/140783) is available on the Flutter master, beta or stable channel.
```
```suggestion
  - This works if the version of Flutter you're using includes [this code](https://github.com/flutter/flutter/pull/140783).
```

buenaflor:
ohh nice! 🥳

dart/lib/src/event_processor/enricher/flutter_runtime.dart
```diff
@@ -0,0 +1,77 @@
...
+      FlutterVersion.channel == null ||
+      FlutterVersion.frameworkRevision == null) {
+    return null;
+  }
+
+  return SentryRuntime(
+    name: 'Flutter',
+    version: '${FlutterVersion.version} (${FlutterVersion.channel})',
+    build: FlutterVersion.frameworkRevision,
+  );
```
ueman: Does it make sense to you to structure it like this?

dart/lib/src/event_processor/enricher/flutter_runtime.dart
```diff
@@ -0,0 +1,77 @@
...
+
+  /// The Flutter channel used to compile the app.
+  static const String? channel = bool.hasEnvironment('FLUTTER_CHANNEL')
+      ? String.fromEnvironment('FLUTTER_CHANNEL')
+      : null;
+
+  /// The URL of the Git repository from which Flutter was obtained.
+  static const String? gitUrl = bool.hasEnvironment('FLUTTER_GIT_URL')
+      ? String.fromEnvironment('FLUTTER_GIT_URL')
+      : null;
```
ueman, line 56: Should this be included in a context somewhere?

ueman:
This PR is ready to review now, but maybe we should wait with merging a little to see whether the changes are sticking this time in the Flutter framework

dart/lib/src/event_processor/enricher/flutter_runtime.dart
```diff
@@ -0,0 +1,77 @@
...
+
+// The Flutter version information can be fetched via Dart defines,
+// see https://github.com/flutter/flutter/pull/140783.
+// This code lives in the Dart only Sentry code, since the code
+// doesn't require any Flutter dependency.
+// Additionally, this makes it work on background isolates in
+// Flutter, where one may not initialize the whole Flutter Sentry
+// SDK.
+
+SentryRuntime? get flutterRuntime {
```
ueman: Note to self: We can probably use lazy initialization here, or even a const property

ueman:
@buenaflor The linked PR is sticking, so it should be safe to merge now.

dart/lib/src/event_processor/enricher/flutter_runtime.dart
```diff
@@ -0,0 +1,76 @@
...
+  /// The Flutter engine revision.
+  static const String? engineRevision =
+      bool.hasEnvironment('FLUTTER_ENGINE_REVISION')
+          ? String.fromEnvironment('FLUTTER_ENGINE_REVISION')
+          : null;
+
+  // This is included since [Platform.version](https://api.dart.dev/stable/dart-io/Platform/version.html)
+  // is not included on web platforms.
+  /// The Dart version used to compile the app.
+  static const String? dartVersion = bool.hasEnvironment('FLUTTER_DART_VERSION')
```
ueman, line 73: It's slightly different than `Platform.version`:
- `Platform.version`: `3.8.0-133.0.dev (dev) (Sun Feb 23 12:02:50 2025 -0800) on "macos_x64"`
- `FlutterVersion.dartVersion`: `3.8.0 (build 3.8.0-133.0.dev)`

ueman:
@denrase Are you able to do a review on this?

dart/lib/src/event_processor/enricher/flutter_runtime.dart
```diff
@@ -0,0 +1,76 @@
...
+// Flutter, where one may not initialize the whole Flutter Sentry
+// SDK.
+// The const-ness of the properties below ensure that the code
+// is tree shaken in a non-Flutter environment.
+
+const _hasFlutterRuntimeInformation = FlutterVersion.version == null ||
+    FlutterVersion.channel == null ||
+    FlutterVersion.frameworkRevision == null;
+
+const SentryRuntime? flutterRuntime = _hasFlutterRuntimeInformation
```
denrase: I'm a bit confused here, maybe i'm just mis-reading. To me `_hasFlutterRuntimeInformation` would indicate that flutterRuntime should be non-null, no?

denrase: It would make sense if it would be called something like "_missingRelevantFlutterVersionInfo", but again, i'm probably misreading the intent.

dart/lib/src/event_processor/enricher/flutter_runtime.dart
```diff
@@ -0,0 +1,76 @@
...
+        name: 'Dart',
+        version: FlutterVersion.dartVersion,
+      );
+
+/// Details about the Flutter version this app was compiled with,
+/// corresponding to the output of `flutter --version`.
+///
+/// When this Flutter version was build from a fork, or when Flutter runs in a
+/// custom embedder, these values might be unreliable.
+abstract class FlutterVersion {
```
denrase, line 40: Since this is code from your Flutter PR, would it also make sense to bring the corresponding tests over?

dart/lib/src/event_processor/enricher/web_enricher_event_processor.dart
```diff
@@ -29,6 +30,7 @@ class WebEnricherEventProcessor implements EnricherEventProcessor {
     final contexts = event.contexts.copyWith(
       device: _getDevice(event.contexts.device),
       culture: _getSentryCulture(event.contexts.culture),
+      runtimes: _getRuntimes(event.contexts.runtimes),
```
denrase, line 33: We should also add this in  'html_enricher_event_processor.dart`. Just ran the example app in chrome and wondered why it was not added. Copied it over to test and it worked as expected.

dart/lib/src/event_processor/enricher/flutter_runtime.dart
denrase: Checkout out dart, there we do not have a build number. Here it makes sense to use the frameworkRevision as well as the channel. Maybe we could also add a rawDescription with more data if there is more useful info? Does that make sense?

Bildschirmfoto 2025-03-06 um 10 53 18

denrase (changes requested):
Nice! \o/ We just need to make sure to also add it in the html enricher, as well as make CI happy.

dart/lib/src/event_processor/enricher/flutter_runtime.dart
ueman: No, you're right. It should be `_missingRelevantFlutterVersionInfo` instead of `_hasFlutterRuntimeInformation`.

dart/lib/src/event_processor/enricher/web_enricher_event_processor.dart
ueman: Good catch!

ueman: Wait, I don't think that makes sense. The `html` enricher will never be used for the Flutter version which includes the Flutter runtime information, since the HTML renderer got removed.

denrase: We diod have a pr where we removed the html renderer, not sure if this is on the main branch or on v9. Depending on that, it would be ok to leave it out. I just had the issue that the new runtime info was not added when i ran the example app with your PR.

ueman: You need to switch to the master channel of Flutter to test this PR. 

The upcoming Flutter version has no HTML renderer, which means the the html enricher will never be executed when the new Flutter information is actually available. Same thing vice versa, if the HTML enricher gets executed, the Flutter version information is not available since it's an old Flutter version.

This is not dependent on the version of the Sentry SDK, but on the version of the Flutter SDK (but only indirectly, since we don't make use of a Flutter API).

denrase: Ah that makes sense, sorry for the confusion.

denrase (approved):
Lookin' good, thank you!

denrase:
@ueman The tests are failing on current dart versions and we have one analyser issue (formatting) left. Could we disable the tests for unsupported dart/flutter versions?

Bildschirmfoto 2025-03-10 um 10 16 50

ueman:
Ahh, good point. I think we can wrap the tests in a group and use `bool.hasEnvironment()` to enable the tests if the environment things are available.

Edit: Nevermind, those test will never work since they're executed in the Dart test runner which will never have those dart defines available, since they're only available in a Flutter environment. Soooo, maybe we should just delete the tests again?

denrase:
@ueman Ah got it, moving them to `sentry_flutter` will probably not work as this is not exposed publicly?

ueman:
> @ueman Ah got it, moving them to `sentry_flutter` will probably not work?

Good idea! I haven't thought about that, but I could imagine that working.

ueman:
@denrase It works 🥳  Everything that's red is unrelated to what I've done, I think.

denrase:
@ueman Cool, i'll check out your branch and see that I fix what's causing the CI to fail.

denrase:
@ueman Think the skip condition needs to be inverted. Also we can ignore failing e2e tests, since the token is missing in your branch, which is also why they fail.

denrase:
@ueman Would you be so kind and sync the main of your fork with sentry? Just landed a fix for CI issues. After that, I'll merge those in here and then we are good to go. 🙇‍♂️

denrase:
@buenaflor FYI,  the integration e2e test is not running because we do not have the token available in forks. Maybe we should consider skipping the test in this case?

buenaflor:
qq, is this important to have in v8? maybe we can just add it to v9 directly

denrase:
@buenaflor This is not breaking anything, just adding additional info about Flutter runtime. Don't see a reason why this could not be merged directly to main. WDYT?

buenaflor:
@denrase nothing really wrong with the PR and adding it to v8 but I'm asking because I was thinking of making a `8.14.0` (non-beta) release but I guess we'd wanna do another beta now if we merge this. or we just include this into the `8.14.0` without a beta, not sure, let's talk about it in the sync

ueman:
FWIW, I don't think the changes in this PR need a beta release

dart/lib/src/event_processor/enricher/io_enricher_event_processor.dart
```diff
@@ -63,12 +64,15 @@ class IoEnricherEventProcessor implements EnricherEventProcessor {
       version: _dartVersion,
       rawDescription: Platform.version,
     );
+    final flRuntime = flutterRuntime;
+
     if (runtimes == null) {
-      return [dartRuntime];
+      return [dartRuntime, if (flRuntime != null) flRuntime];
```
buenaflor, line 70: should we replace the previous `dartRuntime` impl here?

ueman: Nope, the previously existing one is better but not available everywhere.

buenaflor (approved):

buenaflor:
I'll merge this unless there's anything else to be added @denrase @ueman

ueman:
Please go ahead, there's nothing to add from my side

PR #2847, opened 2025-04-08
Set HTTP client breadcrumbs log level based on response status code

### LINKED ISSUE #2689: Set HTTP client breadcrumbs log level based on response status code
### Description

Meta issue: https://github.com/getsentry/team-sdks/issues/100

buenaflor:
https://github.com/getsentry/team-sdks/issues/100

Closes https://github.com/getsentry/sentry-dart/issues/2689

Note: the package-analysis CI failures are currently expected


dart/lib/src/utils/breadcrumb_log_level.dart
```diff
@@ -0,0 +1,16 @@
+import 'package:meta/meta.dart';
+
+import '../../sentry.dart';
+
+/// Determine a breadcrumb's log level (only `warning` or `error`) based on an HTTP status code.
+@internal
+SentryLevel? getBreadcrumbLogLevelFromHttpStatusCode(int statusCode) {
+  // NOTE: null defaults to 'info' in Sentry
```
philipphofmann, line 8: Thanks for the comment.

dart/lib/src/utils/breadcrumb_log_level.dart
```diff
@@ -0,0 +1,16 @@
...
+  if (statusCode >= 400 && statusCode < 500) {
+    return SentryLevel.warning;
+  } else if (statusCode >= 500) {
```
philipphofmann: `m`: I would also add the upper bound for 5xx. There are weird non standard HTTP status codes 599. Please also add a test for that.
```suggestion
  } else if (statusCode >= 500 && statusCode < 600) {
```

philipphofmann (approved):
LGTM, thanks for adding this.

dart/lib/src/utils/breadcrumb_log_level.dart
buenaflor: gotcha, thanks!

PR #2885, opened 2025-04-23
chore: add Gemfile for metrics and lockfile for iOS

philprime:
While refreshing the Apple code signing certificates and profiles I noticed that fastlane is set up at `/metrics/fastlane` but there is not `/metrics/Gemfile`, while there is one for `/flutter/example/ios/Gemfile` implicating the usage of `bundle exec fastlane ...`.

AFAIK the Gemfile.lock should also be checked into version control, because you want to ensure that everyone is using the exact same versions of tooling. As I am not an active maintainer for sentry-dart I'll leave the decision to you if you want them checked in.

#skip-changelog


buenaflor (approved):
thanks!

PR #2911, opened 2025-05-06
refactor(export): add sentry transformer to `sentry_dio` exports

tsinis:
## :scroll: Description

Exports the `SentryTransformer` class from the `sentry_dio` package (`src/sentry_transformer.dart`).

This makes the transformer class available for users to import and utilize directly if needed for custom Dio client transformer configurations.

When integrating Sentry with Dio, the `addSentry()` method overwrites Dio's default transformer with Sentry's own wrapper - `SentryTransformer`. However, in some cases, there are performance benefits of Dio's `BackgroundTransformer` (that could be passed to the `SentryTransformer`'s `transformer` parameter), which processes large payloads (>50KB) in separate isolates. But since `SentryTransformer` is not exported, there is no easy way to do it.

- Verified that the package (`sentry_dio`) compiles successfully with the added export.
- Manually confirmed in a sample project that `SentryTransformer` can be successfully imported via `import 'package:sentry_dio/sentry_dio.dart';`.
- The need for this change was identified through performance analysis in an application experiencing UI freezes when decoding large (multi-MB) JSON responses after integrating `sentry_dio`.

## :pencil: Checklist

- [x] Tests are not needed for exports
- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps

Consider adding documentation explaining what `addSentry()` does under the hood and how it could be customized.

### DIFF
--- a/dio/lib/sentry_dio.dart
+++ b/dio/lib/sentry_dio.dart
@@ -5,3 +5,6 @@ export 'src/sentry_dio_extension.dart';
 // Export the processor in case people want to use it standalone.
 // Normally one doesn't need to use it directly.
 export 'src/dio_event_processor.dart';
+
+// Export the transformer in case people want to use it standalone.
+export 'src/sentry_transformer.dart';


ueman:
This is actually not needed, since you should configure dio first, and as a last step call `addSentry()`. Since Sentry does not override any user configuratioon, it then uses whatever transformer you configured dio to use.

See this 

https://github.com/getsentry/sentry-dart/blob/18b7c08dff4fe3f433650ea774bc5fce91f6dce0/dio/lib/src/sentry_dio_extension.dart#L94-L97

and this

https://github.com/getsentry/sentry-dart/blob/18b7c08dff4fe3f433650ea774bc5fce91f6dce0/dio/lib/src/sentry_dio_extension.dart#L100-L103

This also ensures that integrating Sentry with dio is as seamless as possible, since it's just adding or removing the `addSentry()` call.

So here's how you should use it:

```dart
final dio = Dio();
dio.transformer = ...;
dio.httpClientAdapter = ...;
dio.addSentry(); // Dio is used with whatever transformer is configured two lines above AND also with Sentry
```


Disclaimer: I'm not a Sentry employe, but I wrote the Sentry dio integration and was a former dio maintainer

tsinis:
Hey @ueman I see, yes - make sense, but this is still not that obvious, I will provide documentation update in a separate PR, thanks!

ueman:
It's already stated in the docs here:
https://github.com/getsentry/sentry-dart/blob/18b7c08dff4fe3f433650ea774bc5fce91f6dce0/dio/lib/src/sentry_dio_extension.dart#L17-L18

But yeah, I guess a more exhaustive explanation can not hurt

PR #2917, opened 2025-05-07
feat(web): load injected debug ids

buenaflor:
## :scroll: Description

Adds debug id support in Flutter Web

Symbolication via release/dist association is not stable enough, hard to use and confusing.
Using debug-ids makes it automatically much easier.

Manual test, unit test

## :pencil: Checklist

- [ ] All tests passing
- [ ] No breaking changes


flutter/lib/src/web/web_sentry_js_binding.dart
```diff
@@ -93,6 +93,77 @@ class WebSentryJsBinding implements SentryJsBinding {
...
+        debugIdKeys.length == _lastKeysCount) {
+      // Return a copy
+      return Map.unmodifiable(_cachedFilenameDebugIds!);
+    }
+    _lastKeysCount = debugIdKeys.length;
+
+    // 3) Build a map of filename -> debug_id and refresh cache
+    final options = _client?.getOptions();
+    final Map<String, String> filenameDebugIdMap = {};
+    for (final stackKey in debugIdKeys) {
```
denrase: Maybe we can extract this loop or its main parts into it's own function, as the function is quite long and has side effects like reading/writing to an ivar.

flutter/lib/src/web/web_sentry_js_binding.dart
```diff
@@ -93,6 +93,77 @@ class WebSentryJsBinding implements SentryJsBinding {
...
+    final stackParser = _stackParser();
+    if (stackParser == null) {
+      return null;
+    }
+
+    final debugIdKeys = debugIdMap.keys.toList();
+
+    // 2) Fast path: use cached results if available
+    if (_cachedFilenameDebugIds != null &&
+        debugIdKeys.length == _lastKeysCount) {
```
denrase: I'm assuming we are using the length to check whether the cache is invalidated. Is there a case where the length would be the same, but we'd still need to fetch updated values? I'm missing a bit of context here.

flutter/lib/src/web/web_sentry_js_binding.dart
```diff
@@ -93,6 +93,77 @@ class WebSentryJsBinding implements SentryJsBinding {
       return null;
     }
   }
+
+  Map<String, String>? _cachedFilenameDebugIds;
+  Map<String, String>? get cachedFilenameDebugIds => _cachedFilenameDebugIds;
+  int? _lastKeysCount;
+  Map<String, List<String>>? _parsedStackResults;
```
denrase: We could init this with `{}`, as it is immediately assigned when first used.

flutter/lib/src/web/web_sentry_js_binding.dart
denrase: Maybe we can just compare the debugIdKeys and the keys of the map, so we don't need the additional _lastKeysCount var.

flutter/lib/src/web/web_sentry_js_binding.dart
```diff
@@ -93,6 +93,77 @@ class WebSentryJsBinding implements SentryJsBinding {
...
+
+  @override
+  Map<String, String>? getFilenameToDebugIdMap() {
+    // 1) Read the debug-ID table once
+    final debugIdMap =
+        _globalThis['_sentryDebugIds'].dartify() as Map<dynamic, dynamic>?;
```
denrase: Can those keys be anything other than `string` type? If not, would Map<string, dynamic> also work?

flutter/test/web/web_sentry_js_binding_test.dart
```diff
@@ -0,0 +1,53 @@
...
+/// Most of the other tests that involve the binding live in sentry_web_test.dart.
+void main() {
+  group(WebSentryJsBinding, () {
+    late Fixture fixture;
+
+    setUp(() {
+      fixture = Fixture();
+    });
+
+    test(
```
denrase, line 24: Are there more tests that we need to add here? For example all the cases where we return early.

denrase (changes requested):

flutter/lib/src/web/web_sentry_js_binding.dart
buenaflor: unfortunately doesn't work, Dart won't parse this if I dont use dynamic for the keys

flutter/lib/src/web/web_sentry_js_binding.dart
buenaflor: > I'm assuming we are using the length to check whether the cache is invalidated.

yeah 

> Is there a case where the length would be the same, but we'd still need to fetch updated values? I'm missing a bit of context here.

This is the way javascript does it, I'm not sure either but I assume no since you have to manually inject the debug id after bundling. pretty sure it will never change and it's just small safe guard and to quickly validate that the cached one is correct

flutter/lib/src/web/web_sentry_js_binding.dart
buenaflor: I'll move it to a separate file

flutter/lib/src/web/web_sentry_js_binding.dart
buenaflor: > Maybe we can just compare the debugIdKeys and the keys of the map,

I don't know how long the debug map can be, I think it's more efficient to just quickly check the length

flutter/test/web/web_sentry_js_binding_test.dart
buenaflor: I added some more tests to `sentry_web_test`

flutter/lib/src/web/web_sentry_js_binding.dart
denrase: It would be enough to move it to a private internal method, just to break it up a bit.

buenaflor: I moved the parsing / building to the `debug-ids` file (similar to JS), so the instance variables (in this case now top level) are also in vicinity of where they are most related

denrase: I tried to update the code into something that's easier to understand for me. Feel free to take what makes sense.

- Replaced index based loop with for loop.
- Only hold debug ids we already parse instead of whole unparsed stack.
- Updated naming.

```dart
  final Map<String, String> _filenameToDebugIds = {};
  final Set<String> _debugIdsWithFilenames = {};

  int _lastKeysCount = 0;

  Map<String, String>? get filenameToDebugIds => _filenameToDebugIds;

  @override
  Map<String, String>? getFilenameToDebugIdMap() {
    final debugIdMap =
        _globalThis['_sentryDebugIds'].dartify() as Map<dynamic, dynamic>?;
    if (debugIdMap == null) {
      return null;
    }

    final stackParser = _stackParser();
    if (stackParser == null) {
      return null;
    }

    if (debugIdMap.keys.length != _lastKeysCount) {
      _buildFilenameToDebugIdMap(
        debugIdMap,
        stackParser,
        _options, 
      );
      _lastKeysCount = debugIdMap.keys.length;
    }
    
    return Map.unmodifiable(_filenameToDebugIds);
  }

  void _buildFilenameToDebugIdMap(
    Map<dynamic, dynamic> debugIdMap,
    JSFunction stackParser,
    JSObject? options,
  ) {
    for (final debugIdMapEntry in debugIdMap.entries) {
      final String stackKeyStr = debugIdMapEntry.key.toString();
      final String debugIdStr = debugIdMapEntry.value.toString();

      final debugIdHasCachedFilename = _debugIdsWithFilenames.contains(debugIdStr);

      if (!debugIdHasCachedFilename) {
        final parsedStack = stackParser
            .callAsFunction(options, stackKeyStr.toJS)
            .dartify() as List<dynamic>?;

        if (parsedStack == null) continue;

        for (final stackFrame in parsedStack) {
          final stackFrameMap = stackFrame as Map<dynamic, dynamic>;
          final filename = stackFrameMap['filename']?.toString();
          if (filename != null) {
            _filenameToDebugIds[filename] = debugIdStr;
            _debugIdsWithFilenames.add(debugIdStr);
            break;
          }
        }
      }
    }
  }
```

denrase (approved):
LGTM, just need to make sure CI is happy.

buenaflor:
This PR will stay open until v9 is releasde

flutter/lib/src/integrations/web_load_debug_images_integration.dart
```diff
@@ -0,0 +1,49 @@
...
+import '../sentry_flutter_options.dart';
+
+Integration<SentryFlutterOptions> createLoadDebugImagesIntegration(
+    SentryNativeBinding native) {
+  return LoadNativeDebugImagesIntegration(native);
+}
+
+/// Loads the debug id injected by Sentry tooling e.g Sentry Dart Plugin
+/// This is necessary for symbolication of minified js stacktraces via debug ids.
+class LoadNativeDebugImagesIntegration
```
denrase: We should rename this to `LoadWebDebugImagesIntegration` or similar if possible.

denrase (changes requested):

flutter/lib/src/integrations/web_load_debug_images_integration.dart
buenaflor: done

denrase (approved):

PR #2991, opened 2025-06-14
Report Flutter framework feature flags

ueman:
## :scroll: Description

There's a Flutter framework PR (~https://github.com/flutter/flutter/pull/168437~, https://github.com/flutter/flutter/pull/171545 (reland of the first PR)) that adds feature flags to Flutter itself.
This PR supports that

I want know whether a specific build used one of the feature flags.

Automated tests

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps

I'll keep this a draft until the linked PR is merged for some time. Afterwards this PR can be landed.


flutter/test/integrations/flutter_framework_feature_flag_integration_test.dart
```diff
@@ -0,0 +1,28 @@
...
+          true);
+    });
+
+    test('adds feature flags', () {
+      final options = defaultTestOptions();
+      FlutterFrameworkFeatureFlagIntegration(flags: 'foo,bar,baz')
+          .call(MockHub(), options);
+
+      // TODO what expect to write here?
+    });
```
ueman: @denrase How am supposed to write this test to make sure a feature flag was added via `Sentry.addFeatureFlag(...)`? It's not a method on the hub, which makes that very annoying to test

flutter/lib/src/integrations/flutter_framework_feature_flag_integration.dart
```diff
@@ -0,0 +1,42 @@
...
+  });
+
+  @override
+  FutureOr<void> call(Hub hub, SentryOptions options) {
+    final enabledFeatureFlags = flags.split(',');
+
+    for (final featureFlag in enabledFeatureFlags) {
+      Sentry.addFeatureFlag(featureFlag, true);
+    }
+    options.sdk.addIntegration('FlutterFrameworkFeatureFlagIntegration');
```
buenaflor:
```suggestion
    options.sdk.addIntegration('FlutterFrameworkFeatureFlag');
```

I wanna omit the integration from now in the sdk integration string, I'll also update other integrations to match this

flutter/test/integrations/flutter_framework_feature_flag_integration_test.dart
buenaflor: can't remember why this wasn't added to the hub (cc @denrase)

but ig we can write integration tests instead like the remote config feature flag test

flutter/lib/src/integrations/flutter_framework_feature_flag_integration.dart
ueman: I guess it makes sense to align the naming scheme then too. Right now it's not always `CamelCase` I believe.

flutter/test/integrations/flutter_framework_feature_flag_integration_test.dart
```diff
@@ -0,0 +1,28 @@
...
+void main() {
+  group(FlutterFrameworkFeatureFlagIntegration, () {
+    test('adds sdk integration', () {
+      final options = defaultTestOptions();
+      FlutterFrameworkFeatureFlagIntegration(flags: 'foo,bar,baz')
+          .call(MockHub(), options);
+
+      expect(
+          options.sdk.integrations
+              .contains('FlutterFrameworkFeatureFlagIntegration'),
```
ueman:
```suggestion
              .contains('FlutterFrameworkFeatureFlag'),
```

flutter/lib/src/integrations/flutter_framework_feature_flag_integration.dart
buenaflor: yep I agree, not so urgent because these are mainly used for our internal dashboards but I'll see to it that we'll align here

flutter/lib/src/integrations/flutter_framework_feature_flag_integration.dart
```diff
@@ -0,0 +1,42 @@
...
+  FlutterFrameworkFeatureFlagIntegration({
+    @visibleForTesting this.flags = const String.fromEnvironment(_featureFlag),
+  });
+
+  @override
+  FutureOr<void> call(Hub hub, SentryOptions options) {
+    final enabledFeatureFlags = flags.split(',');
+
+    for (final featureFlag in enabledFeatureFlags) {
+      Sentry.addFeatureFlag(featureFlag, true);
```
ueman: Is there a way to indicate the source of a feature flag? I want to make sure that people know that these flags come from the Flutter framework, and not from Firebase for example.

buenaflor: hm good question, I don't think there are mechanisms ootb to categorize them. 

maybe we can do this ourselves by prefixing them?

```
firebase:feature_flag1: true
firebase:feature_flag2: true
flutter:feature_flag: false

or something like that
```

but I'll ask internally to check if we have anything planned for this

ueman: That's actually a good idea with the prefix. Do those prefixes have an impact on querying?

buenaflor: Asked internally and prefixing would be the only way to go right now

> Do those prefixes have an impact on querying

You'd need to query with the prefix `flags[firebase:feature_flag1]` other than that I dont think it has any other impact

CHANGELOG.md
```diff
@@ -78,7 +80,7 @@ void initState() {
 ```
 - Add `message` parameter to `captureException()` ([#2882](https://github.com/getsentry/sentry-dart/pull/2882))
 - Add module in SentryStackFrame ([#2931](https://github.com/getsentry/sentry-dart/pull/2931))
-  - Set `SentryOptions.includeModuleInStackTrace = true` to enable this. This may change grouping of exceptions.
+  - Set `SentryOptions.includeModuleInStackTrace = true` to enable this. This may change grouping of exceptions.
```
ueman, line 83:
```suggestion
  - Set `SentryOptions.includeModuleInStackTrace = true` to enable this. This may change grouping of exceptions.
```

flutter/lib/src/integrations/flutter_framework_feature_flag_integration.dart
buenaflor: Do you want to add the prefixes now? 

Since this will enabled by default it's possible users might be confused where these feature flags are coming from

ueman: Just added it ✅

ueman:
I would still wait with merging this PR until the linked PR is merged

flutter/lib/src/integrations/flutter_framework_feature_flag_integration.dart
```diff
@@ -0,0 +1,42 @@
...
+
+const _featureFlag = 'FLUTTER_ENABLED_FEATURE_FLAGS';
+
+// The Flutter framework feature flag works like this:
+// An enabled (experimental) feature gets added to the `FLUTTER_ENABLED_FEATURE_FLAGS`
+// dart define. Being in there means the feature is enabled, the feature is disabled
+// if it's not in there.
+// As a result, we also don't know the whole list of flags, but only the active ones.
+//
+// See https://github.com/flutter/flutter/pull/168437
```
loic-sharma: Hello, I'm sure you're aware but this reaches into Flutter internals that aren't meant to be public. This isn't covered by Flutter's breaking change policy. I'd assume the format of `FLUTTER_ENABLED_FEATURE_FLAGS` can and will change, even in a patch release. The current changes seem reasonable, but I'd consider documenting this so that future contributors are aware as well.

ueman:
```suggestion
// The Flutter framework feature flag works like this:
// An enabled (experimental) feature gets added to the `FLUTTER_ENABLED_FEATURE_FLAGS`
// dart define. Being in there means the feature is enabled, the feature is disabled
// if it's not in there.
// As a result, we also don't know the whole list of flags, but only the active ones.
//
// See
// - https://github.com/flutter/flutter/pull/168437
// - https://github.com/flutter/flutter/pull/171545
//
// The Flutter feature flag implementation is not meant to be public and can change in a patch release.
// See this discussion https://github.com/getsentry/sentry-dart/pull/2991/files#r2183105202
```

ueman:
@buenaflor The dependent PR in the Flutter framework has landed, so I would say we can go ahead and merge this PR

buenaflor:
@ueman sounds good

also does the cursor review make sense? if not we can just ignore it

ueman:
I assume those cases should not happen since the flags are supplied by the flutter tool and not user generated, but I added code and tests for them anyway

buenaflor (approved):

PR #3011, opened 2025-06-25
add options to upload more attachments when process android native crash

arc-yong:
## :pencil: Checklist

- [ ] All tests passing
- [ ] No breaking changes

### DIFF
--- a/flutter/android/src/main/kotlin/io/sentry/flutter/AttachLogcatEventProcess.java
+++ b/flutter/android/src/main/kotlin/io/sentry/flutter/AttachLogcatEventProcess.java
@@ -1,0 +1,59 @@
+package io.sentry.flutter;
+import android.content.Context;
+
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+import java.io.BufferedReader;
+import java.io.File;
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.io.InputStreamReader;
+import java.nio.charset.StandardCharsets;
+
+import io.sentry.Attachment;
+import io.sentry.EventProcessor;
+import io.sentry.Hint;
+import io.sentry.SentryEvent;
+
+public class AttachLogcatEventProcess implements EventProcessor {
+    private String getLogcatOutput() {
+        StringBuilder logOutput = new StringBuilder();
+        Process process = null;
+        try {
+            // get most recent 1000 lines of logs and filter warnings.
+            process = Runtime.getRuntime().exec("logcat -d -t 1000 *:W");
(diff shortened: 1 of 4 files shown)


buenaflor (commented):
hey any specific reason why [AttachLogcatEventProcess.java](https://github.com/getsentry/sentry-dart/pull/3011/files#diff-8d4c99c0ce9fa8a010cc9ecb23c11973b675737276763b54a3fed9aacbd5f651) is in Java and not Kotlin?

buenaflor:
@romtsn @markushi can one of y'all have a look here? any thoughts?

arc-yong:
> hey any specific reason why [AttachLogcatEventProcess.java](https://github.com/getsentry/sentry-dart/pull/3011/files#diff-8d4c99c0ce9fa8a010cc9ecb23c11973b675737276763b54a3fed9aacbd5f651) is in Java and not Kotlin?

I will convert it to Kotlin.

romtsn:
@buenaflor I know dotnet (or unity) also provides an integration for this, so I'm thinking if we should rather do this on our side and expose options for the hybrids to enable the integration/processor (make it opt-in by default). Seems like there's demand

bitsandfoxes:
> I know dotnet (or unity) also provides an integration for this

Yes, the .NET SDK does have a [`LogCatAttachmentEventProcessor`](https://github.com/getsentry/sentry-dotnet/blob/dc2b3261c8e475a99f8891bd4e26e4d65ba81e9f/src/Sentry/Platforms/Android/LogCatAttachmentEventProcessor.cs). But that's for the managed layer to include those logs. To make this work from the Android SDK side, we'd need to have mechanisms to sync things both ways.

buenaflor:
> if we should rather do this on our side and expose options for the hybrids to enable the integration/processor (make it opt-in by default)

yeah sounds good, should we close this then and create an issue on sentry-java?

romtsn:
@buenaflor oh you also changed to using `InternalSentrySDK` right? I think that'd be more complicated then, because your events now wouldn't go through our event processors I think (they'd just be send to the transport directly as envelopes).

buenaflor:
> events now wouldn't go through our event processors I think

that is true for Dart events

However I think the purpsoe of this PR is to attach logcat output to native android crashes which still go through the normal android event processors

--- 

Dart events don't go through the native event processors

maybe we can find a middle ground that works for both native android events and dart events.

for Dart events what could work is if Android could expose an api where we can fetch the logcat output, bring them into the Dart layer and then attach it to the event

romtsn:
> However I think the purpsoe of this PR is to attach logcat output to native android crashes which still go through the normal android event processors

ah ok, I see, yeah then we should definitely do that in the Android SDK and just expose an option to enable it or something.

buenaflor:
Closing this then in favour of doing this on the Android SDK side

PR #3037, opened 2025-07-03
fix(replay): dont set `isTerminating` on captureReplay for android

buenaflor:

Fixes a bug where we set the `isTerminating` parameter on `captureReplay` to true if an unhandled error happens.

This is a bug because Android replay will stop capturing segments as it expects the app to terminate - so for flutter we potentially run into missing segments

## :pencil: Checklist

- [ ] All tests passing
- [ ] No breaking changes


flutter/test/replay/replay_event_processor_test.dart
```diff
@@ -17,17 +17,6 @@ void main() {
   });
 
   for (var isHandled in [true, false]) {
```
denrase: Think we don't need the loop anymore if we don't care about the isHandled value, no?

flutter/test/replay/replay_event_processor_test.dart
```diff
@@ -17,17 +17,6 @@ void main() {
...
-      final event = await fixture.apply(isHandled: isHandled);
-      bool isCrash = verify(fixture.binding.captureReplay(captureAny))
-          .captured
-          .single as bool;
-      expect(isCrash, !isHandled);
-      expect(event, isNotNull);
-    });
-
     test(
         'sets scope replay ID for ${isHandled ? 'handled' : 'unhandled'} exceptions',
```
denrase: We can remove the isHandled parameter/mechanism from the fixture setup.

flutter/test/sentry_native_channel_test.dart
```diff
@@ -369,9 +369,8 @@ void main() {
         when(channel.invokeMethod('captureReplay', any))
             .thenAnswer((_) => Future.value(sentryId.toString()));
 
-        final returnedId = await sut.captureReplay(true);
+        final returnedId = await sut.captureReplay();
 
-        verify(channel.invokeMethod('captureReplay', {'isCrash': true}));
```
denrase, line 374: We still need to verify that the method was called, just without the isCrash param.

denrase (changes requested):

buenaflor:
@denrase updated it 👍

flutter/android/src/main/kotlin/io/sentry/flutter/SentryFlutterPlugin.kt
```diff
@@ -619,14 +619,9 @@ class SentryFlutterPlugin :
...
   private fun captureReplay(
-    isCrash: Boolean?,
     result: Result,
   ) {
-    if (isCrash == null) {
-      result.error("5", "Arguments are null", null)
-      return
-    }
-    replay!!.captureReplay(isCrash)
+    replay!!.captureReplay(isTerminating = false)
```
denrase, line 624: Just a small thin, but maybe we could safley unpack the otpional replay. If for some reason this becomes nil, we will crash here.

denrase (approved):
Just a small suggestion.

flutter/android/src/main/kotlin/io/sentry/flutter/SentryFlutterPlugin.kt
buenaflor: if an error happens in the plugin it doesnt crash, Flutter will still catch it and propagate it to the flutter layer (where we catch it with the `SentryNativeSafeInvoker`

image

PR #3039, opened 2025-07-03
Update proguard rules to keep `io.sentry.**`

buenaflor:
Update rules to keep all `io.sentry` classes instead of only `io.sentry.android.replay`

Tested the change locally

### DIFF
--- a/flutter/android/proguard-rules.pro
+++ b/flutter/android/proguard-rules.pro
@@ -1,7 +1,5 @@
 -keep class io.sentry.flutter.** { *; }
-
-# Keep replay integration classes used by JNI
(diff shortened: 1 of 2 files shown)


flutter/android/proguard-rules.pro
```diff
@@ -1,7 +1,5 @@
 -keep class io.sentry.flutter.** { *; }
-
-# Keep replay integration classes used by JNI
--keep class io.sentry.android.replay.** { *; }
+-keep class io.sentry.** { *; }
```
vaind: these could be merged

vaind (approved):

vaind (approved):

Issue #3058, opened 2025-07-10
Critical bug in abortTransaction method: Database operations blocked when no parent span exists


ichiko:
### Platform

Dart

### Obfuscation

Disabled

### Debug Info

Disabled

### Doctor

[✓] Flutter (Channel stable, 3.29.2, on macOS 14.6.1 23G93 darwin-arm64, locale ja-JP) [4.4s]
    • Flutter version 3.29.2 on channel stable at /Users/ichiko-moro/fvm/versions/3.29.2
    • Upstream repository https://github.com/flutter/flutter.git
    • Framework revision c236373904 (4 months ago), 2025-03-13 16:17:06 -0400
    • Engine revision 18b71d647a
    • Dart version 3.7.2
    • DevTools version 2.42.3

[!] Android toolchain - develop for Android devices (Android SDK version 35.0.0) [2.9s]
    • Android SDK at /Users/ichiko-moro/Library/Android/sdk
    ✗ cmdline-tools component is missing
      Run `path/to/sdkmanager --install "cmdline-tools;latest"`
      See https://developer.android.com/studio/command-line for more details.
    ✗ Android license status unknown.
      Run `flutter doctor --android-licenses` to accept the SDK licenses.
      See https://flutter.dev/to/macos-android-setup for more details.

[✓] Xcode - develop for iOS and macOS (Xcode 16.2) [10.1s]
    • Xcode at /Applications/Xcode-16.2.0.app/Contents/Developer
    • Build 16C5032a
    • CocoaPods version 1.16.2

[✓] Chrome - develop for the web [9ms]
    • Chrome at /Applications/Google Chrome.app/Contents/MacOS/Google Chrome

[✓] Android Studio (version 2024.1) [9ms]
    • Android Studio at /Applications/Android Studio.app/Contents
    • Flutter plugin can be installed from:
      🔨 https://plugins.jetbrains.com/plugin/9212-flutter
    • Dart plugin can be installed from:
      🔨 https://plugins.jetbrains.com/plugin/6351-dart
    • Java version openjdk version "17.0.11" 2024-04-16

[✓] Connected device (6 available) [7.1s]
    • Pixel 6 (mobile)                    • 1A151FDF6002G9                       • android-arm64  • Android 14 (API 34)
    • revphone (wireless) (mobile)        • 00008110-000628C82130401E            • ios            • iOS 18.5 22F76
[...]

PR #3062, opened 2025-07-10
fix: abortTransaction does not call provided execute function when parent span is null

ichiko:
## :scroll: Description

This PR fixes a bug in the abortTransaction method where database operations were blocked when no parent span exists. The issue was reported in #3058.

When parentSpan is null in the abortTransaction method (lines 147-153), the method was logging a warning and returning `Future<T>.value()` without executing the provided execute function. This meant the actual database transaction abort operation was never performed, leaving the database in an inconsistent state where:

- Database transactions remained open indefinitely
- Subsequent database operations could be blocked
- Application could hang or fail on subsequent DB access attempts
- Data integrity issues could occur

Added comprehensive tests to verify the fix.

- abortTransaction executes function even when parentSpan is null - Verifies that the execute function is called even when no parent span exists
- abortTransaction with parentSpan executes function and marks span as aborted - Verifies normal behavior when parent span exists

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes


buenaflor:
Thank you for the investigation and bug fix! I'll have a look later today (sry, accidentally clicked the close button)

drift/test/sentry_drift_test.dart
```diff
@@ -603,6 +603,62 @@ void main() {
...
+        return 'test_result';
+      });
+
+      expect(exceptionThrown, true);
+      expect(executeFunctionCalled, true);
+      expect(result, 'test_result');
+
+      expect(sut.spanHelper.transactionStack, isEmpty);
+
+      final abortedSpans =
```
buenaflor: I would replace this test with something like: 

```dart
test('failed Drift transaction is rolled back', () async {
  final db = AppDatabase(
    NativeDatabase.memory().interceptWith(
      SentryQueryInterceptor(databaseName: 'db'),
    ),
  );

  // pre-condition: table empty
  expect(await db.select(db.todoItems).get(), isEmpty);

  // run a transaction that is forced to fail -> should be rolled back
  expect(
    () => db.transaction(() async {
      await _insertRow(db, withError: true);
    }),
    throwsA(isA<Exception>()),
  );

  // if rollback happened the row must be absent
  expect(await db.select(db.todoItems).get(), isEmpty);
});
```

this test times out if the span helper still had the bug therefore indicating that the rollback didn't succeed

buenaflor (commented):
Looks good, I would only change one test case

drift/test/sentry_drift_test.dart
ichiko: Thank you for the great suggestion! I completely agree with your approach. 
So i rewrite tests in [c177342](https://github.com/getsentry/sentry-dart/pull/3062/commits/c1773426db4a5fc0869727e99bcdaaf52fda42c7).

buenaflor:
@ichiko last task is to update the changelog:

```md
### Fixes

- Drift transaction rollback not executed when parent span is null ([#3062](https://github.com/getsentry/sentry-dart/pull/3062))
```

buenaflor (approved):

PR #3088, opened 2025-07-17
Use maybeOf instead of try catch block

ueman:
This is better practice since no errors are thrown

#skip-changelog


buenaflor (approved):
Thx! 🚀

PR #3263, opened 2025-09-19
chore(changelog): Update CHANGELOG for Android 15 page size support

### LINKED ISSUE #3256: 16KB Page Size support
### Problem Statement

I recently reviewed my Flutter project for Android’s new 16KB page size requirement. Most of the native libraries in my APK come from Flutter itself (libflutter.so, libapp.so), but I also noticed that the Sentry SDK ships with its own native libraries (libsentry.so, libsentry-android.so).

Since these .so files need to be compiled with 16KB page size support to avoid runtime issues on newer devices, I wanted to check whether the Sentry Flutter/Dart SDK already ensures this, or if it’s something that still needs to be updated.

Thanks.

### Solution Brainstorm

It would be super helpful if:
Sentry confirms that the bundled native libraries already support 16KB pages, or
Updates the build process to ensure compatibility, and
Mentions this in the release notes/docs so developers know they’re covered.

Thanks.

None

romtsn:
_#skip-changelog_

Closes #3256 

## :pencil: Checklist

- [ ] All tests passing
- [ ] No breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1108,6 +1108,8 @@ This release fixes an issue where Cold starts can be incorrectly reported as War
 
 ### Features
 
+- Android 15: Add support for 16KB page sizes ([#3620](https://github.com/getsentry/sentry-java/pull/3620))
+  - See https://developer.android.com/guide/practices/page-sizes for more details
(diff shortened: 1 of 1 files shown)


kahest (approved):
thanks for adding!

PR #3324, opened 2025-11-06
enh: refactor `init` to use JNI/FFI

buenaflor:
## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes


packages/flutter/android/src/main/kotlin/io/sentry/flutter/SentryFlutterPlugin.kt
```diff
@@ -197,10 +202,6 @@ class SentryFlutterPlugin :
     @Suppress("unused", "ReturnCount", "TooGenericExceptionCaught") // Used by native/jni bindings
     @JvmStatic
     fun fetchNativeAppStartAsBytes(): ByteArray? {
-      if (!sentryFlutter.autoPerformanceTracingEnabled) {
-        return null
-      }
-
```
buenaflor, line 203: removing it here, we have this check in the Dart side

packages/flutter/example/lib/main.dart
```diff
@@ -87,7 +92,7 @@ Future<void> setupSentry(
       options.maxRequestBodySize = MaxRequestBodySize.always;
       options.navigatorKey = navigatorKey;
 
-      options.replay.sessionSampleRate = 0.0;
+      options.replay.sessionSampleRate = 1.0;
```
buenaflor, line 95: it was `1.0` before, it was accidentally changed during the log replay buffering tags impl

packages/flutter/lib/src/native/cocoa/sentry_native_cocoa.dart
```diff
@@ -321,7 +302,7 @@ class SentryNativeCocoa extends SentryNativeChannel {
 // The default conversion does not handle bool so we will add it ourselves
 final ObjCObjectBase Function(Object) _defaultObjcConverter = (obj) {
   return switch (obj) {
-    bool b => b ? 1.toNSNumber() : 0.toNSNumber(),
+    bool b => NSNumberCreation.numberWithBool(b),
```
buenaflor, line 297: previous convert was wrong, it should've been using `NSNumberCreation`

packages/flutter/ios/sentry_flutter/Sources/sentry_flutter/SentryFlutterPlugin.swift
```diff
@@ -266,6 +108,202 @@ public class SentryFlutterPlugin: NSObject, FlutterPlugin {
...
+        "nativeSdkVersion": PrivateSentrySDKOnly.getSdkVersionString()
+      ]
+      for (key, value) in tags {
+        newTags[key] = value
+      }
+      PrivateSentrySDKOnly.setReplayTags(newTags)
+    #endif
+  }
+
+  @objc(setBeforeSend:packages:integrations:)
```
denrase, line 132: Don't think  we need different naming for objc, as this will never be called from Swift directly.

packages/flutter/ios/sentry_flutter/Sources/sentry_flutter/SentryFlutterPlugin.swift
```diff
@@ -266,6 +108,202 @@ public class SentryFlutterPlugin: NSObject, FlutterPlugin {
...
+        }
+        event.sdk = sdk
+      }
+
+      return event
+    }
+  }
+
+  @objc(setAutoPerformanceFeatures:)
+  public class func setAutoPerformanceFeatures(enableAutoPerformanceTracing: Bool) {
```
denrase: If we only do something if `enableAutoPerformanceTracing` is true, we might as well just remove it and just handle this on the call side with an if.

packages/flutter/ios/sentry_flutter/Sources/sentry_flutter/SentryFlutterPlugin.swift
```diff
@@ -266,6 +108,202 @@ public class SentryFlutterPlugin: NSObject, FlutterPlugin {
...
+    // We send a SentryHybridSdkDidBecomeActive to the Sentry Cocoa SDK, to mimic
+    // the didBecomeActiveNotification notification. This is needed for session, OOM tracking, replays, etc.
+    if appIsActive {
+      NotificationCenter.default.post(name: Notification.Name("SentryHybridSdkDidBecomeActive"), object: nil)
+    }
+  }
+
+  @objc(setSdkMetaData:packages:integrations:)
+  public class func setSdkMetaData(event: Event, packages: [[String: String]], integrations: [String]) {
+    if var sdk = event.sdk, self.isValidSdk(sdk: sdk) {
```
denrase, line 165: This code looks similar/the same as above (beforeSend). Can we re-use it?

packages/flutter/ios/sentry_flutter/Sources/sentry_flutter/SentryFlutterPlugin.swift
```diff
@@ -266,6 +108,202 @@ public class SentryFlutterPlugin: NSObject, FlutterPlugin {
...
+    return !name.isEmpty
+  }
+
+  @objc(setProxyOptions:user:pass:host:port:type:)
+  public class func setProxyOptions(
+    options: Options,
+    user: String?,
+    pass: String?,
+    host: String,
+    port: String,
```
denrase: `port` should be a number representation. we cast it to string on the dart side and cast it back here.

packages/flutter/ios/sentry_flutter/Sources/sentry_flutter_objc/SentryFlutterPlugin.h
```diff
@@ -3,11 +3,44 @@
 #if __has_include(<sentry_flutter/sentry_flutter-Swift.h>)
 #import <sentry_flutter/sentry_flutter-Swift.h>
 #else
+typedef void (^SentryReplayCaptureCallback)(
```
denrase: Do we need this additional typedef? Isn't it the same as the other one?

packages/flutter/ios/sentry_flutter/Sources/sentry_flutter_objc/SentryFlutterReplayScreenshotProvider.m
```diff
@@ -9,13 +9,12 @@
...
-  FlutterMethodChannel *channel;
+  SentryReplayCaptureCallback callback;
 }
 
-- (instancetype _Nonnull)initWithChannel:
-    (FlutterMethodChannel *_Nonnull)channel {
+- (instancetype _Nonnull)initWithCallback:(SentryReplayCaptureCallback)callbackParam {
   if (self = [super init]) {
-    self->channel = channel;
+    self->callback = [callbackParam copy];
```
denrase, line 17: I'm not this works if we copy the callback.

packages/flutter/ios/sentry_flutter/Sources/sentry_flutter_objc/SentryFlutterReplayScreenshotProvider.m
```diff
@@ -9,13 +9,12 @@
 #import <Flutter/Flutter.h>
 
 @implementation SentryFlutterReplayScreenshotProvider {
-  FlutterMethodChannel *channel;
+  SentryReplayCaptureCallback callback;
```
denrase, line 12: I'm not sure here, but please check if we need to pass this by pointer reference `SentryReplayCaptureCallback *callback`

packages/flutter/ios/sentry_flutter/Sources/sentry_flutter_objc/SentryFlutterReplayScreenshotProvider.m
```diff
@@ -28,74 +27,73 @@ - (void)imageWithView:(UIView *_Nonnull)view
...
+    NSLog(@"SentryFlutterReplayRecorderFFI has no callback. "
+          @"Cannot capture a replay screenshot.");
+    return;
+  }
+
+  self->callback(replayId, replayIsBuffering, ^(id value) {
+    if (value == nil) {
+      NSLog(@"SentryFlutterReplayRecorderFFI received null result. "
+            @"Cannot capture a replay screenshot.");
+    } else if ([value isKindOfClass:[NSDictionary class]]) {
```
denrase, line 41: We can have two params with explicit types of `NSDictionary *` and `FlutterError`. But see this as optional if you just want to port this 1:1 for now.

packages/flutter/lib/src/native/cocoa/sentry_native_cocoa.dart
```diff
@@ -30,44 +35,20 @@ class SentryNativeCocoa extends SentryNativeChannel {
   @visibleForTesting
   CocoaReplayRecorder? get testRecorder => _replayRecorder;
 
+  @visibleForTesting
+  cocoa.SentryOptions? get testNativeOptions =>
```
denrase: Do we even need this if we can just as well call `cocoa.PrivateSentrySDKOnly.getOptions();` statically in options? It's not bound to this instance.

packages/flutter/lib/src/native/cocoa/sentry_native_cocoa.dart
```diff
@@ -30,44 +35,20 @@ class SentryNativeCocoa extends SentryNativeChannel {
...
+      cocoa.PrivateSentrySDKOnly.getOptions();
+
+  @visibleForTesting
+  cocoa.SentryReplayOptions? get testNativeReplayOptions =>
```
denrase: Same here, no need for this helper getter.

packages/flutter/lib/src/native/cocoa/sentry_native_cocoa.dart
```diff
@@ -331,13 +312,29 @@ NSDictionary _dartToNSDictionary(Map<String, dynamic> json) {
...
 }
 
+List<Object> _deepConvertListNonNull(List<dynamic> list) => [
+      for (final e in list)
+        if (e case Map<String, dynamic> m)
+          _deepConvertMapNonNull(m)
+        else if (e case List<dynamic> l)
+          _deepConvertListNonNull(l)
+        else if (e case Object o)
+          o,
```
denrase, line 327: Don't we need `toObjCObject` here?

packages/flutter/lib/src/native/cocoa/sentry_native_cocoa.dart
```diff
@@ -349,13 +346,7 @@ Map<Object, Object> _deepConvertMapNonNull(Map<String, dynamic> input) {
...
       Map<String, dynamic> m => _deepConvertMapNonNull(m),
-      List<dynamic> l => [
-          for (final e in l)
-            if (e != null)
-              e is Map<String, dynamic>
-                  ? _deepConvertMapNonNull(e)
-                  : e as Object
-        ],
+      List<dynamic> l => _deepConvertListNonNull(l),
       _ => value as Object,
```
denrase, line 342: Also here, `toObjCObject` not needed?

packages/flutter/lib/src/native/cocoa/sentry_native_cocoa_init.dart
```diff
@@ -0,0 +1,176 @@
...
+    quality: options.replay.quality.level,
+    sessionSampleRate: options.replay.sessionSampleRate ?? 0,
+    onErrorSampleRate: options.replay.onErrorSampleRate ?? 0,
+    sdkName: options.sdk.name.toNSString(),
+    sdkVersion: options.sdk.version.toNSString(),
+  );
+
+  if (options.proxy != null) {
+    final host = options.proxy!.host?.toNSString();
+    final port = options.proxy!.port?.toString().toNSString();
```
denrase: We can just pass the int, if it is int.

packages/flutter/lib/src/native/cocoa/sentry_native_cocoa_init.dart
```diff
@@ -0,0 +1,176 @@
...
+}
+
+/// Builds the Replay capture callback that bridges native replay requests
+/// to the Dart-side screenshot recorder.
+cocoa.DartSentryReplayCaptureCallback createReplayCaptureCallback({
+  required SentryFlutterOptions options,
+  required Hub hub,
+  required SentryNativeCocoa owner,
+}) {
+  return cocoa.ObjCBlock_ffiVoid_NSString_bool_ffiVoidobjcObjCObject.listener(
```
denrase, line 154: `cocoa.ObjCBlock_ffiVoid_NSString_bool_ffiVoidobjcObjCObject.listener` wow, such readable generate names 😅

packages/flutter/ffi-cocoa.yaml
```diff
@@ -1,20 +1,32 @@
...
+    - '**/Sentry.framework/**.h'
+    - '**/SentryFlutterPlugin.h'
+
 compiler-opts:
   - -DSENTRY_TARGET_PROFILING_SUPPORTED=1
   - -DCOCOAPODS=1
-  - '-I./temp/Sentry.framework/Headers/'
-  - '-I./temp/Sentry.framework/PrivateHeaders/'
-  - '-I./ios/sentry_flutter/Sources/sentry_flutter_objc/'
+  - -DFFIGEN=1
```
denrase, line 21: What does this compiler option do?

packages/flutter/lib/src/replay/replay_quality.dart
```diff
@@ -2,12 +2,16 @@ import 'package:meta/meta.dart';
 
 /// The quality of the captured replay.
 enum SentryReplayQuality {
-  high(resolutionScalingFactor: 1.0),
-  medium(resolutionScalingFactor: 1.0),
-  low(resolutionScalingFactor: 0.8);
+  high(resolutionScalingFactor: 1.0, level: 2),
```
denrase: Why the addition of level?

denrase (changes requested):

packages/flutter/ios/sentry_flutter/Sources/sentry_flutter/SentryFlutterPlugin.swift
buenaflor: This is so it's more explicit how the name is generated for objc

without it it's `setBeforeSendWithOptions(...)` by default

this took a lot of time away from me because I couldn't figure out why the app was crashing and it turns out I should've used `setBeforeSendWithOptions` in `SentryFlutterPlugin.h` when creating the bindings instead of only `setBeforeSend`.

In the end I decided to add this so it's explicit how it's converted

I don't mind reverting it, though I think it would be helpful for the future to be more specific

packages/flutter/ios/sentry_flutter/Sources/sentry_flutter/SentryFlutterPlugin.swift
buenaflor: perhaps I can even port this to Dart directly instead of setting it in native

packages/flutter/lib/src/native/cocoa/sentry_native_cocoa.dart
buenaflor: no this is only to convert a `List<dynamic>` into a `List<Object>`, there is no objc conversion done here

buenaflor: this is so we can access the objective_c package extensions which aer only available for `List<Object>`

packages/flutter/lib/src/native/cocoa/sentry_native_cocoa.dart
buenaflor: same comment as above, this function doesn't do any objc conversions

packages/flutter/lib/src/sentry_privacy_options.dart
```diff
@@ -95,9 +95,9 @@ class SentryPrivacyOptions {
                   'widget comes from a third-party plugin or your code, Sentry '
                   "doesn't recognize it and can't reliably mask it in release "
                   'builds (due to obfuscation). '
-                  'Please mask it explicitly using options.privacy.mask<$type>(). '
+                  'Please mask it explicitly using mask<$type>(). '
                   'If you want to silence this warning and keep the widget '
-                  'visible in captures, you can use options.privacy.unmask<$type>(). '
+                  'visible in captures, you can use unmask<$type>(). '
```
buenaflor: oops, will revert this

packages/flutter/lib/src/native/cocoa/sentry_native_cocoa_init.dart
```diff
@@ -0,0 +1,176 @@
...
+    options.enableAutoPerformanceTracing,
+  );
+
+  final version = cocoa.PrivateSentrySDKOnly.getSdkVersionString();
+  cocoa.PrivateSentrySDKOnly.setSdkName(
+    cocoaSdkName.toNSString(),
+    andVersionString: version,
+  );
+
+  // Use native beforeSend to avoid crashes when capturing a native event.
```
buenaflor: will add another comment why this crashes

packages/flutter/ios/sentry_flutter/Sources/sentry_flutter/SentryFlutterPlugin.swift
buenaflor: yeah makes sense

packages/flutter/ios/sentry_flutter/Sources/sentry_flutter/SentryFlutterPlugin.swift
buenaflor: definitely, good catch

packages/flutter/lib/src/replay/replay_quality.dart
buenaflor: so we can map it directly to cocoa quality:

```
quality: options.replay.quality.level,
```

alternatively we could remove it and just do it on the call site

packages/flutter/lib/src/native/cocoa/sentry_native_cocoa.dart
buenaflor: true, let's do that instead

packages/flutter/ios/sentry_flutter/Sources/sentry_flutter_objc/SentryFlutterPlugin.h
buenaflor: yeah, there was trouble when trying to generate binding without adding it here as well. I can try again and see if it works

buenaflor: ok got it working

packages/flutter/ios/sentry_flutter/Sources/sentry_flutter_objc/SentryFlutterReplayScreenshotProvider.m
buenaflor: I'm not an expert in objc but according to cursor it's fine as it is because the typedef is already a block pointer with `^`

`typedef void (^SentryReplayCaptureCallback)`

packages/flutter/ios/sentry_flutter/Sources/sentry_flutter_objc/SentryFlutterReplayScreenshotProvider.m
buenaflor: apparently I have to copy to the heap via copy otherwise the pointer stays in the stack and will eventually crash if the block goes out of scope.

Also I have tested this and replays work well

packages/flutter/ffi-cocoa.yaml
buenaflor: See `ffigen_objc_imports.h`

```objc
#ifdef FFIGEN
// Only included while running ffigen (provide paths in compiler-opts).
#import "PrivateSentrySDKOnly.h"
#import "Sentry-Swift.h"
#import "SentryOptions.h"
#import "SentryScope.h"
#endif
```

if we dont have this it will complain when compiling the app. We only need these headers when we use ffigen

buenaflor:
I'm currently working on optimizing the JNI bindings so it only creates bindings for fields and methods we specify

Edit: we don't need to optimize generated bindings, the compiler will tree shake unused ones anyway so the result  is the same size diff

packages/flutter/ios/sentry_flutter/Sources/sentry_flutter/SentryFlutterPlugin.swift
buenaflor: Okay don't think it's feasible/worth the time to try and port this to Dart directly, I'll change it to number instead

.github/workflows/flutter_test.yml
```diff
@@ -146,19 +146,6 @@ jobs:
...
-        if: ${{ matrix.target != 'macos' }}
-        working-directory: packages/flutter/example/${{ matrix.target }}
-        # For some reason running native unit tests directly after Flutter integration tests fails
-        # running flutter build ios before works: https://stackoverflow.com/a/77487525/22813624
-        run: |
-          flutter build ios --no-codesign
-          xcodebuild test -workspace Runner.xcworkspace -scheme Runner -configuration Debug -destination "platform=$DEVICE_PLATFORM" -allowProvisioningUpdates CODE_SIGNING_ALLOWED=NO
-        env:
-          DEVICE_PLATFORM: ${{ steps.device.outputs.platform }}
-
```
buenaflor, line 161: we dont need those native tests anymore

packages/flutter/example/integration_test/integration_test.dart
```diff
@@ -877,69 +1079,69 @@ void main() {
...
-
-      final dist = tags.firstWhere((element) => element['key'] == 'dist');
-      expect('1', dist['value']);
-
-      final environment =
-          tags.firstWhere((element) => element['key'] == 'environment');
-      expect('integration', environment['value']);
-    });
-  });
+  // group('e2e', () {
```
denrase: Removing the e2e test here is intentional?

packages/flutter/example/integration_test/platform_integrations_test.dart
```diff
@@ -0,0 +1,349 @@
...
+          expect(names.contains('LoadContextsIntegration'), isTrue);
+          expect(names.contains('ReplayIntegration'), isTrue);
+          expect(names.contains('ReplayLogIntegration'), isTrue);
+        } else if (isIOS) {
+          expect(names.contains('LoadContextsIntegration'), isTrue);
+          expect(names.contains('ReplayIntegration'), isTrue);
+          expect(names.contains('ReplayLogIntegration'), isTrue);
+        } else if (isMacOS) {
+          expect(names.contains('LoadContextsIntegration'), isTrue);
+          // Replay not supported on macOS by default
```
denrase, line 177: Nice catch!

packages/flutter/ios/sentry_flutter/Sources/sentry_flutter_objc/SentryFlutterReplayScreenshotProvider.m
denrase: Ah yeah, this is not an object we are passing, but a block. Block syntax, always a banger: https://fuckingblocksyntax.com

packages/flutter/lib/src/replay/replay_quality.dart
denrase: If this is public API i'd rather leave as is and do this where needed. Don't need to expose a cocoa implementation detail here.

denrase (approved):
Just the `SentryReplayQualit`, think we should keep this as it was.

packages/flutter/example/integration_test/integration_test.dart
buenaflor: yup was trying to make it pass a bit faster, will smoothen it all out once the tests pass

packages/flutter/lib/src/native/java/sentry_native_java.dart
```diff
@@ -33,74 +35,20 @@ class SentryNativeJava extends SentryNativeChannel {
   @visibleForTesting
   AndroidReplayRecorder? get testRecorder => _replayRecorder;
 
+  @visibleForTesting
+  native.SentryAndroidOptions? get testNativeOptions {
+    // ignore: invalid_use_of_internal_member
+    final ref = native.ScopesAdapter.getInstance()?.getOptions().reference;
+    if (ref == null) return null;
+    return native.SentryAndroidOptions.fromReference(ref);
+  }
```
buenaflor: I'll remove this

packages/flutter/ios/sentry_flutter/Sources/sentry_flutter_objc/SentryFlutterReplayScreenshotProvider.m
```diff
@@ -28,74 +27,73 @@ - (void)imageWithView:(UIView *_Nonnull)view
...
+    } else if ([value isKindOfClass:[FlutterError class]]) {
+      FlutterError *error = (FlutterError *)value;
+      NSLog(@"SentryFlutterReplayRecorderFFI received an error: %@. "
+            @"Cannot capture a replay screenshot.",
+            error.message);
+      return;
+    } else {
+      NSLog(@"SentryFlutterReplayRecorderFFI received an unexpected result. "
+            @"Cannot capture a replay screenshot.");
+    }
```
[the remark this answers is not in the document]
buenaflor: > Bug: Performance Tracking Ignores Disabled State

happens on Dart side now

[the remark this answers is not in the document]
buenaflor: > Bug: API contract broken: Missing callback causes hangs.

Not an issue

packages/flutter/example/lib/main.dart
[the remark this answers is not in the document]
buenaflor: This is an example app and this value has been accidentally changed in a previous PR

not an issue

packages/flutter/example/integration_test/integration_test.dart
```diff
@@ -754,10 +956,10 @@ void main() {
     if (Platform.isIOS) {
       expect(values['key1'], {'value': 'randomValue'}, reason: 'key1 mismatch');
       expect(values['key2'],
-          {'String': 'Value', 'Bool': 1, 'Int': 123, 'Double': 12.3},
+          {'String': 'Value', 'Bool': true, 'Int': 123, 'Double': 12.3},
           reason: 'key2 mismatch');
       // bool values are mapped to num values of 1 or 0 during objc conversion
-      expect(values['key3'], {'value': 1}, reason: 'key3 mismatch');
+      expect(values['key3'], {'value': true}, reason: 'key3 mismatch');
```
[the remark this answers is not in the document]
buenaflor: comment is outdated

Issue #3331, opened 2025-11-11
[Span First]: Add new Span APIs


buenaflor:
[https://develop.sentry.dev/sdk/telemetry/spans/span-api/](https://develop.sentry.dev/sdk/telemetry/spans/span-api/)

PR #3360, opened 2025-11-24
[Span First #1]: Add simple span API

buenaflor:
## :scroll: Description

Span First initiative

Part of #3331 

This is only basic API scaffolding

No tests yet since there is no specific logic yet 

## :pencil: Checklist

- [ ] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps

#skip-changelog


packages/dart/lib/src/protocol/simple_span.dart
```diff
@@ -0,0 +1,30 @@
+import '../../sentry.dart';
+import 'span.dart';
+import 'span_v2_status.dart';
+
+class SimpleSpan implements Span {
```
buenaflor, line 3: Also open to calling this `SpanImpl` or something different. The gist is that this is the default span impl

packages/dart/lib/src/sentry.dart
```diff
@@ -379,6 +381,15 @@ class Sentry {
...
       );
 
+  static Span startSpan(
+    String name, {
+    SpanParent parentSpan = const SpanParent.active(),
+    bool active = true,
+    Map<String, SentryAttribute>? attributes,
+  }) =>
+      _hub.startSpan(name,
+          attributes: attributes, parentSpan: parentSpan, active: active);
```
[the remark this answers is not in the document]
buenaflor: Will fix this once we decide on the namings

packages/dart/lib/src/protocol/noop_span.dart
```diff
@@ -0,0 +1,25 @@
...
+import 'span_v2_status.dart';
+
+class NoOpSpan implements Span {
+  const NoOpSpan();
+
+  @override
+  void end({DateTime? endTimestamp}) {}
+
+  @override
+  void setAttribute(String key, SentryAttribute value) {}
```
denrase, line 10: We can also set `undefined` per API: `setAttribute(key: string, value: SpanAttributeValue | undefined): this;`. Should make this optional, so users can remove keys.

packages/dart/lib/src/span_v2/span_parent.dart
```diff
@@ -0,0 +1,27 @@
+import 'span.dart';
+
+sealed class SpanParent {
```
denrase: Nice! 👍

packages/dart/lib/src/hub.dart
```diff
@@ -573,6 +576,25 @@ class Hub {
     return NoOpSentrySpan();
   }
 
+  Span startSpan(
+    String name, {
+    SpanParent parentSpan = const SpanParent.active(),
```
denrase: Since the API has the name as `parentSpan`, maybe we should also call the type `ParentSpan` to be more consistent.

packages/dart/lib/src/hub.dart
```diff
@@ -573,6 +576,25 @@ class Hub {
...
+    bool active = true,
```
denrase, line 581: Also having both `SpanParent.active()` and `active` parameter might be confusing. WDYT?

packages/dart/lib/src/hub_adapter.dart
```diff
@@ -161,6 +163,16 @@ class HubAdapter implements Hub {
...
       );
 
+  @override
+  Span startSpan(
+    String name, {
+    SpanParent parentSpan = const SpanParent.active(),
+    bool active = true,
+    Map<String, SentryAttribute>? attributes,
+  }) =>
+      Sentry.currentHub.startSpan(name,
```
denrase: Should call `attributes` as last parameter, just like the method definition. Just o be consistent.

packages/dart/lib/src/sentry.dart
```diff
@@ -379,6 +381,15 @@ class Sentry {
         onFinish: onFinish,
       );
 
+  static Span startSpan(
+    String name, {
+    SpanParent parentSpan = const SpanParent.active(),
+    bool active = true,
+    Map<String, SentryAttribute>? attributes,
+  }) =>
+      _hub.startSpan(name,
```
denrase: Dito: Should call `attributes` as last parameter, just like the method definition. Just o be consistent.

denrase (changes requested):

packages/dart/lib/src/hub.dart
buenaflor: I was also thinking of `.inherit`

packages/dart/lib/src/hub.dart
buenaflor: my issue with this is that ParentSpan kind of infers that it's a type of span, but its not at all (for us in Dart at least)

packages/dart/lib/src/hub.dart
buenaflor: or `.fromActiveSpan`?

packages/dart/lib/src/hub.dart
buenaflor: I'm thinking of calling it something like ParentSpanOption or ParentSpanSelector

packages/dart/lib/src/protocol/noop_span.dart
buenaflor: This is typescript syntax, no?

buenaflor: Also we decided not to support null values and instead have a separate `removeAttribute` API (we added that in the global scope too). This is probably just an outdated part of the doc - I guess we may add that here as well at some point

buenaflor:
As discussed we will change this to using an internal `UnsetSpan` which should serve as a marker for not provided span

packages/dart/lib/src/hub.dart
```diff
@@ -573,6 +575,25 @@ class Hub {
...
+        SentryLevel.warning,
+        "Instance is disabled and this 'startSpan' call is a no-op.",
+      );
+    } else if (_options.isTracingEnabled()) {
+      // TODO: implementation of span api behaviour according to https://develop.sentry.dev/sdk/telemetry/spans/span-api/
+      return NoOpSpan();
+    }
+
+    return NoOpSpan();
+  }
```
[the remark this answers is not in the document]
buenaflor: This is just scaffolding, real impl will follow in the next PR

denrase (approved):

PR #3362, opened 2025-11-24
Add crashpad path to Windows

9AZX:
## :scroll: Description

On Windows we don’t get any native crash reporting, which is a real issue since a significant part of our users are on this platform. We receive many crash reports from the native side, but we have no visibility into them. 

> According to the documentation, this isn’t supported on Windows.

@buenaflor : this doc is outdated and needs to be updated

When testing, we saw that the crashpad handler is packaged with the app, but it seems to have never launched. So I dug into the code and saw it is not launched.

No unit test for now, will add in the PR later. I tested with 2 Windows machine and errors was reported to Sentry dashboard.

## :pencil: Checklist

- [ ] All tests passing
- [x] No breaking changes

### DIFF
--- a/packages/flutter/lib/src/native/c/sentry_native.dart
+++ b/packages/flutter/lib/src/native/c/sentry_native.dart
@@ -436,15 +436,20 @@ extension on List<dynamic> {
 }
 
 String? _getDefaultCrashpadPath() {
-  if (Platform.isLinux) {
-    final lastSeparator =
-        Platform.resolvedExecutable.lastIndexOf(Platform.pathSeparator);
-    if (lastSeparator >= 0) {
-      final appDir = Platform.resolvedExecutable.substring(0, lastSeparator);
+  final lastSeparator =
+      Platform.resolvedExecutable.lastIndexOf(Platform.pathSeparator);
+  if (lastSeparator >= 0) {
+    final appDir = Platform.resolvedExecutable.substring(0, lastSeparator);
+    if (Platform.isLinux) {
       final candidates = [
         '$appDir${Platform.pathSeparator}crashpad_handler',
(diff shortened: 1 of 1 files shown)


buenaflor:
@9AZX did you check if crashpad exe is in the same dir as your app's executable?

buenaflor:
@vaind do we need this change?

afaik if no handler path is set sentry-native looks for crashpad_handler.exe in the same directory as the running executable right?

9AZX:
> @9AZX did you check if crashpad exe is in the same dir as your app's executable?

Yes it is: 
image

When launching the app, i don't see it, but with my change it does appear in the task manager

vaind:
> https://docs.sentry.io/platforms/native/configuration/backends/crashpad/
> The SDK will automatically look for a crashpad_handler executable in the same directory as the running application.

@JoshuaMoelans can you confirm this works on Windows?

--- 

FWIW I've checked the [tests](https://github.com/getsentry/sentry-dart/blob/main/packages/flutter/test/sentry_native/sentry_native_test_ffi.dart) and they specify the path exactly because it's not alongside the executable (as there's none because this is running in the context of the current flutter test process, not a separate executable)

JoshuaMoelans:
@vaind yeah if the `handler_path` option is not set, we look for the `crashpad_handler.exe` in the same directory as the current executable https://github.com/getsentry/sentry-native/blob/0.12.1/src/backends/sentry_backend_crashpad.cpp#L437-L455

buenaflor:
> yeah if the handler_path option is not set, we look for the crashpad_handler.exe in the same directory as the current executable

I can try it out on my windows machine but tbh not sure why it's not working in this case

buenaflor:
user reported that the issue may have been their anti-cheat which prevented crashpad from running. I advised them to use `breakpad` instead which works for them.

PR #3372, opened 2025-11-27
fix(jni): dart to native type conversion

buenaflor:
Fixes conversion of non-supported types in Dart conversion to JNI types.

What we want: if we try to convert an unsupported type we just fallback to its `.toString()` representation

Integration test. manual test

## :pencil: Checklist

- [ ] All tests passing
- [x] No breaking changes


packages/flutter/lib/src/native/java/sentry_native_java.dart
```diff
@@ -390,35 +388,37 @@ class SentryNativeJava extends SentryNativeChannel {
...
+  for (final entry in json.entries.where((e) => e.value != null)) {
     final jk = entry.key.toJString();
-    final jv = _dartToJObject(entry.value);
+    final jv = dartToJObject(entry.value);
     jMap[jk] = jv;
     jk.release();
-    jv?.release();
+    jv.release();
   }
   return jMap;
```
buenaflor, line 421: java drops null values so we can just directly filter null values out

packages/flutter/lib/src/native/java/sentry_native_java.dart
```diff
@@ -390,35 +388,37 @@ class SentryNativeJava extends SentryNativeChannel {
...
       String s => s.toJString(),
       bool b => b.toJBoolean(),
       int i => i.toJLong(),
       double d => d.toJDouble(),
-      List<dynamic> l => _dartToJList(l),
-      Map<String, dynamic> m => _dartToJMap(m),
-      _ => null
+      List<dynamic> l => dartToJList(l),
+      Map<String, dynamic> m => dartToJMap(m),
+      _ => value.toString().toJString()
```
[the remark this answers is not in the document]
buenaflor: fixed

packages/flutter/example/integration_test/native_jni_utils_test.dart
```diff
@@ -0,0 +1,172 @@
...
+
+import 'dart:io';
+
+import 'package:flutter_test/flutter_test.dart';
+import 'package:jni/jni.dart';
+import 'package:sentry_flutter/src/native/java/sentry_native_java.dart';
+
+import 'utils.dart';
+
+final _customObject = CustomObject();
```
denrase: Most (if not all) other tests have their varsity in the main function scope. Any reason to do this outside?

packages/flutter/lib/src/native/java/sentry_native_java.dart
```diff
@@ -233,15 +233,14 @@ class SentryNativeJava extends SentryNativeChannel {
 
   @override
   void setContexts(String key, value) => tryCatchSync('setContexts', () {
+        if (value == null) return;
```
denrase: We should check if setContexts on the dart side is resetting/removing the value. If yes, we'll be out of sync, as we are just returning here.

denrase (approved):
Just the one comment that we need to check.

packages/flutter/example/integration_test/native_jni_utils_test.dart
buenaflor: Initially I had FFI tests in there as well so both JNI and FFI used the same test objects and I found it to be extremely cluttered because of indentation and I had trouble finding the test sections

so the only reason it's on the top scope is for a bit more clarity since the inputs don't change anyway

but I can move it down to the main function as well to be consistent

EDIT: updated it

packages/flutter/lib/src/native/java/sentry_native_java.dart
buenaflor: removed the null checks. on native scope sync it will both send null values with the key on Dart and native events. (Sentry will complain though but thats fine)

PR #3393, opened 2025-12-10
Bump iOS version example

itaybre:
Fixes some of the issues when building with Cocoa v9

### DIFF
--- a/min_version_test/ios/Podfile
+++ b/min_version_test/ios/Podfile
@@ -1,5 +1,5 @@
 # Uncomment this line to define a global platform for your project
-platform :ios, '12.0'
+platform :ios, '15.0'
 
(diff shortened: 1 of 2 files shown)


buenaflor:
I think we can skip work on integrating Cocoa v9 for now, we won't be able to use it on Flutter anytime soon since it bumps the min iOS version

denrase:
@buenaflor If the cocoa SDK bumps the iOS version, we should consider doing this sooner. But as this is a breaking change, you probably want to do this in a major version, right?

buenaflor:
@denrase yea we don't wanna do this now, only in a new major

buenaflor:
Let's revisit this when it's time for a new major for Dart

PR #3400, opened 2025-12-15
[DRAFT] feat(span-first): telemetry processor

buenaflor:
## :pencil: Checklist

- [ ] All tests passing
- [ ] No breaking changes


packages/dart/test/sentry_client_test.dart
```diff
@@ -1734,41 +1751,32 @@ void main() {
...
     }
 
-    test('sets log batcher on options when logs are enabled', () async {
-      expect(fixture.options.logBatcher is NoopLogBatcher, true);
-
-      fixture.options.enableLogs = true;
-      fixture.getSut();
-
-      expect(fixture.options.logBatcher is NoopLogBatcher, false);
-    });
```
buenaflor, line 1744: this is tested in telemetry processor test file

packages/dart/lib/src/telemetry_processing/telemetry_buffer.dart
```diff
@@ -0,0 +1,111 @@
...
+/// Holds both raw item and encoded bytes for size tracking and grouping.
+class EncodedTelemetryItem<T extends TelemetryItem> {
+  final T item;
+  final List<int> encoded;
+
+  EncodedTelemetryItem(this.item, this.encoded);
+}
+
+/// In-memory buffer with time and size-based flushing.
+class InMemoryTelemetryBuffer<T extends TelemetryItem>
```
buenaflor, line 21: pretty much copy pasted the sentry log batcher

packages/dart/lib/src/protocol/simple_span.dart
```diff
@@ -18,7 +21,14 @@ class SimpleSpan implements Span {
     Hub? hub,
   })  : _spanId = SpanId.newId(),
         _hub = hub ?? HubAdapter(),
-        _name = name;
+        _startTimestamp = DateTime.now().toUtc(),
```
denrase, line 24: Don't we have some date wrapper, also used in testing?

packages/dart/lib/src/telemetry_processing/envelope_builder.dart
```diff
@@ -0,0 +1,84 @@
...
+              items.map((i) => i.encoded).toList(),
+              options.sdk,
+            ));
+}
+
+/// Groups spans by segment, one envelope per segment.
+class SpanEnvelopeBuilder implements EnvelopeBuilder<Span> {
+  final SentryOptions _options;
+
+  SpanEnvelopeBuilder(this._options);
```
denrase, line 37: Could we inject `publicKey`, `release` & 'environment'  instead of all options, or are they subject to change after init? Could bundle ein `Metadata` object.

packages/dart/lib/src/telemetry_processing/envelope_builder.dart
```diff
@@ -0,0 +1,84 @@
...
+
+  SingleEnvelopeBuilder(this._create);
+
+  @override
+  List<SentryEnvelope> build(List<EncodedTelemetryItem<T>> items) =>
+      items.isEmpty ? [] : [_create(items)];
+}
+
+class LogEnvelopeBuilder extends SingleEnvelopeBuilder<SentryLog> {
+  LogEnvelopeBuilder(SentryOptions options)
```
denrase, line 26: Could we inject `sdk` instead of all options, or is it subject to change after init?

packages/dart/lib/src/telemetry_processing/telemetry_buffer.dart
```diff
@@ -0,0 +1,111 @@
...
+    } catch (error) {
+      _logger(
+        SentryLevel.error,
+        '$InMemoryTelemetryBuffer for $T: Failed to encode item: $error',
+      );
+    }
+  }
+
+  @override
+  FutureOr<void> flush() => _performFlush();
```
denrase, line 68: There was some naming discussion over at Cocoa, I think the gist was to call this `clear` instead of flush. Not super important now, just something to consider.

packages/dart/lib/src/telemetry_processing/telemetry_item.dart
```diff
@@ -0,0 +1,6 @@
+import 'package:meta/meta.dart';
+
+abstract class TelemetryItem {
```
denrase, line 3: This is so generic, we should probably call it something like `Encodable`

packages/dart/lib/src/telemetry_processing/telemetry_buffer.dart
denrase, line 21: Cocoa has additionaly seperated the InMemory Storage from the logic when to dispatch the buffered elements. Not sure this is needed here, here's the [PR](https://github.com/getsentry/sentry-cocoa/pull/7017)

buenaflor: dont think we'll need this, we won't implement file-based buffering or any other storage other than in-memory fwiw - we'll rely on the native SDKs and when we replace it with the native file-based ones in the future we will swap the whole impl, not only the storage

so imo doesn't make sense to separate them for Dart (currently)

when we do have a use case later we can just adjust the impl

packages/dart/lib/src/telemetry_processing/telemetry_item.dart
buenaflor: Wasn't sure if we're gonna add more things but yea `SentryEncodable` would be better I think

packages/dart/lib/src/telemetry_processing/envelope_builder.dart
buenaflor: don't think it changes, I'll update

buenaflor:
this PR will be split up

Issue #3333, opened 2025-11-11
[Span First]: Add telemetry buffer


buenaflor:
For now a simple in-memory buffer will suffice - we need it for Flutter Desktop and Web anyway. In the future we want to reuse the native telemetry buffers in iOS and Android since we also want to make sure data in the buffer is not lost on crashes.

buenaflor:
Buffer is done

PR #3409, opened 2025-12-17
Telemetry Processor (2): In memory buffer implementation

buenaflor:
## :scroll: Description
Adds in-memory buffer implementation

#skip-changelog

Part of #3333 

Unit test

## :pencil: Checklist

- [ ] All tests passing
- [x] No breaking changes


packages/dart/lib/src/telemetry_processing/telemetry_buffer.dart
```diff
@@ -12,7 +13,7 @@ abstract class TelemetryBuffer<T> {
   void add(T item);
 
   /// When executed immediately sends all buffered items to Sentry and clears the buffer.
-  FutureOr<void> clear();
+  FutureOr<void> flush();
```
buenaflor, line 17: @denrase I doubled checked cocoa and imo flush would fit better here. 

their architecture is a bit different so they call it clear because it actually only clears the buffer and not send it and their "flush" is done through the delegate

our 'flush' actually sends it and clears as well

packages/dart/lib/src/telemetry_processing/telemetry_buffer.dart
```diff
@@ -23,20 +25,92 @@ class BufferedItem<T> {
...
+      return null;
+    }
+
+    return _buildAndSend(itemsToSend);
+  }
+
+  Future<void> _buildAndSend(List<BufferedItem<T>> items) async {
+    try {
+      final envelopes = _envelopeBuilder.build(items);
+      await Future.wait(envelopes.map((envelope) => _transport.send(envelope)));
```
denrase, line 123: Will this await all envelopes?

packages/dart/lib/src/telemetry_processing/telemetry_buffer_config.dart
```diff
@@ -0,0 +1,15 @@
+final class TelemetryBufferConfig {
```
denrase, line 1: ❤️

packages/dart/lib/src/telemetry_processing/telemetry_buffer.dart
```diff
@@ -23,20 +25,92 @@ class BufferedItem<T> {
...
   void add(T item) {
-    final encoded = utf8JsonEncoder.convert(item.toJson());
-    final _ = BufferedItem(item, encoded);
-    // TODO(next-pr): finish this impl
+    try {
+      final encoded = utf8JsonEncoder.convert(item.toJson());
+      _items.add(BufferedItem(item, encoded));
+      _bufferSize += encoded.length;
+
+      if (_bufferSize >= _config.maxBufferSizeBytes ||
```
denrase, line 73: I'm not sure if we need to handle this on the client, but users could add single logs which are larger 3 MB. With this check, we'd still send it, and the server would then reject it. Leading to endless re-tries (currently on cocoa, this part will be fixed). Maybe worth checking how the behaviour of this case is on iOS/Android/Flutter Web.

denrase (approved):
Only one thing regarding large items, not directly linked to this PR, but worth checking/debugging.

packages/dart/lib/src/telemetry_processing/telemetry_buffer.dart
buenaflor: We've already checked this: https://github.com/getsentry/sentry-dart/issues/3356

packages/dart/lib/src/telemetry_processing/telemetry_buffer.dart
buenaflor: Only in the callee does `await flush()`

packages/dart/lib/src/telemetry_processing/telemetry_buffer.dart
buenaflor: Actually we do have to reject it in the SDK because the fix for cocoa is only in v9

buenaflor: Will update the impl and add a note

PR #3420, opened 2025-12-19
fix(android): Envelope worker not starting when `autoInitializeNativeSdk` is disabled

### LINKED ISSUE #3416: Sentry does not capture flutter layer exceptions on the dashboard when using native initialization
### Platform

Flutter Mobile Android

### Obfuscation

Disabled

### Debug Info

Disabled

### Doctor

[✓] Flutter (Channel stable, 3.29.3, on macOS 26.0 25A354 darwin-arm64, locale en-CA) [10.7s]
    • Flutter version 3.29.3 on channel stable
    • Framework revision ea121f8859 (8 months ago), 2025-04-11 19:10:07 +0000
    • Engine revision cf56914b32
    • Dart version 3.7.2
    • DevTools version 2.42.3

### Version

9.8.0

### Steps to Reproduce

Follow the native initialization setup docs.sentry.io/platforms/dart/guides/flutter/native-init/ including setting `autoInitializeNative` to `false`

Throw an exception in the flutter layer

I have setup a minimal github project to reproduce the issue we are encountering in our app
https://github.com/divine-edung/blanksentry

### Expected Result

After flutter exception is thrown or Sentry.captureException is called, it should be reported on sentry dashboard

### Actual Result

Flutter exceptions are **not** reported. Only crashes that start at the native layer are captured. Interestingly if I add breadcrumbs on the flutter side, they appear as part of the bread crumbs after the native crash

Note: After both initializations, I checked Sentry.isEnabled and it returned true.

None

buenaflor:
## :scroll: Description

Currently the worker only starts within the `init` method which is not triggered when `autoInitializeNativeSdk` is disabled.

Fixes #3416 

Manual test, unit test

## :pencil: Checklist

- [ ] All tests passing
- [x] No breaking changes


packages/flutter/lib/src/native/java/sentry_native_java.dart
```diff
@@ -23,7 +23,12 @@ class SentryNativeJava extends SentryNativeChannel {
   AndroidEnvelopeSender? _envelopeSender;
   native.ReplayIntegration? _nativeReplay;
 
-  SentryNativeJava(super.options);
+  SentryNativeJava(super.options) {
+    // Initialize envelope sender here instead of init() to ensure it starts
+    // in both autoInitializeNativeSdk enabled and disabled cases.
+    _envelopeSender = AndroidEnvelopeSender.factory(options);
+    _envelopeSender!.start();
```
denrase: We don't need to use the force unwrap here. Also should look into the cursor bot comment.

`_envelopeSender?.start();`

packages/flutter/test/native/android_envelope_sender_test_real.dart
```diff
@@ -40,7 +40,7 @@ void main() {
       expect(() => sender.close(), returnsNormally);
     });
 
-    test('warns and drops after close', () async {
+    test('buffers envelopes after close', () async {
```
denrase: Is buffering correct after we close the sender, or should we just handle the deferred init case?

buenaflor: I realized we dont need the buffer, when the background isolate hasnt spawned yet we can just send envelopes in the main isolate like we normally do

packages/flutter/test/native/android_envelope_sender_test_real.dart
```diff
@@ -12,7 +12,7 @@ import 'package:sentry_flutter/src/isolate/isolate_worker.dart';
 
 void main() {
   group('AndroidEnvelopeSender host behavior', () {
-    test('warns and drops when not started', () {
+    test('sends envelopes in main isolated when not started', () {
```
denrase: I don't see where we assert that the envelope was send in the main envelope, we just assert the log.

denrase (approved):
Just one test to update, rest is LGTM.

packages/flutter/test/native/android_envelope_sender_test_real.dart
buenaflor: yeah I'll update the test name, the actual sending cannot really be tested since it uses JNI

PR #3422, opened 2025-12-22
feat(span-first): Refactor telemetry spans to `SentrySpanV2`

buenaflor:
## :scroll: Description

This is mostly a refactor PR but I will go into details what has been changed

Span:
- Renames:
  - `Span` -> `SentrySpanV2`
  - `UnsetSpan`  -> `UnsetSentrySpanV2`
  - `NoOpSpan` -> `NoOpSentrySpanV2`
- Refactor to sealed class
  - exhaustive switch
  - prevents implementing the public Span API

File structure:
- created a new `telemetry` directory
  - `telemetry/span` where span telemetry is
  - `telemetry/processing` where processor and buffer is
  - future: `telemetry/log`, `telemetry/metric`, etc...

Telemetry Processor:
- Moved the creation of the telemetry processor from `sentry_client` to separate integration

Buffer:
- Now separated in normal `InMemoryTelemetryBuffer` and `GroupedInMemoryTelemetryBuffer`
- Removed the generic dependency to `JsonEncodable` and instead use a generic type encoding function that's passed in the constructor

Span First

Unit tests

## :pencil: Checklist

- [ ] All tests passing
- [ ] No breaking changes

## :crystal_ball: Next steps

#skip-changelog


buenaflor:
cursor review

packages/dart/lib/src/telemetry/span/recording_sentry_span_v2.dart
```diff
@@ -0,0 +1,97 @@
...
+  })  : _spanId = SpanId.newId(),
+        _parentSpan = parentSpan,
+        _name = name,
+        _clock = clock,
+        _onSpanEnded = onSpanEnded,
+        _log = log,
+        _startTimestamp = clock(),
+        _traceId = parentSpan?.traceId ?? defaultTraceId {
+    _segmentSpan = parentSpan?.segmentSpan ?? this;
+  }
```
buenaflor: instead of a hub we should instead opt for passing only the required dependencies

buenaflor:
cursor review

packages/dart/lib/src/telemetry/processing/in_memory_buffer.dart
```diff
@@ -0,0 +1,145 @@
...
+    try {
+      encoded = _encoder(item);
+    } catch (exception, stackTrace) {
+      _logger(SentryLevel.error, '$runtimeType: Failed to encode item $T',
+          exception: exception, stackTrace: stackTrace);
+      return;
+    }
+
+    if (encoded.length > _config.maxBufferSizeBytes) {
+      _logger(
```
denrase: Nice

packages/dart/lib/src/telemetry/processing/processor.dart
```diff
@@ -2,15 +2,14 @@ import 'dart:async';
...
-import 'telemetry_buffer.dart';
+import '../../../sentry.dart';
+import '../span/sentry_span_v2.dart';
+import 'buffer.dart';
 
 /// Manages buffering and sending of telemetry data to Sentry.
 abstract class TelemetryProcessor {
   /// Adds a span to the buffer.
-  void addSpan(Span span);
+  void addSpan(RecordingSentrySpanV2 span);
```
denrase, line 15: Only the concrete type here? If so, have you thought about naming the protocol `ISpanV2` and the `RecordingSentrySpanV2` just `SentrySpanV2`? This would be in line with how the SDK was structured before, so it's easier to understand for users.

denrase (changes requested):

packages/dart/lib/src/telemetry/processing/processor.dart
buenaflor: the user still only interacts with the interface `SentrySpanV2` e.g startSpan only returns and accepts `SentrySpanV2` - all concrete types are marked as internal

internally for the telemetry processing, etc... allowing only the real recording span makes sense imo since supporting the whole interface with other concrete types (noop, unset) just adds extra if/switch checks everywhere that don't really add much value - imo we should just handle them at the startSpan level which we already do

we'd just litter the processor/buffer/scope and any other place that requires a 'real' span with unnecessary checks

e.g  noop spans should not be serialized and be processed to the buffer so there is no reason to support this use case

we only have the span interface because of the requirement to have the UnsetSpan for startSpan otherwise I'd just model it as a concrete class SentrySpanV2 because realistically there is and will only be one "real" span implementation

> have you thought about naming the protocol ISpanV2 and the RecordingSentrySpanV2 just SentrySpanV2

ideally I'd like to have the overall interface name just be SentrySpan - but we can't use it right now since it's already used in the transaction models
I matched the 'recording span' with what Sentry JS and general OTEL implementations name these "real" spans.

fwiw the user won't see RecordingSentrySpanV2 anyway

since it's also a brand new API we can deviate a little bit - personally I'm not a big fan of putting I in front of interfaces and rather have the concrete implementations be specific about what it is. 

RecordingSentrySpan -> sentry span that is actively recording data

denrase (approved):

PR #3425, opened 2025-12-23
enh(internal): debug log api

buenaflor:
## :scroll: Description

Improves the debug logging dx

**Currently the debug logging api is tied to the options which has few caveats**
- by design not usable in isolates
- cumbersome to use, need to inject the logger or the options instance everywhere
- no clear logger name separation per package since there is only one instance

**New design:**
- no need to inject
- one top-level instance per package with the proper name: 
  - e.g `final debugLogger = SentryDebugLogger('sentry.dio')` that is then used within the dio package
- isolate-compatible, you can continue using the `debugLogger` instance, you only need to run `SentryDebugLogger.configure` in the new isolate

Manual and unit test

## :crystal_ball: Next steps

Step-by-step replace existing API

#skip-changelog


packages/dart/lib/src/utils/debug_logger.dart
```diff
@@ -0,0 +1,106 @@
...
+
+  @pragma('vm:prefer-inline')
+  void _log(
+    SentryLevel level,
+    String message, {
+    Object? error,
+    StackTrace? stackTrace,
+  }) {
+    if (!_isEnabled) return;
+    if (level.ordinal < _minLevel.ordinal) return;
```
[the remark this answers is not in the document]
buenaflor: This doesn't make sense anymore
- we already automatically enable `debug` anyway.
- if someone explicitly turns off debug it should not log at all

packages/dart/lib/src/utils/debug_logger.dart
```diff
@@ -0,0 +1,106 @@
...
+/// Each package should have at least one top-level instance.
+///
+/// Example:
+/// ```dart
+/// const debugLogger = SentryDebugLogger('sentry.flutter');
+///
+/// debugLogger.warning('My Message')
+///```
+@internal
+class SentryDebugLogger {
```
denrase: I think `SentryInternalLogger` would be a more appropriate name.

packages/dart/lib/src/utils/debug_logger.dart
```diff
@@ -0,0 +1,106 @@
...
+  @visibleForTesting
+  static bool get isEnabled => _isEnabled;
+
+  @visibleForTesting
+  static SentryLevel get minLevel => _minLevel;
+
+  /// Configure logging for the current isolate.
+  ///
+  /// This needs to be called for each new spawned isolate before logging.
+  static void configure({
```
denrase: Is this ever changing during runtime? I don't understand why we now have both an instance and static vars. If we go the instance, can't we just call `Logger(name, minLevel)'? If it needs to be enabled later, we can just set it on the instance.

So either static only, or instance only, no need to mix both.

packages/dart/lib/src/utils/debug_logger.dart
```diff
@@ -0,0 +1,106 @@
...
+    required bool isEnabled,
+    SentryLevel minLevel = SentryLevel.warning,
+  }) {
+    _isEnabled = isEnabled;
+    _minLevel = minLevel;
+  }
+
+  void debug(
+    String message, {
+    Object? error,
```
denrase: Is there a case where we would log an error + stacktrace as debug? Same for `info` and probably also `warning`.

denrase (changes requested):

packages/dart/lib/src/utils/debug_logger.dart
buenaflor: >  I don't understand why we now have both an instance and static vars

the reason why it's an instance is so we have the benefit of creating multiple logger instance with the same configuration within the same isolate

the configuration is static per isolate, I don't think we will ever have a use case where we want multiple loggers with different minLevels

Example: Flutter
```dart
const logger = SentryDebugLogger(sentry.flutter)

// SentryDebugLogger.configure(...) called in SentryOptions
logger.info('info in main isolate)' // by default uses the main isolate static config

// spawn a new isolate
// within that isolate
// SentryDebugLogger.configure(...) called directly after isolate has spawned
logger.info('info in other isolate') // you can still use that same logger instance
```
---

- if we went the route with static only we would lose the ability to directly associate a package to a specific logger
  - `SentryDebugLogger.info(name: 'sentry.flutter', message: 'xyz')`
  - and repeating `name: 'sentry.flutter',` hundreds of time is error prone

- if we went the route with instance only we would need to create a new logger instance in each new isolate which can get messy quickly because then you would have to differentiate between the main isolate logger and the current isolate logger and the code interchanges heavily where it's easy to use the wrong one (take a look at android replay recorder and envelope sender)

> Logger(name, minLevel)

this doesn't work because you can only configure the logger through the static function

packages/dart/lib/src/utils/debug_logger.dart
buenaflor: hm not sure, just added it for consistency

packages/dart/lib/src/utils/debug_logger.dart
buenaflor: what we could do is to separate the config and have only the config be static so we don't mix it into the log class

```dart
/// Per-isolate logging configuration.
@internal
class SentryLogConfig {
  static bool isEnabled = false;
  static SentryLevel minLevel = SentryLevel.warning;

  /// Configure logging for the current isolate.
  static void configure({
    required bool isEnabled,
    SentryLevel minLevel = SentryLevel.warning,
  }) {
    SentryLogConfig.isEnabled = isEnabled;
    SentryLogConfig.minLevel = minLevel;
  }
}


// within the logger:
@pragma('vm:prefer-inline')
void _log(SentryLevel level, String message, {Object? error, StackTrace? stackTrace}) {
  if (!SentryLogConfig.isEnabled) return;
  if (level.ordinal < SentryLogConfig.minLevel.ordinal) return;
  
  dev.log(
    '[${level.name}] $message',
    name: _name,
    level: level.toDartLogLevel(),
    error: error,
    stackTrace: stackTrace,
    time: DateTime.now(),
  );
}
```

**alternative 1 - update the config on the instance:**

```dart
const logger = SentryDebugLogger(sentry.flutter);

// later at some point
logger.updateConfig(isEnabled: xyz, name: 'sentry.flutter');
```

caveat: we would need to update the config for every new logger

**alternative 2 - update the config using a callback on a global config instance:**
```dart
const loggerConfig = LoggerConfig() // updated later at some point when Sentry initializes
final logger = SentryDebugLogger(sentry.flutter, () => loggerConfig);
```

denrase: > if we went the route with static only we would lose the ability to directly associate a package to a specific logger
> SentryDebugLogger.info(name: 'sentry.flutter', message: 'xyz')
> and repeating name: 'sentry.flutter', hundreds of time is error prone

We always could add the name to the static config call in this case.

> the reason why it's an instance is so we have the benefit of creating multiple logger instance with the same configuration within the same isolate

Whats the use-case for this in internal only logging?

```dart

const sentryPackageName = 'sentry.flutter'

// Once in main Isolate 
InternalLogger.configure(name:  sentryPackageName, ...)
InternalLogger.debug("i'm in main")

// Once in Isolate
InternalLogger.configure(name:  sentryPackageName, ...)
InternalLogger.debug("i'm in an isolate")
```

buenaflor: if we have a static only api  like you showed here we lose the ability to associate the logging names with the packages in a convenient way. it will always be sentry.flutter even if the logging came from sentry dio. of course you can always add the names directly to the log calls but that requires maintenance, discipline etc. imo its just easier if you can use the logger and it automatically uses the right package name to associate the log to

thats why my proposal is
-> each package creates their own logger instance
-> logger config should be global to the isolate (minLevel and isEnabled realistically will never change so thats why I propose its global)

overall there is nothing wrong with using static only, but since we have 10 packages at this point (and more to come) it would be a nice additiom to know exactly and quickly where specific logs came from and imo its easier for each package to have its own logger instance with the right config rather than trying to enforce discipline to add the package name to the log call sites

denrase (approved):

PR #3426, opened 2025-12-29
chore(agents): Add `AGENTS.md`

buenaflor:
#skip-changelog

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,13 +1,12 @@
-# Sentry Development Guide for AI Agents
+# Sentry Dart/Flutter SDK - Agent Guide
 
 ## Overview
 
 Sentry is a developer-first error tracking and performance monitoring platform.
-This repository contains the Sentry Dart/Flutter SDK and integrations with third party libraries.
+This repository contains the Sentry Dart/Flutter SDK and integrations.
 
 ## Project Structure
 
-- `packages/` - SDK packages (Melos monorepo)
 - `packages/dart/` - Core Sentry Dart SDK
 - `packages/flutter/` - Sentry Flutter SDK (includes native integrations)
 - `packages/dio/` - Dio HTTP client integration
@@ -20,7 +19,7 @@ This repository contains the Sentry Dart/Flutter SDK and integrations with third
 - `packages/supabase/` - Supabase integration
 - `packages/firebase_remote_config/` - Firebase Remote Config integration
 - `packages/link/` - Deep linking integration
-- `docs/` - Documentation and release checklists
+- `docs/` - Documentation
 - `e2e_test/` - End-to-end test suite
 - `min_version_test/` - Minimum SDK version compatibility tests
 - `metrics/` - Size and performance metrics tooling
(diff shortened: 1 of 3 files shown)


AGENTS.md
```diff
@@ -0,0 +1,460 @@
...
+- `docs/` - Documentation and release checklists
+- `e2e_test/` - End-to-end test suite
+- `min_version_test/` - Minimum SDK version compatibility tests
+- `metrics/` - Size and performance metrics tooling
+- `scripts/` - Build, release, and utility scripts
+- `melos.yaml` - Melos monorepo configuration
+
+## Environment
+
+- Flutter Version: `3.24.0`
```
denrase: Should we add to the update/version script so this is auto synced to the correct version, or is it enough to update this on the rare occasions we update? Also, it should be noted that these are min supported versions.

AGENTS.md
```diff
@@ -0,0 +1,460 @@
...
+
+Check the package's `pubspec.yaml` to determine if it's a Dart-only or Flutter package:
+
+- Dart packages - run `(fvm) dart test`
+- Flutter packages - run `(fvm) flutter test`
+- Web tests - run `(fvm) flutter test -d chrome`
+
+Run tests from within the package directory (e.g., `packages/dart/` or `packages/flutter/`).
+
+### Formatting & Analysis
```
denrase: Does it make sense to mention `dart fix --apply`?

AGENTS.md
```diff
@@ -0,0 +1,460 @@
...
+
+```dart
+SentryOptions defaultTestOptions() {
+  return SentryOptions(dsn: testDsn)..automatedTestMode = true;
+}
+```
+
+### Test File Organization
+
+When a test file grows large, consider splitting it into sub-files by context:
```
denrase: Probably more of a code-smell if single classes have so many responsibilities what we need to split up test files just to keep it understandable, no?

AGENTS.md
```diff
@@ -0,0 +1,460 @@
...
+- [DO sort sections alphabetically.](https://dart.dev/effective-dart/style#do-sort-sections-alphabetically)
+
+### Formatting
+
+- [DO format your code using `dart format`.](https://dart.dev/effective-dart/style#do-format-your-code-using-dart-format)
+- [CONSIDER changing your code to make it more formatter-friendly.](https://dart.dev/effective-dart/style#consider-changing-your-code-to-make-it-more-formatter-friendly)
+- [PREFER lines 80 characters or fewer.](https://dart.dev/effective-dart/style#prefer-lines-80-characters-or-fewer)
+- [DO use curly braces for all flow control statements.](https://dart.dev/effective-dart/style#do-use-curly-braces-for-all-flow-control-statements)
+
+### Comments
```
denrase: LLMS tend to overdo it with comments, often describing very obvious code-behaviour, or injecting comments about the iterative process during prompting ("this has been refactored to..." or similar). Maybe we could instruct it to create readable code/variables with clear intent and only add comments where we need additional context (why was this made) or public API docs.

AGENTS.md
```diff
@@ -0,0 +1,460 @@
...
+- [DO make your `==` operator obey the mathematical rules of equality.](https://dart.dev/effective-dart/design#do-make-your-operator-obey-the-mathematical-rules-of-equality)
+- [AVOID defining custom equality for mutable classes.](https://dart.dev/effective-dart/design#avoid-defining-custom-equality-for-mutable-classes)
+- [DON'T make the parameter to `==` nullable.](https://dart.dev/effective-dart/design#dont-make-the-parameter-to-nullable)
+
+## Development
+
+### When Stuck
+
+- Ask a clarifying question or propose a short plan
+- Do not push large speculative changes without confirmation
```
denrase: Id' also like to mention that the agent should never commit, unless explicitly asked to.

denrase (changes requested):

AGENTS.md
buenaflor: yeah definitely, I'll rework this a bit

AGENTS.md
buenaflor: usually yea, but we do have examples of those like Hub, Client where the test file itself is 1k LoC

we can probably remove this, makes more sense to judge individually

AGENTS.md
buenaflor: I'll also remove the stuff that is lintable

buenaflor:
putting this into draft, I saw some blog posts about writing better `AGENTS.md` files, gonna rework this a bit

AGENTS.md
buenaflor: I dont think we need it, let's just update when its actually time to do it

AGENTS.md
buenaflor: Updated

AGENTS.md
buenaflor: Updated

buenaflor:
@denrase updated it

denrase (approved):

PR #3429, opened 2025-12-30
feat(span-first): Add `traceLifecycle` option, dsc to span and sampling to `startSpan`

buenaflor:
### New Features

- **Dynamic Sampling Context (DSC)** for the new SpanV2 system
  - DSC is created lazily and frozen on first access
  - All spans in a trace segment share the same DSC
  - New factory: `SentryTraceContextHeader.fromRecordingSpan()`

- **Sampling for SpanV2**
  - Root spans now get a sampling decision (sampled/not sampled)
  - Child spans automatically inherit the root span's sampling decision
  - Uses the existing `tracesSampleRate` and `tracesSampler` options

- **New `traceLifecycle` option**
  - `SentryTraceLifecycle.streaming` - spans sent individually as they complete
  - `SentryTraceLifecycle.static` - spans sent together when (legacy) transaction ends (default)

### Code Changes

- **RecordingSentrySpanV2** now uses factory constructors:
  - `RecordingSentrySpanV2.root()` - creates root span with sampling decision
  - `RecordingSentrySpanV2.child()` - creates child span inheriting from parent

- **SentrySamplingContext** updated:
  - New `forSpanV2()` constructor for streaming mode
  - New `forTransaction()` constructor for static mode (legacy)

- **Hub.captureSpan()** now forwards spans to `SentryClient.captureSpan()`

- **SentryClient.captureSpan()** adds spans to the telemetry processor

Part of https://github.com/getsentry/sentry-dart/issues/3331

Unit tests

## :pencil: Checklist

- [ ] All tests passing
- [ ] No breaking changes

## :crystal_ball: Next steps

#skip-changelog


buenaflor:
bugbot review

packages/dart/lib/src/sentry_sampling_context.dart
```diff
@@ -1,21 +1,91 @@
...
+  // TODO: Remove these placeholders once legacy transaction API is removed.
 
-  /// The Transaction context
-  SentryTransactionContext get transactionContext => _transactionContext;
+  /// Placeholder for streaming mode where transaction context is not used.
+  static final _unusedTransactionContext =
+      SentryTransactionContext('unused', 'unused');
+
+  /// Placeholder for static mode where span context is not used.
+  static final _unusedSpanContext = SentrySpanSamplingContextV2('unused', {});
```
buenaflor, line 29: we can't remove/change `transactionContext` because it's a breaking change so we have to add some mechanism for `spanContext` to co-exist with it but only have one active at a time

buenaflor: LMK if there is a better option here

packages/dart/lib/src/hub.dart
```diff
@@ -609,13 +612,52 @@ class Hub {
...
+
+      span = RecordingSentrySpanV2.root(
         traceId: scope.propagationContext.traceId,
         name: name,
-        parentSpan: resolvedParentSpan,
-        log: options.log,
         clock: options.clock,
-        onSpanEnd: captureSpan);
+        dscCreator: (span) =>
+            SentryTraceContextHeader.fromRecordingSpan(span, this),
```
denrase: The `dscCreator` is always called with (this), so the only thing it does not have access to is the `Hub`. Will this be different in the future or would it be enough to provide the Span just the hub though the callback, letting it create the trace context header internally?

packages/dart/lib/src/sentry_sampling_context.dart
```diff
@@ -1,21 +1,91 @@
...
+import 'telemetry/span/sentry_span_sampling_context.dart';
 
 /// Context used by [TracesSamplerCallback] to determine if transaction
 /// is going to be sampled.
+///
+/// Note: This class currently supports both static (transaction-based) and
+/// streaming (span-based) modes for backwards compatibility. The dual-mode
+/// design with placeholder values and runtime checks is a temporary solution.
+/// Once the legacy transaction API is removed, this class should be simplified
+/// to only support the streaming mode with [SentrySpanSamplingContextV2].
```
denrase, line 14: Thx for the explanation why this is here 👍

packages/dart/lib/src/sentry_sampling_context.dart
denrase: I think it's fine, it's a straight forward solution as long as we have to support both.

packages/dart/lib/src/sentry_sampling_context.dart
```diff
@@ -1,21 +1,91 @@
...
+
+  /// The Transaction context.
+  ///
+  /// Throws [StateError] if accessed in streaming mode.
+  ///
+  /// TODO: Remove this getter once legacy transaction API is removed.
+  /// This runtime check is a temporary solution for backwards compatibility.
+  SentryTransactionContext get transactionContext {
+    if (_traceLifecycle != SentryTraceLifecycle.static) {
+      throw StateError('transactionContext is only available in static mode. '
```
denrase: I'm not sure if we should throw if this is accessible by SDK users. They might want to try the new approach, while still having code that is accessing this. Maybe it's less disruptive if we just log an error when accessing this?

packages/dart/lib/src/sentry_sampling_context.dart
```diff
@@ -1,21 +1,91 @@
...
+
+  /// The Span V2 sampling context.
+  ///
+  /// Throws [StateError] if accessed in static mode.
+  ///
+  /// TODO: Remove the runtime check once legacy transaction API is removed.
+  /// This runtime check is a temporary solution for backwards compatibility.
+  SentrySpanSamplingContextV2 get spanContext {
+    if (_traceLifecycle != SentryTraceLifecycle.streaming) {
+      throw StateError('spanContext is only available in streaming mode. '
```
denrase: Same as above.

packages/dart/lib/src/telemetry/span/recording_sentry_span_v2.dart
```diff
@@ -1,43 +1,100 @@
 part of 'sentry_span_v2.dart';
 
+/// Factory for creating a [SentryTraceContextHeader] from a [RecordingSentrySpanV2].
+typedef DscCreator = SentryTraceContextHeader Function(
```
denrase: We could think about adding 'Callback` as a suffix (with updated name), so it's clear that this is a function, like `OnSpanEndCallback`.

packages/dart/lib/src/telemetry/span/sentry_span_sampling_context.dart
```diff
@@ -0,0 +1,18 @@
...
+class SentrySpanSamplingContextV2 {
+  final String name;
+  final Map<String, SentryAttribute> attributes;
+
+  SentrySpanSamplingContextV2(this.name, this.attributes);
+
+  factory SentrySpanSamplingContextV2.fromSpan(SentrySpanV2 span) =>
+      SentrySpanSamplingContextV2(
+        span.name,
+        Map.unmodifiable(span.attributes),
```
denrase, line 18: These attributes will never be mutated, right? We should add a comment for this to the public `attributes`, because if users/we try to mutate it, we'd "crash", right?

denrase (changes requested):

packages/dart/lib/src/sentry_sampling_context.dart
buenaflor: yeah wasn't sure what's appropriate here since we are very strict that it's an either/or situation. but thinking about it I think logging the error makes more sense true

packages/dart/lib/src/hub.dart
buenaflor: specifically it's options and replayId which I should update since we wanna generally be more specific and not just always rely on Hub

The span knows when to create DSC but not how so I think this abstraction is nice since tests are also easy to write

denrase (approved):

Issue #3332, opened 2025-11-11
[Span First]: Add new telemetry pipeline (including filtering and mutation)


buenaflor:
Docs TBD but the idea is that different events (errors, logs, spans, etc…) go through different pipelines

For spans specifically we want to use:

* `scope.applyToSpan` -> specifically applies attributes (in the future only attributes will exist)
  * we will need `Sentry.setAttribute` for that
* `beforeSendSpan` -> will not allow to drop spans but mutate
* `ignoreSpan` -> ignore at start
* client reports must be added if a span is dropped

PR #3441, opened 2026-01-12
feat(span-first): telemetry enrichment

buenaflor:
## :scroll: Description

Generic telemetry enrichment

Part of #3332 

Unit test, manual tests

## :pencil: Checklist

- [ ] All tests passing
- [ ] No breaking changes

## :crystal_ball: Next steps

#skip-changelog


packages/dart/lib/src/telemetry/enricher/common_attributes_provider.dart
```diff
@@ -0,0 +1,67 @@
...
+
+    if (_options.sendDefaultPii) {
+      final user = _scope.user;
+      if (user != null) {
+        if (user.id != null) {
+          attributes[SemanticAttributesConstants.userId] =
+              SentryAttribute.string(user.id!);
+        }
+        if (user.name != null) {
+          attributes[SemanticAttributesConstants.userUsername] =
```
denrase: Shouldn't it be `SemanticAttributesConstants.userName` instead of `SemanticAttributesConstants.userUsername`?

packages/dart/lib/src/telemetry/enricher/common_attribute_provider.dart
```diff
@@ -0,0 +1,67 @@
+import 'dart:async';
+
+import '../../../sentry.dart';
+import '../../utils/os_utils.dart';
+import 'enricher.dart';
+
+class CommonAttributesProvider implements TelemetryAttributesProvider {
```
denrase: Maybe even `SharedAttributesProvider`? No strong opinion here though.

packages/dart/lib/src/telemetry/enricher/common_attributes_provider.dart
```diff
@@ -0,0 +1,67 @@
...
+
+  late final operatingSystem = getSentryOperatingSystem();
+
+  CommonAttributesProvider(this._scope, this._options);
+
+  @override
+  String get name => 'CommonAttributesProvider';
+
+  @override
+  FutureOr<Map<String, SentryAttribute>> provide() {
```
denrase: Could name this as what it returns `getAttributes()`/`attributes`. No strong opinion.

packages/dart/lib/src/telemetry/enricher/enricher_registry.dart
```diff
@@ -0,0 +1,94 @@
...
+        AttributesEnricher<SentryLog>(
+          providers: _attributeProviders,
+          apply: (log, attrs) => log.attributes.addAll(attrs),
+        );
+    _spanEnricher = spanEnricher ??
+        AttributesEnricher<RecordingSentrySpanV2>(
+          providers: _attributeProviders,
+          apply: (span, attrs) =>
+              span.attributes.addAll(attrs), // TODO: this needs to be updated
+        );
```
buenaflor: currently it's fine to have the enricher be only a specific attribute enricher.

when we have more use cases we can compose a (Telemetry)Enricher to be of multiple enrichers

packages/dart/lib/src/telemetry/enricher/enricher_registry.dart
```diff
@@ -0,0 +1,94 @@
...
+
+  /// The single span enricher that uses all registered providers.
+  late final TelemetryEnricher _spanEnricher;
+
+  TelemetryEnricherRegistry(
+      {TelemetryEnricher? logEnricher, TelemetryEnricher? spanEnricher}) {
+    _logEnricher = logEnricher ??
+        AttributesEnricher<SentryLog>(
+          providers: _attributeProviders,
+          apply: (log, attrs) => log.attributes.addAll(attrs),
```
denrase: Is the applier interface/implementation necessary when all we do is `addAll`?

packages/dart/lib/src/telemetry/enricher/enrichment_pipeline.dart
```diff
@@ -0,0 +1,50 @@
...
+
+  final TelemetryEnricherRegistry registry;
+
+  @override
+  FutureOr<void> enrichLog(SentryLog log) => registry.logEnricher.enrich(log);
+
+  @override
+  FutureOr<void> enrichSpan(RecordingSentrySpanV2 span) =>
+      registry.spanEnricher.enrich(span);
+}
```
buenaflor: may not be needed, it's jsut abstraction for easy testing

packages/dart/lib/src/telemetry/enricher/enricher_registry.dart
```diff
@@ -0,0 +1,94 @@
...
+    if (_attributeProviders.contains(provider)) {
+      return;
+    }
+    _attributeProviders.insert(0, provider);
+  }
+
+  /// Removes a [TelemetryAttributesProvider] by name.
+  ///
+  /// Returns `true` if a provider was removed, `false` otherwise.
+  bool removeProviderByName(String name) {
```
denrase: Will this be user facing api where we need `removeProviderByName`, `removeProvider` and `clearProviders`? Do we ever need them?

packages/dart/lib/src/telemetry/enricher/enrichment_pipeline.dart
```diff
@@ -0,0 +1,50 @@
...
+  @override
+  FutureOr<void> enrichLog(SentryLog log) {}
+
+  @override
+  FutureOr<void> enrichSpan(RecordingSentrySpanV2 span) {}
+}
+
+// TODO: we may remove this and use the registry directly, this is just PoC
+@internal
+class DefaultTelemetryEnrichmentPipeline
```
denrase: What's the reason we have this pipeline implementation? If there only is this one, we might as well just call `registry.logEnricher.enrich(log);` directly.

packages/dart/lib/src/telemetry/enricher/enricher_registry.dart
buenaflor: Not really I would also remove it

packages/dart/lib/src/telemetry/enricher/enricher_registry.dart
buenaflor: span attributes are unmodifiable so we have to treat it a bit different, havent pushed the impl yet (we need to use setAttributes)

also we need it because the attributes enricher has no concept of a Telemetry

alternative is to use dynamic and cast

buenaflor: span attributes are unmodifiable so we have to treat it a bit different, havent pushed the impl yet

also we need it because the attributes enricher has no concept of a Telemetry

alternative is to use dynamic and cast

packages/dart/lib/src/telemetry/enricher/enrichment_pipeline.dart
buenaflor: I'm also looking into removing it

buenaflor: Or the registry could live in the pipeline instead of the options

buenaflor: I dont want to pollute the options too much

packages/dart/lib/src/telemetry/enricher/common_attribute_provider.dart
buenaflor: I think I'll keep it as common because Sentry docs multiple times call them Common attributes

packages/dart/lib/src/telemetry/enricher/attributes_provider.dart
```diff
@@ -0,0 +1,124 @@
...
+import '../../../sentry.dart';
+
+/// Base interface for attribute providers used in telemetry enrichment.
+///
+/// Providers compute attributes that are added to telemetry items (spans, logs).
+/// Each provider can filter which items it supports via [supports].
+@internal
+abstract interface class TelemetryAttributesProvider {
+  /// Returns true if this provider should contribute attributes for [item].
+  bool supports(Object item);
```
buenaflor: I use Object here because telemetry currently don't share a common interface but maybe we should do that instead?

packages/dart/lib/src/telemetry/enricher/attributes_aggregator.dart
```diff
@@ -0,0 +1,95 @@
...
+          }
+        }
+      } catch (error, stackTrace) {
+        internalLogger.error(
+          'Provider "$provider" failed: $error',
+          error: error,
+          stackTrace: stackTrace,
+        );
+      }
+    }
```
buenaflor: maybe we should just return Future because I don't know if the perf improvements are worth the effort of maintaining this extra code

packages/dart/lib/src/telemetry/enricher/attributes_provider.dart
buenaflor: I would consider adding a `sealed class Telemetry` with an empty body to constraint this.

that way we can also constraint other places like the telemetry processor to only accept telemetry and nothing else - but not in this PR though

packages/dart/lib/src/telemetry/enricher/attributes_aggregator.dart
```diff
@@ -0,0 +1,95 @@
...
+  /// logged but don't stop aggregation.
+  FutureOr<Map<String, SentryAttribute>> build(
+      Object item, TelemetryAttributesProviderContext context) {
+    final out = <String, SentryAttribute>{};
+
+    for (int i = 0; i < _providers.length; i++) {
+      final provider = _providers[i];
+      if (!provider.supports(item)) continue;
+
+      try {
```
denrase: If i recall correctly we'd only do try/catch around code where SDK users could hook into, do we need this for out own code we have under control?

packages/dart/lib/src/telemetry/enricher/attributes_aggregator.dart
denrase: Yeah i'd rather have this just be a simple loop where we just await individual providers.

```dart
final aggregated = <String, SentryAttribute>{};
for (final provider in providers) {
  aggregated.addAll(await provider(item, context));
}
return aggregated;
```

packages/dart/lib/src/telemetry/enricher/attributes_provider.dart
```diff
@@ -0,0 +1,124 @@
...
+
+  /// Computes attributes for the given [item].
+  ///
+  /// Return a [Future] only if async work is required.
+  FutureOr<Map<String, SentryAttribute>> call(
+      Object item, TelemetryAttributesProviderContext context);
+}
+
+@immutable
+final class TelemetryAttributesProviderContext {
```
denrase: Does it make sense to just inject the needed context into individual provider implementations? That way we don't have to have a collection of all data that all providers need which would pile up here.

packages/dart/lib/src/telemetry/enricher/attributes_provider.dart
```diff
@@ -0,0 +1,124 @@
...
+    required this.scope,
+  });
+}
+
+/// Fully caches the result of any [TelemetryAttributesProvider].
+///
+/// The first call to [call] delegates to the wrapped provider
+/// and caches the result. Subsequent calls return the cached value.
+@internal
+final class CachedTelemetryAttributesProvider
```
denrase: We can do this in the individual providers as well, no need to generalize. I reckon there won't be that many providers to begin with anyway.

packages/dart/lib/src/telemetry/enricher/common_attributes_provider.dart
```diff
@@ -0,0 +1,71 @@
...
+
+import '../../../sentry.dart';
+import '../../utils/os_utils.dart';
+
+/// Provides common attributes for all telemetry items.
+///
+/// Includes SDK metadata, environment/release, user identity (gated by PII),
+/// and operating system information.
+@internal
+final class CommonTelemetryAttributesProvider
```
denrase, line 10: Can we just inject the needed dependencies that we are now holding in context? Other providers might not need them.

packages/flutter/lib/src/telemetry/enricher/native_contexts_attributes_provider.dart
```diff
@@ -0,0 +1,67 @@
...
+
+  NativeContextsTelemetryAttributesProvider(this._nativeBinding);
+
+  @override
+  bool supports(Object item) => true;
+
+  @override
+  FutureOr<Map<String, SentryAttribute>> call(
+      Object item, TelemetryAttributesProviderContext context) async {
+    final infos = await _nativeBinding.loadContexts() ?? {};
```
denrase: Are we loading contexts for every span/log that this is called with? Seeing that we have vlaues like `os.name`, 'os.version', 'device.brand', 'device.model' and 'device.family', which should not change during runtime, I guess we could get away with only doing this once.

denrase (changes requested):
One idea would be to separate loading of dependencies and actual application of attributes. That way we can easily cache what is needed for multiple logs/spans/etc. Also we might not need futures when applying attributes to individual item. WDYT?

```dart

class Aggregator {
  final List<AttributeProvider> providers;

  Aggregator({required this.providers});

  Future<void> install(SentryOptions options) async {
    for (final provider in providers) {
      await provider.install(options);
    }
  }

  Map<String, SentryAttribute> getAttributes(Object item) {
    final attributes = <String, SentryAttribute>{};
    for (final provider in providers) {
      attributes.addAll(provider.getAttributes(item));
    }
    return attributes;
  }
}

abstract class AttributeProvider {
  FutureOr<void> install(SentryOptions options) {};
  Map<String, SentryAttribute> getAttributes();
}

class DefaultAttributeProvider implements AttributeProvider {

  SentryOptions? options;
  
  @override
  FutureOr<void> install(SentryOptions options) async {
    self.options = options;
  }

  @override
  Map<String, SentryAttribute> provide() {
    if (options == null) {
      return {};
    }
    return {
      "sentry.sdk.name": SentryAttribute.string(options!.sdk.name),
    };
  }
}

class NativeAttributeProvider implements AttributeProvider {
  Contexts? contexts;

  @override
  FutureOr<void> install(SentryOptions options) async {
    contexts = await options.native.loadContexts();
  }

  @override
  Map<String, SentryAttribute> provide(Object _) {
    if (contexts == null) {
      return {};
    }
    return {
      "os.name": SentryAttribute.string(contexts?.operatingSystem?.name ?? ''),
      "os.version": SentryAttribute.string(contexts?.operatingSystem?.version ?? ''),
      "device.brand": SentryAttribute.string(contexts?.device?.brand ?? ''),
      "device.model": SentryAttribute.string(contexts?.device?.model ?? ''),
      "device.family": SentryAttribute.string(contexts?.device?.family ?? ''),
    };
  }
}
```

packages/flutter/lib/src/telemetry/enricher/native_contexts_attributes_provider.dart
buenaflor: yeah had the generalized caching for this but we acn also do it within the provider

packages/dart/lib/src/telemetry/enricher/attributes_provider.dart
buenaflor: I refactored it a bit but Scope and Telemetry must be part of the parameters because they depend on the context at capture time

packages/dart/lib/src/telemetry/enricher/attributes_provider.dart
buenaflor: done

packages/dart/lib/src/telemetry/enricher/common_attributes_provider.dart
buenaflor: done

packages/flutter/lib/src/telemetry/enricher/native_contexts_attributes_provider.dart
buenaflor: updated it

packages/dart/lib/src/telemetry/enricher/attributes_aggregator.dart
buenaflor: I dont think we really have a common convention unfortunately, however I think catching still makes sense because I think we dont it will not continue to enrich the non-failing attributes

packages/dart/lib/src/telemetry/enricher/attributes_provider.dart
```diff
@@ -0,0 +1,21 @@
...
+/// Each provider can filter which items it supports via [supports].
+@internal
+abstract interface class TelemetryAttributesProvider {
+  /// Returns true if this provider should contribute attributes for [item].
+  bool supports(Object item);
+
+  /// Computes attributes for the given [item].
+  ///
+  /// Return a [Future] only if async work is required.
+  FutureOr<Map<String, SentryAttribute>> attributes(Object item,
```
buenaflor: using `attributes` instead of `getAttributes` here since `get` prefixes are [discouraged](https://dart.dev/effective-dart/design#avoid-starting-a-method-name-with-get)

packages/dart/lib/src/sentry_client.dart
```diff
@@ -518,54 +522,7 @@ class SentryClient {
...
-    final propagationContext = scope?.propagationContext;
-    if (propagationContext != null) {
-      log.traceId = propagationContext.traceId;
-    }
-    final span = scope?.span;
-    if (span != null) {
-      log.attributes['sentry.trace.parent_span_id'] = SentryAttribute.string(
-        span.context.spanId.toString(),
-      );
-    }
```
buenaflor, line 554: from internal conversations:

_both versions work in Relay - but the canonical/correct way is to set it as the top level field in the log/metric_

so I'll update this impl

buenaflor:
bugbot review

packages/flutter/lib/src/telemetry/enricher/native_contexts_attributes_provider.dart
```diff
@@ -0,0 +1,63 @@
...
+  NativeContextsTelemetryAttributesProvider(this._nativeBinding);
+
+  @override
+  Future<Map<String, SentryAttribute>> attributes(Object item,
+      {Scope? scope}) async {
+    if (_cachedAttributes != null) {
+      return _cachedAttributes!;
+    }
+
+    final nativeContexts = await _nativeBinding.loadContexts() ?? {};
```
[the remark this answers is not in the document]
buenaflor: Enricher takes care of that

Issue #3334, opened 2025-11-11
[Span First]: Port auto instrumentation


buenaflor:
Auto instrumentation should work ootb with the new Span model when the flag to use the new Span model is enabled

buenaflor:
mobile vitals dashboard has been ported

PR #3447, opened 2026-01-14
feat(span-first): add span v2 support for frames tracking

buenaflor:
## :scroll: Description

Part of span-first

Part of #3334 

Manual, unit/widget test

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps

#skip-changelog


packages/flutter/lib/src/frames_tracking/span_frame_metrics_collector_v2.dart
```diff
@@ -0,0 +1,95 @@
...
+import 'package:meta/meta.dart';
+
+import '../../sentry_flutter.dart';
+import '../utils/internal_logger.dart';
+import 'sentry_delayed_frames_tracker.dart';
+
+/// Collects frames from [SentryDelayedFramesTracker], calculates the metrics
+/// and attaches them to spans.
+@internal
+class SpanFrameMetricsCollectorV2 implements PerformanceContinuousCollectorV2 {
```
buenaflor: this class is mostly copied from the transaction version, just adapted to span v2

denrase: Would it make sense to use the abstraction `InstrumentationSpan` instead, or did the logic also change for v2?

packages/flutter/lib/src/frames_tracking/span_frame_metrics_collector_v2.dart
```diff
@@ -0,0 +1,95 @@
...
+
+  /// Stores the spans that are actively being tracked.
+  /// After the frames are calculated and stored in the span the span is removed from this list.
+  @visibleForTesting
+  final List<SentrySpanV2> activeSpans = [];
+
+  @override
+  Future<void> onSpanStarted(SentrySpanV2 span) async {
+    return _tryCatch('onSpanStarted', () async {
+      if (span is NoOpSentrySpan) {
```
denrase: Yup, should be NoOpSentryV2Span

packages/flutter/lib/src/frames_tracking/span_frame_metrics_collector_v2.dart
```diff
@@ -0,0 +1,95 @@
...
+
+      activeSpans.add(span);
+      _resumeFrameTracking();
+    });
+  }
+
+  @override
+  Future<void> onSpanFinished(SentrySpanV2 span, DateTime endTimestamp) async {
+    return _tryCatch('onSpanFinished', () async {
+      if (span is NoOpSentrySpan) {
```
denrase: Same as above, probably also a good idea to check for the UnsetSpan and throw, if we do this elsewhere.

packages/flutter/lib/src/integrations/frames_tracking_integration.dart
```diff
@@ -50,11 +51,29 @@ class FramesTrackingIntegration implements Integration<SentryFlutterOptions> {
...
-        resumeFrameTracking: () => widgetsBinding.resumeTrackingFrames(),
-        pauseFrameTracking: () => widgetsBinding.pauseTrackingFrames());
-    options.addPerformanceCollector(collector);
-    _collector = collector;
+    switch (options.traceLifecycle) {
+      case SentryTraceLifecycle.streaming:
+        final collector = SpanFrameMetricsCollectorV2(framesTracker,
+            resumeFrameTracking: () => widgetsBinding.resumeTrackingFrames(),
+            pauseFrameTracking: () => widgetsBinding.pauseTrackingFrames());
+        _collector = collector;
```
denrase: We did implement a V2 perfomrce collector, but are not adding it to options, is this correct?

denrase (changes requested):

packages/flutter/lib/src/integrations/frames_tracking_integration.dart
buenaflor: it's a behaviour difference. 

for transactions we add it to the options so we can manually trigger onSpanFinish and onSpanEnded via options.collector.onSpanFinished() but now that we use lifecycle callbacks we dont need to do that anymroe for spanv2 so we dont need to add it to the options

I'll refactor the static version to use lifeecycle callbacks as well

packages/flutter/lib/src/frames_tracking/span_frame_metrics_collector_v2.dart
buenaflor: Updated: I updated the impl to make the unified

packages/flutter/lib/src/frames_tracking/span_frame_metrics_collector_v2.dart
buenaflor: Rather than checking both let's check `if (span is! RecordingSentrySpanV2)`

buenaflor:
> Would it make sense to use the abstraction InstrumentationSpan instead, or did the logic also change for v2?

I checked and probably not, this specific integration requires lots of concrete specific usages, the abstraction wouldn't help here

packages/dart/lib/src/protocol/sentry_span.dart
```diff
@@ -207,7 +211,7 @@ class SentrySpan extends ISentrySpan {
   }
 
   @override
-  bool get finished => _endTimestamp != null;
```
buenaflor, line 210: this change is necessary for the lifecycle callbacks to function correctly. we were triggering the callbacks before the endTimestamp was set (which means we never had access to the endTimestamp) but we could not trigger it after because then the span would be immutable to change in `setData` for example. so this check alone is not enough

packages/flutter/lib/src/frames_tracking/span_frame_metrics_collector_v2.dart
```diff
@@ -0,0 +1,95 @@
...
+        attributes.putIfAbsent(SemanticAttributesConstants.framesTotal,
+            () => SentryAttribute.int(metrics.totalFrameCount));
+        attributes.putIfAbsent(SemanticAttributesConstants.framesSlow,
+            () => SentryAttribute.int(metrics.slowFrameCount));
+        attributes.putIfAbsent(SemanticAttributesConstants.framesFrozen,
+            () => SentryAttribute.int(metrics.frozenFrameCount));
+        attributes.putIfAbsent(SemanticAttributesConstants.framesDelay,
+            () => SentryAttribute.int(metrics.framesDelay));
+        span.setAttributesIfAbsent(attributes);
+      }
```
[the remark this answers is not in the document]
buenaflor: This is indeed an issue, I will fix this in another PR (in the main branch) to keep this streamlined

packages/dart/lib/src/tracing/instrumentation/instrumentation_span.dart
```diff
@@ -180,4 +202,20 @@ class StreamingInstrumentationSpan implements InstrumentationSpan {
     }
     return SentrySpanStatusV2.error;
   }
+
+  @override
+  bool get isNoop => _span is! RecordingSentrySpanV2;
```
buenaflor: I think this check suffices

packages/flutter/lib/src/frames_tracking/span_frame_metrics_collector.dart
```diff
@@ -60,28 +63,47 @@ class SpanFrameMetricsCollector implements PerformanceContinuousCollector {
...
+      internalLogger.warning(
+        'Unknown InstrumentationSpan type: ${span.runtimeType}',
+      );
+    }
+  }
+
   Future<void> _tryCatch(String methodName, Future<void> Function() fn) async {
     try {
-      return fn();
+      return await fn();
```
buenaflor, line 75: await here otherwise errors won't be caught in the catch block

packages/dart/lib/src/protocol/sentry_span.dart
```diff
@@ -51,50 +53,59 @@ class SentrySpan extends ISentrySpan {
...
-    if (finished) {
+    // Prevent concurrent or duplicate finish() calls
+    if (_isFinished || _isFinishing) {
       return;
     }
 
-    if (status != null) {
-      _status = status;
-    }
+    _isFinishing = true;
```
denis-meisterlabs: Do we need a separate `_isFinishing` flag? Isn't this just the state where we have a timestamp, but did not set the `_isFinished` flag yet? If yes, we can just derive it from this condition.

packages/dart/lib/src/tracing/instrumentation/instrumentation_span.dart
```diff
@@ -65,6 +71,22 @@ class LegacyInstrumentationSpan implements InstrumentationSpan {
...
   SentryBaggageHeader? toBaggageHeader() => _span.toBaggageHeader();
+
+  @override
+  bool get isNoop => _span is NoOpSentrySpan;
+
+  @override
+  DateTime get startTimestamp => _span.startTimestamp;
+
+  @override
+  bool operator ==(Object other) =>
```
denis-meisterlabs: Whats with the equality override? Could add some claifying comments here.

packages/flutter/lib/src/frames_tracking/span_frame_metrics_collector.dart
denrase: Good "catch" 🥁

packages/flutter/lib/src/frames_tracking/span_frame_metrics_collector.dart
```diff
@@ -3,34 +3,30 @@
...
   final SentryDelayedFramesTracker _frameTracker;
   final void Function() _resumeFrameTracking;
   final void Function() _pauseFrameTracking;
 
-  /// Stores the spans that are actively being tracked.
-  /// After the frames are calculated and stored in the span the span is removed from this list.
-  @visibleForTesting
-  final List<ISentrySpan> activeSpans = [];
+  /// Spans currently being tracked. Frame tracking pauses when empty.
+  final List<InstrumentationSpan> activeSpans = [];
```
denrase: No more `@visibleForTesting`? I guess it does not matter much, as this is an internal class?

packages/dart/lib/src/tracing/instrumentation/instrumentation_span.dart
denrase, line 84: We should add a short comment on why we are doing this, it's not immediately clear to me. Also for the other span.

packages/dart/lib/src/tracing/instrumentation/instrumentation_span.dart
denrase: Hmm, naming doesn't fit it then, as in this case could also be Unset span, which could be accessed and throw, which we "disguise" a bit here.

What would be a name where we have for legacy: It's noop, and for this: it is not recording (so either noop or unset).

Not sure, maybe just flip it and call it `isvalid/isReccoding/isBeingSmapled` and check if it is false?

No super strong opinion here though.

packages/dart/lib/src/protocol/sentry_span.dart
```diff
@@ -51,50 +53,59 @@ class SentrySpan extends ISentrySpan {
...
-        _hub.options.lifecycleRegistry.dispatchCallback(OnSpanFinish(this));
-    if (callback is Future) {
-      await callback;
-    }
+      _endTimestamp = endTimestamp;
+
+      // ignore: deprecated_member_use_from_same_package
+      for (final collector in _hub.options.performanceCollectors) {
+        if (collector is PerformanceContinuousCollector) {
+          await collector.onSpanFinished(this, endTimestamp);
```
denrase, line 82: Technically incorrect now, we could call this `onSpanwillFinish/onSpanIsFinishing`, so the state is also reflected in the naming.

packages/dart/lib/src/protocol/sentry_span.dart
```diff
@@ -14,6 +14,8 @@ class SentrySpan extends ISentrySpan {
   late final DateTime _startTimestamp;
   final Hub _hub;
 
+  bool _isFinished = false;
+  bool _isFinishing = false;
```
denrase: Do we need a separate `_isFinishing` flag? Isn't this just the state where we have a `timestamp`, but did not set the `_isFinished` flag yet? If yes, we can just derive it from this condition.

packages/dart/lib/src/telemetry/span/unset_sentry_span_v2.dart
```diff
@@ -26,6 +26,9 @@ final class UnsetSentrySpanV2 implements SentrySpanV2 {
   @override
   SentrySpanV2? get parentSpan => _throw();
 
+  @override
+  DateTime get startTimestamp => _throw();
```
denrase, line 30: Seeing that we are not only checking with a bool, could it become an issue in the SDK that every access to this class throws?

Maybe we could just subclass NoOpSpan, or even use NoOp span for the one case where we need the unset check?

denrase (changes requested):

packages/dart/lib/src/protocol/sentry_span.dart
buenaflor: I guess we dont need the collector to be depnedent on these anymore, we could just call it 

start / collectFramesForSpan or something

buenaflor: I guess we dont need the collector to be depnedent on these anymore, we could just call it 

start / collectFrames or something

packages/dart/lib/src/telemetry/span/unset_sentry_span_v2.dart
buenaflor: imo we should at least alert somehow that something went wrong because the unset span should never be exposed like this

instead of throw we could also instead run an assert so it just errors on debug

Noop instead is expected, e.g unsampled spans so it shouldnt error at all - imo we shouldn't subclass from noop since unset != noop

packages/flutter/lib/src/frames_tracking/span_frame_metrics_collector.dart
buenaflor: yeah its internal anyway

packages/dart/lib/src/protocol/sentry_span.dart
buenaflor: I renamed it

packages/dart/lib/src/telemetry/span/unset_sentry_span_v2.dart
buenaflor: I would address this in another PR

packages/dart/lib/src/tracing/instrumentation/instrumentation_span.dart
buenaflor: I think `isRecording` is good, that's what JS uses as well (and I think OTEL in general)

denrase (approved):

PR #3448, opened 2026-01-14
Move `TelemetryProcessor` from span-first branch and replace `LogBatcher`

buenaflor:
No implementations added, only removed things that dont exist yet in the main branch which is the new span model

this allows us to work on metrics sooner


packages/dart/lib/src/sentry_client.dart
```diff
@@ -75,9 +74,6 @@ class SentryClient {
     if (enableFlutterSpotlight) {
       options.transport = SpotlightHttpTransport(options, options.transport);
     }
-    if (options.enableLogs) {
-      options.logBatcher = SentryLogBatcher(options);
```
denrase, line 79: We can probably remove the `SentryLogBatcher` and tests. If we need to restore, we can go back to this PR. WDYT?

denrase (approved):
Lookin good, just one suggestion.

denrase:
We also need to check why some CI checks are red.

packages/dart/lib/src/sentry_client.dart
buenaflor: yeah let's do it

buenaflor: I'll remove it

buenaflor:
@denrase double checked and red CI is unrelated to this PR

PR #3450, opened 2026-01-15
feat: trace connected metrics

### LINKED ISSUE #3341: [Metrics]: Add new Metrics API
[https://develop.sentry.dev/sdk/telemetry/metrics/#public-api](https://develop.sentry.dev/sdk/telemetry/metrics/#public-api)

buenaflor:
## :scroll: Description

Adds trace connected metrics feature to Dart

Closes #3341 

Manual and unit tests

image

image

### Protocol ✅
| Requirement | Status |
|-------------|--------|
| Envelope type `trace_metric` | ✅ `SentryItemType.metric = 'trace_metric'` |
| Content-type `application/vnd.sentry.items.trace-metric+json` | ✅ |
| Format `{"items": [...]}` | ✅ |
| Header `item_count` | ✅ |

### Metric Payload ✅
| Field | Status |
|-------|--------|
| `timestamp` (seconds) | ✅ `millisecondsSinceEpoch / 1000.0` |
| `trace_id` (32-char hex) | ✅ `SentryId.toString()` |
| `span_id` (16-char hex, optional) | ✅ `SpanId.toString()` |
| `name`, `value`, `type` | ✅ |
| `unit` (optional) | ✅ |
| `attributes` (typed, optional) | ✅ |

### Public API ✅
| Method | Status |
|--------|--------|
| `count(name, value, {attributes, scope})` | ✅ |
| `gauge(name, value, {unit, attributes, scope})` | ✅  |
| `distribution(name, value, {unit, attributes, scope})` | ✅ |

### Configuration ✅
| Option | Spec | Implementation |
|--------|------|----------------|
| `enableMetrics` default | `true` | ✅ |
| `beforeSendMetric` | optional | ✅ |

### Default Attributes ✅
- `sentry.environment`, `sentry.release`, `sentry.sdk.name`, `sentry.sdk.version` ✅
- `sentry.replay_id`, `sentry._internal.replay_is_buffering` ✅ (via lifecycle hooks)
- User attributes (`user.id`, `user.name`, `user.email`) guarded by `sendDefaultPii` ✅

### SDK Behavior ✅
- Check disabled → scope attributes → lifecycle dispatch → default attributes → beforeSendMetric → buffer ✅

### Data Category & Rate Limiting ✅
- `trace_metric` → `DataCategory.metric` ✅

### Buffering & Debug Logging ✅
- Uses `InMemoryTelemetryBuffer` ✅
- Debug logging via `internalLogger.debug()` ✅


packages/dart/lib/src/sentry_envelope.dart
```diff
@@ -160,6 +175,20 @@ class SentryEnvelope {
...
+    builder.add(utf8.encode('{"items":['));
+    for (int i = 0; i < encodedItems.length; i++) {
+      if (i > 0) {
+        builder.add(utf8.encode(','));
+      }
+      builder.add(encodedItems[i]);
+    }
+    builder.add(utf8.encode(']}'));
+    return builder.takeBytes();
+  }
```
buenaflor, line 174: this is taken from the span-first branch

packages/dart/lib/src/utils.dart
```diff
@@ -34,3 +36,66 @@ Object? jsonSerializationFallback(Object? nonEncodable) {
...
 }
+
+@internal
+extension AddAllAbsentX<K, V> on Map<K, V> {
+  void addAllIfAbsent(Map<K, V> other) {
+    for (final e in other.entries) {
+      putIfAbsent(e.key, () => e.value);
+    }
+  }
+}
```
buenaflor, line 47: from span-first telemetry enrichment

packages/dart/lib/src/telemetry/metric/default_metrics.dart
```diff
@@ -0,0 +1,100 @@
...
+
+  @override
+  void count(
+    String name,
+    int value, {
+    Map<String, SentryAttribute>? attributes,
+    Scope? scope,
+  }) {
+    internalLogger.debug(() =>
+        'Sentry.metrics.count("$name", $value) called with attributes ${_formatAttributes(attributes)}');
```
buenaflor, line 29: I'm using lazy evaluation `(() =>` for these since these formattings can be potentially heavy when the user has many attributes

packages/dart/lib/src/telemetry/metric/metrics_setup_integration.dart
```diff
@@ -0,0 +1,31 @@
...
+    }
+
+    if (options.metrics is! NoOpSentryMetrics) {
+      internalLogger.debug(
+          '$integrationName: Custom metrics already configured, skipping setup');
+      return;
+    }
+
+    options.metrics = DefaultSentryMetrics(
+        captureMetricCallback: hub.captureMetric,
```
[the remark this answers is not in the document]
buenaflor: added try catch to captureMetric

packages/dart/lib/src/telemetry/default_attributes.dart
```diff
@@ -0,0 +1,55 @@
+import '../../sentry.dart';
+import '../utils/os_utils.dart';
+
+final _operatingSystem = getSentryOperatingSystem();
+
+Map<String, SentryAttribute> defaultAttributes(SentryOptions options,
```
buenaflor, line 6: from span-first branch, this is used by all telemetry pipelines

packages/dart/lib/src/telemetry/metric/metric_capture_pipeline.dart
```diff
@@ -0,0 +1,73 @@
...
+    if (!_options.enableMetrics) {
+      internalLogger.debug(
+          '$MetricCapturePipeline: Metrics disabled, dropping ${metric.name}');
+      return;
+    }
+
+    try {
+      if (scope != null) {
+        metric.attributes.addAllIfAbsent(scope.attributes);
+      }
```
buenaflor, line 25: merge scope attributes before proceeding with the rest

packages/dart/lib/src/telemetry/metric/metric.dart
```diff
@@ -0,0 +1,81 @@
...
+/// The metrics telemetry.
+abstract class SentryMetric {
+  final String type;
+
+  DateTime timestamp;
+  String name;
+  num value;
+  SentryId traceId;
+  SpanId? spanId;
+  String? unit;
```
buenaflor, line 31: according to spec we must not constraint this with an enum and instead the user needs to know which unit types we support

packages/dart/lib/src/transport/data_category.dart
```diff
@@ -29,6 +30,8 @@ enum DataCategory {
         return DataCategory.metricBucket;
       case 'log':
         return DataCategory.logItem;
+      case 'trace_metric':
+        return DataCategory.metric;
```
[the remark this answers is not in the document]
buenaflor: Fixed

packages/dart/lib/src/client_reports/discarded_event.dart
```diff
@@ -70,6 +70,8 @@ extension _DataCategoryExtension on DataCategory {
         return 'feedback';
       case DataCategory.metricBucket:
         return 'metric_bucket';
+      case DataCategory.metric:
```
denrase, line 73: All other types are called like the raw value, so in this case it should be 'traceMetric'

packages/dart/lib/src/telemetry/metric/default_metrics.dart
```diff
@@ -0,0 +1,100 @@
...
+        attributes: attributes ?? {});
+
+    _captureMetricCallback(metric);
+  }
+
+  @override
+  void gauge(
+    String name,
+    num value, {
+    String? unit,
```
denrase, line 46: Is unit a closed set of values? If so, we could provide an enum or similar.

packages/dart/lib/src/telemetry/metric/metric.dart
denrase: Then ignore/resolve my previous comment. 🙇‍♂️

packages/dart/lib/src/telemetry/metric/metrics.dart
```diff
@@ -0,0 +1,26 @@
+import '../../../sentry.dart';
+
+/// Interface for emitting custom metrics to Sentry.
+///
+/// Access via [Sentry.metrics].
+abstract interface class SentryMetrics {
+  /// Increments a counter metric by the given [value].
+  ///
+  /// Use counters to track the number of times an event occurs.
+  void count(String name, int value,
```
denrase: Does it make sense to also expose the Future, or should this API be strictly fire & forget?

packages/dart/lib/src/telemetry/metric/noop_metrics.dart
```diff
@@ -0,0 +1,19 @@
+import '../../../sentry.dart';
+
+final class NoOpSentryMetrics implements SentryMetrics {
+  const NoOpSentryMetrics();
+
+  static const instance = NoOpSentryMetrics();
```
denrase: Do we need to always keep this instance around?

packages/dart/lib/src/telemetry/processing/processor.dart
```diff
@@ -27,46 +31,70 @@ abstract class TelemetryProcessor {
...
-    if (logBuffer == null) {
+    if (_logBuffer == null) {
       internalLogger.warning(
         '$runtimeType: No buffer registered for ${log.runtimeType} - item was dropped',
       );
       return;
     }
 
-    logBuffer!.add(log);
+    _logBuffer.add(log);
```
denrase, line 60: Nice that the compiler knowns that this property value is non-null because of the check, when we'd have to force unwrap before with the computed property.

packages/dart/lib/src/telemetry/default_attributes.dart
denrase: I was already wondering why i'm seeing this added so often. :D

packages/dart/lib/src/sentry_envelope.dart
```diff
@@ -160,6 +175,20 @@ class SentryEnvelope {
     }
   }
 
+  /// Builds a payload in the format {"items": [item1, item2, ...]}
+  static Uint8List _buildItemsPayload(List<List<int>> encodedItems) {
```
denrase, line 163: This should also be shared by `SentryEnvelope.fromLogsData` method.

packages/dart/lib/src/sentry_item_type.dart
```diff
@@ -6,5 +6,6 @@ class SentryItemType {
   static const String profile = 'profile';
   static const String statsd = 'statsd';
   static const String log = 'log';
+  static const String metric = 'trace_metric';
```
denrase, line 9: If we want to update naming, this should also be `traceMetric`?

packages/dart/lib/src/sentry_options.dart
```diff
@@ -545,13 +551,21 @@ class SentryOptions {
   /// Disabled by default.
   bool enableLogs = false;
 
+  /// Enable to capture and send metrics to Sentry.
+  ///
+  /// Enabled by default.
+  bool enableMetrics = true;
```
denrase, line 556: I'm not sure if we did this in Dart, but I remember that we moved logs to experimental options before making it GA. We should consider this here as well.

packages/dart/lib/src/sentry_options.dart
```diff
@@ -545,13 +551,21 @@ class SentryOptions {
...
+
   /// Enables adding the module in [SentryStackFrame.module].
   /// This option only has an effect in non-obfuscated builds.
   /// Enabling this option may change grouping.
   bool includeModuleInStackTrace = false;
 
   late final SentryLogger logger = SentryLogger(clock);
 
+  @internal
+  SentryMetrics metrics = NoOpSentryMetrics.instance;
```
denrase: Might as well just create the instance here once, then we don't have to keep it around when it gets replaced. Could also consider making this `late`.

packages/dart/test/mocks/mock_metric_capture_pipeline.dart
```diff
@@ -0,0 +1,18 @@
+import 'package:sentry/sentry.dart';
+import 'package:sentry/src/telemetry/metric/metric.dart';
+import 'package:sentry/src/telemetry/metric/metric_capture_pipeline.dart';
+
+class FakeMetricCapturePipeline extends MetricCapturePipeline {
```
denrase: Naming: MockMetricCapturePipeline.

packages/dart/test/mocks/mock_metric_capture_pipeline.dart
```diff
@@ -0,0 +1,18 @@
...
+  FakeMetricCapturePipeline(super.options);
+
+  int callCount = 0;
+  SentryMetric? capturedMetric;
```
denrase: Would make sense to do an array of `CaptureMetricCall(metric, scope)`, that way we don't love all previous to the current one.

denrase: Not super important if we don't use it in any test case.

packages/dart/test/mocks/mock_sentry_client.dart
```diff
@@ -11,6 +12,7 @@ class MockSentryClient with NoSuchMethodProvider implements SentryClient {
   List<CaptureTransactionCall> captureTransactionCalls = [];
   List<CaptureFeedbackCall> captureFeedbackCalls = [];
   List<CaptureLogCall> captureLogCalls = [];
+  List<CaptureMetricCall> captureMetricCalls = [];
```
denrase, line 14: Could use this type in the Fake/Mock pipeline.

packages/dart/test/sentry_options_test.dart
```diff
@@ -183,4 +183,10 @@ void main() {
 
     expect(() => options.parsedDsn, throwsA(isA<StateError>()));
   });
+
+  test('enableMetrics is true by default', () {
```
denrase, line 187: No experimental phase or opt-ointment like with logs?

packages/flutter/lib/src/integrations/load_contexts_integration.dart
```diff
@@ -54,55 +55,78 @@ class LoadContextsIntegration extends Integration<SentryFlutterOptions> {
...
               stackTrace: stackTrace,
             );
           }
         },
       );
     }
+
+    if (options.enableMetrics) {
+      options.lifecycleRegistry
+          .registerCallback<OnProcessMetric>((event) async {
```
denrase: For logs we are registering for OnBefore capture, why do we do it for onProcess for metrics? Shouldn't both add attributes at the same time in their journey through the SDK?

packages/flutter/lib/src/integrations/replay_telemetry_integration.dart
```diff
@@ -0,0 +1,97 @@
...
+@internal
+class ReplayTelemetryIntegration implements Integration<SentryFlutterOptions> {
+  static const String integrationName = 'ReplayTelemetry';
+
+  final SentryNativeBinding? _native;
+  ReplayTelemetryIntegration(this._native);
+
+  SentryFlutterOptions? _options;
+  SdkLifecycleCallback<OnBeforeCaptureLog>? _onBeforeCaptureLog;
+  SdkLifecycleCallback<OnProcessMetric>? _onProcessMetric;
```
denrase, line 19: Same here, why does log use OnBeforeCapture (waiting after default auttributes were added), on metrics use `OnProcess`?

packages/dart/lib/src/telemetry/metric/default_metrics.dart
```diff
@@ -0,0 +1,100 @@
+import '../../../sentry.dart';
+import '../../utils/internal_logger.dart';
+import 'metric.dart';
+
+typedef CaptureMetricCallback = void Function(SentryMetric metric);
+typedef ScopeProvider = Scope Function();
+
+final class DefaultSentryMetrics implements SentryMetrics {
```
denrase, line 9: We could also think about refactoring log class `SentryLog` into an interface (abstract class) with noop and default implementation to have an aligned architecture. no need to do this here though.

denrase (changes requested):
Looks great overall, some comments regarding naming/api/callbacks timing. 🥳

packages/dart/lib/src/client_reports/discarded_event.dart
buenaflor: We should keep it as metrics as per spec:

> The envelope item type is named trace_metric for internal usage to avoid naming collisions with other metric systems within Sentry's infrastructure. From an SDK perspective, these are simply referred to as "metrics".

packages/dart/lib/src/telemetry/metric/metrics.dart
buenaflor: imo it should just be a fire and forget (many other SDKs do it like that as well) 

Future adds a lot of noise since it could get flagged for some people as a warning with the analyzer

packages/dart/lib/src/telemetry/metric/default_metrics.dart
buenaflor: according to spec the SDKs shouldn't constraint them

> SDKs must not restrict what can be sent in (e.g. by using an enum) but should offer constants or similar that help customers send in [units we support](https://develop.sentry.dev/sdk/telemetry/attributes/#units).

buenaflor: but yeah I think we should add these constants to the public api

packages/flutter/lib/src/integrations/load_contexts_integration.dart
buenaflor: the `OnBeforeCapture` usage is actually not correct for logs since it runs after beforeSend

in the next PR I would also replace that with an `OnProcessLog` 

OnProcess would run before the default telemetry attributes enrichment so more specific enrichments can be added first and we dont need to do this removal of integrations anymore

packages/dart/test/mocks/mock_metric_capture_pipeline.dart
buenaflor: tbh it's a bit of a pet peeve since these are actually not mocks but fakes, maybe I should just rename the file?

buenaflor: but whatever, I'll just do it as mock it's for test code only anyway

packages/dart/lib/src/sentry_item_type.dart
buenaflor: I think we can keep it as metric:

> The envelope item type is named trace_metric for internal usage to avoid naming collisions with other metric systems within Sentry's infrastructure. From an SDK perspective, these are simply referred to as "metrics".

packages/dart/test/mocks/mock_metric_capture_pipeline.dart
buenaflor: I'll update it

packages/dart/lib/src/sentry_options.dart
buenaflor: Was following the specs but yeah we can publish it as experimental first

packages/flutter/lib/src/integrations/load_contexts_integration.dart
buenaflor: I didn't wanna update that for logs right now so I don't explode this PR

packages/flutter/lib/src/integrations/replay_telemetry_integration.dart
buenaflor: answered above

packages/dart/lib/src/sentry_envelope.dart
buenaflor: Done

packages/dart/test/mocks/mock_sentry_client.dart
buenaflor: thx, updated

packages/dart/lib/src/sentry_options.dart
buenaflor: double checked and we can define it as non-experiemntal

packages/dart/lib/src/telemetry/metric/default_metrics.dart
buenaflor: added the constants and updated the public API docs to reference them

packages/dart/lib/src/telemetry/metric/default_metrics.dart
```diff
@@ -0,0 +1,100 @@
...
+
+    final metric = SentryCounterMetric(
+        timestamp: _clockProvider(),
+        name: name,
+        value: value,
+        spanId: _activeSpanIdFor(scope),
+        traceId: _traceIdFor(scope),
+        attributes: attributes ?? {});
+
+    _captureMetricCallback(metric);
```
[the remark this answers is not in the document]
buenaflor: no

packages/flutter/lib/src/integrations/load_contexts_integration.dart
buenaflor: Opened another PR for log specific changes

packages/dart/lib/src/telemetry/metric/default_metrics.dart
buenaflor: Opened a new PR for that

packages/dart/lib/src/telemetry/metric/metric.dart
```diff
@@ -0,0 +1,182 @@
...
+    super.unit,
+    super.attributes,
+  }) : super(type: 'distribution');
+}
+
+/// String constants for metric units.
+///
+/// These constants represent the API names of measurement units that can be
+/// used with metrics.
+abstract final class SentryMetricUnit {
```
denrase, line 110: ❤️

packages/flutter/lib/src/integrations/load_contexts_integration.dart
denrase: That makes sense, thank you!

denrase (approved):
LGTM 👍

PR #3453, opened 2026-01-15
fix(ci): failing package-analysis

buenaflor:
issues:
- the run script that's supposed to swap the sentry dart path did not trigger; happened when we moved the packages from root to `/packages`
- invalid dependency was not ignored for all analysis rules
- the run script should execute for all targets that are not sentry-dart, not only for flutter

#skip-changelog

### DIFF
--- a/.github/workflows/analyze.yml
+++ b/.github/workflows/analyze.yml
@@ -12,7 +12,7 @@ on:
         description: Minumum percentage of Dart Package Analyzer score that must be achieved.
         required: false
         type: number
-        default: 100
+        default: 160
 
 jobs:
   analyze:
@@ -74,8 +74,8 @@ jobs:
       - name: Check scores
         env:
           TOTAL: ${{ steps.analysis.outputs.total }}
-          TOTAL_MAX: 150
-          PANA_THRESHOLD: 100
+          TOTAL_MAX: ${{ steps.analysis.outputs.total_max }}
+          PANA_THRESHOLD: ${{ inputs.panaThreshold }}
         run: |
           PERCENTAGE=$(( $TOTAL * 100 / $TOTAL_MAX ))
           if (( $PERCENTAGE < $PANA_THRESHOLD ))
--- a/.github/workflows/analyze.yml
+++ b/.github/workflows/analyze.yml
@@ -80,8 +80,8 @@ jobs:
       - name: Check scores
         env:
           TOTAL: ${{ steps.analysis.outputs.total }}
-          TOTAL_MAX: ${{ steps.analysis.outputs.total_max }}
-          PANA_THRESHOLD: ${{ inputs.panaThreshold }}
+          TOTAL_MAX: 70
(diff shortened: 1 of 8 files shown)


.github/workflows/analyze.yml
```diff
@@ -9,10 +9,10 @@ on:
         type: string
         default: dart
       panaThreshold:
-        description: Minumum percentage of Dart Package Analyzer score that must be achieved.
+        description: Minimum number of points that must be achieved.
         required: false
         type: number
-        default: 100
+        default: 140
```
buenaflor, line 15: use total number of points instead of percentages, at least 140 out of 160

.github/workflows/analyze.yml
```diff
@@ -77,9 +77,10 @@ jobs:
           TOTAL_MAX: ${{ steps.analysis.outputs.total_max }}
           PANA_THRESHOLD: ${{ inputs.panaThreshold }}
         run: |
-          PERCENTAGE=$(( $TOTAL * 100 / $TOTAL_MAX ))
-          if (( $PERCENTAGE < $PANA_THRESHOLD ))
-          then
-            echo "Score too low ($PERCENTAGE % is less than the expected $PANA_THRESHOLD %)!"
+          MISSING_POINTS=$(( TOTAL_MAX - TOTAL ))
+          if (( MISSING_POINTS > PANA_THRESHOLD )); then
```
denrase: The way I understand it, the PANA_THRESHOLD ist the upper bound we have to reach.
`Overall = 160`
`Threshold = 140`

Say we have only reached 60 Points, we'd be missing 100 (160-60).
`Missing = 100`

But the check is looking if missing is larger than the threshold, which would not trigger here.

`100 > 140 == false`

Is this what we want, shouldn't it be something like this:

```yaml  
MISSING_POINTS=$(( TOTAL_MAX - TOTAL ))
ALLOWED_MISSING=$(( TOTAL_MAX - PANA_THRESHOLD ))
if (( MISSING_POINTS > ALLOWED_MISSING )); then
  echo "Pana score check failed: ${TOTAL}/${TOTAL_MAX} points (missing ${MISSING_POINTS}; allowed missing <= ${ALLOWED_MISSING}, minimum required: ${PANA_THRESHOLD})."
  exit 1
else
  echo "Pana score check passed: ${TOTAL}/${TOTAL_MAX} points (missing ${MISSING_POINTS}; allowed missing <= ${ALLOWED_MISSING}, minimum required: ${PANA_THRESHOLD})."
fi
````

denrase (changes requested):

.github/workflows/analyze.yml
buenaflor: yeah that's wrong, I'll update it, bugbot should've flagged this

buenaflor: we can just compare the TOTAL directly, no need to calculate any missing points

denrase (approved):

PR #3457, opened 2026-01-19
feat(span-first): telemetry enrichment

buenaflor:
## :scroll: Description

Part of span first

#3332 

Unit test, manual tests

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps
#skip-changelog


buenaflor:
bugbot review

buenaflor:
bugbot review

packages/dart/lib/src/telemetry/span/span_capture_pipeline.dart
```diff
@@ -0,0 +1,48 @@
...
+@internal
+class SpanCapturePipeline {
+  final SentryOptions _options;
+
+  SpanCapturePipeline(this._options);
+
+  Future<void> captureSpan(SentrySpanV2 span, Scope scope) async {
+    if (_options.traceLifecycle == SentryTraceLifecycle.static) {
+      internalLogger.warning(
+        'captureSpan: invalid usage with traceLifecycle static, skipping capture.',
```
buenaflor, line 17: why another abstraction
- sentry client is **huge**, the test file has 3k loc, this allow us to decouple testing the impl and have sentry_client only be the orchestrator
- this will take care of beforeSend, client reports and eventually other things related to capturing
- with this we then want to have separate capture pipelines -> LogCapturePipeline, MetricCapturePipeline, ErrorCapturePipeline (in the future)

packages/dart/lib/src/sdk_lifecycle_hooks.dart
```diff
@@ -96,3 +97,13 @@ class OnSpanFinish extends SdkLifecycleEvent {
...
 }
+
+/// Dispatched when span is ready for processing (before default enrichment).
+///
+/// This is useful for integrations to hook into e.g for enriching with attributes.
+@internal
+class ProcessSpan extends SdkLifecycleEvent {
+  final SentrySpanV2 span;
+
+  ProcessSpan(this.span);
```
buenaflor: inspiration from JS SDK where they have `ProcessMetric`, `ProcessLog`

packages/dart/lib/src/hub.dart
```diff
@@ -685,7 +685,7 @@ class Hub {
     return span;
   }
 
-  void captureSpan(SentrySpanV2 span) {
+  Future<void> captureSpan(SentrySpanV2 span) async {
```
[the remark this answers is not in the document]
buenaflor: fixed

packages/dart/lib/src/sdk_lifecycle_hooks.dart
```diff
@@ -96,3 +97,13 @@ class OnSpanFinish extends SdkLifecycleEvent {
 
   final ISentrySpan span;
 }
+
+/// Dispatched when span is ready for processing (before default enrichment).
+///
+/// This is useful for integrations to hook into e.g for enriching with attributes.
+@internal
+class ProcessSpan extends SdkLifecycleEvent {
```
denrase: For the other events we can figure out fron where/when it is dispatched by naming. So here it would be something like `OnBeforeSpanProcessed`, or do you want to keep this aligned with other SDKs?

packages/dart/lib/src/sentry_client.dart
```diff
@@ -43,14 +44,18 @@ String get defaultIpAddress => _defaultIpAddress;
...
   static final _emptySentryId = Future.value(SentryId.empty());
 
   SentryExceptionFactory get _exceptionFactory => _options.exceptionFactory;
   SentryStackTraceFactory get _stackTraceFactory => _options.stackTraceFactory;
 
   /// Instantiates a client using [SentryOptions]
-  factory SentryClient(SentryOptions options) {
+  factory SentryClient(
+    SentryOptions options, {
+    SpanCapturePipeline? spanCapturePipeline,
```
denrase, line 57: What's the case where this can be optional, or tis this just for testing?

packages/dart/test/mocks/mock_sentry_client.dart
```diff
@@ -93,7 +93,7 @@ class MockSentryClient with NoSuchMethodProvider implements SentryClient {
   }
 
   @override
-  void captureSpan(SentrySpanV2 span, {Scope? scope}) {
+  Future<void> captureSpan(SentrySpanV2 span, Scope scope) async {
```
denrase: Every other capture call has optional scope, do we want this to be non-optional so users calling this on client will always have to provide a scope param?

denrase (changes requested):

packages/dart/lib/src/sentry_client.dart
buenaflor: testing but we can make it a required parameter as well, I don't mind either

packages/dart/lib/src/sdk_lifecycle_hooks.dart
buenaflor: we don't need to be strictly aligned, I don't think `OnBeforeSpanProcessed` is totally right either though because the way we use it is the processing itself. 

maybe `OnSpanProcessReady` or something? indicating that it's ready for processing

otherwise I think `ProcessXYZ` is not too bad, to be consistent we could also do `OnProcessXYZ`

packages/dart/test/mocks/mock_sentry_client.dart
buenaflor: the general idea is that at this point there has to be a resolved scope of some sorts, but on the other hand we can just make it nullable and keep it consistent

packages/dart/lib/src/sdk_lifecycle_hooks.dart
buenaflor: for more context we want to generally move away from event processor and utilize these hooks more

packages/dart/lib/src/sentry_client.dart
buenaflor: but imo probably it's easier to make it nullable and have a default impl in sentry_client since it's a lot of noise in tests

packages/dart/lib/src/telemetry/span/span_capture_pipeline.dart
denrase: We should handle nullable scope here as well.

denrase (approved):
Looks good, pls check the checks (🥁 ) and CI 🙇

packages/dart/lib/src/telemetry/span/span_capture_pipeline.dart
buenaflor: yup updated it 👍

packages/dart/lib/src/sdk_lifecycle_hooks.dart
denrase: Whatever you think is best, as long as it's consistent with other naming we're using so far. 👍

PR #3463, opened 2026-01-20
refactor(log): make implementation consistent with metrics and span-first

buenaflor:
## :scroll: Description
Refactor the implementations to use the new abstractions

- Add new capture pipeline
- Use `OnProcessLog` instead of `OnBeforeCaptureLog` for the attribute enrichments
- Remove `LogsEnricherIntegration` -> moved to defaultAttributes

Manual and unit tests

## :pencil: Checklist

- [ ] All tests passing
- [ ] No breaking changes


buenaflor:
bugbot review

packages/dart/lib/src/protocol/sentry_log.dart
```diff
@@ -14,17 +16,19 @@ class SentryLog {
   /// by the time processing completes, it is guaranteed to be a valid non-empty trace id.
   SentryLog({
     required this.timestamp,
-    SentryId? traceId,
+    required this.traceId,
```
denrase: Woudn't this be a breaking change?

packages/dart/lib/src/telemetry/log/default_logger.dart
```diff
@@ -0,0 +1,302 @@
...
+        attributes: attributes, scope: scope);
+  }
+
+  @override
+  FutureOr<void> fatal(
+    String body, {
+    Map<String, SentryAttribute>? attributes,
+    Scope? scope,
+  }) {
+    internalLogger.debug(() =>
```
denrase: We could do these debug logger calls once in `_captureLog` and discriminate by the type.

packages/dart/lib/src/telemetry/log/log_capture_pipeline.dart
```diff
@@ -0,0 +1,79 @@
...
+      if (processedLog == null) {
+        _options.recorder
+            .recordLostEvent(DiscardReason.beforeSend, DataCategory.logItem);
+        internalLogger.debug(
+            '$LogCapturePipeline: Log "${log.body}" dropped by beforeSendLog');
+        return;
+      }
+
+      _options.telemetryProcessor.addLog(processedLog);
+      internalLogger.debug(
```
denrase: Should we also debug with a closure here, as this migth get called very often?

packages/dart/lib/src/telemetry/log/logger.dart
```diff
@@ -0,0 +1,108 @@
...
+
+/// Interface for emitting custom logs to Sentry.
+///
+/// Access via [Sentry.logger].
+abstract interface class SentryLogger {
+  /// Logs a message at TRACE level.
+  FutureOr<void> trace(
+    String body, {
+    Map<String, SentryAttribute>? attributes,
+    Scope? scope,
```
denrase: Adding scope to all log methods is intentional? This is new API, right?

denrase (changes requested):
We should check wether we want to change the cor of a log, since it's public API AFAIK.

packages/dart/lib/src/protocol/sentry_log.dart
buenaflor: technically yes, but on the other side creating logs manually is not a typical use case and it's technically wrong since traceIds should never be null in the first place

not sure about this one, I guess if we wanna be strict I can just mark this as a TODO for v10

packages/dart/lib/src/telemetry/log/log_capture_pipeline.dart
buenaflor: updated it, I moved the log instead to the processor (and with a closure)

packages/dart/lib/src/telemetry/log/logger.dart
buenaflor: ~It should have been part of the API - checked other SDKs and noticed we didn't have it~

I double checked, some SDKs have it, some don't

I removed it for now for the sake of "when we have a use case we will add it"

packages/dart/lib/src/telemetry/log/log.dart
```diff
@@ -14,17 +16,20 @@ class SentryLog {
...
     required this.attributes,
+    this.spanId,
     this.severityNumber,
   }) : traceId = traceId ?? SentryId.empty();
 
   Map<String, dynamic> toJson() {
     return {
       'timestamp': timestamp.toIso8601String(),
       'trace_id': traceId.toString(),
+      if (spanId != null) 'span_id': spanId.toString(),
```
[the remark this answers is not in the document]
buenaflor: it's guaranteed to be not null

denrase (approved):

packages/dart/lib/sentry.dart
```diff
@@ -60,7 +60,6 @@ export 'src/utils/tracing_utils.dart';
 export 'src/utils/url_details.dart';
 // ignore: invalid_export_of_internal_element
 export 'src/utils/breadcrumb_log_level.dart';
-export 'src/sentry_logger.dart';
```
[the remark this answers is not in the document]
buenaflor: it's fine, this use case is rare and we mark the logger instance internal anyway

PR #3472, opened 2026-01-22
feat(flutter): Add captureNativeFailedRequests option for iOS/macOS

### LINKED ISSUE #3321: How to disable enableCaptureFailedRequests for iOS
### Problem Statement

I would like to disable the option "enableCaptureFailedRequests" on the iOS SDK when using Flutter. My guess is that because the option is not proxied to Dart, I need to rely on `autoInitializeNativeSdk=false`. 
My concern is that I haven't found any example app that uses the auto init as false, so I'm not sure how should I structure the app to ensure the native SDK is running before the Flutter engine starts (this is a warning in the option documentation)

buenaflor:
## :scroll: Description

Add a new `captureNativeFailedRequests` option to control native HTTP failed request capturing independently from `captureFailedRequests`.

This separation allows users to:
- Keep Dart-side failed request capturing enabled while disabling native
- Or vice versa, based on their needs

For backwards compatibility, `captureNativeFailedRequests` defaults to `null`, which falls back to the value of `captureFailedRequests`. This ensures existing users who set `captureFailedRequests = false` will still have native capturing disabled as expected.

Closes #3321 

Unit test

## :pencil: Checklist

- [ ] All tests passing
- [x] No breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@
     - `Sentry.metrics.gauge(...)`
     - `Sentry.metrics.count(...)`
     - `Sentry.metrics.distribution(...)`
+- Add `captureNativeFailedRequests` option for iOS/macOS ([#3472](https://github.com/getsentry/sentry-dart/pull/3472))
+  - This option allows controlling native HTTP error capturing independently from `captureFailedRequests`.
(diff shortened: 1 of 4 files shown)


denrase (approved):
LGTM 👍

PR #3483, opened 2026-01-27
feat(span): add beforeSendSpan callback for span data scrubbing

buenaflor:
## Summary
- Specs: https://develop.sentry.dev/sdk/telemetry/spans/scrubbing-data/
- Add `beforeSendSpan` callback to `SentryOptions` for modifying span data before sending
- Unlike other `beforeSend` callbacks, this cannot drop spans - it's intended for scrubbing sensitive data or PII
- Add tests for the callback behavior
- Add example usage in the Flutter example app

## Test plan
- [x] Unit tests verify callback is called with span
- [x] Unit tests verify span attributes can be modified
- [x] Unit tests verify sensitive attributes can be removed
- [x] Unit tests verify async callbacks work
- [x] Unit tests verify callback ordering (after lifecycle callbacks, after default attributes)
- [x] Example app compiles with the new callback

Part of #3332 

#skip-changelog


packages/dart/lib/src/sentry_options.dart
```diff
@@ -734,6 +742,10 @@ typedef BeforeSendLogCallback = FutureOr<SentryLog?> Function(SentryLog log);
 typedef BeforeSendMetricCallback = FutureOr<SentryMetric?> Function(
     SentryMetric metric);
 
+/// This function is called right before a span is about to be sent.
+typedef BeforeSendSpanCallback = FutureOr<SentrySpanV2> Function(
+    SentrySpanV2 span);
```
buenaflor: alternatively we can just not have the user return a span because the user can modify the span directly but then we'd kinda go against the conventions for `beforeSendXYZ`, the specs are not very strict here. we only need to ensure that the user cannot drop the span

denrase: I think just having the span, without returning it would be more suitable for this API, since user could just return a completely different span, which would break the spec.

> This callback MUST NOT allow the removal of any spans from the span tree. It receives a deep copy of all spans in the span tree and their attributes.

I also worry that users could change status of individual spans, as this is allowed in the span v2 API.

packages/dart/test/telemetry/span/span_capture_pipeline_test.dart
```diff
@@ -146,6 +146,111 @@ void main() {
 
         expect(fixture.processor.addedSpans, isEmpty);
       });
+
```
denrase, line 148: We should also test what users are not allowed to do, eg return unrelated spans, change properties that are not allowed. Should also document this the best we can. Do other SDKs provide API to traverse the span tree (access children, not just parent span)?

denrase (changes requested):

packages/dart/lib/src/sentry_options.dart
denrase: Maybe we could introduce a container, where users can only access what they are allowed, and we can merge it into the span hierarchy after, like the spec describes.

```dart
// Terrible naming, but you get the gist.
class ScrubbableSpan {
  ScrubbableSpan get parent;
  List<ScrubbableSpan> children;
  
  void setAttribute(String key, SentryAttribute value);
  void removeAttribute(String key);
}

File: packages/dart/test/telemetry/span/span_capture_pipeline_test.dart (hunk 2 above)
Comment (buenaflor) [reply packages/dart/test/telemetry/span/span_capture_pipeline_test.dart:148]: >  Do other SDKs provide API to traverse the span tree (access children, not just parent span)?

This span tree traversal and deep span copy section from the spec is actually quite old and not up-to-date anymore, we don't need that

We should just expose the individual span

File: packages/dart/test/telemetry/span/span_capture_pipeline_test.dart [hunk 3]
```diff
@@ -146,6 +146,111 @@ void main() {
...
+        };
+
+        final span = fixture.createRecordingSpan(name: 'original-name');
+        await fixture.pipeline.captureSpan(span, scope: fixture.scope);
+
+        expect(span.attributes['modified-by-callback']?.value, true);
+        expect(span.name, 'modified-span-name');
+      });
+
+      test('beforeSendSpan callback can remove sensitive attributes', () async {
```
denrase, line 175: Are users allowed to remove/modify SDK attributes?

packages/dart/test/telemetry/span/span_capture_pipeline_test.dart
buenaflor: In general they should be able to freely modify the individual span

denrase (changes requested):

packages/dart/lib/src/sentry_options.dart
buenaflor: >  It receives a deep copy of all spans in the span tree and their attributes.

this is old and doesn't apply here. we just expose the individual span

buenaflor: But yeah, I agree we should just not allow returning a span

packages/dart/test/telemetry/span/span_capture_pipeline_test.dart
buenaflor: yeah this is the main use case of this callback, to modify but not drop the span

buenaflor: for example what JS does is they pass the serialized JSON to this callback where the user can make adjustments but we don't need to do that

packages/dart/lib/src/telemetry/span/span_capture_pipeline.dart
```diff
@@ -27,24 +27,48 @@ class SpanCapturePipeline {
...
 
-        await _options.lifecycleRegistry
-            .dispatchCallback<OnProcessSpan>(OnProcessSpan(span));
+          span.addAttributesIfAbsent(defaultAttributes(_options, scope: scope));
+          span.addAttributesIfAbsent({
+            SemanticAttributesConstants.sentrySegmentName:
+                SentryAttribute.string(span.segmentSpan.name),
+            SemanticAttributesConstants.sentrySegmentId:
+                SentryAttribute.string(span.segmentSpan.spanId.toString()),
+          });
```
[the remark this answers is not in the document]
buenaflor: not really a bug
- the user cannot modify the segmetn anyway from this callback
- if the segment name changes we cannot do anything because this span is already sent off
- this sort of inconsistency is handled in the backend

denrase (approved):

PR #3488, opened 2026-01-28
ref(tracing-instrumentation): Introduce internal instrumentation abstraction for packages

buenaflor:
## :scroll: Description

Introduce internal instrumentation abstraction for packages

This PR will also serve as a umbrella branch where we will push all the migrations onto

Prepare for span-first migration

Existing tests should not fail and manual check

## :pencil: Checklist

- [ ] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps

#skip-changelog


buenaflor:
cursor review

buenaflor:
cursor review

packages/drift/lib/src/sentry_span_helper.dart
```diff
@@ -11,25 +11,35 @@ import 'constants.dart';
...
   final Hub _hub;
   final String _origin;
+  late final InstrumentationSpanFactory _factory;
 
   /// Represents a stack of Drift transaction spans.
   /// These are used to allow nested spans if the user nests Drift transactions.
   /// If the transaction stack is empty, the spans are attached to the
   /// active span in the Hub's scope.
-  final ListQueue<ISentrySpan?> _transactionStack = ListQueue();
+  final ListQueue<InstrumentationSpan> _transactionStack = ListQueue();
```
buenaflor, line 21: previous impl was nullable but no need for it to be null

packages/dart/lib/src/tracing/instrumentation/instrumentation_span.dart
```diff
@@ -0,0 +1,62 @@
...
+  @override
+  set origin(String? origin) => _span.origin = origin;
+
+  @override
+  Future<void> finish({
+    SpanStatus? status,
+    DateTime? endTimestamp,
+    Hint? hint,
+  }) =>
+      _span.finish(status: status, endTimestamp: endTimestamp);
```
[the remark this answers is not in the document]
buenaflor: we dont need it right now

packages/drift/lib/src/sentry_span_helper.dart
```diff
@@ -127,6 +148,8 @@ class SentrySpanHelper {
       return execute();
     }
 
+    final parentSpan = _transactionStack.removeLast();
```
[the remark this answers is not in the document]
buenaflor: will fix it

packages/dart/lib/src/tracing/instrumentation/instrumentation_span.dart
```diff
@@ -0,0 +1,61 @@
+import 'package:meta/meta.dart';
+
+import '../../hint.dart';
+import '../../protocol.dart';
+import '../../sentry_span_interface.dart';
+
+/// Opaque span handle enabling swappable tracing backends.
+@internal
+abstract class InstrumentationSpan {
```
denrase, line 8: Could we use SentrySpanV2 and wrap the logacy span instead of introducing another abstraction?

packages/dart/lib/src/tracing/instrumentation/instrumentation_span.dart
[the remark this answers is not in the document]
denrase: We could at least add a comment as to why we are not passing this.

packages/sqflite/lib/src/utils/sentry_database_span_attributes.dart
```diff
@@ -2,9 +2,14 @@ import 'package:sentry/sentry.dart';
 
 import '../../sentry_sqflite.dart';
 
-/// Sets the database attributes on the [span].
+/// Sets the database attributes on the [span] using InstrumentationSpan.
 /// It contains the database system and the database name.
-void setDatabaseAttributeData(ISentrySpan? span, String? dbName) {
+// ignore: invalid_use_of_internal_member
+void setDatabaseAttributeDataOnInstrumentationSpan(
```
denrase: IHMO no need to rename the method.

packages/drift/lib/src/sentry_span_helper.dart
```diff
@@ -84,11 +99,19 @@ class SentrySpanHelper {
...
+      parentSpan,
       SentrySpanOperations.dbSqlTransaction,
       description: SentrySpanDescriptions.dbTransaction,
     );
 
+    // If span creation failed (e.g., hit span limit), just execute without
+    // affecting the stack. Operations inside can still attach to whatever
+    // active span exists via _getParent() fallback.
+    if (newParent == null) {
+      return execute();
```
denrase, line 106: We should add a test case for this and also address the issue of stack imbalance (no more span, previous one gets removed).

denrase (approved):
Looking good, should add a test for the new behaviour where we early return.

packages/dart/lib/src/tracing/instrumentation/instrumentation_span.dart
buenaflor: Probably also possible but I'd rather not try to shoehorn one type into another and only expose what we actually need to adapt via this InstrumentationSpan

buenaflor: I'll try it out and check which one is better

buenaflor: I think the current approach is good, if we write an adapter around `ISentrySpan`
- it's harder to phase out when we later remove transactions, with the new abstraction we just switch the factory on the options and thats it
- not all fields and functions map nicely into SpanV2 like startChild (and nested startChild) so we'd need to add some custom logic that tracks parents etc...

PR #3490, opened 2026-01-28
fix(dart): Catch client exceptions in HttpTransport.send

### LINKED ISSUE #3455: Connection closed before full header was received Windows crash
### Platform

Flutter Desktop Windows

### Obfuscation

Enabled

### Debug Info

Enabled

### Doctor

Flutter 3.35.7
Windows 10

### Version

9.9.0

### Steps to Reproduce

I'm getting crash reports in Sentry that look like this:

```
ClientException: ClientException: Connection closed before full header was received, uri=https://o4508484832460800.ingest.us.sentry.io/api/4508484839145472/envelope/
#0      IOClient.send (package:http/src/io_client.dart:229)
#1 (<asynchronous suspension>)
#2      Response.fromStream (package:http/src/response.dart:58)
#3 (<asynchronous suspension>)
#4      HttpTransport.send (package:sentry/src/transport/http_transport.dart:42)
#5 (<asynchronous suspension>)
#6      ParsedPath._splitExtension.<fn> (package:path/src/parsed_path.dart:194)
```

This seems to happen only on Windows and not on the other platforms.

### Expected Result

I don't want the app to crash if Sentry communication does not work well for various reasons.

### Actual Result

App crashes with 

```
ClientException: ClientException: Connection closed before full header was received, uri=https://o4508484832460800.ingest.us.sentry.io/api/4508484839145472/envelope/
#0      IOClient.send (package:http/src/io_client.dart:229)
#1 (<asynchronous suspension>)
#2      Response.fromStream (package:http/src/response.dart:58)
#3 (<asynchronous suspension>)
#4      HttpTransport.send (package:sentry/src/transport/http_transport.dart:42)
#5 (<asynchronous suspension>)
#6      ParsedPath._splitExtension.<fn> (package:path/src/parsed_path.dart:194)
```

None

buenaflor:
## Summary
- Wraps the HTTP client send operation in a try/catch block to handle exceptions like `ClientException: Connection closed before full header was received`
- Logs errors via `internalLogger` for debugging visibility
- Returns `SentryId.empty()` on failure (graceful degradation) unless `automatedTestMode` is true, in which case it rethrows

Closes #3455 

## Test plan
- [x] Added test `returns empty SentryId when client throws exception`
- [x] All existing tests pass

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/packages/dart/lib/src/utils/transport_utils.dart
+++ b/packages/dart/lib/src/utils/transport_utils.dart
@@ -21,7 +21,7 @@ class TransportUtils {
     }
   }
 
-  static void recordLostEvent(
+  static void recordLostEvents(
       SentryOptions options, SentryEnvelope envelope, DiscardReason reason) {
     for (final item in envelope.items) {
       options.recorder.recordLostEvent(
--- a/packages/dart/lib/src/utils/transport_utils.dart
+++ b/packages/dart/lib/src/utils/transport_utils.dart
@@ -14,10 +14,8 @@ class TransportUtils {
     if (response.statusCode != 200) {
       if (options.debug) {
         internalLogger.error(() =>
-            'Error, statusCode = ${response.statusCode}, body = ${response.body}');
+            'Failed to send envelope, statusCode = ${response.statusCode}, body = ${response.body}');
       }
-    } else if (response.statusCode == 429) {
-      internalLogger.warning('Rate limit reached, failed to send envelope');
     } else {
       internalLogger.debug(
         () =>
--- a/packages/dart/lib/src/utils/transport_utils.dart
+++ b/packages/dart/lib/src/utils/transport_utils.dart
@@ -8,14 +8,11 @@ import '../transport/data_category.dart';
 import 'internal_logger.dart';
 
 class TransportUtils {
-  static void logResponse(
-      SentryOptions options, SentryEnvelope envelope, Response response,
+  static void logResponse(SentryEnvelope envelope, Response response,
       {required String target}) {
     if (response.statusCode != 200) {
-      if (options.debug) {
-        internalLogger.error(() =>
-            'Failed to send envelope, statusCode = ${response.statusCode}, body = ${response.body}');
-      }
+      internalLogger.error(() =>
+          'Failed to send envelope, statusCode = ${response.statusCode}, body = ${response.body}');
     } else {
       internalLogger.debug(
         () =>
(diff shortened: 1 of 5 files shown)


packages/dart/test/transport/http_transport_test.dart
```diff
@@ -289,18 +301,4 @@ class Fixture {
...
-  void mockLogger(
-    SentryLevel level,
-    String message, {
-    String? logger,
-    Object? exception,
-    StackTrace? stackTrace,
-  }) {
-    loggedLevel = level;
-    loggedMessage = message;
-  }
```
buenaflor, line 305: imo no need to assert log messages

packages/dart/lib/src/utils/transport_utils.dart
```diff
@@ -5,41 +5,38 @@ import '../protocol.dart';
...
+    }
+  }
 
-      if (response.statusCode >= 400 && response.statusCode != 429) {
-        for (final item in envelope.items) {
-          options.recorder.recordLostEvent(
-            DiscardReason.networkError,
-            DataCategory.fromItemType(item.header.type),
-          );
+  static void recordLostEvent(
```
denrase: Naming: recordLostEvent -> recordLostEvents, as we are recording one for every item.

packages/dart/lib/src/utils/transport_utils.dart
```diff
@@ -5,41 +5,38 @@ import '../protocol.dart';
 import '../sentry_envelope.dart';
 import '../sentry_options.dart';
 import '../transport/data_category.dart';
+import 'internal_logger.dart';
 
 class TransportUtils {
-  static void logResponse(
-      SentryOptions options, SentryEnvelope envelope, Response response,
+  static void logResponse(SentryEnvelope envelope, Response response,
```
denrase, line 11: Did you check all occurrences where `logResponse` was previously called? Just want to make sure we are not missing to record lost events, as this was a side-effect of this function.

denrase (changes requested):

packages/dart/lib/src/utils/transport_utils.dart
buenaflor: yep only used in HTTP and SpotlightHTTP so this behaviour was actually wrong since we'd be counting lost events twice because spotlight http sits on top of the normal http (spotlight http shouldn't record lost events anyway)

denrase (approved):

Issue #3487, opened 2026-01-28
Major v10 Task Collection


buenaflor:
### Description

This issue serves as a way to roughly collect ideas and tasks for the next major v10.

## Remove deprecated API

* Update deprecated constants in `SemanticAttributesConstants` (#3818)
* Remove `copyWith` deprecations (#3823)
* Set `clone()` to internal (#3824)
* Replace naming `SentryFeedbackWidget` in favour of `SentryFeedbackForm` -> remove `SentryFeedbackWidget` (#3825)
* Remove `SentryLogAttribute` in favour of `SentryAttribute` (#3826)

## Experimental → Stable APIs

* Remove experimental APIs from V10 (graduate the ones worth keeping to stable, drop the rest) (#3827)

## API Unification

* Unify setting options on integrations such as User interaction, navigator observer. It is extremely split right now. (#3828)
* Unify setting proper semantic attributes on the auto instrumented spans (#3818)
* Idea: Remove `enableNativeCrashHandling` in favour of single options (#3829)

## Removals & replacements (behavior-affecting)

* Remove `file.async` attributes from instrumentation -> it's something we invented and there is no product feature around this attribute so it's essentially useless. (#3830)
* Remove performance collectors from options, SDK should use trace lifecycles instead (#3831)
* Drop `sendDefaultPii` in favour of the new upcoming data collection spec ([https://develop.sentry.dev/sdk/foundations/client/data-collection/](<https://develop.sentry.dev/sdk/foundations/client/data-collection/>)) (#3832)

## Default & behavior changes

* Add `enableLogs` default change to true (1.17.0 docs version) (#3682)
* Add auto-emitted logs opt-in requirements for integrations (1.17.0 docs version) (#3833)
* Consider disabling `captureNativeFailedRequests` by default on iOS, so far it didnt yield good results and people generally wanted to disable it. (#3834)

## API additions / hardening

* Make `traceId` required on `SentryLog` (#3835)
* Add hint to `beforeSend` (log, metrics, span) (#3836)

## Platform / dependency / build

* ✅  Drop iOS 13 and 14 on Flutter #3564 for Sentry Cocoa v9
* ✅ Use `jni_flutter` instead of `jni` for the Flutter pkg and update to 1.0.0
* ✅ If we update Flutter to 3.41.0+ then we must add this to Package.swift : `.package(name: "FlutterFramework", path: "../FlutterFramework")`

## Refactoring / internal

[...]

PR #3496, opened 2026-02-02
feat(span-first): Add streaming instrumentation span implementation and tests

buenaflor:
## :scroll: Description

Implement streaming span infrastructure to support SpanV2 telemetry:

The streaming implementation enables packages to emit SpanV2 telemetry while maintaining backward compatibility with legacy transaction-based tracing.

Part of #3334 

## :pencil: Checklist

- [ ] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps

#skip-changelog

### DIFF
--- a/packages/_sentry_testing/lib/_sentry_testing.dart
+++ b/packages/_sentry_testing/lib/_sentry_testing.dart
@@ -1,0 +1,7 @@
+/// Internal testing utilities for Sentry Dart/Flutter SDK.
+///
+/// This library is for internal use only and should not be used outside
+/// the Sentry SDK monorepo.
+library _sentry_testing;
+
+export 'src/fake_telemetry_processor.dart';
--- a/packages/_sentry_testing/lib/src/fake_telemetry_processor.dart
+++ b/packages/_sentry_testing/lib/src/fake_telemetry_processor.dart
@@ -4,7 +4,7 @@ import 'package:collection/collection.dart';
 import 'package:sentry/sentry.dart';
 import 'package:sentry/src/telemetry/processing/processor.dart';
 
-/// Fake telemetry processor that captures spans for test assertions.
+/// Fake telemetry processor that captures telemetry data for test assertions.
 ///
 /// Usage:
 /// ```dart
--- a/packages/_sentry_testing/lib/src/fake_telemetry_processor.dart
+++ b/packages/_sentry_testing/lib/src/fake_telemetry_processor.dart
@@ -1,3 +1,5 @@
+// ignore_for_file: invalid_use_of_internal_member
+
 import 'dart:async';
 
 import 'package:collection/collection.dart';
@@ -39,7 +41,9 @@ class FakeTelemetryProcessor implements TelemetryProcessor {
 
   RecordingSentrySpanV2? findSpanByOperation(String operation) {
     return capturedSpans.firstWhereOrNull(
-      (span) => span.attributes['sentry.op']?.value == operation,
+      (span) =>
+          span.attributes[SemanticAttributesConstants.sentryOp]?.value ==
+          operation,
     );
   }
 
--- a/packages/_sentry_testing/lib/src/fake_telemetry_processor.dart
+++ b/packages/_sentry_testing/lib/src/fake_telemetry_processor.dart
@@ -1,0 +1,54 @@
+import 'dart:async';
+
+import 'package:collection/collection.dart';
+import 'package:sentry/sentry.dart';
+import 'package:sentry/src/telemetry/processing/processor.dart';
+
+/// Fake telemetry processor that captures spans for test assertions.
+///
+/// Usage:
+/// ```dart
+/// final processor = FakeTelemetryProcessor();
+/// final options = SentryOptions()
+///   ..telemetryProcessor = processor;
+/// ```
+class FakeTelemetryProcessor implements TelemetryProcessor {
+  final List<RecordingSentrySpanV2> capturedSpans = [];
+
+  @override
+  void addSpan(RecordingSentrySpanV2 span) {
+    capturedSpans.add(span);
+  }
+
+  @override
+  void addLog(SentryLog log) {}
+
+  @override
+  void addMetric(SentryMetric metric) {}
+
+  @override
+  FutureOr<void> flush() {}
+
+  void clear() {
+    capturedSpans.clear();
+  }
+
+  List<RecordingSentrySpanV2> getChildSpans() {
+    return capturedSpans.where((s) => s.parentSpan != null).toList();
+  }
+
+  RecordingSentrySpanV2? findSpanByOperation(String operation) {
+    return capturedSpans.firstWhereOrNull(
+      (span) => span.attributes['sentry.op']?.value == operation,
+    );
+  }
+
+  List<RecordingSentrySpanV2> findChildrenOf(RecordingSentrySpanV2 parent) {
+    return capturedSpans.where((s) => s.parentSpan == parent).toList();
+  }
+
+  Future<void> waitForProcessing() async {
+    // Small delay to ensure async telemetry callbacks complete.
+    await Future.delayed(Duration(milliseconds: 10));
+  }
+}
--- a/packages/_sentry_testing/pubspec.yaml
+++ b/packages/_sentry_testing/pubspec.yaml
@@ -10,7 +10,7 @@ dependencies:
   sentry:
     path: ../dart
   meta: ^1.11.0
-  collection: ^1.18.0  # For firstWhereOrNull
+  collection: ^1.18.0
 
 dev_dependencies:
   test: ^1.24.0
--- a/packages/_sentry_testing/pubspec.yaml
+++ b/packages/_sentry_testing/pubspec.yaml
@@ -1,0 +1,16 @@
+name: _sentry_testing
+version: 0.1.0
+description: Testing utilities for Sentry Dart/Flutter SDK (internal use only)
+publish_to: none
+
+environment:
+  sdk: '>=3.5.0 <4.0.0'
+
+dependencies:
+  sentry:
+    path: ../dart
+  meta: ^1.11.0
+  collection: ^1.18.0  # For firstWhereOrNull
+
+dev_dependencies:
+  test: ^1.24.0
--- a/packages/dart/lib/src/constants.dart
+++ b/packages/dart/lib/src/constants.dart
@@ -85,10 +85,12 @@ abstract class SemanticAttributesConstants {
   /// The replay ID.
   static const sentryReplayId = 'sentry.replay_id';
 
+  /// The operation name of a span.
+  static const sentryOp = 'sentry.op';
+
   /// Whether the replay is buffering (onErrorSampleRate).
   static const sentryInternalReplayIsBuffering =
       'sentry._internal.replay_is_buffering';
-
 
   /// The user ID (gated by `sendDefaultPii`).
   static const userId = 'user.id';
@@ -116,4 +118,31 @@ abstract class SemanticAttributesConstants {
 
   /// The device family (e.g., "iOS", "Android").
   static const deviceFamily = 'device.family';
+
+  /// The HTTP request method (e.g., "GET", "POST").
+  static const httpRequestMethod = 'http.request.method';
+
+  /// The URL of an HTTP request.
+  // TODO: this needs to be updated to use the new url attributes e.g url.full, etc...
+  static const url = 'url';
+
+  /// The HTTP query string (e.g., "foo=bar").
+  static const httpQuery = 'http.query';
+
+  /// The HTTP fragment (e.g., "section").
+  static const httpFragment = 'http.fragment';
+
+  /// The HTTP response status code.
+  static const httpResponseStatusCode = 'http.response.status_code';
+
+  /// The HTTP response content length.
+  static const httpResponseContentLength = 'http.response_content_length';
+
+  /// The database system identifier.
+  // TODO: deprecated, needs to be replaced later by db.system.name
+  static const dbSystem = 'db.system';
+
+  /// The database name.
+  // TODO: deprecated, needs to be replaced later by db.namespace
+  static const dbName = 'db.name';
 }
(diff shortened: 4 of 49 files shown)


packages/_sentry_testing/pubspec.yaml
```diff
@@ -0,0 +1,16 @@
...
+  sdk: '>=3.5.0 <4.0.0'
+
+dependencies:
+  sentry:
+    path: ../dart
+  meta: ^1.11.0
+  collection: ^1.18.0  # For firstWhereOrNull
+
+dev_dependencies:
+  test: ^1.24.0
```
buenaflor, line 16: I added a private/internal dev only library for sharing test utilities, there are some classes that are used everywhere such as TelemetryProcessor and I'd rather not duplicate a Mock/Fake processor for every single package and importing via relative paths is problematic for JS aka doesn't work well

packages/dart/lib/src/tracing/instrumentation/instrumentation_span.dart
```diff
@@ -66,3 +66,116 @@ class LegacyInstrumentationSpan implements InstrumentationSpan {
...
+    if (status != null) {
+      _span.status = _convertToV2Status(status);
+    }
+  }
+
+  @override
+  dynamic get throwable => _throwable;
+
+  @override
+  set throwable(dynamic throwable) => _throwable = throwable;
```
buenaflor, line 109: we dont use these throwables in Span V2 so we just kinda noop here

packages/dart/lib/src/tracing/instrumentation/instrumentation_span.dart
```diff
@@ -66,3 +66,116 @@ class LegacyInstrumentationSpan implements InstrumentationSpan {
...
+        return SpanStatus.unknownError();
+    }
+  }
+
+  SentrySpanStatusV2 _convertToV2Status(SpanStatus status) {
+    if (status.toString() == 'ok') {
+      return SentrySpanStatusV2.ok;
+    }
+    return SentrySpanStatusV2.error;
+  }
```
buenaflor: conversion is very simple since we only support ok and error in spanv2

packages/dart/lib/src/tracing/instrumentation/span_factory.dart
```diff
@@ -53,3 +51,43 @@ class LegacyInstrumentationSpanFactory implements InstrumentationSpanFactory {
...
+
+      if (childSpan is NoOpSentrySpanV2) return null;
+
+      childSpan.setAttribute('sentry.op', SentryAttribute.string(operation));
+
+      return StreamingInstrumentationSpan(childSpan);
+    }
+
+    return null;
+  }
```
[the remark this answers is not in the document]
buenaflor: Removed it, we dont use startTimestamp anyway

packages/flutter/example/lib/main.dart
```diff
@@ -93,8 +93,8 @@ Future<void> setupSentry(
       options.navigatorKey = navigatorKey;
       options.traceLifecycle = SentryTraceLifecycle.streaming;
 
-      options.replay.sessionSampleRate = 1.0;
-      options.replay.onErrorSampleRate = 1.0;
+      // options.replay.sessionSampleRate = 1.0;
+      // options.replay.onErrorSampleRate = 1.0;
```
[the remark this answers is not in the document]
buenaflor: Reverted it

packages/dart/lib/src/tracing/instrumentation/instrumentation_span.dart
```diff
@@ -66,3 +66,118 @@ class LegacyInstrumentationSpan implements InstrumentationSpan {
...
+      _span.setAttribute(key, SentryAttribute.double(value));
+    } else if (value is bool) {
+      _span.setAttribute(key, SentryAttribute.bool(value));
+    } else if (value is SentryAttribute) {
+      _span.setAttribute(key, value);
+    } else {
+      internalLogger.warning(
+          '$StreamingInstrumentationSpan: Unsupported data type in setData: $value');
+    }
+  }
```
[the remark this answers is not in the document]
buenaflor: will set it to info instead

packages/dart/lib/src/tracing/instrumentation/instrumentation_span.dart
```diff
@@ -66,3 +66,118 @@ class LegacyInstrumentationSpan implements InstrumentationSpan {
...
+
+  SpanStatus _convertFromV2Status(SentrySpanStatusV2 status) {
+    switch (status) {
+      case SentrySpanStatusV2.ok:
+        return SpanStatus.ok();
+      case SentrySpanStatusV2.error:
+        return SpanStatus.unknownError();
+    }
+  }
```
buenaflor, line 175: we dont have a spec on how to convert these transaction status into spanv2 status and vice versa so I'd just go for this until someone complains (no one will though)

packages/dart/lib/src/tracing/instrumentation/span_factory.dart
```diff
@@ -13,7 +11,6 @@ abstract class InstrumentationSpanFactory {
     InstrumentationSpan? parent,
     String operation, {
     String? description,
-    DateTime? startTimestamp,
```
buenaflor, line 16: we don't use custom start timestamps for now. we can add it later when there's a use case

packages/flutter/example/lib/main.dart
```diff
@@ -593,17 +591,10 @@ class MainScaffold extends StatelessWidget {
               text: 'Demonstrates the logging with Sentry Log.',
               buttonTitle: 'Sentry Log with Attribute',
             ),
-            if (Sentry.currentHub.options.traceLifecycle ==
```
buenaflor, line 596: accidental removal, will put it back

packages/flutter/example/lib/main.dart
```diff
@@ -823,67 +814,6 @@ Future<void> asyncThrows() async {
...
-
-  childSpan.status = SentrySpanStatusV2.ok;
-  childSpan.end();
-
-  await Future.delayed(const Duration(milliseconds: 10));
-
-  rootSpan.status = SentrySpanStatusV2.ok;
-  rootSpan.end();
-}
-
```
buenaflor, line 886: accidental removal, will put it back

packages/_sentry_testing/lib/_sentry_testing.dart
```diff
@@ -0,0 +1,7 @@
+/// Internal testing utilities for Sentry Dart/Flutter SDK.
```
denrase, line 1: ❤️

packages/hive/test/spanv2_integration_test.dart
```diff
@@ -0,0 +1,164 @@
...
+      );
+      expect(
+        span.attributes[SemanticAttributesConstants.dbSystem]?.value,
+        equals('flutter_hive'),
+      );
+      expect(
+        span.attributes[SemanticAttributesConstants.dbName]?.value,
+        equals('test-box'),
+      );
+      expect(span.attributes['sync']?.value, equals(true));
```
buenaflor, line 69: sync is not part of the semantic attributes constants, we invented this back then so it's not an official standard

packages/dart/lib/src/tracing/instrumentation/instrumentation_span.dart
```diff
@@ -66,3 +66,118 @@ class LegacyInstrumentationSpan implements InstrumentationSpan {
...
+
+  @override
+  SpanStatus? get status => _convertFromV2Status(_span.status);
+
+  @override
+  set status(SpanStatus? status) {
+    if (status != null) {
+      _span.status = _convertToV2Status(status);
```
denrase, line 101: null params should just be ignored? What did we fall back in the old span to when status was set to null?

packages/dart/lib/src/tracing/instrumentation/instrumentation_span.dart
```diff
@@ -66,3 +66,118 @@ class LegacyInstrumentationSpan implements InstrumentationSpan {
...
+
+  @override
+  dynamic get throwable => _throwable;
+
+  @override
+  set throwable(dynamic throwable) => _throwable = throwable;
+
+  @override
+  Future<void> finish({SpanStatus? status, DateTime? endTimestamp}) async {
+    if (status != null) {
```
denrase, line 113: Same here, don't we need a sane fallback to something reasonable when the passed status is null?

packages/_sentry_testing/pubspec.yaml
buenaflor: in the future we should add common test utils like `defaultTestOptions` there

packages/_sentry_testing/lib/src/fake_telemetry_processor.dart
```diff
@@ -0,0 +1,54 @@
...
+  @override
+  void addSpan(RecordingSentrySpanV2 span) {
+    capturedSpans.add(span);
+  }
+
+  @override
+  void addLog(SentryLog log) {}
+
+  @override
+  void addMetric(SentryMetric metric) {}
```
buenaflor, line 29: not yet used so I added it as noop for now

packages/dart/lib/src/tracing/instrumentation/span_factory.dart
```diff
@@ -53,3 +48,42 @@ class LegacyInstrumentationSpanFactory implements InstrumentationSpanFactory {
...
+
+  StreamingInstrumentationSpanFactory(this._hub);
+
+  @override
+  InstrumentationSpan? createSpan(
+    InstrumentationSpan? parent,
+    String operation, {
+    String? description,
+  }) {
+    if (parent == null) return null;
```
denrase: If we return null in both legacy/streaming, wouldn't it make sense to have non-null parent?

packages/hive/test/spanv2_integration_test.dart
buenaflor: we can probably remove this in the next major since we already have the `ThreadInfoIntegration`, there's also no product specific UI for this attribute so it's kinda useless. Added it to https://github.com/getsentry/sentry-dart/issues/3487

packages/dart/lib/src/constants.dart
```diff
@@ -116,4 +118,31 @@ abstract class SemanticAttributesConstants {
...
+  static const httpFragment = 'http.fragment';
+
+  /// The HTTP response status code.
+  static const httpResponseStatusCode = 'http.response.status_code';
+
+  /// The HTTP response content length.
+  static const httpResponseContentLength = 'http.response_content_length';
+
+  /// The database system identifier.
+  // TODO: deprecated, needs to be replaced later by db.system.name
```
denrase, line 142: Should we add a @deprecated annotation instead?

packages/dart/lib/src/constants.dart
```diff
@@ -116,4 +118,31 @@ abstract class SemanticAttributesConstants {
...
+  static const dbSystem = 'db.system';
+
+  /// The database name.
+  // TODO: deprecated, needs to be replaced later by db.namespace
+  static const dbName = 'db.name';
```
denrase, line 147: dito

packages/dio/test/spanv2_integration_test.dart
```diff
@@ -0,0 +1,174 @@
...
+import 'package:sentry_dio/src/tracing_client_adapter.dart';
+import 'package:test/test.dart';
+
+import 'mocks.dart';
+import 'mocks/mock_http_client_adapter.dart';
+
+final requestUri = Uri.parse('https://example.com/api/users');
+
+void main() {
+  group('Dio SpanV2 Integration', () {
```
denrase, line 16: I'm not sure we even need to test the concrete implementation of `InstrumentationSpan`, but rather against a `Mock/FakeInstrumentationSpan`, as we don't care about the implementation, but only the interface and which parts of the interface are called/set. Same goes for any other test we have, as long as the packages do no direct access to `Legacy/Streaming` classes. WDYT?

denrase (changes requested):

packages/dart/lib/src/tracing/instrumentation/instrumentation_span.dart
buenaflor: the status will just default to `ok` when the span is ended in both cases

packages/dart/lib/src/tracing/instrumentation/instrumentation_span.dart
buenaflor: it defaults to `ok` in both cases

packages/dart/lib/src/tracing/instrumentation/span_factory.dart
buenaflor: I'll check, I think we can

packages/dio/test/spanv2_integration_test.dart
buenaflor: I think we can keep this as it deeply asserts the flow until the end of the capture pipeline. your proposed solution would be very fitting as well for unit tests but I think we can skip that for now. 

my idea was to have a high level integration test that ideally tests as much as possible in one go

packages/dart/lib/src/constants.dart
buenaflor: I think I'd rather not since it would flag a warning everywhere and we'd just ignore it anyway, I'll add it as a v10 task https://github.com/getsentry/sentry-dart/issues/3487

packages/dart/lib/src/tracing/instrumentation/span_factory.dart
buenaflor: updated it

packages/dart/lib/src/tracing/instrumentation/instrumentation_span.dart
```diff
@@ -66,3 +66,118 @@ class LegacyInstrumentationSpan implements InstrumentationSpan {
...
+      _span.setAttribute(key, SentryAttribute.double(value));
+    } else if (value is bool) {
+      _span.setAttribute(key, SentryAttribute.bool(value));
+    } else if (value is SentryAttribute) {
+      _span.setAttribute(key, value);
+    } else {
+      internalLogger.info(
+          '$StreamingInstrumentationSpan: Unsupported data type in setData: $value');
+    }
+  }
```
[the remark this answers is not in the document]
buenaflor: We don't support this use case. attribute arrays are generally homogenous so we would be able to support something like `List<String>` but not `Map<String, dynamic>`

let's not do this for now due to limitations

denrase (approved):

PR #3505, opened 2026-02-09
feat(span-first): `startSpan` callback api

buenaflor:
## :scroll: Description

Reworks `startSpan` from a simple span-factory method into a callback-based API that automatically manages span lifecycle and scope propagation via Dart zones.

The previous `startSpan` returned a bare span object, leaving callers responsible for ending it and with no automatic scope propagation. This made it easy to forget to end spans and impossible to automatically nest child spans under a parent.

The callback-based API (modeled after the JS SDK's `Sentry.startSpan`) solves both problems: the span is guaranteed to end when the callback completes, and zone-local scope forking provides implicit parent-child relationships for nested `startSpan` calls without any manual wiring.

If more use cases arise in the future we can add more APIs but for span-first this will suffice for now.

```dart
// This will be the default startSpan API for Dart

Sentry.startSpan('span1', (_) async {
  // Due to zones this span knows span1 is the parent
  await Sentry.startSpan('span1 child1', (_) async {
    await Future.delayed(const Duration(milliseconds: 50));
  });

  // Due to zones this span knows span1 is the parent
  await Sentry.startSpan('span1 child2', (_) async {
    await Future.delayed(const Duration(milliseconds: 50));
  });
});
```

- Updated existing tests in `hub_span_test.dart` to cover the callback API, including sync/async callbacks, error propagation, and nested span parenting via zones
- Updated integration tests across `dio`, `drift`, `file`, `hive`, `isar`, `link`, `sqflite`, and `supabase` packages to use the new `startInactiveSpan` API
- Verified `scope_test.dart` reflects the simplified single-`_activeSpan` field

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps

#skip-changelog


buenaflor:
bugbot review

packages/dart/lib/src/hub.dart
```diff
@@ -616,16 +616,90 @@ class Hub {
...
+          'Hub: _createSpan returned UnsetSentrySpanV2 for \'$name\'. '
+          'This is a bug — the sentinel should never leak out of _createSpan.',
+        );
+        return callback(span);
+      case RecordingSentrySpanV2():
+        break;
+    }
+
+    final parentScope = (Zone.current[_scopeKey] as Scope?) ?? scope;
+    final forkedScope = parentScope.clone()..setActiveSpan(span);
```
buenaflor: same impl as Sentry JS on NodeJS -> async context with scope forking

packages/dart/lib/src/hub.dart
```diff
@@ -616,16 +616,90 @@ class Hub {
...
+
+    // Error handling is split into sync and async paths to preserve the
+    // FutureOr<T> return type — callers with a sync callback get a sync result.
+    FutureOr<T> result;
+    try {
+      result = runZoned(
+        () => callback(span),
+        zoneValues: {_scopeKey: forkedScope},
+      );
```
buenaflor, line 683: this is the bread and butter of this new API which allows to do this:

```dart
Sentry.startSpan('span1 test', (_) async {
  // Due to zones this span knows span1 test is the parent
  await Sentry.startSpan('span1 child1', (_) async {
    await Future.delayed(const Duration(milliseconds: 50));
  });

  // Due to zones this span knows span1 test is the parent
  await Sentry.startSpan('span1 child2', (_) async {
    await Future.delayed(const Duration(milliseconds: 50));
  });
});
```

it's a similar API used in JS

packages/dart/lib/src/hub.dart
```diff
@@ -616,16 +616,102 @@ class Hub {
     return NoOpSentrySpan();
   }
 
-  SentrySpanV2 startSpan(
+  static final _scopeKey = Object();
+
+  @internal
+  RecordingSentrySpanV2? getActiveSpan() {
+    final zoneScope = Zone.current[_scopeKey] as Scope?;
```
denrase: Could we create an internal getter for `Zone.current[_scopeKey] as Scope?`?

packages/dart/lib/src/hub.dart
```diff
@@ -616,16 +616,102 @@ class Hub {
...
+
+    FutureOr<T> result;
+    try {
+      result = runZoned(
+        () => callback(span),
+        zoneValues: {_scopeKey: forkedScope},
+      );
+    } catch (_) {
+      endSpan(isError: true);
+      rethrow;
```
denrase, line 686: Below we do `return Future<T>.error(error, stackTrace);` and here we rethrow, as this expected behaviour?

buenaflor:
e2e tests are failing due to 403

packages/dart/lib/src/scope.dart
```diff
@@ -44,41 +44,34 @@ class Scope {
...
+    _activeSpan = span;
   }
 
-  /// Removes the given recording [span] from the active spans list.
-  ///
-  /// This should be called when a span ends to remove it from the active
-  /// span list.
+  /// Clears the active span if it matches the given [span].
   @internal
   void removeActiveSpan(RecordingSentrySpanV2 span) {
```
denrase, line 71: A `removeActiveSpan` with span parameter is redundant, no? As there is only one at a time.

denrase (approved):
Only smaller suggestions for internal API and one question.

packages/dart/lib/src/hub.dart
buenaflor: yes because if it was caught directly it was a sync function therefore we need to call rethrow, otherwise it's a Future then we need Future.error

packages/dart/lib/src/scope.dart
buenaflor: It's a defensive approach in case it tries to remove a span that does not match e.g., someone accidentally calls endSpan twice, or a future change sets a different span on the same scope

Issue #3406, opened 2025-12-16
Synchronize `PropagationContext` across layers


bitsandfoxes:
### Description

The native SDKs support receiving a trace ID. This would enhance the debug experience by sharing trace across the hybrid applications layers.

See 
- https://github.com/getsentry/sentry-java/pull/4137
- https://github.com/getsentry/sentry-java/pull/4188
- https://github.com/getsentry/sentry-cocoa/pull/5081

PR #3507, opened 2026-02-11
feat(flutter): Synchronize `traceId` to native SDKs

buenaflor:
## :scroll: Description

Synchronize the Dart trace to native Android and iOS SDKs so that native events share the same trace as Dart events.

This adds:
- `OnTraceReset` lifecycle event dispatched from `Hub.generateNewTrace()`
- `NativeTraceSyncIntegration` that subscribes to `OnTraceReset` and calls the platform-specific native `setTrace` API
- `setTrace` implementation on all platform bindings (JNI for Android, method channel for Cocoa, no-op for Native/Web)
- `supportsTraceSync` capability flag on `SentryNativeBinding`
- Disables native auto trace ID generation on Android so Flutter is the single source of truth
- `setTrace` handler in iOS `SentryFlutterPlugin.swift`
- Registration of `NativeTraceSyncIntegration` in default integrations

Implements #3406

Native crashes/errors on Android and iOS carry their own independently-generated `traceId`, disconnecting them from the Dart-side trace in the Sentry UI. The native SDKs already expose APIs to accept a trace from hybrid SDKs — this PR wires them up.

- Unit tests for `OnTraceReset` dispatch from `Hub.generateNewTrace()`
- Mock regenerated to include `setTrace` and `supportsTraceSync`
- Integration unit test for `NativeTraceSyncIntegration`
- E2E integration tests on Android emulator
- E2E integration test for iOS is a lot of time effort so we will just settle for the android test for now and continue manually testing ios

**Manual Android Test:**

image

--

**Manual iOS Test:**

image

## :pencil: Checklist
- [ ] All tests passing
- [x] No breaking changes


packages/flutter/lib/src/native/java/sentry_native_java_init.dart
```diff
@@ -239,6 +239,11 @@ void configureAndroidOptions({
     }
     androidOptions.setSendDefaultPii(options.sendDefaultPii);
     androidOptions.setEnableScopeSync(options.enableNdkScopeSync);
+    // Disable native auto trace ID generation if trace sync is enabled
```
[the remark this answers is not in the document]
buenaflor: it's already guarded

packages/flutter/example/integration_test/integration_test.dart
```diff
@@ -1065,6 +1066,45 @@ void main() {
...
+      await setupSentryAndApp(tester);
+    });
+
+    final dartTraceId =
+        Sentry.currentHub.scope.propagationContext.traceId.toString();
+
+    final traceParent = jni.Sentry.getTraceparent();
+    expect(traceParent, isNotNull,
+        reason: 'Native traceparent should not be null');
+    final traceHeader = traceParent!.getValue().toDartString();
```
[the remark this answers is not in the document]
buenaflor: Not really. This is an integration test that runs once per CI job, not a tight loop so JNI local ref leaks won't accumulate enough to matter

packages/dart/lib/src/hub.dart
```diff
@@ -610,6 +610,11 @@ class Hub {
     // Create a brand-new trace and reset the sampling flag and sampleRand so
     // that the next root transaction can set it again.
     scope.propagationContext.resetTrace();
+    // Fire-and-forget the callback
+    // Native SDK synchronization over async method channels may be slightly delayed, but this is not problematic in practice.
+    _options.lifecycleRegistry.dispatchCallback(OnTraceReset(
```
denrase: From a naming perspective we now have `new` and `reset`. Is there anything we can rename without changing public API to align this?

packages/dart/lib/src/hub.dart
```diff
@@ -610,6 +610,11 @@ class Hub {
...
+        scope.propagationContext.traceId,
+        getSpan()?.context.spanId ?? SpanId.newId()));
```
denrase, line 617: Is the fallback to a new span id here correct, or should the event have optional span?

packages/dart/lib/src/sentry_options.dart
```diff
@@ -397,6 +397,13 @@ class SentryOptions {
   /// Send statistics to sentry when the client drops events.
   bool sendClientReports = true;
 
+  /// Whether to synchronize the Dart trace to the native SDK.
+  ///
+  /// Allows native events to share the same trace as Dart events.
+  ///
+  /// Supported on Android and iOS/macOS.
+  bool enableNativeTraceSync = true;
```
buenaflor: this is wrong here, this should be in `SentryFlutterOptions`. I'll update that

packages/dart/lib/src/sdk_lifecycle_hooks.dart
```diff
@@ -103,3 +103,12 @@ class OnProcessMetric extends SdkLifecycleEvent {
 
   OnProcessMetric(this.metric);
 }
+
+/// Dispatched when a new trace is generated via [Hub.generateNewTrace].
+@internal
+class OnTraceReset extends SdkLifecycleEvent {
+  final SentryId traceId;
+  final SpanId spanId;
```
denrase, line 111: See first comment about fallback new span when this event is dispatched.

packages/dart/lib/src/sentry_options.dart
denrase: This should be opt-out?

packages/flutter/example/integration_test/integration_test.dart
```diff
@@ -1065,6 +1066,45 @@ void main() {
...
+
+    final nativeTraceId = traceHeader.split('-').first;
+    expect(nativeTraceId, dartTraceId,
+        reason: 'Native traceId should match Dart traceId after initial sync');
+
+    Sentry.currentHub.generateNewTrace();
+    final newDartTraceId =
+        Sentry.currentHub.scope.propagationContext.traceId.toString();
+    expect(newDartTraceId, isNot(dartTraceId),
```
denrase: I'm sure we tested this before, no need to re-test existing behaviour (Sentry.currentHub.generateNewTrace creates a new trace id).

packages/flutter/ios/sentry_flutter/Sources/sentry_flutter/SentryFlutterPlugin.swift
```diff
@@ -696,6 +702,17 @@ public class SentryFlutterPlugin: NSObject, FlutterPlugin {
         result("")
     }
 
+    private func setTrace(traceId: String?, spanId: String?, result: @escaping FlutterResult) {
+        guard let traceId = traceId, let spanId = spanId else {
+            result("")
```
denrase: Shouldn't we report an error here?

packages/flutter/lib/src/integrations/native_trace_sync_integration.dart
```diff
@@ -0,0 +1,46 @@
...
+      return;
+    }
+
+    _options = options;
+    options.lifecycleRegistry
+        .registerCallback<OnTraceReset>(_syncTraceToNative);
+    options.sdk.addIntegration(integrationName);
+
+    final traceId = hub.scope.propagationContext.traceId;
+    final spanId = hub.getSpan()?.context.spanId ?? SpanId.newId();
```
denrase, line 32: Again a fallback to a new span id, that is just propagated to native sie, wihtout flutter/dart having a reference.

packages/flutter/lib/src/integrations/native_trace_sync_integration.dart
```diff
@@ -0,0 +1,46 @@
...
+
+    // Sync the initial PropagationContext created at Hub construction.
+    _syncTraceToNative(OnTraceReset(traceId, spanId));
+  }
+
+  @override
+  void close() {
+    _options?.lifecycleRegistry
```
denrase, line 40: Do other integrations nil options here?

packages/flutter/lib/src/native/java/sentry_native_java.dart
```diff
@@ -388,6 +388,20 @@ class SentryNativeJava extends SentryNativeChannel {
...
+  @override
+  bool get supportsTraceSync => true;
+
+  @override
+  void setTrace(SentryId traceId, SpanId spanId) {
+    tryCatchSync('setTrace', () {
+      using((arena) {
+        final jTraceId = traceId.toString().toJString()..releasedBy(arena);
+        final jSpanId = spanId.toString().toJString()..releasedBy(arena);
+        native.InternalSentrySdk.setTrace(jTraceId, jSpanId, null, null);
```
denrase, line 407: Does it make sense to add a comment what the null params are and why we are not using them?

denrase (changes requested):

packages/dart/lib/src/sentry_options.dart
buenaflor: the default experience should be that traces are synced by default (same way contexts etc. are synced by default) - I wouldn't make it opt-in

majority of users will benefit from this

packages/dart/lib/src/hub.dart
buenaflor: `OnNewTrace` maybe?

buenaflor: in any case the whole trace / propagation context impl is internal api so we can change it without worries

packages/dart/lib/src/hub.dart
buenaflor: yeah the fallback span id is fine here and is defined the same way across other SDKs. 

the fallback doesn't map to a real measured span, it's just a placeholder identifier for propagation purposes

packages/flutter/lib/src/integrations/native_trace_sync_integration.dart
buenaflor: that's expected behaviour, this is used for propagating traces in distributed tracing if tracing is disabled or there is no active span then we generate a new span id.

it doesn't map to a real measured span,  it's just a placeholder identifier for propagation purposes

packages/flutter/lib/src/native/java/sentry_native_java.dart
buenaflor: done

packages/flutter/lib/src/integrations/native_trace_sync_integration.dart
buenaflor: yup, would you set it as late instead?

packages/dart/lib/src/hub.dart
buenaflor: updated the namings

packages/flutter/lib/src/integrations/native_trace_sync_integration.dart
denrase: no preference really, just to be consistent in cleanup like we do elsewhere

denrase (approved):

PR #3508, opened 2026-02-11
internal(flutter): Add SDK features metadata for SPM vs CocoaPods tracking

### LINKED ISSUE #3492: Investigate if it's possible to track SPM vs Cocoapods usage
### Description

For internal stats it would be good if it's possible

buenaflor:
## Summary
- Adds a `features` list to `SdkVersion` (alongside `integrations` and `packages`) for additional SDK metadata
- Reports `SwiftPackageManager` or `CocoaPods` as an SDK feature on iOS based on build type
- Differentiates package name as `spm:sentry-cocoa` vs `cocoapods:sentry-cocoa`

This allows us to internally query usage data

Note: features metadata was introduced here https://github.com/getsentry/team-sdks/issues/83 and is currently used in sentry cocoa as well

## Motivation

Closes #3492 

## Test plan
- [x] `SdkVersion` serialization roundtrip (toJson/fromJson) includes features
- [x] `addFeature()` deduplication works correctly
- [x] `LoadContextsIntegration` merges native features into event SDK
- [x] Feature deduplication in event processor verified
- [x] Manual verification: SPM build shows `"features": ["SwiftPackageManager"]` in event payload

image


packages/flutter/ios/sentry_flutter/Sources/sentry_flutter/SentryFlutterPlugin.swift
```diff
@@ -319,6 +331,13 @@ public class SentryFlutterPlugin: NSObject, FlutterPlugin {
                                 sdk["integrations"] = integrations
                             }
                         }
+                        if let features = flutterSdk!["features"] as? [String] {
+                            if let sdkFeatures = sdk["features"] as? [String] {
+                                sdk["features"] = sdkFeatures + features
```
denrase, line 336: Not sure if we even need this, but shouldn't we do de-duplication here as well?

buenaflor:
e2e tests failing with 403 still, probably the auth token is invalid ..

packages/flutter/lib/src/integrations/load_contexts_integration.dart
```diff
@@ -260,6 +260,18 @@ class _LoadContextsIntegrationEventProcessor implements EventProcessor {
         event.sdk = sdk;
       }
 
+      final featuresList = infos['features'] as List?;
+      if (featuresList != null && featuresList.isNotEmpty) {
+        final features = List<String>.from(featuresList);
+        final sdk = event.sdk ?? _options.sdk;
+
+        for (final feature in features) {
+          sdk.addFeature(feature);
```
denrase, line 269: Here should the merged de-duplication be done, right?

packages/flutter/test/integrations/load_contexts_integration_test.dart
```diff
@@ -291,6 +292,37 @@ void main() {
...
+        await fixture.registerIntegration();
+
+        final e = getEvent();
+        final event =
+            await fixture.options.eventProcessors.first.apply(e, Hint());
+
+        expect(event?.sdk?.features.contains('SwiftPackageManager'), true);
+      });
+
+      test('does not duplicate feature if already present', () async {
```
denrase, line 307: 👍

denrase (approved):
Just some questions regarding dedupe.

packages/flutter/ios/sentry_flutter/Sources/sentry_flutter/SentryFlutterPlugin.swift
buenaflor: we can, I just didnt add it because theres almost no chance for a dupe to happen - imo we don't really need it

packages/flutter/lib/src/integrations/load_contexts_integration.dart
buenaflor: yeah when adding/merging to the dart features it already dedupes (addFeature only adds if unique)

Issue #3494, opened 2026-02-02
Span-first: Parent to scope if explicit parent is non-recording (noop)


buenaflor:
### Description

It's possible that ignoreSpans can ignore the "middle" of a span so therefore we need to parent correctly. e.g

ignoreSpan: Span2
Span1 -> Span2 -> Span3

which means Span3 needs to be parented to Span1 (implicitly and explicitly)

buenaflor:
already implemented in #3517

PR #3517, opened 2026-02-16
feat(span-first): `ignoreSpans`

### LINKED ISSUE #3332: [Span First]: Add new telemetry pipeline (including filtering and mutation)

buenaflor:
## :scroll: Description

Adds `ignoreSpans` option which allows to add rules for ignoring spanv2 spans.

Span first

Closes #3332 #3494 

Unit tests

## :pencil: Checklist

- [ ] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps

#skip-changelog


buenaflor:
@cursor review

packages/dart/lib/src/telemetry/span/ignore_span_rule.dart
```diff
@@ -0,0 +1,51 @@
...
+sealed class IgnoreSpanRule {
+  const IgnoreSpanRule();
+
+  factory IgnoreSpanRule.nameContains(Pattern pattern) = _NameContains;
+  factory IgnoreSpanRule.nameEquals(String value) = _NameEquals;
+  factory IgnoreSpanRule.nameStartsWith(Pattern pattern) = _NameStartsWith;
+  factory IgnoreSpanRule.nameEndsWith(String value) = _NameEndsWith;
+
+  @internal
+  bool nameMatches(String name);
```
denrase: I'm not sure about the naming here, `nameMatches` signals an exact match to me, but the implementation do different thing. Maybe something like "test", "validate" or similar?

packages/dart/lib/src/sentry_options.dart
```diff
@@ -266,6 +266,15 @@ class SentryOptions {
...
   List<String> ignoreTransactions = [];
 
+  /// Rules that determine which spans are ignored.
+  ///
+  /// Rules are evaluated when a span starts, in the order they are added.
+  /// The first rule that matches the span name determines whether the span is
+  /// ignored.
+  ///
+  /// If empty, all spans are processed.
+  List<IgnoreSpanRule> ignoreSpans = [];
```
denrase, line 276: nit: Naming, could call this `ignoreSpanRules` instead of `ignoreSpans`, the latter reads like a collection of ignored spans, not rules.

denrase (approved):
Just some nits regarding naming.

packages/dart/lib/src/sentry_options.dart
buenaflor: not sure how much we can deviate from the spec, this pattern is already established in JS and I think we generalyl want to align as much as possible

https://develop.sentry.dev/sdk/telemetry/spans/filtering/

```dart
options.ignoreSpans = [
  .nameContains('ignore'),
];
```

I guess you can also interpret this as: Ignore spans whose names contain 'ignore'. This becomes much more readable with the dot shorthand

PR #3521, opened 2026-02-17
feat(span-first): Add idle span support and support user interaction tracing

buenaflor:
## :scroll: Description

Adds support for idle spans which is going to be the main driver for the instrumentations:
- App starts (will be added in another PR)
- TTID/TTFD (will be added in another PR)
- User Interactions

Not using instrumentation span since instrumentation spans are not meant to represent idle spans also there are only few places with custom logic so imo we dont need to abstract it

Part of #3334 

Unit, manual

## :pencil: Checklist

- [ ] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps

#skip-changelog


buenaflor:
@cursor review

buenaflor:
@cursor review

buenaflor:
@cursor review

buenaflor:
@cursor review

packages/dart/lib/src/hub.dart
```diff
@@ -40,6 +40,11 @@ class Hub {
 
   late final _WeakMap _throwableToSpan;
 
+  IdleRecordingSentrySpanV2? get _currentIdleSpan =>
+      _idleSpan?.isEnded == false ? _idleSpan : null;
+
+  IdleRecordingSentrySpanV2? _idleSpan;
```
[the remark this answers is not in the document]
buenaflor: this is intended and not a critical memory leak, captureSpan is async so it could lead to race conditions when ending a span and then a new one starts.

it is cleaned up in close anyway

packages/dart/lib/src/hub.dart
```diff
@@ -751,58 +796,80 @@ class Hub {
...
+      trimEndTimestamp: trimIdleSpanEndTimestamp,
+      lifecycleRegistry: _options.lifecycleRegistry,
+    );
+
     if (attributes != null) {
       span.setAttributes(attributes);
     }
 
     _options.lifecycleRegistry.dispatchCallback(OnSpanStartV2(span));
+    _idleSpan = span;
```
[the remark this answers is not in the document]
buenaflor: not an issue

packages/flutter/lib/src/user_interaction/sentry_user_interaction_widget.dart
```diff
@@ -406,6 +407,63 @@ class _SentryUserInteractionWidgetState
...
+
+    _hub.startIdleSpan(
+      widgetKey,
+      attributes: {
+        SemanticAttributesConstants.sentryOp:
+            SentryAttribute.string(SentrySpanOperations.uiActionClick),
+        SemanticAttributesConstants.sentryOrigin:
+            SentryAttribute.string(SentryTraceOrigins.autoUiInteraction),
+      },
+    );
```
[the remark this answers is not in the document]
buenaflor: not an issue

packages/dart/lib/src/telemetry/span/idle_recording_sentry_span_v2.dart
```diff
@@ -0,0 +1,214 @@
...
+    );
+  }
+
+  void _onSpanStartEvent(OnSpanStartV2 event) {
+    if (_isEnded) return;
+
+    if (event.span case final RecordingSentrySpanV2 child
+        when _shouldTrackDescendant(child)) {
+      _activeDescendants[child.spanId] = child;
+      _hadActivity = true;
```
denrase: Do we even need _hadActivity as a separate flag? If any child was added, _activeDescendants is not empty.

packages/dart/lib/src/telemetry/span/idle_recording_sentry_span_v2.dart
```diff
@@ -0,0 +1,214 @@
...
+        super._(parentSpan: null) {
+    _lifecycleRegistry.registerCallback<OnSpanStartV2>(_onSpanStartEvent);
+    _lifecycleRegistry.registerCallback<OnSpanEndV2>(_onSpanEndEvent);
+    _startIdleTimer();
+    _startFinalTimer();
+  }
+
+  bool get hadActivity => _hadActivity;
+
+  void resetIdleTimer() {
```
denrase, line 53: Why do we need a mehtod to do this? I though only creation and finishing a child would stat a new idle timer?

packages/dart/lib/src/telemetry/span/idle_recording_sentry_span_v2.dart
```diff
@@ -0,0 +1,214 @@
...
+
+/// Recording span with idle behavior built into the span itself.
+final class IdleRecordingSentrySpanV2 extends RecordingSentrySpanV2 {
+  final Duration idleTimeout;
+  final Duration finalTimeout;
+  final bool trimEndTimestamp;
+  final SdkLifecycleRegistry _lifecycleRegistry;
+
+  final Map<SpanId, RecordingSentrySpanV2> _activeDescendants = {};
+  bool _isEnded = false;
```
denrase: We could probably use `RecordingSentrySpanV2` isEnded, no need for a duplicated flag.

packages/dart/lib/src/telemetry/span/idle_recording_sentry_span_v2.dart
```diff
@@ -0,0 +1,214 @@
...
+  }
+
+  void _trackLatestChildEnd(DateTime childEndTimestamp) {
+    if (_latestChildEndTimestamp == null ||
+        childEndTimestamp.isAfter(_latestChildEndTimestamp!)) {
+      _latestChildEndTimestamp = childEndTimestamp;
+    }
+  }
+
+  DateTime _computeFinalEndTimestamp(DateTime idleEndTimestamp) {
```
denrase, line 194: Isn't this enough or do you fear that child spans come around with weird start/end times so that we do all this extra guarding?

```dart
DateTime _computeFinalEndTimestamp(DateTime idleEndTimestamp) {
  if (!trimEndTimestamp || _latestChildEndTimestamp == null) {
    return idleEndTimestamp;
  }
  // Clamp: don't extend beyond when the idle span actually ended.
  return _latestChildEndTimestamp!.isBefore(idleEndTimestamp)
    ? _latestChildEndTimestamp!
    : idleEndTimestamp;
}
```

packages/dart/lib/src/hub.dart
```diff
@@ -751,58 +796,80 @@ class Hub {
...
+      finalTimeout: finalTimeout,
+      trimEndTimestamp: trimIdleSpanEndTimestamp,
+      lifecycleRegistry: _options.lifecycleRegistry,
+    );
+
     if (attributes != null) {
       span.setAttributes(attributes);
     }
 
     _options.lifecycleRegistry.dispatchCallback(OnSpanStartV2(span));
```
denrase, line 885: Will the idle san receive itself from this callback?

packages/flutter/lib/src/user_interaction/sentry_user_interaction_widget.dart
```diff
@@ -406,6 +407,63 @@ class _SentryUserInteractionWidgetState
     return path;
   }
 
+  void _startSpanOnTap(UserInteractionInfo info, String? widgetKey) {
+    if (widgetKey == null ||
```
denrase: Something like this would be easier to reason about:

 ```dart
   final options = _options;
   if (widgetKey == null || options == null) return;
   if (!options.isTracingEnabled() || !options.enableUserInteractionTracing) return;
 ```

packages/flutter/lib/src/user_interaction/sentry_user_interaction_widget.dart
```diff
@@ -406,6 +407,63 @@ class _SentryUserInteractionWidgetState
...
+      if (activeSpan is IdleRecordingSentrySpanV2) {
+        final lastElement = _lastTappedWidget?.element;
+        final isSameWidget = _isElementMounted(lastElement) &&
+            _isElementMounted(info.element) &&
+            lastElement?.widget == info.element.widget;
+
+        if (isSameWidget) {
+          // Same widget tapped again — reset the idle timer to keep the span
+          // alive instead of starting a new one.
+          activeSpan.resetIdleTimer();
```
denrase, line 439: Ah, now i understand, do we do this in other SDKs as well?

packages/flutter/lib/src/user_interaction/sentry_user_interaction_widget.dart
```diff
@@ -406,6 +407,63 @@ class _SentryUserInteractionWidgetState
...
+          return;
+        }
+
+        // If the active idle span had no descendant activity, skip — the
+        // previous tap produced no meaningful work.
+        if (!activeSpan.hadActivity) {
+          return;
```
denrase: If user taps widget A, then taps on widget B, without anything happening in the meantime, wouldn't we completley block out the idle start of widget B? Shouldn't we stop A at this point and let B start?

packages/flutter/test/user_interaction/sentry_user_interaction_widget_test.dart
```diff
@@ -532,6 +532,141 @@ void main() {
...
+        await tapMe(tester, sut, 'btn_1');
+
+        // Should still be the same span (not a new one)
+        expect(fixture.hub.getActiveSpan(), same(firstSpan));
+        expect(firstSpan?.isEnded, isFalse);
+      });
+    });
+
+    testWidgets(
+        'does not start new span when tapping different widget if previous had no activity',
```
denrase: So in this case, widget A wins and we consider everything tha comes after widget B tap to be part of widget A idle span?

denrase (changes requested):

packages/flutter/lib/src/user_interaction/sentry_user_interaction_widget.dart
buenaflor: yup this is also specced like this in the Sentry docs

packages/dart/lib/src/hub.dart
buenaflor: yup, because it should also be collected via e.g the frame tracking

packages/dart/lib/src/telemetry/span/idle_recording_sentry_span_v2.dart
buenaflor: I guess we can expose them and check directly

packages/flutter/lib/src/user_interaction/sentry_user_interaction_widget.dart
buenaflor: I think this is a thinking error of my side

the sequence must be: 

Tap A - start idle span

Nothing happens

Tap B - Do not send idle span with no children from A, start idle span B

buenaflor: so yea this is wrong, maybe this should be baked into the idle span class, dont capture if no children

packages/dart/lib/src/telemetry/span/idle_recording_sentry_span_v2.dart
buenaflor: initially yes but since we have pretty strict apis with the callback based spans I think your method is fine

packages/flutter/test/user_interaction/sentry_user_interaction_widget_test.dart
buenaflor: See my other comment, this is wrong bebaviour

packages/dart/lib/src/telemetry/span/idle_recording_sentry_span_v2.dart
buenaflor: We shouldn't remove it because `_end()` needs to do teardown work (cancel timers, finish active descendants, compute timestamps) before it can call super.end(). Without it, a timer callback or end() call could re-enter _end() during that teardown window, since isEnded is still false until super.end() completes.

Instead I renamed it to `_isEnding` which reflects its purpose better imo

packages/dart/lib/src/telemetry/span/idle_recording_sentry_span_v2.dart
buenaflor: Okay I double checked and we do need this. activeDescendants is not enough because by the time we check the children might have already finished cleared from the list. 

e.g

IdleSpan start 
  -> child 1
  -> child 2

child1 finishes, child2 finishes, descendants list is cleared -> now we never know if the span had any children

packages/flutter/lib/src/user_interaction/sentry_user_interaction_widget.dart
denrase: what does the spec saya bout idle spans that capture nothing? I mean users can still end them and they'd get the duration of the operation in question, which is valuable i guess 🤔

packages/dart/lib/src/telemetry/span/idle_recording_sentry_span_v2.dart
```diff
@@ -0,0 +1,215 @@
...
+  final Duration finalTimeout;
+  final bool trimEndTimestamp;
+  final SdkLifecycleRegistry _lifecycleRegistry;
+
+  final Map<SpanId, RecordingSentrySpanV2> _activeDescendants = {};
+
+  /// Separate from [isEnded] which only becomes true after [super.end()].
+  /// This flag is set at the start of [_end] to guard against re-entrant calls
+  /// while teardown (cancelling timers, finishing descendants) is still in progress.
+  bool _isEnding = false;
```
denrase, line 29: Maybe i forgot, or i can't find my comment, but isn't _isEnding a duplicate, since we have simethin like this in the superclass?

denrase (approved):
Lookin godd, i'm not sure if i forgot a comment or just didn't saw it, added again. Otherwise pls check if ci is happy. 🙇

packages/dart/lib/src/telemetry/span/idle_recording_sentry_span_v2.dart
buenaflor: I think you're thinking of another PR where we had a similar issue (in SentryTransaction)

buenaflor: but no we only have isEnded in the superclass

packages/flutter/lib/src/user_interaction/sentry_user_interaction_widget.dart
buenaflor: the main issue is the duration is not useful at all since it would just end when the idle timer ends, 

if not ended manually then all taps with no children would have a e.g 3 second duration which doesn't really say anything.

idle spans are only "synthetic" spans so the user can attach to it

to be pragmatic here I'd say let's keep sending them (would also reduce some code) and revisit when it's an issue

packages/dart/lib/src/hub_adapter.dart
```diff
@@ -242,4 +242,20 @@ class HubAdapter implements Hub {
     return Sentry.currentHub.startSpan(name, callback,
         attributes: attributes, parentSpan: parentSpan);
   }
+
+  @override
+  SentrySpanV2 startIdleSpan(
+    String name, {
+    Duration idleTimeout = const Duration(seconds: 5),
```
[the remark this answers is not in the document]
buenaflor: it's not public API, this will be unified later

buenaflor:
seems like something is up with github's infra, lots of failures due to authentication issues

packages/dart/lib/src/hub_adapter.dart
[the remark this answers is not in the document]
buenaflor: that's fine, we'll unify later

PR #3524, opened 2026-02-17
fix(dart): Dont guard user behind `sendDefaultPii`

### LINKED ISSUE #3504: Dont guard all user info with `sendDefaultPii`
### Description

If a user manually sets User on the scope, we should send it with logs and metrics, regardless of sendDefaultPii

buenaflor:
## :scroll: Description

The user API is manual (not inferred) so that's why we do not need to gate it behind `sendDefaultPii`

This has recently been discovered to be a develop documentation issue and therefore a bug.

See https://develop.sentry.dev/sdk/telemetry/metrics/#user-attributes

Closes https://github.com/getsentry/sentry-dart/issues/3504

Unit tests, manual

## :pencil: Checklist

- [ ] All tests passing
- [x] No breaking changes

### DIFF
--- a/packages/dart/lib/src/constants.dart
+++ b/packages/dart/lib/src/constants.dart
@@ -89,16 +89,16 @@ abstract class SemanticAttributesConstants {
   static const sentryInternalReplayIsBuffering =
       'sentry._internal.replay_is_buffering';
 
-  /// The user ID (gated by `sendDefaultPii`).
+  /// The user ID.
+  /// Users are manually set on the scope, therefore this is not gated by `sendDefaultPii`.
   static const userId = 'user.id';
 
-  /// The user email (gated by `sendDefaultPii`).
+  /// The user email.
+  /// Users are manually set on the scope, therefore this is not gated by `sendDefaultPii`.
   static const userEmail = 'user.email';
 
-  /// The user IP address (gated by `sendDefaultPii`).
-  static const userIpAddress = 'user.ip_address';
-
-  /// The user username (gated by `sendDefaultPii`).
+  /// The user username.
+  /// Users are manually set on the scope, therefore this is not gated by `sendDefaultPii`.
   static const userName = 'user.name';
 
   /// The operating system name.
--- a/packages/dart/lib/src/constants.dart
+++ b/packages/dart/lib/src/constants.dart
@@ -90,15 +90,18 @@ abstract class SemanticAttributesConstants {
       'sentry._internal.replay_is_buffering';
 
   /// The user ID.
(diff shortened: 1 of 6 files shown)


packages/dart/lib/src/constants.dart
```diff
@@ -89,16 +89,16 @@ abstract class SemanticAttributesConstants {
...
+  /// Users are manually set on the scope, therefore this is not gated by `sendDefaultPii`.
   static const userId = 'user.id';
 
-  /// The user email (gated by `sendDefaultPii`).
+  /// The user email.
+  /// Users are manually set on the scope, therefore this is not gated by `sendDefaultPii`.
   static const userEmail = 'user.email';
 
-  /// The user IP address (gated by `sendDefaultPii`).
-  static const userIpAddress = 'user.ip_address';
```
buenaflor, line 99: we dont use this currently so we can remove it

packages/flutter/example/lib/main.dart
```diff
@@ -114,6 +114,15 @@ Future<void> setupSentry(
...
   );
+
+  Sentry.configureScope((scope) {
+    final user = SentryUser(
+      id: SentryId.newId().toString(),
+      name: 'J. Smith',
+      email: 'j.smith@example.com',
+    );
+    scope.setUser(user);
+  });
```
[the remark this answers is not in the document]
buenaflor: that's fine we dont need to

packages/dart/lib/src/telemetry/default_attributes.dart
```diff
@@ -23,21 +23,20 @@ Map<String, SentryAttribute> defaultAttributes(SentryOptions options,
...
-      }
-      if (user.name != null) {
-        attributes[SemanticAttributesConstants.userName] =
-            SentryAttribute.string(user.name!);
-      }
-      if (user.email != null) {
-        attributes[SemanticAttributesConstants.userEmail] =
-            SentryAttribute.string(user.email!);
-      }
+  // Users are manually set on the scope, therefore this is not gated by `sendDefaultPii`.
```
denrase: We could also add this comment in the user setter on scope.

denrase (approved):
Looks good, just one suggestion.

packages/dart/lib/src/telemetry/default_attributes.dart
buenaflor: I think we can leave it commented in the default attributes and the constants file - in the future we will only us attributes so this is the source of truth anyway

the setUser API doesn't really know about these attributes

PR #3534, opened 2026-02-20
feat(span-first): Port native app start integration to V2 span API

buenaflor:
## :scroll: Description

Ports the native app start and generic app start integrations to the V2 streaming span API, gated by `SentryTraceLifecycle.streaming`.

Span-first and closes https://github.com/getsentry/sentry-dart/issues/3334

Unit tests, manual tests

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps

Make sure mobile vitals insights product page works

#skip-changelog


packages/flutter/lib/src/integrations/native_app_start_data.dart
```diff
@@ -0,0 +1,122 @@
...
+  final DateTime start;
+  final DateTime end;
+  final String description;
+}
+
+/// Parses and validates native app start data.
+///
+/// Returns `null` if validation fails (e.g. >60s duration, missing setup time).
+@internal
+AppStartInfo? parseNativeAppStart(
```
buenaflor, line 66: pull out the common things we can to share logic with native app start handler v2 and v1

buenaflor:
@cursor review

buenaflor:
@sentry review

packages/flutter/lib/src/integrations/native_app_start_handler_v2.dart
```diff
@@ -0,0 +1,116 @@
...
+      }
+    }
+
+    pluginRegistrationSpan.end(endTimestamp: appStartInfo.pluginRegistration);
+    sentrySetupSpan.end(endTimestamp: appStartInfo.sentrySetupStart);
+    firstFrameRenderSpan.end(endTimestamp: appStartEnd);
+
+    appStartSpan.end(endTimestamp: appStartEnd);
+
+    // TODO(next-pr): add mobile vitals specific attributes later
```
buenaflor, line 97: spec is still tbd

packages/flutter/lib/src/navigation/time_to_display_tracker_v2.dart
```diff
@@ -6,46 +6,117 @@ import '../../sentry_flutter.dart';
...
 
-  void trackRoute(String routeName) {
+  /// Prepares the root navigation span for native app start.
+  ///
+  /// Creates an idle span so child spans can attach before navigation fires.
+  /// Also creates the TTFD span so [ttfdSpanId] is available for
+  /// [SentryFlutter.currentDisplay] before [trackRootNavigation] fires.
+  /// Timestamps are backdated later in [trackRootNavigation].
+  void prepareRootNavigation() {
+    assert(_preparedRootNavigationSpan == null,
```
buenaflor, line 39: realistically won't happen but let's assert so it shows up in debug at least as an error

packages/dart/lib/src/telemetry/span/recording_sentry_span_v2.dart
```diff
@@ -42,12 +42,13 @@ base class RecordingSentrySpanV2 implements SentrySpanV2 {
...
     required DscCreatorCallback dscCreator,
     required SentryTracesSamplingDecision samplingDecision,
+    DateTime? startTimestamp,
   })  : _traceId = parentSpan?.traceId ?? traceId,
         _name = name,
         _parentSpan = parentSpan,
         _clock = clock,
         _onSpanEnd = onSpanEnd,
-        _startTimestamp = clock(),
+        _startTimestamp = (startTimestamp ?? clock()).toUtc(),
```
denrase, line 51: We call `.toUtc()` in the setter. Would it also make sense to do this here in the ctor?

packages/flutter/lib/src/integrations/native_app_start_handler_v2.dart
```diff
@@ -0,0 +1,116 @@
...
+      return;
+    }
+
+    final appStartInfo = parseNativeAppStart(nativeAppStart, appStartEnd);
+    if (appStartInfo == null) {
+      options.timeToDisplayTrackerV2.cancelCurrentRoute();
+      return;
+    }
+
+    final tracker = options.timeToDisplayTrackerV2;
```
denrase: Could do this at the start of the method, so we don't need multiple `options.timeToDisplayTrackerV2` calls.

packages/flutter/lib/src/integrations/native_app_start_handler_v2.dart
```diff
@@ -0,0 +1,116 @@
...
+            SentryAttribute.string(appStartInfo.appStartTypeOperation),
+        SemanticAttributesConstants.sentryOrigin:
+            SentryAttribute.string(SentryTraceOrigins.autoUiTimeToDisplay),
+      },
+    );
+
+    final pluginRegistrationSpan = hub.startInactiveSpan(
+      appStartInfo.pluginRegistrationDescription,
+      parentSpan: appStartSpan,
+      startTimestamp: appStartInfo.start,
```
denrase, line 58: We have introduced `startTimeStamp` in ctors, but also added it as a setter. Could we do with just one of them or do we need both API surfaces?

packages/flutter/lib/src/integrations/native_app_start_handler_v2.dart
```diff
@@ -0,0 +1,116 @@
...
+          },
+        );
+        nativeSpan.end(endTimestamp: timeSpan.end);
+      } catch (error, stackTrace) {
+        internalLogger.error('Failed to attach native span to app start',
+            error: error, stackTrace: stackTrace);
+      }
+    }
+
+    pluginRegistrationSpan.end(endTimestamp: appStartInfo.pluginRegistration);
```
denrase, line 91: Does it make sense to also add EndTimestamp to `startInactiveSpan`?

packages/flutter/lib/src/integrations/native_app_start_handler_v2.dart
```diff
@@ -0,0 +1,116 @@
...
+            SentryAttribute.string(SentryTraceOrigins.autoUiTimeToDisplay),
+      },
+    );
+
+    final pluginRegistrationSpan = hub.startInactiveSpan(
+      appStartInfo.pluginRegistrationDescription,
+      parentSpan: appStartSpan,
+      startTimestamp: appStartInfo.start,
+      attributes: {
+        SemanticAttributesConstants.sentryOp:
```
denrase: We could DRY the attributes if they are the same for every individual span here.

packages/flutter/lib/src/integrations/native_app_start_handler_v2.dart
```diff
@@ -0,0 +1,116 @@
...
+        );
+        nativeSpan.end(endTimestamp: timeSpan.end);
+      } catch (error, stackTrace) {
+        internalLogger.error('Failed to attach native span to app start',
+            error: error, stackTrace: stackTrace);
+      }
+    }
+
+    pluginRegistrationSpan.end(endTimestamp: appStartInfo.pluginRegistration);
+    sentrySetupSpan.end(endTimestamp: appStartInfo.sentrySetupStart);
```
denrase, line 92: The end of `sentrySetupSpan` is `appStartInfo.sentrySetupStart`?

packages/flutter/lib/src/integrations/native_app_start_integration.dart
```diff
@@ -49,12 +63,22 @@ class NativeAppStartIntegration extends Integration<SentryFlutterOptions> {
...
+            await _nativeAppStartHandlerV2.call(
+              hub,
+              options,
+              appStartEnd: appStartEnd,
+            );
+          case SentryTraceLifecycle.static:
+            await _nativeAppStartHandler.call(
+              hub,
+              options,
+              context: context!,
```
denrase: Can we avoid the force unwrap here?

packages/flutter/lib/src/navigation/sentry_navigator_observer.dart
```diff
@@ -250,7 +250,7 @@ class SentryNavigatorObserver extends RouteObserver<PageRoute<dynamic>> {
     }
 
     if (_hub.options.traceLifecycle == SentryTraceLifecycle.streaming) {
-      _timeToDisplayTrackerV2?.trackRoute(routeName);
+      _timeToDisplayTrackerV2?.trackNonRootNavigation(routeName);
```
denrase: trackNonRootNavigation seems verbose. Why not keep trackRoute?

packages/flutter/lib/src/navigation/time_to_display_tracker_v2.dart
```diff
@@ -72,11 +170,11 @@ class TimeToDisplayTrackerV2 {
     _ttfdSpan = null;
   }
 
+  /// Cancels all spans for the current route and resets tracker state.
   void cancelCurrentRoute() {
     _ttfdSpan = null;
```
denrase, line 175: In reportFullyDisplayed we end the _ttfdSpan, here we just set it to null. Is this valid?

packages/flutter/test/navigation/sentry_navigator_observer_test.dart
```diff
@@ -1152,65 +1152,49 @@ void main() {
       streamingFixture = _StreamingFixture();
     });
 
-    test('didPush calls trackRoute with route name', () {
+    test('didPush calls trackNonRootNavigation with route name', () {
```
denrase: It's hard to do reviews if we do renamings like trackRoute to trackNonRootNavigation, as it's more difficult to spot real logic/expectation changes in the diff. 😅

denrase (changes requested):

packages/flutter/lib/src/navigation/time_to_display_tracker_v2.dart
buenaflor: yeah because the idle span of v2 now properly finishes all child spans automatically so we don't need to manually finish it

packages/flutter/test/navigation/sentry_navigator_observer_test.dart
buenaflor: for this case `trackRoute` was not really the best choice since it meant `track all navigation except root`

packages/dart/lib/src/telemetry/span/recording_sentry_span_v2.dart
buenaflor: we already do set it here  in the ctor or do you mean a different path?

packages/flutter/lib/src/integrations/native_app_start_handler_v2.dart
buenaflor: what do you mean by that? `startInactiveSpan` already supports endTimestamp

packages/flutter/lib/src/integrations/native_app_start_handler_v2.dart
buenaflor: yup, sentry setup span is the span duration from plugin registration to sentry setup start

packages/flutter/lib/src/integrations/native_app_start_handler_v2.dart
buenaflor: we need both because (internally) we need to be able to backdate here because we only get the startTimestamp of the native app start later (due to async)

packages/flutter/lib/src/navigation/sentry_navigator_observer.dart
buenaflor: To me it wasn't really clear if trackRoute tracks everything, and it does in fact not cover root navigation because these are separately tracked through app start integrations

packages/flutter/lib/src/integrations/native_app_start_handler_v2.dart
buenaflor: what do you mean by that? this is function is only called once

packages/dart/lib/src/telemetry/span/recording_sentry_span_v2.dart
denrase: ah, did't see it, sry

packages/flutter/lib/src/integrations/native_app_start_handler_v2.dart
denrase:
```dart
final tracker = options.timeToDisplayTrackerV2;

final nativeAppStart = await _native.fetchNativeAppStart();
if (nativeAppStart == null) {
  tracker.cancelCurrentRoute();
  return;
}

final appStartInfo = parseNativeAppStart(nativeAppStart, appStartEnd);
if (appStartInfo == null) {
  tracker.cancelCurrentRoute();
  return;
}
...
```

packages/flutter/lib/src/integrations/native_app_start_handler_v2.dart
denrase: I mean in the method.

```dart
@override
  SentrySpanV2 startInactiveSpan(
    String name, {
    SentrySpanV2? parentSpan = const UnsetSentrySpanV2(),
    Map<String, SentryAttribute>? attributes,
    DateTime? startTimestamp,
    DateTime? endTimestamp,
  })
  ```

denrase: So if we know both values beforehand, we can just pass them.

packages/flutter/lib/src/integrations/native_app_start_handler_v2.dart
buenaflor: ah yeah lol, didn't see that

packages/flutter/lib/src/navigation/sentry_navigator_observer.dart
denrase: Since we only ever call it for non-route routes and there is a separete one for the root span, my prefernce would have been to leave the short name. No need to change back though if you want to keep it.

packages/flutter/lib/src/integrations/native_app_start_handler_v2.dart
buenaflor: hm I don't think this has a purpose outside of obscure internal things like in this case 

most users won't interact with the API like this (since we'll be making it public)

denrase (approved):
Added some more context to the questions, feel free to pick up if they make sense.

packages/flutter/lib/src/navigation/sentry_navigator_observer.dart
buenaflor: no you're right, after thjinking about this I'll keep this as `trackRoute`

PR #3539, opened 2026-02-24
fix(sqflite): Implement SqfliteDatabaseExecutor to prevent TypeError on getVersion/setVersion

### LINKED ISSUE #3530: SentryDatabase crashes with TypeError when sqflite extension calls getVersion()
## Environment

| Package | Version |
|---------|---------|
| sentry_sqflite | 9.10.0 |
| sqflite | 2.4.2 |
| sqflite_common | 2.5.5 |
| Flutter | 3.32.7 |
| Dart | 3.8.1 |
| Device | Galaxy Note8 (SM-N950F), Android 9 |

## Description

`SentryDatabase` crashes at runtime when `getVersion()` is called via sqflite_common's `SqfliteDatabaseExecutorExt` extension. The extension performs a hard cast `(this as SqfliteDatabaseExecutor)` which fails because `SentryDatabase` is a decorator that `implements Database` but does not extend `SqfliteDatabaseExecutor`.

## Error

```
TypeError: type 'SentryDatabase' is not a subtype of type 'SqfliteDatabaseExecutor' in type cast
```

## Stack Trace

```
at SqfliteDatabaseExecutorExt._db (sqlite_api.dart:348)
at SqfliteDatabaseExecutorExt.getVersion (sqlite_api.dart:364)
at MeditationDatabase.createSession (meditation_database.dart)
at <asynchronous suspension>
at EnhancedMeditationService.startSession (enhanced_meditation_service.dart)
at <asynchronous suspension>
at EnhancedMeditationProvider.startSession (enhanced_meditation_provider.dart)
at <asynchronous suspension>
```

## Root Cause

In `sqflite_common` 2.5.5, [`SqfliteDatabaseExecutorExt`](https://github.com/tekartik/sqflite/blob/master/sqflite_common/lib/sqlite_api.dart) is an extension on `DatabaseExecutor` that provides `getVersion()` and `setVersion()`:

```dart
extension SqfliteDatabaseExecutorExt on DatabaseExecutor {
  SqfliteDatabase get _db => (this as SqfliteDatabaseExecutor).db;  // hard cast
  
  Future<int> getVersion() {
    _db.checkNotClosed();
    return _db.txnGetVersion(_txn);
  }
}
```

`SentryDatabase` extends `SentryDatabaseExecutor` and `implements Database`, but does **not** implement `SqfliteDatabaseExecutor`. Since `getVersion()` is not an abstract method on `Database` (it's only an extension method on `DatabaseExecutor`), `SentryDatabase` doesn't override it, so the extension method fires and the cast fails.

## Reproduction

1. Open a database with `openDatabaseWithSentry()`
2. Use the returned `SentryDatabase` reference in any code path that triggers `getVersion()` via the extension

## Suggested Fix

Either:
1. Have `SentryDatabase` override `getVersion()` and `setVersion()` as instance methods that delegate to the wrapped database (takes priority over the extension)
2. Or have `SentryDatabaseExecutor` implement `SqfliteDatabaseExecutor` so the cast succeeds

Option 1 is the minimal fix:

```dart
class SentryDatabase extends SentryDatabaseExecutor implements Database {
  // ... existing code ...

  Future<int> getVersion() => _database.getVersion();
  Future<void> setVersion(int version) => _database.setVersion(version);
}
```

Note: the same issue likely affects `SentrySqfliteTransaction` and `SentryDatabaseExecutor` if any code path calls `getVersion()`/`setVersion()` on those wrappers.

denrase:
## :scroll: Description

Implement `SqfliteDatabaseExecutor` on Sentry wrapper classes so the hard cast in sqflite_common's getVersion()/setVersion() extension methods succeeds.

Fixes #3530 

Added tests (were failing before implementing solution)

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

### DIFF
--- a/packages/sqflite/lib/src/sentry_database_executor.dart
+++ b/packages/sqflite/lib/src/sentry_database_executor.dart
@@ -15,6 +15,9 @@ import 'utils/sentry_database_span_attributes.dart';
 
 @internal
 // ignore: public_member_api_docs
+// Workaround: sqflite hard-casts DatabaseExecutor to SqfliteDatabaseExecutor
+// (see sqlite_api.dart SqfliteDatabaseExecutorExt), causing a TypeError when wrapped.
+// https://github.com/tekartik/sqflite/blob/c28b35b17f98e714cee6e540ef6a96e48642aafd/sqflite_common/lib/sqlite_api.dart#L349
 class SentryDatabaseExecutor
     implements DatabaseExecutor, SqfliteDatabaseExecutor {
   final DatabaseExecutor _executor;
--- a/packages/sqflite/lib/src/sentry_database_executor.dart
+++ b/packages/sqflite/lib/src/sentry_database_executor.dart
@@ -3,7 +3,11 @@ import 'package:sentry/sentry.dart';
 import 'package:sqflite/sqflite.dart';
 
 // ignore: implementation_imports
+import 'package:sqflite_common/src/database.dart';
+// ignore: implementation_imports
 import 'package:sqflite_common/src/sql_builder.dart';
+// ignore: implementation_imports
+import 'package:sqflite_common/src/transaction.dart';
 
 import 'sentry_batch.dart';
 import 'sentry_database.dart';
@@ -11,7 +15,7 @@ import 'utils/sentry_database_span_attributes.dart';
 
 @internal
 // ignore: public_member_api_docs
-class SentryDatabaseExecutor implements DatabaseExecutor {
+class SentryDatabaseExecutor implements DatabaseExecutor, SqfliteDatabaseExecutor {
   final DatabaseExecutor _executor;
   // ignore: invalid_use_of_internal_member
   final InstrumentationSpan? _parentSpan;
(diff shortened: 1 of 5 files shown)


packages/sqflite/lib/src/sentry_database_executor.dart
```diff
@@ -41,6 +46,16 @@ class SentryDatabaseExecutor implements DatabaseExecutor {
...
+  // Implement SqfliteDatabaseExecutor so that the cast in
+  // SqfliteDatabaseExecutorExt (this as SqfliteDatabaseExecutor) succeeds.
+  // Without this, calling getVersion()/setVersion() on a SentryDatabaseExecutor
+  // crashes with a TypeError.
+  @override
+  SqfliteDatabase get db => (_executor as SqfliteDatabaseExecutor).db;
+
+  @override
+  SqfliteTransaction? get txn => (_executor as SqfliteDatabaseExecutor).txn;
+
```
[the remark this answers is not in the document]
denrase: we're not making anythingworse — we're just forwarding the same cast that sqflite itself does

packages/sqflite/lib/src/sentry_sqflite_transaction.dart
```diff
@@ -18,7 +22,8 @@ import 'sentry_batch.dart';
 /// });
 /// ```
 @experimental
```
buenaflor, line 24: not relevant to this pr but I didn't know this is still marked as experimental, I'll update the v10 task tracking issue for removing this

packages/sqflite/lib/src/sentry_database_executor.dart
```diff
@@ -2,16 +2,21 @@ import 'package:meta/meta.dart';
...
 
 import 'sentry_batch.dart';
 import 'sentry_database.dart';
 import 'utils/sentry_database_span_attributes.dart';
 
 @internal
 // ignore: public_member_api_docs
-class SentryDatabaseExecutor implements DatabaseExecutor {
+class SentryDatabaseExecutor
+    implements DatabaseExecutor, SqfliteDatabaseExecutor {
```
buenaflor, line 22: maybe we should document this behaviour here as well as it might break in the future again if they decide to change it (not ideal but it's actually an upstream issue with the dependency, they really shouldn't hard cast to a separate type ...)

buenaflor (commented):
Overall looks good, I think we should still document this in some way here

buenaflor (approved):

PR #3542, opened 2026-02-27
Sync JNI branch with 9.14.0

buenaflor:
#skip-changelog


packages/flutter/pubspec.yaml
```diff
@@ -46,7 +46,7 @@ dev_dependencies:
     git:
       url: https://github.com/getsentry/ffigen
       ref: 6aa2c2642f507eab3df83373189170797a9fa5e7
-  jnigen: ^0.15.0
+  jnigen: 0.14.2
```
Danyalo, line 49: @buenaflor I think this change defeats the point of this branch

buenaflor: yeah that's wrong.. this seems to be a merge issue

we'll fix this asap on monday

you can pin to the previous commit if you need a correct build right now

denrase: @Danyalo This is fixed now, we set `0.15.0` again and re-generated the binding.

Danyalo: Thanks a lot, that was quick!

Issue #3527, opened 2026-02-18
[sentry_logging] Support custom objects for messages


shaakunuigurumi:
### Problem Statement

It feels like a trade off choosing between using Sentry SDK directly for logging or using the supported `logging` package. I'd like to keep using logging for the official Dart feeling, but I do want rich logs later in Sentry.

I noticed that `logging` actually accepts `Object?` not just `String`. In the [documentation](https://pub.dev/documentation/logging/latest/logging/Logger/log.html), it states that the `message` will be the `toString()` result of the object given and its original instance is available in [`LogRecord`](https://pub.dev/documentation/logging/latest/logging/LogRecord-class.html).

### Solution Brainstorm

Could be there an option to have an intermediary class that can be used to utilize the most of Sentry (e.g. placeholders/formatting, attributes, etc.) while playing nicely with the `logging` package? It could be beneficial if some project wants to swap between packages or approaches.

If we were to consider the main point of this feature request being Sentry possibly being removed or deactivated, an alternative could be a hook/callback, that allows us to transform/translate the log message ourselves (allowing for custom classes). Or does that explode the scope of the `sentry_logging` package, and a custom integration is warranted?

### Are you willing to submit a PR?

Yes

buenaflor:
Thanks for the issue! 

I agree the experience should be more integrated with logging. 

I'll have a look

denrase:
@shaakunuigurumi We've added `Hint` support to `beforeSendLog` in an experimental branch. I can't guarantee that this will land on main, for now it's only an exploration if this would cover your use-case. It would also be a breaking API change.

`LoggingIntegration` would passes the full `LogRecord` as a hint, so you can keep using `logging` normally and derive Sentry attributes in one place:

```dart
await SentryFlutter.init((options) {
  options.dsn = '...';
  options.enableLogs = true;
  options.addIntegration(LoggingIntegration(minSentryLogLevel: Level.INFO));

  options.beforeSendLog = (log, {Hint? hint}) {
    final record = hint?.get(TypeCheckHint.record);
    if (record is LogRecord && record.object is PaymentInfo) {
      final info = record.object as PaymentInfo;
      log.attributes['user'] = SentryAttribute.string(info.user);
      log.attributes['amount'] = SentryAttribute.double(info.amount);
    }
    return log;
  };
});

// Standard logging — no Sentry-specific code at call sites
final log = Logger('PaymentService');
[...]

PR #3549, opened 2026-03-10
feat: Add Hint support to `beforeSendLog` and pass LogRecord from sentry_logging

denrase:
## :scroll: Description

Adds an optional Hint parameter to BeforeSendLogCallback and threads it through the entire log capture pipeline. The LoggingIntegration now passes the full LogRecord as a hint, allowing users to access the original object in beforeSendLog and derive custom attributes from it — without needing Sentry-specific code at log call sites.

For now this in only an exploration of a possible solution to #3527

- Added tests

## :pencil: Checklist

- [x] All tests passing
- [ ] No breaking changes

## :crystal_ball: Next steps

This is a breaking change and should be part of a major version.


denrase:
@buenaflor AFAIK, only [python SDK](https://github.com/getsentry/sentry-python/blob/bde87ac5f197bc33683929269593d7b9b360549d/sentry_sdk/utils.py#L2027) offers a hint in the before send log method.

packages/logging/lib/src/logging_integration.dart
```diff
@@ -108,25 +108,32 @@ class LoggingIntegration implements Integration<SentryOptions> {
         'sentry.origin': SentryAttribute.string(origin),
       };
 
+      final hint = Hint.withMap({TypeCheckHint.record: record});
```
buenaflor, line 111: We can consider making this more convenient and easy to discover for users.

something like `hint.logging.record` 

logging -> refers to the logging integration / package
record -> give me the record associated with this log if there is one

```dart
/// Logging-specific hint data.
class LoggingHintData {
  final Hint _hint;

  LoggingHintData(this._hint);

  /// The original [LogRecord] from the `logging` package, if this
  /// log originated from [LoggingIntegration].
  LogRecord? get record => _hint.get(TypeCheckHint.record) as LogRecord?;
}

extension SentryLoggingHint on Hint {
  /// Access logging-specific hint data.
  LoggingHintData get logging => LoggingHintData(this);
}
```

But this is just suggestion, we don't have to do this but I think it would be nice to have.

denrase: Great idea, since this is in the logging package and the pattern is not used elsewhere, i added it directly to hint as an extension.

buenaflor:
Converting this back to draft until we get an idea on the usage of beforeSend(log)

buenaflor:
We are planning on Major v10 soon so we won't go ahead with this at this time

PR #3574, opened 2026-03-18
feat(span-first): Add `startSpan` and `startSpanSync`

buenaflor:
## :scroll: Description

Updates from a single `FutureOr<T> startSpan` to two distinct ones.

`FutureOr` gets dropped because it pushes ambiguity onto every caller.

It's ergonomically also not well to work with and we want to maximize dev experience as much as possible

- Manually, existing tests and update `hub_span_test`

## :pencil: Checklist
- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps

#skip-changelog

### DIFF
--- a/packages/dart/lib/src/hub.dart
+++ b/packages/dart/lib/src/hub.dart
@@ -669,11 +669,10 @@ class Hub {
         return () async {
           final forkedScope = _forkScopeWithActiveSpan(recordingSpan);
           try {
-            final result = await runZoned(
+            return await runZoned(
               () => callback(recordingSpan),
               zoneValues: {_scopeKey: forkedScope},
             );
-            return result;
           } catch (_) {
             span.status = SentrySpanStatusV2.error;
             rethrow;
@@ -712,11 +711,10 @@ class Hub {
       case RecordingSentrySpanV2 recordingSpan:
         final forkedScope = _forkScopeWithActiveSpan(recordingSpan);
         try {
-          final result = runZoned(
+          return runZoned(
             () => callback(recordingSpan),
             zoneValues: {_scopeKey: forkedScope},
           );
-          return result;
         } catch (_) {
           span.status = SentrySpanStatusV2.error;
           rethrow;
--- a/packages/dart/lib/src/hub.dart
+++ b/packages/dart/lib/src/hub.dart
@@ -627,23 +627,24 @@ class Hub {
 
   static final _scopeKey = Object();
 
-  /// The scope forked by the innermost [startSpan] call on the current zone
-  /// chain, or `null` if we are outside any [startSpan] callback.
+  /// The scope forked by the innermost [startSpan] or [startSpanSync] call on
+  /// the current zone chain, or `null` if we are outside any span callback.
   Scope? get _zoneScope => Zone.current[_scopeKey] as Scope?;
 
   /// Returns the active span from the current zone's forked scope, or `null`
   /// if no span is active.
   ///
-  /// Only zone-forked scopes (created by [startSpan]) carry an active span.
+  /// Only zone-forked scopes (created by [startSpan] or [startSpanSync]) carry
+  /// an active span.
   /// The hub's top-level scope is never mutated, so calling this outside a
   /// [startSpan] callback always returns `null`.
   @internal
   RecordingSentrySpanV2? getActiveSpan() =>
       _zoneScope?.getActiveSpan() ?? _currentIdleSpan;
 
-  T startSpan<T>(
+  Future<T> startSpan<T>(
     String name,
-    T Function(SentrySpanV2 span) callback, {
+    Future<T> Function(SentrySpanV2 span) callback, {
     Map<String, SentryAttribute>? attributes,
     SentrySpanV2? parentSpan = const UnsetSentrySpanV2(),
     DateTime? startTimestamp,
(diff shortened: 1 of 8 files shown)


packages/dart/lib/src/sentry.dart
```diff
@@ -411,14 +413,34 @@ class Sentry {
...
+    Future<T> Function(SentrySpanV2 span) callback, {
     Map<String, SentryAttribute>? attributes,
     SentrySpanV2? parentSpan = const UnsetSentrySpanV2(),
+    DateTime? startTimestamp,
   }) =>
       _hub.startSpan(name, callback,
-          attributes: attributes, parentSpan: parentSpan);
+          attributes: attributes,
+          parentSpan: parentSpan,
+          startTimestamp: startTimestamp);
```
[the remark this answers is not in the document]
buenaflor: it's not public yet, it's fine

packages/dart/lib/src/hub.dart
```diff
@@ -662,52 +671,67 @@ class Hub {
...
-        return Future<T>.error(error, stackTrace);
-      });
+  T startSpanSync<T>(
+    String name,
+    T Function(SentrySpanV2 span) callback, {
+    Map<String, SentryAttribute>? attributes,
+    SentrySpanV2? parentSpan = const UnsetSentrySpanV2(),
+    DateTime? startTimestamp,
+  }) {
+    final span = _createSpan(
```
denrase, line 698: Looks like we have some duplicated code here, would it make sense to do one internal function that both the asynchronous and sync call?

packages/dart/lib/src/hub.dart
```diff
@@ -662,52 +671,67 @@ class Hub {
...
+      case RecordingSentrySpanV2 recordingSpan:
+        final forkedScope = _forkScopeWithActiveSpan(recordingSpan);
+        try {
+          return runZoned(
+            () => callback(recordingSpan),
+            zoneValues: {_scopeKey: forkedScope},
+          );
+        } catch (_) {
+          span.status = SentrySpanStatusV2.error;
+          rethrow;
```
denrase, line 724: I think we have some kind of flag if we re-throw in testing.

packages/dart/lib/src/hub.dart
```diff
@@ -662,52 +671,67 @@ class Hub {
...
+      case NoOpSentrySpanV2():
+        internalLogger
+            .info('Hub: startSpanSync returning no-op for \'$name\'.');
+        return callback(span);
+      case UnsetSentrySpanV2():
+        internalLogger.error(
+          'Hub: _createSpan returned UnsetSentrySpanV2 for \'$name\'. '
+          'This is a bug — the sentinel should never leak out of _createSpan.',
+        );
+        return callback(span);
```
denrase, line 714: I remembered that we have some kind of handling user provided callbacks in the case that they throw. Worth checking what we du in other SDK places and handle it the same way, including a test case. Same for the async method.

packages/dart/test/hub_span_test.dart
```diff
@@ -514,8 +514,220 @@ void main() {
...
+        test(
+          'provides NoOpSentrySpanV2 and still calls callback when hub is closed',
+          () async {
+            final hub = fixture.getSut();
+            await hub.close();
+            var callbackInvoked = false;
+
+            final result = hub.startSpanSync('test-span', (span) {
+              callbackInvoked = true;
+              expect(span, isA<NoOpSentrySpanV2>());
```
denrase, line 527: If we do this, is NoOpSentrySpanV2 supposed to be a public type that users can check against. If not, how can they know that the span in question is not being send? If yes, do we document this behaviour for them?

We definitely always have to call it, just wondering how to surface that the span will not be recorded.

packages/dart/test/hub_span_test.dart
```diff
@@ -514,8 +514,220 @@ void main() {
...
+
+          hub.startSpanSync('test-span', (span) {});
+
+          expect(fixture.client.captureSpanCalls, hasLength(1));
+        });
+
+        test('captures all nested spans', () {
+          final hub = fixture.getSut();
+
+          hub.startSpanSync('outer', (span) {
```
denrase, line 697: Is it possible to mix & match the sync/async api? If yes, we should have tests (ignore if there are ones already). If no, we should clearly document this.

denrase (changes requested):

packages/dart/test/hub_span_test.dart
buenaflor: some SDKs provide a `isNoop` for convenience but yeah the noop class is public and can be checked against.

I don't think we have to really surface it any further. we already log if it's being dropped. imo that should be enough

packages/dart/lib/src/hub.dart
buenaflor: imo this case is different since we never want to swallow user code here and this only wraps user code (callback)

so if for some reason user code errors we want to bubble it back up to them

packages/dart/test/hub_span_test.dart
buenaflor: yeah it's possible to mix and match, will update docs and tests

packages/dart/test/hub_span_test.dart
buenaflor: I'll just update the function docs and refer to the noop type

packages/dart/lib/src/hub.dart
buenaflor: I don't think we can further meaningfully dedupe here because of behaviour differences of try/catch/finally differences for sync and async

packages/dart/lib/src/hub.dart
denrase: Please just check if it would make sense to log/rethrow an error here.

denrase (approved):
Just one comment for the non-wrapped comments.

packages/dart/lib/src/hub.dart
buenaflor: hmm I'm not sure I understand, the exception already propagates to the user since we don't try/catch around it

or do you mean something else?

buenaflor: resolved internally

Issue #3373, opened 2025-11-27
Bump JNI to 0.15.0 to be compatible with other packages


fluttersiast:
edit from @buenaflor: until the next major we are unable to bump JNI due to higher min versions used in the new JNI versions, you can use instead the branch: `deps/jni-0.15.x`if you require jni 0.15.x support

Note: as we keep the branch updated with the latest Sentry Flutter version we highly recommend pinning the dependency to commits to prevent possible regressions.

```yaml
# in pubspec.yaml
sentry_flutter:
  git:
    url: git@github.com:getsentry/sentry-dart.git
    ref: deps/jni-0.15.x
    path: packages/flutter
sentry:
  git:
    url: git@github.com:getsentry/sentry-dart.git
    ref: deps/jni-0.15.x
    path: packages/dart
```

---

### Platform

Flutter Mobile iOS, Dart

### Obfuscation

Enabled

### Debug Info

Enabled

### Doctor

Please plan to bump version of JNI to 0.15.x to be compatible with other packages.

### Version

9.8.0

### Steps to Reproduce

Use latest version and cronet_http(or any other package which is using latest jni)

### Expected Result

flutter pub get works as expected

### Actual Result

packages has incompatible jni versions

### Are you willing to submit a PR?

None

buenaflor:
hey, we will bump the jni version soon

dickermoshe:
@buenaflor Can this be done with objective_c too?

buenaflor:
> Can this be done with objective_c too?

Unfortunately no because objc 9.0.0 has a much higher min Dart version than we support (and they update it quite frequently 😅 )

However, we are planning on reverting our FFI usage on iOS/macOS due to issues and therefore will also be removing the objective_c package - once that's released there won't be any conflicts anymore in that regard

dickermoshe:
@buenaflor I also added ffi/jni to my app too early, had some nasty segfualts.

fluttersiast:
Hi!

Soon new cronet version with some critical bugfixes will be released https://github.com/dart-lang/http/pull/1862
Don't want to push you guys, but do you have any timelines or beta?

buenaflor:
> Don't want to push you guys, but do you have any timelines or beta?

all good! I'm doing a `9.9.1` release today and then the plan is to update JNI. I'll prepare a beta release for that JNI bump in `9.10.0-beta.0`

buenaflor:
[...]

PR #3593, opened 2026-03-26
ci: Automate JNI branch sync on release

buenaflor:
Automates the sync of the `deps/jni-0.15.x` branch when a new release is published.

**How it works:**

1. On release → merges the release tag into a `sync/jni-{tag}` branch, regenerates JNI bindings, opens a PR targeting `deps/jni-0.15.x`
2. On PR merge (push to `deps/jni-0.15.x`) → updates the version tracking table on #3373 with the merge commit SHA

The regeneration script (`safe-regenerate-jni.sh`) has a two-phase approach:
- **Normal path**: `flutter build apk` + jnigen (works when `binding.dart` is compatible)
- **Fallback path**: compiles only the Java/Kotlin plugin code via Gradle, creates a stub `libs.jar`, and runs jnigen directly — bypassing Dart compilation. This handles the case where a jni version bump breaks the existing generated bindings.

The fallback was validated locally by deliberately corrupting `binding.dart` and confirming the script recovers and produces valid bindings.

**One-time setup needed:** after merging this PR, cherry-pick the workflow file to `deps/jni-0.15.x` so the `push` trigger for the version tracker works from day one. Subsequent syncs will keep it up to date automatically.

Refs GH-3373


.github/workflows/sync-jni-branch.yml
```diff
@@ -0,0 +1,217 @@
...
+        id: token
+        uses: actions/create-github-app-token@f8d387b68d61c58ab83c6c016672934102569859 # v3.0.0
+        with:
+          app-id: ${{ vars.SENTRY_RELEASE_BOT_CLIENT_ID }}
+          private-key: ${{ secrets.SENTRY_RELEASE_BOT_PRIVATE_KEY }}
+
+      - name: Determine release tag
+        id: release
+        run: |
+          TAG="${{ github.event.release.tag_name }}"
```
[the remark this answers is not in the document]
denrase: We also addressed these on the cocoa side, worth doing it here as well.

.github/workflows/sync-jni-branch.yml
```diff
@@ -0,0 +1,217 @@
...
+        run: |
+          git merge "${{ steps.release.outputs.tag }}" --no-edit \
+            -m "Merge release ${{ steps.release.outputs.tag }} into ${{ env.JNI_BRANCH }}"
+
+      - name: Regenerate JNI bindings
+        id: regenerate
+        continue-on-error: true
+        working-directory: packages/flutter
+        run: |
+          scripts/safe-regenerate-jni.sh 2>&1 | tee /tmp/jni-regen.log
```
denrase, line 95: I remember you mentioning that some files need to be deleted or commented out before re-generation, is this still true?

packages/flutter/scripts/safe-regenerate-jni.sh
```diff
@@ -0,0 +1,106 @@
...
+(
+    cd example
+    $FLUTTER build apk
+)
+verify_result=$?
+set -e
+
+if [ $verify_result -ne 0 ]; then
+    echo ""
+    echo "=== ERROR: Regenerated bindings do not compile ==="
```
denrase, line 100: Maybe work logging the error code or failure logs.

denrase (approved):
Copilot comments are worth checking out.

.github/workflows/sync-jni-branch.yml
buenaflor: safe-regenerate-jni.sh handles this now - it's not strictly needed anymore since the script will create a stub so it still compiles

.github/workflows/sync-jni-branch.yml
```diff
@@ -0,0 +1,247 @@
...
+          git merge "$TAG" --no-edit \
+            -m "Merge release $TAG into $BRANCH"
+
+      - name: Regenerate JNI bindings
+        id: regenerate
+        if: steps.merge.outcome == 'success'
+        continue-on-error: true
+        working-directory: packages/flutter
+        run: |
+          scripts/safe-regenerate-jni.sh 2>&1 | tee /tmp/jni-regen.log
```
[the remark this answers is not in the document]
buenaflor: not an issue - pipefail is already enabled by default. if the script fails, the pipe's exit code will reflect that

packages/flutter/scripts/safe-regenerate-jni.sh
buenaflor: afaict the errors should be visible in the logs when running flutter build apk so we dont need to log more here

PR #3608, opened 2026-03-30
chore: Add SDK features for beforeSend callbacks

buenaflor:
## :scroll: Description

Register configured `beforeSend` callbacks as SDK features at `SentryClient` construction time. When a user configures any of the beforeSend callbacks, the corresponding feature string is added via `sdk.addFeature()`:

- `beforeSend` → `beforeSendEvent`
- `beforeSendTransaction` → `beforeSendTransaction`
- `beforeSendFeedback` → `beforeSendFeedback`
- `beforeSendLog` → `beforeSendLog`
- `beforeSendMetric` → `beforeSendMetric`

Since `addFeature` deduplicates, re-creating the client is safe.

This lets us track which beforeSend APIs users have configured as SDK metadata, without needing to know how often the callbacks are actually invoked.

Added 6 unit tests in `sentry_client_test.dart` covering:
- Each callback type adds its corresponding feature when configured
- No features are added when no callbacks are set

## :pencil: Checklist
- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps


packages/dart/lib/src/sentry_client.dart
```diff
@@ -81,6 +82,22 @@ class SentryClient {
     if (enableFlutterSpotlight) {
       options.transport = SpotlightHttpTransport(options, options.transport);
     }
+    if (options.beforeSend != null) {
```
denrase: Any particular reason we do this in the client and not somewhere else?

denrase (approved):

packages/dart/lib/src/sentry_client.dart
buenaflor: yeah thinking about it I think it could be a `TrackFeaturesIntegration` since we should opt for integrations when possible

PR #3643, opened 2026-04-16
test(supabase): Fix flaky error client test for postgrest retry

buenaflor:
## :scroll: Description

Change the `should capture error if send throws` test assertion from exact count (`1`) to `greaterThanOrEqualTo(1)` to decouple it from postgrest's internal retry behavior.

`postgrest` 2.7.0 added automatic retry logic for GET/HEAD requests on transient failures (up to 3 retries). Since the test's mock client always throws, each retry passes through `SentrySupabaseErrorClient.send()` and triggers `captureEvent` — resulting in 4 calls instead of the expected 1.

This is not a bug in our code: duplicates are deduplicated by event processors in the full pipeline, which the unit test doesn't exercise (it mocks the Hub directly).

`dart test packages/supabase/test/sentry_supabase_error_client_test.dart` — all 13 tests pass.

## :pencil: Checklist
- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps

### DIFF
--- a/packages/supabase/test/sentry_supabase_error_client_test.dart
+++ b/packages/supabase/test/sentry_supabase_error_client_test.dart
@@ -77,7 +77,8 @@ void main() {
         expect(e, error); // Error is rethrown
       }
 
-      expect(fixture.mockHub.captureEventCalls.length, 1);
+      // Supabase retries the request up to 3 times, so we expect at least 1 event.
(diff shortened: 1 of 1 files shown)


denrase (approved):
Should we consider handling these retries in our supabase package so we do not report them, even if they are de-duplicated?

buenaflor:
imo no need to since it's handled in the pipeline - maybe if there are specific issues that users come across we can take a look but for now no need for any handling here

PR #3678, opened 2026-04-30
feat(span-first): Add transaction and app start type span attributes

buenaflor:
## :scroll: Description

Add streaming span attributes for Sentry transaction/segment naming and app start classification.

Streaming spans now include the deprecated `sentry.transaction` attribute alongside `sentry.segment.name`, both populated from the local segment name. V2 native app-start child spans also receive `app.vitals.start.type` so each app-start span carries its cold/warm classification.

`sentry.transaction` is needed for the mobile vitals page to work correctly

This keeps span streaming payloads compatible with consumers that still look for the transaction attribute while also exposing app-start type on every app-start child span, not only the top-level app-start span.

- `fvm dart test test/telemetry/span/span_capture_pipeline_test.dart`
- `fvm flutter test test/integrations/native_app_start_handler_v2_test.dart`
- `fvm dart analyze lib/src/constants.dart lib/src/telemetry/span/span_capture_pipeline.dart test/telemetry/span/span_capture_pipeline_test.dart`
- `fvm flutter analyze lib/src/integrations/native_app_start_handler_v2.dart test/integrations/native_app_start_handler_v2_test.dart`
- Pre-commit formatting and package analyze checks

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps

N/A

### DIFF
--- a/packages/dart/lib/src/constants.dart
+++ b/packages/dart/lib/src/constants.dart
@@ -84,6 +84,11 @@ abstract class SemanticAttributesConstants {
   /// The segment name (e.g., "GET /users")
   static const sentrySegmentName = 'sentry.segment.name';
 
+  /// The Sentry transaction name, also known as the segment name.
+  /// This is deprecated in favour of [sentrySegmentName].
+  /// We should remove this later when Sentry itself supports this deprecation.
+  static const sentryTransaction = 'sentry.transaction';
+
   /// The span id of the segment that this span belongs to.
   static const sentrySegmentId = 'sentry.segment.id';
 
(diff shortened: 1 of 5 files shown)


packages/dart/lib/src/constants.dart
```diff
@@ -84,6 +84,11 @@ abstract class SemanticAttributesConstants {
   /// The segment name (e.g., "GET /users")
   static const sentrySegmentName = 'sentry.segment.name';
 
+  /// The Sentry transaction name, also known as the segment name.
+  /// This is deprecated in favor of [sentrySegmentName].
```
denrase, line 88: Should we add a `@Deprecated` annotation here instead? Might get lost if it's just in the comments.

denrase (approved):
Just one comment regarding the deprecation.

packages/dart/lib/src/constants.dart
buenaflor: I'm not adding it because I don't wanna get flagged by the linter that we're using deprecated variables

PR #3680, opened 2026-04-30
build: Remove collection runtime dependency

### LINKED ISSUE #3679: Remove `collection` dependency
### Description

We can remove collection from runtime dependencies because production code only uses a couple of simple nullable iterable lookups (`firstWhereOrNull` / `firstOrNull`). Those are trivial to implement locally, and the repo already has a small `IterableUtils.firstWhereOrNull` helper. The heavier collection APIs are only used in tests, so they can remain as dev-only usage.

buenaflor:
## :scroll: Description
Removes `collection` from runtime dependencies in `sentry`, `sentry_flutter`, and `_sentry_testing` by replacing production nullable iterable lookups with a small `SentryIterableUtils` extension in the core SDK.

`collection` remains available as a dev dependency where tests still use its equality helpers.

Production code only relied on `firstOrNull` and `firstWhereOrNull`, so shipping the full `collection` package for runtime use is unnecessary. Keeping those helpers in the core SDK avoids an extra runtime dependency while preserving existing behavior.

- `fvm dart analyze` on changed Dart SDK files
- `fvm flutter analyze` on changed Flutter SDK files
- `fvm dart analyze` in `_sentry_testing`
- `fvm dart test test/utils/iterable_utils_test.dart test/sentry_event_test.dart test/event_processor/exception/io_exception_event_processor_test.dart test/sentry_test.dart`
- `fvm flutter test test/integrations/load_contexts_integration_test.dart test/navigation/time_to_display_tracker_test.dart test/navigation/sentry_navigator_observer_test.dart test/feedback/sentry_feedback_widget_test.dart`
- pre-commit analyze hooks

Closes #3679 

## :pencil: Checklist
- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps
None.

### DIFF
--- a/packages/dart/lib/src/utils/iterable_utils.dart
+++ b/packages/dart/lib/src/utils/iterable_utils.dart
@@ -1,11 +1,14 @@
-import 'package:meta/meta.dart';
+extension SentryIterableUtils<T> on Iterable<T>? {
+  T? get firstOrNull {
+    final iterator = this?.iterator;
+    if (iterator == null || !iterator.moveNext()) {
+      return null;
+    }
+    return iterator.current;
+  }
 
-@internal
-class IterableUtils {
-  static T? firstWhereOrNull<T>(
-    Iterable<T>? iterable,
-    bool Function(T item) test,
-  ) {
+  T? firstWhereOrNull(bool Function(T item) test) {
+    final iterable = this;
     if (iterable == null) {
       return null;
     }
--- a/packages/dart/lib/src/utils/iterable_utils.dart
+++ b/packages/dart/lib/src/utils/iterable_utils.dart
@@ -1,3 +1,6 @@
+import 'package:meta/meta.dart';
+
+@internal
 extension SentryIterableUtils<T> on Iterable<T>? {
   T? get firstOrNull {
     final iterator = this?.iterator;
--- a/packages/dart/lib/src/utils/iterable_utils.dart
+++ b/packages/dart/lib/src/utils/iterable_utils.dart
@@ -1,3 +1,6 @@
+import 'package:meta/meta.dart';
+
+@internal
 abstract final class SentryIterableUtils {
   static T? firstOrNull<T>(Iterable<T> iterable) {
     final iterator = iterable.iterator;
--- a/packages/dart/lib/src/utils/iterable_utils.dart
+++ b/packages/dart/lib/src/utils/iterable_utils.dart
@@ -1,20 +1,17 @@
 import 'package:meta/meta.dart';
 
 @internal
-abstract final class SentryIterableUtils {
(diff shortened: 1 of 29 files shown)


packages/flutter/lib/src/event_processor/android_platform_exception_event_processor.dart
```diff
@@ -89,7 +89,7 @@ class AndroidPlatformExceptionEventProcessor implements EventProcessor {
     MapEntry<SentryException, List<SentryThread>>? detailsStackTrace,
   ) {
     _markDartThreadsAsNonCrashed(event.threads);
-    final exception = event.exceptions?.firstOrNull;
+    final exception = event.exceptions.firstOrNull;
```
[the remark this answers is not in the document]
buenaflor: Not an issue.

packages/dart/lib/src/utils/iterable_utils.dart
```diff
@@ -1,11 +1,17 @@
 import 'package:meta/meta.dart';
 
 @internal
-class IterableUtils {
-  static T? firstWhereOrNull<T>(
-    Iterable<T>? iterable,
-    bool Function(T item) test,
-  ) {
+extension SentryIterableUtils<T> on Iterable<T>? {
```
denrase: We should consider only doing this on `Iterable<T>` instead of `Iterable<T>?`.

We'd immediately see if a collection was nullable at the call site, now we have hidden this in the extensions.

IMHO `exceptions?.firstOrNull` gives more context than `exceptions.firstOrNull`. With the latter i'd immediately assume exceptions is non-null. This is also the same behaviour as before with `collection` dependency.

```dart  
extension SentryIterableUtils<T> on Iterable<T> {
  T? get firstOrNull {
    final iterator = this.iterator;
    if (!iterator.moveNext()) {
      return null;
    }
    return iterator.current;
  }
}
```

packages/dart/lib/src/utils/iterable_utils.dart
```diff
@@ -1,11 +1,17 @@
...
+  T? get firstOrNull {
+    final iterator = this?.iterator;
+    if (iterator == null || !iterator.moveNext()) {
+      return null;
+    }
+    return iterator.current;
+  }
+
+  T? firstWhereOrNull(bool Function(T item) test) {
```
denrase, line 13: Same for firstWhereOrNull.

denrase (changes requested):

packages/dart/lib/src/utils/iterable_utils.dart
buenaflor: makes sense, let's do it

denrase (approved):

PR #3697, opened 2026-05-12
fix(dart): Make sentryOnError synchronous in runZonedGuarded

### LINKED ISSUE #3541: runZonedGuarded error handler is async, causing errors to be silently swallowed
## Description

The `sentryOnError` handler in `SentryRunZonedGuarded.sentryRunZonedGuarded` is declared as `async`, but Dart's `runZonedGuarded` expects a synchronous `void Function(Object, StackTrace)` callback.

**File:** `lib/src/sentry_run_zoned_guarded.dart:19`

```dart
final sentryOnError = (exception, stackTrace) async {
  await _captureError(hub, options, exception, stackTrace);
  if (onError != null) {
    onError(exception, stackTrace);
  }
};
```

This causes two problems:

### 1. The `onError` callback runs inside an unawaited Future

Since `sentryOnError` is `async`, it returns a `Future<void>` which Dart silently coerces to `void`. The `runZonedGuarded` implementation never awaits this future. This means:

- `_captureError` may not complete before the process exits — events can be lost
- The user's `onError` callback executes inside the unawaited future rather than synchronously

### 2. Rethrowing in `onError` creates an infinite loop that Dart silently drops

If the user's `onError` callback rethrows the error (a common pattern to preserve normal crash behavior):

```dart
Sentry.runZonedGuarded(() async {
  await Sentry.init(...);
}, (error, stackTrace) {
  throw error; // intended to let the error propagate
});
```

The `throw error` happens inside the unawaited `async` future, which makes it a new uncaught error in the same zone — triggering `sentryOnError` again in an infinite loop. Dart breaks this loop by silently dropping the error.

**Net effect:** The unhandled exception is completely swallowed — no stack trace is printed, and the Sentry event may not be sent either.

### Expected behavior

Errors should propagate normally after being captured by Sentry. The user's `onError` callback should be able to rethrow errors to preserve the default unhandled exception output.

### Reproduction

```dart
Future<void> main() async {
  Sentry.runZonedGuarded(() async {
    await Sentry.init(
      (options) {
        options.dsn = '...';
      },
      callAppRunnerInRunZonedGuarded: false,
      appRunner: () async {
        throw Exception('some test');
      },
    );
  }, (error, stackTrace) {
    throw error;
  });
}
```

**Without Sentry:** prints `Unhandled exception: Exception: some test` with full stack trace.
**With Sentry:** silently exits with no output.

### Suggested fix

Make `sentryOnError` synchronous and call `_captureError` without awaiting (or use `unawaited`), then invoke the user's `onError` synchronously:

```dart
final sentryOnError = (exception, stackTrace) {
  unawaited(_captureError(hub, options, exception, stackTrace));
  if (onError != null) {
    onError(exception, stackTrace);
  }
};
```

This preserves the synchronous contract expected by `runZonedGuarded` and allows the user's `onError` to rethrow normally.

theprantadutta:
## :scroll: Description

`SentryRunZonedGuarded.sentryRunZonedGuarded` wraps the user's `onError` in `sentryOnError`, which is passed to Dart's `runZonedGuarded`. The wrapper was declared `async`, so it returned a `Future<void>` that the framework silently coerced to `void` and discarded.

Two consequences:

1. `_captureError(...)` was awaited inside an unawaited future, which meant the user-supplied `onError` only ran after the awaited microtask resolved (rather than synchronously during error dispatch).
2. If the user's `onError` rethrew the error — a common pattern to preserve normal crash output — the throw became an uncaught async error in the same zone and recursively re-entered `sentryOnError` until Dart silently dropped it. The unhandled exception was swallowed entirely (no stack trace printed, no Sentry event guaranteed to have been sent).

This PR makes the handler synchronous, fires `_captureError` via `unawaited`, and calls the user's `onError` synchronously so a rethrow propagates out through `runZonedGuarded` instead of looping back into itself.

```dart
// Before
final sentryOnError = (exception, stackTrace) async {
  await _captureError(hub, options, exception, stackTrace);
  if (onError != null) onError(exception, stackTrace);
};

// After
final sentryOnError = (Object exception, StackTrace stackTrace) {
  unawaited(_captureError(hub, options, exception, stackTrace));
  if (onError != null) onError(exception, stackTrace);
};
```

The synchronous prefix of `_captureError` — the `options.log` call, span-status update via `hub.configureScope`, and the call into `hub.captureEvent` — all still run before the first `await`, so the existing tests that observe scope and event state continue to pass.

Closes #3541

- Added a regression test (`invokes user onError synchronously and captures the event (regression for #3541)`) that observes the user's `onError` synchronously from the caller frame. With the broken `async` handler this snapshot was `false` because the user callback only ran after `await _captureError(...)`; with the fix it is `true`.
- Verified the regression test **fails** when the source fix is reverted (`Expected: true / Actual: <false>`), then passes once restored.
- The existing 4 tests (`calls onError`, `calls zoneSpecification print`, `marks transaction as internal error if no status`, `sets level to error instead of fatal`) all continue to pass — the synchronous prefix of `_captureError` still sets the scope/span state before the existing tests observe it.
- `dart analyze --fatal-warnings` on `packages/dart/` — clean.
- Ran `hub_test.dart` alongside to check for nearby regressions — all 72 tests pass.

## :pencil: Checklist

- [ ] I updated the docs if needed (no doc changes — internal behaviour)
- [x] All tests passing
- [x] No breaking changes (public API unchanged; the handler's external contract was already `void Function(Object, StackTrace)`, the async return was being discarded)

### DIFF
--- a/packages/dart/test/sentry_run_zoned_guarded_test.dart
+++ b/packages/dart/test/sentry_run_zoned_guarded_test.dart
@@ -90,6 +90,57 @@ void main() {
       await span?.finish();
     });
 
+    test(
+        'invokes user onError synchronously and captures the event '
+        '(regression for #3541)', () async {
+      // Before the fix `sentryOnError` was declared `async`, returning a
+      // `Future<void>` that `runZonedGuarded` discarded. The user's
+      // `onError` only ran after `await _captureError(...)` resolved on
+      // a later microtask, which meant a rethrow from `onError` became
+      // an uncaught async error of the same zone — recursively
+      // re-entering `sentryOnError` until Dart silently dropped it.
+      //
+      // After the fix `sentryOnError` is synchronous: `_captureError` is
+      // fire-and-forget via `unawaited`, and the user's `onError` runs
+      // (and may rethrow cleanly) before `sentryRunZonedGuarded`
+      // returns. The clearest sync-vs-async discriminator is whether
+      // the user's onError has been invoked by the time control returns
+      // to the caller.
+      final client = MockSentryClient();
+      final hub = Hub(fixture.options);
+      hub.bindClient(client);
+
+      var userOnErrorCalled = false;
+      var userOnErrorCalledSyncFromCaller = false;
+
+      SentryRunZonedGuarded.sentryRunZonedGuarded(
+        hub,
+        () => throw StateError('boom'),
+        (error, stackTrace) {
+          userOnErrorCalled = true;
+        },
+      );
+      // Synchronous snapshot: with the sync handler this must already
+      // be `true`; with the broken async handler it would still be
+      // `false` because `await _captureError(...)` had not yet
+      // resolved.
+      userOnErrorCalledSyncFromCaller = userOnErrorCalled;
+
+      expect(userOnErrorCalledSyncFromCaller, isTrue,
+          reason: "sentryOnError must invoke the user's onError synchronously, "
+              "not after an awaited microtask. Otherwise a rethrow from "
+              "onError becomes an unhandled async error of the same zone "
+              "and recursively re-enters sentryOnError until Dart drops "
+              "the error.");
+
+      // The unawaited `_captureError` still reaches the client; we just
+      // observe it after microtasks drain.
+      await Future<void>.delayed(const Duration(milliseconds: 50));
+      expect(client.captureEventCalls, hasLength(1),
+          reason: 'The Sentry event must still be captured even though '
+              '`_captureError` is now fire-and-forget.');
+    });
+
     test('sets level to error instead of fatal', () async {
       final client = MockSentryClient();
       final hub = Hub(fixture.options);
--- a/packages/dart/test/sentry_run_zoned_guarded_test.dart
+++ b/packages/dart/test/sentry_run_zoned_guarded_test.dart
@@ -90,28 +90,13 @@ void main() {
       await span?.finish();
     });
 
-    test(
-        'invokes user onError synchronously and captures the event '
-        '(regression for #3541)', () async {
-      // Before the fix `sentryOnError` was declared `async`, returning a
-      // `Future<void>` that `runZonedGuarded` discarded. The user's
-      // `onError` only ran after `await _captureError(...)` resolved on
-      // a later microtask, which meant a rethrow from `onError` became
-      // an uncaught async error of the same zone — recursively
-      // re-entering `sentryOnError` until Dart silently dropped it.
-      //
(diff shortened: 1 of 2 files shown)


packages/dart/test/sentry_run_zoned_guarded_test.dart
```diff
@@ -90,6 +90,57 @@ void main() {
       await span?.finish();
     });
 
+    test(
```
buenaflor: you can add the reference to the issue regression as a comment instead

packages/dart/test/sentry_run_zoned_guarded_test.dart
```diff
@@ -90,6 +90,57 @@ void main() {
...
+      // a later microtask, which meant a rethrow from `onError` became
+      // an uncaught async error of the same zone — recursively
+      // re-entering `sentryOnError` until Dart silently dropped it.
+      //
+      // After the fix `sentryOnError` is synchronous: `_captureError` is
+      // fire-and-forget via `unawaited`, and the user's `onError` runs
+      // (and may rethrow cleanly) before `sentryRunZonedGuarded`
+      // returns. The clearest sync-vs-async discriminator is whether
+      // the user's onError has been invoked by the time control returns
+      // to the caller.
```
buenaflor: don't think all this comment is needed

packages/dart/test/sentry_run_zoned_guarded_test.dart
```diff
@@ -90,6 +90,57 @@ void main() {
...
+
+      expect(userOnErrorCalledSyncFromCaller, isTrue,
+          reason: "sentryOnError must invoke the user's onError synchronously, "
+              "not after an awaited microtask. Otherwise a rethrow from "
+              "onError becomes an unhandled async error of the same zone "
+              "and recursively re-enters sentryOnError until Dart drops "
+              "the error.");
+
+      // The unawaited `_captureError` still reaches the client; we just
+      // observe it after microtasks drain.
```
buenaflor: same here, we can remove these comments, they're quite verbose imo

theprantadutta:
Thanks for the review @buenaflor! Addressed all three notes in `f2f0e7b6`:

- moved the "regression for #3541" out of the test name into a leading code comment
- dropped the long bug-history paragraph at the top of the test
- removed the "we just observe it after microtasks drain" comment
- also trimmed the `reason:` text and the redundant inline variable comment

All 5 tests still pass locally.

buenaflor (approved):

PR #3698, opened 2026-05-12
fix(dart): Route SDK diagnostic logs to browser console on web

### LINKED ISSUE #3043: [Flutter Web] Debug logs do not show up in the browser dev console
On web we could forward log calls to `print` instead of `developer.log`

theprantadutta:
## :scroll: Description

Splits the default output for `SentryInternalLogger` into platform-specific implementations selected via the existing `_io_*` / `_web_*` companion-file pattern (already used by `_io_platform.dart` / `_web_platform.dart`, `origin_io.dart` / `origin_web.dart`, `dart_exception_type_identifier_io.dart` / `_web.dart`, etc.):

- `_io_default_log_output.dart` keeps the existing `dart:developer.log` behaviour on the VM, so DevTools and IDE log views are unchanged.
- `_web_default_log_output.dart` forwards to `window.console.*` via `package:web`, mapping `SentryLevel` to the matching console method:

  | `SentryLevel`     | `console.*`     |
  | ----------------- | --------------- |
  | `fatal` / `error` | `console.error` |
  | `warning`         | `console.warn`  |
  | `info`            | `console.info`  |
  | `debug`           | `console.debug` |

`dart:developer.log` does not surface in the browser dev console on Flutter Web, so SDK diagnostic messages (the ones gated by `options.debug = true`) were silently dropped during web development.

I deliberately used `package:web` `console.*` instead of `print`, to address @buenaflor's concern from [the issue thread](https://github.com/getsentry/sentry-dart/issues/3043#issuecomment-4423246375):

> "Something we need to be careful of is that we have an integration that reports prints as breadcrumbs, I want to make sure we don't spill out internal prints as breadcrumbs."

Sentry's print-breadcrumb integration installs a `ZoneSpecification(print:)` hook inside `runZonedGuarded` and records every `print` call as a `Breadcrumb`. If the SDK's own internal diagnostic logger went through `print`, every internal line would feed back as a user-visible breadcrumb on captured events. Calling `console.*` directly bypasses that hook entirely.

Additionally, `RuntimeChecker.kDebugMode` already short-circuits the logger at the top of `_log()`, so the whole code path is tree-shaken from release builds regardless of platform — providing a second layer of safety against the breadcrumb-feedback scenario even if a future integration changes how `print` is intercepted.

Closes #3043

- Added `packages/dart/test/utils/default_log_output_test.dart` with two tests for the IO path: (a) it never produces output via `print` (so the same property holds for the web path, which uses `console.*`); (b) it accepts all `SentryLevel`s without throwing.
- Added a routing test to `internal_logger_test.dart` that verifies the conditional import resolves to the IO file on the VM (no `print` output when logging through the public `SentryInternalLogger`).
- Verified the web target compiles cleanly: `dart compile js -o build/main.dart.js web/main.dart` in `packages/dart/example_web/` succeeds with the new conditional import in place.
- The existing 25 `internal_logger_test.dart` tests still pass.
- `dart analyze --fatal-warnings` on `packages/dart/` — clean.

The web `console.*` behaviour itself is not unit-tested on the VM (the web file can't be exercised there); the `example_web` compile path provides the static-typecheck guarantee, and a Flutter Web app with `options.debug = true` would surface the diagnostic logs at the appropriate severity in the dev console.

## :pencil: Checklist

- [ ] I updated the docs if needed (no doc changes — internal logger format)
- [x] All tests passing (existing + new)
- [x] No breaking changes (purely an additive platform split; VM behaviour unchanged)

### DIFF
--- a/packages/dart/lib/src/utils/_io_default_log_output.dart
+++ b/packages/dart/lib/src/utils/_io_default_log_output.dart
@@ -1,0 +1,22 @@
+import 'dart:developer' as dev;
+
+import '../protocol/sentry_level.dart';
+
+/// Default log output for non-web platforms. Uses `dart:developer.log` which
+/// integrates with Dart DevTools and the Flutter / IDE log views.
+void defaultLogOutput({
+  required String name,
+  required SentryLevel level,
+  required String message,
+  Object? error,
+  StackTrace? stackTrace,
+}) {
+  dev.log(
+    '[${level.name}] $message',
+    name: name,
+    level: level.toDartLogLevel(),
+    error: error,
+    stackTrace: stackTrace,
+    time: DateTime.now(),
+  );
+}
--- a/packages/dart/lib/src/utils/_web_default_log_output.dart
+++ b/packages/dart/lib/src/utils/_web_default_log_output.dart
@@ -1,0 +1,51 @@
+import 'dart:js_interop';
+
+import 'package:web/web.dart' as web;
+
+import '../protocol/sentry_level.dart';
+
+/// Default log output for web. `dart:developer.log` calls do not surface in
+/// the browser dev console on Flutter Web, so diagnostic SDK messages were
+/// otherwise silently dropped.
+///
+/// We deliberately forward to `window.console.*` (via `package:web`) instead
+/// of using Dart's top-level `print`. Sentry installs a print-zone hook in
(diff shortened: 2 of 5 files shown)


buenaflor (approved):
lgtm 👍

buenaflor:
I'll rerun some of the tests later, seems like CI is acting up

PR #3699, opened 2026-05-12
fix(flutter): Release replay JNI refs

### LINKED ISSUE #3633: fix: Release leaked JNI references in _AndroidReplayHandler
## Summary

`_AndroidReplayHandler` (the isolate-side replay screenshot handler) leaks JNI object references in three places.

### 1. `Bitmap$Config.ARGB_8888` leaks on every bitmap creation

https://github.com/getsentry/sentry-dart/blob/a7f011a3/packages/flutter/lib/src/native/java/android_replay_recorder.dart#L134

```dart
_bitmap ??= native.Bitmap.createBitmap$10(
    item.width, item.height, native.Bitmap$Config.ARGB_8888);
```

`Bitmap$Config.ARGB_8888` is a static getter that returns a **new JNI reference** each call (the generated binding comments say "must be released after use"). This reference is passed to `createBitmap$10` and then abandoned — it leaks every time a new bitmap is created (first capture, and on every dimension change).

### 2. `_bitmap` never released on isolate shutdown

https://github.com/getsentry/sentry-dart/blob/a7f011a3/packages/flutter/lib/src/native/java/android_replay_recorder.dart#L98

The cached `Bitmap` is released when dimensions change (line 126), but when the worker isolate shuts down the bitmap is abandoned without releasing.

### 3. `_nativeReplay` never released on isolate shutdown

https://github.com/getsentry/sentry-dart/blob/a7f011a3/packages/flutter/lib/src/native/java/android_replay_recorder.dart#L99-L103

The `ReplayIntegration` JObject obtained in the constructor is stored as a field but never released.

## Root cause

`WorkerHandler` has no `close()` lifecycle callback. When the isolate receives the shutdown command, it closes the `ReceivePort` without giving the handler a chance to release resources.

## Suggested fix

1. Add a `FutureOr<void> close()` method to `WorkerHandler` (default no-op)
2. Call `handler.close()` in `runWorker` before closing the inbox on shutdown
3. Override `close()` in `_AndroidReplayHandler` to release `_bitmap` and `_nativeReplay`
4. Release the `Bitmap$Config.ARGB_8888` reference after passing it to `createBitmap$10`

buenaflor:
## :scroll: Description
Release JNI references held by the Android replay screenshot path without changing the isolate worker shutdown behavior.

The replay worker now releases the temporary `Bitmap$Config.ARGB_8888` reference, the per-capture `Bitmap` reference, and the isolate-side replay integration handle after recording each screenshot. The native Java owner also releases replay integration handles before replacing or clearing them.

Fixes #3633.

The issue reports leaked JNI references in the Android replay path. This keeps the fix scoped to JNI handle ownership instead of adding a generic worker lifecycle hook.

- `fvm flutter analyze lib/src/native/java/android_replay_recorder.dart`
- `fvm flutter analyze lib/src/native/java/android_replay_recorder.dart lib/src/native/java/sentry_native_java.dart lib/src/native/java/sentry_native_java_init.dart`
- `fvm flutter test test/isolate/isolate_worker_test.dart`
- Pre-commit hooks ran formatter and `flutter analyze --fatal-warnings` for `sentry_flutter` successfully.

## :pencil: Checklist
- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps

### DIFF
--- a/packages/flutter/lib/src/native/java/android_replay_recorder.dart
+++ b/packages/flutter/lib/src/native/java/android_replay_recorder.dart
@@ -113,6 +113,7 @@ class _AndroidReplayHandler extends WorkerHandler {
   final WorkerConfig _config;
   // Android Bitmap creation is a bit costly so we reuse it between captures.
   native.Bitmap? _bitmap;
+  native.ReplayIntegration? _nativeReplay;
 
   _AndroidReplayHandler(this._config);
 
@@ -126,7 +127,7 @@ class _AndroidReplayHandler extends WorkerHandler {
   @override
   FutureOr<Object?> onRequest(Object? payload) {
     if (payload is _CloseRequest) {
-      _releaseBitmap();
+      _releaseNativeRefs();
       return null;
     }
 
@@ -140,7 +141,6 @@ class _AndroidReplayHandler extends WorkerHandler {
     final item = payload;
     JByteBuffer? jBuffer;
     native.Bitmap$Config? bitmapConfig;
-    native.ReplayIntegration? nativeReplay;
 
     try {
       var bitmap = _bitmap;
@@ -173,9 +173,9 @@ class _AndroidReplayHandler extends WorkerHandler {
       bitmap.copyPixelsFromBuffer(jBuffer);
 
       // TODO timestamp is currently missing in onScreenshotRecorded()
-      nativeReplay =
+      _nativeReplay ??=
           native.SentryFlutterPlugin.privateSentryGetReplayIntegration();
-      nativeReplay?.onScreenshotRecorded(bitmap);
+      _nativeReplay?.onScreenshotRecorded(bitmap);
 
       return null;
     } catch (exception, stackTrace) {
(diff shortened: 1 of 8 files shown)


packages/flutter/lib/src/native/java/android_replay_recorder.dart
```diff
@@ -94,62 +97,70 @@ class AndroidReplayRecorder extends ScheduledScreenshotRecorder {
...
-          .warning('${_config.debugName}: Unexpected payload type: $payload');
+      internalLogger.warning(
+        '${_config.debugName}: Unexpected payload type: $payload',
+      );
       return null;
     }
 
     final item = payload;
     JByteBuffer? jBuffer;
+    native.Bitmap? bitmap;
```
denrase: `Android Bitmap creation is a bit costly so we reuse it between captures.` Is this not true/relevant anymore?

denrase (approved):

packages/flutter/lib/src/native/java/android_replay_recorder.dart
buenaflor: hm no youre right, we should keep the existing behaviour, I'll double check

PR #3700, opened 2026-05-13
enh(dart): Add span v2 envelope `ingest_settings` metadata

### LINKED ISSUE #3690: Span V2 envelope container should have `ingest_settings`
### Description

https://develop.sentry.dev/sdk/telemetry/spans/span-protocol/#span-v2-envelope-item-payload

buenaflor:
## :scroll: Description
Add the span v2 envelope item container metadata required by Relay for pre-encoded span batches.

Span v2 envelope item payloads now include `version: 2` and `ingest_settings` with automatic IP and user-agent inference.

See: https://develop.sentry.dev/sdk/telemetry/spans/span-protocol/#span-v2-envelope-item-payload

Fixes #3690.

The span v2 envelope item payload contract requires top-level `version` and `ingest_settings` fields in addition to the `items` array. Without these fields, span v2 batches do not fully match the expected ingest payload shape.

- `fvm dart format lib/src/sentry_envelope.dart test/sentry_envelope_test.dart`
- `fvm dart analyze lib/src/sentry_envelope.dart test/sentry_envelope_test.dart`
- `fvm dart test test/sentry_envelope_test.dart`
- Pre-commit hooks ran staged Dart formatting and package analyze successfully.

## :pencil: Checklist
- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps


packages/dart/lib/src/sentry_envelope.dart
```diff
@@ -177,17 +197,43 @@ class SentryEnvelope {
     }
   }
 
-  /// Builds a payload in the format {"items": [item1, item2, ...]}
-  static Uint8List _buildItemsPayload(List<List<int>> encodedItems) {
+  // Pre-encoded JSON tokens, shared across calls. Safe to reuse with
```
denrase: What is the benefit of this? If there is no measurabel perfomance benefit, i'd rather have we do this inline like before, as it is more readable.

packages/dart/lib/src/sentry_envelope.dart
```diff
@@ -177,17 +197,43 @@ class SentryEnvelope {
...
+  // `BytesBuilder(copy: false)` because they are never mutated.
+  static final _openBrace = utf8.encode('{');
+  static final _comma = utf8.encode(',');
+  static final _colon = utf8.encode(':');
+  static final _itemsKey = utf8.encode('"items":[');
```
denrase: `_itemsKey` and `_closeItems` are not symetrical. We should split this differently, maybe in itemsOpen `'"items":[` and itemsClose `]`.

denrase (approved):
Looks good, just some comments about the shared encoded items.

packages/dart/lib/src/sentry_envelope.dart
buenaflor: I benchmarked this and just caching one `,` (comma) speeds up the payload encoding by about ~1.3–1.6×

buenaflor: It's a hot path since it's called extremely frequently for all telemetry so imo worth it to sacrifice readability here

PR #3701, opened 2026-05-13
docs(repo): Add branch naming guidance

buenaflor:
Document the preferred branch naming format for contributors.

This adds a short `git-username/type/short-description` convention near the existing contributor workflow docs so branch names stay consistent without adding extra process.

This makes it easy to filter branches by username or type

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -103,7 +103,7 @@ Only packages with staged changes are analyzed — no need to wait for the full 
 Use the format `git-username/type/short-description` for branch names.
 
 * `git-username` should match your Git username or GitHub handle.
-* `type` should be a conventional change type such as `feat`, `fix`, `docs`, `test`, or `chore`.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -98,6 +98,16 @@ automatically:
 
 Only packages with staged changes are analyzed — no need to wait for the full monorepo.
 
+## Branch Naming
+
+Use the format `git-username/type/short-description` for branch names.
+
+* `git-username` should match your Git username or GitHub handle.
+* `type` should be a conventional change type such as `feat`, `fix`, `docs`, `test`, or `chore`.
```
denrase: Is this the whole set of change types that are allowed? If already described elsewhere, we could link to it.

denrase (approved):

CONTRIBUTING.md
buenaflor: we don't have a fixed list, generally whatever fits in https://www.conventionalcommits.org/

buenaflor: maybe I'll just link this

PR #3708, opened 2026-05-19
perf(flutter): Optimize Android scope sync

### LINKED ISSUE #3709: Large Android scope payloads can cause JNI overhead
Large arbitrary maps in Flutter scope data, such as breadcrumb data, contexts, or user data, can be expensive on Android when each nested value is converted into Java objects through JNI.

This can add avoidable overhead on lower-end devices. Scope payloads that are arbitrary or potentially large should cross the JNI boundary as a coarse JSON byte payload and be deserialized on the Kotlin/Java side.

buenaflor:
## :scroll: Description

Optimize Android scope sync by sending breadcrumb, user, and context payloads over JNI as UTF-8 JSON bytes instead of recursively constructing Java `Map`/`List` objects from Dart.

This keeps primitive JNI calls direct, preserves structured nested user data through native deserialization, and reduces per-entry JNI object churn for large user-controlled scope payloads.

Large arbitrary maps in breadcrumbs, contexts, or user data can be expensive on lower-end Android devices when each nested value crosses JNI as an object. Passing the serialized payload once and deserializing on the Kotlin/Java side keeps the bridge coarse-grained and aligns with the existing byte-based native context/debug-image paths.

This lead to ANRs or even crashes in lower end devices.

Fixes #3709

- `fvm flutter analyze lib/src/native/utils/utf8_json.dart lib/src/native/java/sentry_native_java.dart example/integration_test/native_jni_utils_test.dart`
- `JAVA_HOME="$(/usr/libexec/java_home -v 17)" PATH="$(/usr/libexec/java_home -v 17)/bin:$PATH" ./gradlew :sentry_flutter:compileReleaseKotlin`
- `fvm flutter test test/native/sentry_native_java_test.dart`
- `fvm flutter test -d emulator-5554 integration_test/native_jni_utils_test.dart`
- `fvm flutter test -d emulator-5554 integration_test/integration_test.dart --plain-name "syncs large scope maps to native on Android"`
- `fvm flutter test -d emulator-5554 integration_test/integration_test.dart --plain-name "setUser syncs to native"`
- Full `fvm flutter test -d emulator-5554 integration_test` was also run locally; the token-gated `e2e captureException` test failed because `SENTRY_AUTH_TOKEN_E2E` was unset.

## :pencil: Checklist

- [ ] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps

### DIFF
--- a/packages/flutter/android/src/main/kotlin/io/sentry/flutter/SentryFlutterPlugin.kt
+++ b/packages/flutter/android/src/main/kotlin/io/sentry/flutter/SentryFlutterPlugin.kt
@@ -15,6 +15,7 @@ import io.flutter.plugin.common.MethodChannel.MethodCallHandler
 import io.flutter.plugin.common.MethodChannel.Result
 import io.sentry.Breadcrumb
 import io.sentry.DateUtils
+import io.sentry.JsonObjectReader
 import io.sentry.ScopesAdapter
 import io.sentry.Sentry
 import io.sentry.SentryOptions
@@ -31,8 +32,11 @@ import io.sentry.protocol.DebugImage
 import io.sentry.protocol.SdkVersion
 import io.sentry.protocol.User
 import io.sentry.transport.CurrentDateProvider
-import org.json.JSONObject
 import org.json.JSONArray
+import org.json.JSONObject
+import org.json.JSONTokener
+import java.io.ByteArrayInputStream
+import java.io.InputStreamReader
 import java.lang.ref.WeakReference
 import java.net.Proxy.Type
 import kotlin.math.roundToInt
@@ -169,6 +173,57 @@ class SentryFlutterPlugin :
     ) {
       Sentry.configureScope { scope ->
         scope.setContexts(key, value)
+      }
+    }
+
+    @Suppress("unused", "TooGenericExceptionCaught") // Used by native/jni bindings
+    @JvmStatic
+    fun addBreadcrumbFromJsonBytes(bytes: ByteArray) {
+      try {
+        val options = ScopesAdapter.getInstance().options
+        val breadcrumb =
+          jsonObjectReader(bytes).use { reader ->
+            Breadcrumb.Deserializer().deserialize(reader, options.logger)
+          }
+        Sentry.addBreadcrumb(breadcrumb)
+      } catch (e: Exception) {
+        Log.e("Sentry", "Failed to add breadcrumb from JSON bytes", e)
+      }
+    }
+
+    @Suppress("unused", "TooGenericExceptionCaught") // Used by native/jni bindings
+    @JvmStatic
+    fun setUserFromJsonBytes(bytes: ByteArray?) {
+      try {
+        if (bytes == null) {
+          Sentry.setUser(null)
+          return
+        }
+
+        val options = ScopesAdapter.getInstance().options
+        val user =
+          jsonObjectReader(bytes).use { reader ->
+            User.Deserializer().deserialize(reader, options.logger)
+          }
+        Sentry.setUser(user)
+      } catch (e: Exception) {
+        Log.e("Sentry", "Failed to set user from JSON bytes", e)
+      }
+    }
+
+    @Suppress("unused", "TooGenericExceptionCaught") // Used by native/jni bindings
+    @JvmStatic
+    fun setContextFromJsonBytes(
+      key: String,
+      bytes: ByteArray,
+    ) {
+      try {
+        val value = parseJsonBytes(bytes)
+        Sentry.configureScope { scope ->
+          scope.setContexts(key, value)
+        }
+      } catch (e: Exception) {
+        Log.e("Sentry", "Failed to set context from JSON bytes", e)
       }
     }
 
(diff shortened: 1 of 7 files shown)


packages/flutter/lib/src/native/java/sentry_native_java.dart
```diff
@@ -255,9 +234,9 @@ class SentryNativeJava extends SentryNativeChannel {
   void setContexts(String key, value) => tryCatchSync('setContexts', () {
         using((arena) {
           final jKey = key.toJString()..releasedBy(arena);
-          final jVal = dartToJObject(value)..releasedBy(arena);
+          final jBytes = jsonToJByteArray(value)..releasedBy(arena);
 
-          native.SentryFlutterPlugin.setContext(jKey, jVal);
+          native.SentryFlutterPlugin.setContextFromJsonBytes(jKey, jBytes);
         });
       });
```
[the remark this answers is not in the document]
buenaflor: roundtrip test is done in integration_test

packages/flutter/android/src/main/kotlin/io/sentry/flutter/SentryFlutterPlugin.kt
```diff
@@ -172,6 +176,57 @@ class SentryFlutterPlugin :
...
     }
 
+    @Suppress("unused", "TooGenericExceptionCaught") // Used by native/jni bindings
+    @JvmStatic
+    fun addBreadcrumbFromJsonBytes(bytes: ByteArray) {
+      try {
+        val options = ScopesAdapter.getInstance().options
+        val breadcrumb =
+          jsonObjectReader(bytes).use { reader ->
+            Breadcrumb.Deserializer().deserialize(reader, options.logger)
```
denrase: We always create `Breadcrumb.Deserializer()` instances for every individual breadcrumb. Does it make sense to lazy init the deserializer once?

packages/flutter/android/src/main/kotlin/io/sentry/flutter/SentryFlutterPlugin.kt
```diff
@@ -172,6 +176,57 @@ class SentryFlutterPlugin :
...
+      try {
+        if (bytes == null) {
+          Sentry.setUser(null)
+          return
+        }
+
+        val options = ScopesAdapter.getInstance().options
+        val user =
+          jsonObjectReader(bytes).use { reader ->
+            User.Deserializer().deserialize(reader, options.logger)
```
denrase: Same as above for `User.Deserializer()` instances.

packages/flutter/android/src/main/kotlin/io/sentry/flutter/SentryFlutterPlugin.kt
```diff
@@ -395,6 +450,45 @@ class SentryFlutterPlugin :
...
+      val json = String(bytes, Charsets.UTF_8)
+      return JSONTokener(json).nextValue().toKotlinJsonValue()
+    }
+
+    private fun jsonObjectReader(bytes: ByteArray): JsonObjectReader =
+      JsonObjectReader(
+        InputStreamReader(ByteArrayInputStream(bytes), Charsets.UTF_8),
+      )
+
+    private fun Any?.toKotlinJsonValue(): Any? =
```
denrase: We could turn this into a method taking a parameter instead of an extension on `Any?`, which is very broad.

packages/flutter/lib/src/native/utils/utf8_json.dart
```diff
@@ -3,6 +3,12 @@ import 'dart:typed_data';
 
 import 'package:meta/meta.dart';
 
+@internal
+Uint8List encodeUtf8Json(Object? data) {
+  final jsonString = json.encode(data);
```
denrase, line 8: The current path now is:

`Object -> String -> Bytes -> String -> Object`

What if we'd pass the payloads as String instead?

`Object -> String -> Object`

Might as well just skip the byte conversion.

denrase (changes requested):

packages/flutter/lib/src/native/utils/utf8_json.dart
buenaflor: we could, I was mostly following existing conventions which is using bytes as handles, I'll check if there are cons to using string directly

packages/flutter/android/src/main/kotlin/io/sentry/flutter/SentryFlutterPlugin.kt
buenaflor: Ah makes sense, I thought since the Deserializer is a static class we'd not allocate new objects every time but I think we are since we're using a constructor

packages/flutter/lib/src/native/utils/utf8_json.dart
buenaflor: the main benefit is we can use Breadcrumb/User JSON readers read directly from bytes/streams

buenaflor: and also afaik Strings in Java internally are set as UTF-16 so we'd be better of with utf8 encoding here memory wise (although it's a micro optimization)

packages/flutter/android/src/main/kotlin/io/sentry/flutter/SentryFlutterPlugin.kt
```diff
@@ -172,6 +184,57 @@ class SentryFlutterPlugin :
...
+    @JvmStatic
+    fun setContextFromJsonBytes(
+      key: String,
+      bytes: ByteArray,
+    ) {
+      try {
+        val value = parseJsonBytes(bytes)
+        Sentry.configureScope { scope ->
+          scope.setContexts(key, value)
+        }
```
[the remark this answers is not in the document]
buenaflor: not an issue

packages/flutter/lib/src/native/java/sentry_native_java.dart
```diff
@@ -423,6 +407,10 @@ JObject dartToJObject(Object? value) => switch (value) {
       _ => value.toString().toJString()
     };
 
+@visibleForTesting
+JByteArray jsonToJByteArray(Object? value) =>
+    JByteArray.from(encodeUtf8Json(normalize(value)));
```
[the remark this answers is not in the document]
buenaflor: that's fine, we shouldn't support those anyway

packages/flutter/android/src/main/kotlin/io/sentry/flutter/SentryFlutterPlugin.kt
```diff
@@ -395,6 +457,16 @@ class SentryFlutterPlugin :
         "debug_file" to debugFile,
       )
 
+    private fun parseJsonBytes(bytes: ByteArray): Any? =
+      jsonObjectReader(bytes).use { reader ->
+        reader.nextObjectOrNull()
+      }
```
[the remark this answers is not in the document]
buenaflor: wrong, the reader supports reading primitives as well

packages/flutter/android/src/main/kotlin/io/sentry/flutter/SentryFlutterPlugin.kt
```diff
@@ -395,6 +447,37 @@ class SentryFlutterPlugin :
...
       )
 
+    private fun setContextValue(
+      scope: IScope,
+      key: String,
+      value: Any?,
+    ) {
+      // Force sentry-java's typed overloads so primitive contexts are wrapped
+      // as {"value": ...} instead of being stored as invalid raw values.
+      when (value) {
```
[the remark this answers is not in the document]
buenaflor: not an issue

denrase (approved):

PR #3710, opened 2026-05-19
ci(flutter): Align CI with stable SwiftPM defaults

buenaflor:
Align Flutter CI with stable Flutter enabling Swift Package Manager by default for iOS and macOS builds. The dedicated main-channel SPM job is removed, while the regular build and native integration workflows now carry the SwiftPM setup needed on macOS runners.

**SwiftPM CI Setup**

The iOS and macOS CI paths now select the expected Xcode version, clear SwiftPM and Xcode package caches around package resolution, and rename the local Flutter package for macOS so SwiftPM sees the `sentry_flutter` package identity.

**Cocoa Integration Tests**

The cocoa integration workflow no longer runs an explicit `pod install` before Flutter tests. Stable Flutter resolves this path through SwiftPM, and macOS tests run from the renamed `packages/sentry_flutter/example` directory.

**Related PR**

The newer Dart analyzer export fix was split out into #3711 so this PR stays scoped to SPM CI changes.

### DIFF
--- a/.github/workflows/flutter.yml
+++ b/.github/workflows/flutter.yml
@@ -98,10 +98,6 @@ jobs:
           TARGET=${{ matrix.target }}
           flutter pub get
 
-      - name: Clear SwiftPM cache
-        if: matrix.target == 'ios' || matrix.target == 'macos'
-        run: bash ./.github/scripts/clear-swiftpm-cache.sh
-
       - name: Build example for ${{ matrix.target }}
         # The example currently doesn't support compiling for WASM. Should be OK once we add package:web in v9.
         if: matrix.target != 'wasm'
--- a/.github/workflows/flutter.yml
+++ b/.github/workflows/flutter.yml
@@ -78,7 +78,6 @@ jobs:
       # Flutter derives the local SwiftPM package identity from the plugin path basename on macOS.
       # Our package lives in packages/flutter, which makes Xcode resolve it as "flutter" and fail with:
       # "unable to override package 'sentry_flutter' because its identity 'flutter' doesn't match override's identity (directory name) 'sentry_flutter'"
-      # See also: spm job below.
       - name: Rename sentry_flutter package for macOS SwiftPM
         if: matrix.target == 'macos'
         run: mv packages/flutter packages/sentry_flutter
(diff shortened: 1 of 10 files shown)


.github/workflows/flutter.yml
```diff
@@ -94,12 +93,29 @@ jobs:
...
+            clear_swiftpm_cache
+            flutter build ios --no-codesign
+          }
           ;;
           macos)
-          flutter build macos
+          flutter build macos || {
+            clear_swiftpm_cache
+            flutter build macos
+          }
```
buenaflor, line 118: if it fails, try to clear cache then retry

.github/workflows/flutter.yml
```diff
@@ -52,9 +52,21 @@ jobs:
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
 
+      # QuickFix for failing iOS 18.0 builds https://github.com/actions/runner-images/issues/12758#issuecomment-3187115656
```
denrase, line 55: We were also logging Flutter version in the SPM job. Might also be useful here.

.github/workflows/flutter.yml
```diff
@@ -115,49 +131,6 @@ jobs:
           ;;
           esac
 
-  spm:
```
denrase: Do we have a dedicated `cocoapods` test as well? Might make sense if we still support it.

packages/flutter/lib/src/event_processor/widget_event_processor.dart
denrase, line 6: Aren't these fixes already on main?

denrase (approved):
Just 2 things

packages/flutter/lib/src/event_processor/widget_event_processor.dart
buenaflor: yeah they should be, something messed up with merging

.github/workflows/flutter.yml
buenaflor: the flutter version from the setup flutter step also has the info about the version, we don't necessarily need to log it explicitly

packages/flutter/lib/src/event_processor/widget_event_processor.dart
buenaflor: okay everything was good from our side, first time this happens but this was a GitHub issue, it used a stale commit as base

.github/workflows/flutter.yml
buenaflor: added one to be safe

Issue #3348, opened 2025-11-18
Android apps crash when try add big json response in breadcrumbs


GalimzyanovDanil:
### Platform

Flutter Mobile Android

### Obfuscation

Enabled

### Debug Info

Disabled

### Doctor

[✓] Flutter (Channel stable, 3.35.7, on macOS 15.6.1 24G90 darwin-arm64, locale ru-RU) [246ms]
    • Flutter version 3.35.7 on channel stable at /Users/danil/development/flutter
    • Upstream repository https://github.com/flutter/flutter.git
    • Framework revision adc9010625 (4 weeks ago), 2025-10-21 14:16:03 -0400
    • Engine revision 035316565a
    • Dart version 3.9.2
    • DevTools version 2.48.0
    • Feature flags: enable-web, enable-linux-desktop, enable-macos-desktop, no-enable-windows-desktop, enable-android, enable-ios, cli-animations,
      enable-lldb-debugging

[✓] Android toolchain - develop for Android devices (Android SDK version 35.0.0) [887ms]
    • Android SDK at /Users/danil/Library/Android/sdk
    • Emulator version 35.2.10.0 (build_id 12414864) (CL:N/A)
    • Platform android-36, build-tools 35.0.0
    • ANDROID_HOME = /Users/danil/Library/Android/sdk
    • Java binary at: /Applications/Android Studio.app/Contents/jbr/Contents/Home/bin/java
      This is the JDK bundled with the latest Android Studio installation on this machine.
      To manually set the JDK path, use: `flutter config --jdk-dir="path/to/jdk"`.
    • Java version OpenJDK Runtime Environment (build 17.0.11+0-17.0.11b1207.24-11852314)
    • All Android licenses accepted.

[✓] Xcode - develop for iOS and macOS (Xcode 16.4) [1 200ms]
    • Xcode at /Applications/Xcode.app/Contents/Developer
    • Build 16F6
    • CocoaPods version 1.16.2

[✓] Chrome - develop for the web [63ms]
    • Chrome at /Applications/Google Chrome.app/Contents/MacOS/Google Chrome

[✓] Android Studio (version 2024.1) [63ms]
    • Android Studio at /Applications/Android Studio.app/Contents
    • Flutter plugin can be installed from:
      🔨 https://plugins.jetbrains.com/plugin/9212-flutter
    • Dart plugin can be installed from:
      🔨 https://plugins.jetbrains.com/plugin/6351-dart
[...]

Issue #2440, opened 2024-11-22
ANRs when fetching debug images on Android


buenaflor:
### Description

![Image](https://github.com/user-attachments/assets/de91aaa3-192f-4579-8d3f-c561ce650ba1)

Couldn't reproduce locally so far but prolly doesn't hurt to look into ways to improve the fetching or caching

Maybe https://github.com/getsentry/sentry-java/issues/3925 can help us mitigate this problem as we send much less data over the method channels

```[tasklist]
### Tasks
- [ ] https://github.com/getsentry/sentry-java/issues/3925
- [ ] https://github.com/getsentry/sentry-dart/pull/2706
```

PR #3713, opened 2026-05-19
perf(flutter): Move Android JNI work to core worker to avoid work on main isolate

### LINKED ISSUE #3536: ANR: loadDebugImages runs synchronously on main thread causing ANR on slow I/O devices
## Problem

`LoadNativeDebugImagesIntegration` causes ANR on Android devices with slow I/O (observed on Realme RMX3830, RMX3834, RMX3940 across Android 13-15).

The root cause is that `SentryNativeJava.loadDebugImages()` performs a **synchronous JNI call** to `SentryFlutterPlugin.loadDebugImagesAsBytes()` on the **main thread**. The native function `sentry_get_modules_list` reads `/proc/self/maps` via `syscall`, which blocks the UI thread for >5 seconds on affected devices.

## Stacktrace

```
at io.sentry.flutter.SentryFlutterPlugin.loadDebugImagesAsBytes
at io.sentry.flutter.SentryFlutterPlugin$Companion.loadDebugImagesAsBytes
at io.sentry.android.ndk.DebugImagesLoader.loadDebugImagesForAddresses
at io.sentry.android.ndk.DebugImagesLoader.loadDebugImages
at io.sentry.ndk.NativeModuleListLoader.loadModuleList
at io.sentry.ndk.NativeModuleListLoader.nativeLoadModuleList
at sentry_get_modules_list
at syscall
```

## Call chain

```
EventProcessor.apply() [main isolate]
  → SentryNativeJava.loadDebugImages() [synchronous, main isolate]
    → SentryFlutterPlugin.loadDebugImagesAsBytes() [JNI, main thread]
      → DebugImagesLoader.loadDebugImagesForAddresses() [Java, main thread]
        → NativeModuleListLoader.loadModuleList() [Java, main thread]
          → sentry_get_modules_list() [native, main thread]
            → read /proc/self/maps [syscall, blocking I/O]
```

## Environment

- `sentry_flutter`: 9.13.0
- Flutter: 3.35.0
- Devices: Realme RMX3830, RMX3834, RMX3940 (budget devices with slow I/O)
- OS: Android 13, 14, 15
- Mechanism: AppExitInfo

## Previous optimizations (already in 9.13.0)

- PR #2706 — load only relevant debug images by addresses (instead of all ~82)
- PR #3224 — refactor from Method Channel to JNI (~2x faster)

These optimizations helped, but `sentry_get_modules_list` → `syscall` still runs **synchronously on the main thread**, which is too slow on budget devices.

## Expected behavior

`loadDebugImages` should run on a **background thread/isolate** to avoid blocking the UI thread.

## Workaround

We currently remove `LoadNativeDebugImagesIntegration` from integrations list to prevent the ANR:

```dart
options.integrations.removeWhere(
  (i) => i.runtimeType.toString() == 'LoadNativeDebugImagesIntegration',
);
```

This sacrifices NDK stacktrace symbolication but eliminates the ANR.

buenaflor:
## :scroll: Description

Moves the Android envelope sender into a broader `AndroidCoreWorker` that can also handle native debug images, native contexts, breadcrumbs, users, and context updates. Value-returning APIs use worker request/response when available

Now running on background isolate (also confirmed these are safe to be called from non-main threads):
- loadContexts
- loadDebugImages
- setContexts
- removeContexts
- addBreadcrumb
- setUser

This keeps more Android JNI work off the main isolate when the worker is available, while preserving current-isolate fallbacks for calls made before worker startup. The worker serializes JNI work to keep native calls ordered.

Relevant #2440, https://github.com/getsentry/sentry-dart/issues/3348
Fixes https://github.com/getsentry/sentry-dart/issues/3536

- `../../.fvm/flutter_sdk/bin/flutter analyze lib/src/native/java/android_core_worker.dart lib/src/native/java/sentry_native_java.dart test/native/android_core_worker_test_real.dart test/native/sentry_native_java_test_real.dart`
- `../../.fvm/flutter_sdk/bin/flutter test test/native/android_core_worker_test_real.dart test/native/sentry_native_java_test_real.dart`
- Commit hook ran `melos exec -- dart analyze --fatal-warnings` and `melos exec -- flutter analyze --fatal-warnings`

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps

None.


packages/flutter/lib/src/native/java/android_core_worker.dart
```diff
@@ -0,0 +1,462 @@
...
+  }
+
+  FutureOr<List<DebugImage>?> loadDebugImages(SentryStackTrace stackTrace) {
+    final instructionAddresses =
+        stackTrace.frames.map((f) => f.instructionAddr).nonNulls.toList(
+              growable: false,
+            );
+
+    final client = _worker;
+    if (client == null) {
```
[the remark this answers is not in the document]
buenaflor: no, we dont want to block those if the worker is closed

packages/flutter/lib/src/native/java/android_core_worker.dart
```diff
@@ -0,0 +1,546 @@
...
+    } catch (exception, stackTrace) {
+      internalLogger.error(
+        'Android core worker failed to $operation',
+        error: exception,
+        stackTrace: stackTrace,
+      );
+      if (_config.automatedTestMode) {
+        rethrow;
+      }
+    }
```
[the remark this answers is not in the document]
buenaflor: not an issue, the work still is offloaded to another isolate

packages/flutter/lib/src/native/java/android_core_worker.dart
```diff
@@ -0,0 +1,644 @@
...
+  @internal
+  static AndroidCoreWorker Function(SentryFlutterOptions) factory =
+      AndroidCoreWorker.new;
+
+  FutureOr<void> start() {
+    if (_isClosed) return null;
+    if (_worker != null) return null;
+    if (_startFuture != null) return _startFuture;
+    _startFuture = _start();
+    return _startFuture;
```
buenaflor, line 43: Note: `_startFuture` deduplicates concurrent start() calls so multiple callers share a single in-flight spawn rather than racing to create multiple workers.

packages/flutter/lib/src/native/java/android_core_worker.dart
```diff
@@ -0,0 +1,737 @@
...
+        error: exception,
+        stackTrace: stackTrace,
+      );
+    } finally {
+      _startFuture = null;
+    }
+  }
+
+  FutureOr<void> close() async {
+    _isClosed = true;
```
denrase, line 67: Do we need `_isClosed`? Since we are awaiting `_startFuture` anyway. Also if we early return in `_startFuture`, we can never call close on `_worker`, as it was never assigned.

denrase: I see we are guarding the other methods with it.

packages/flutter/lib/src/native/java/android_core_worker.dart
```diff
@@ -0,0 +1,737 @@
...
+      containsUnhandledException,
+    );
+  }
+
+  void _captureEnvelopeFromWorker(
+    Worker client,
+    Uint8List envelopeData,
+    bool containsUnhandledException,
+  ) {
+    client.send(
```
denrase, line 104: No special handling here, we could inline this, but seeing that all other methods follow the same pattern, we might as well leave this to be consistent. Your call.

packages/flutter/lib/src/native/java/android_core_worker.dart
```diff
@@ -0,0 +1,737 @@
...
+    }
+
+    return _loadDebugImagesFromWorker(client, instructionAddresses);
+  }
+
+  Future<List<DebugImage>?> _loadDebugImagesFromWorker(
+    Worker client,
+    List<String> instructionAddresses,
+  ) async {
+    try {
```
denrase, line 135: All async `*FromWorker` methods follow the same try/chatch/check testmode pattern. Does it make sense to extract this into a higher order function?

packages/flutter/lib/src/native/java/android_core_worker.dart
```diff
@@ -0,0 +1,737 @@
...
+    return _addBreadcrumbFromWorker(client, breadcrumb);
+  }
+
+  Future<void> _addBreadcrumbFromWorker(
+    Worker client,
+    Breadcrumb breadcrumb,
+  ) async {
+    try {
+      await client.request(
+        _AddBreadcrumbRequest(normalizeMap(breadcrumb.toJson())!),
```
denrase: We should not do force unwraps.

packages/flutter/lib/src/native/java/android_core_worker.dart
```diff
@@ -0,0 +1,737 @@
...
+      }
+    }
+  }
+
+  FutureOr<void> clearBreadcrumbs() {
+    if (_isClosed) return null;
+
+    final client = _worker;
+    if (client == null) {
+      _clearBreadcrumbs(automatedTestMode: _config.automatedTestMode);
```
denrase, line 229: `_config.automatedTestMode` is an instance var, no need to pass as parameter to the method, we don't do this in the other calls.

packages/flutter/lib/src/native/java/android_core_worker.dart
```diff
@@ -0,0 +1,737 @@
...
+      if (_config.automatedTestMode) {
+        rethrow;
+      }
+    }
+  }
+
+  FutureOr<void> setContexts(String key, dynamic value) {
+    if (_isClosed) return null;
+
+    final normalizedValue = normalize(value);
```
denrase, line 287: Should we do normalization in the worker if possible?

packages/flutter/lib/src/native/java/android_core_worker.dart
```diff
@@ -0,0 +1,737 @@
...
+      return null;
+    }
+
+    return _setUserFromWorker(client, user);
+  }
+
+  Future<void> _setUserFromWorker(Worker client, SentryUser? user) async {
+    try {
+      await client.request(
+        _SetUserRequest(user == null ? null : normalizeMap(user.toJson())),
```
denrase: If possible, we could move normalization and json mapping to the worker.

packages/flutter/lib/src/native/java/sentry_native_java.dart
```diff
@@ -198,56 +115,27 @@ class SentryNativeJava extends SentryNativeChannel {
   @override
   Future<void> close() async {
     await _replayRecorder?.stop();
-    await _envelopeSender?.close();
+    await _coreWorker?.close();
```
denrase, line 118: Since we are not setting the `_coreWorker` to null, could we make it non-optional?

denrase (changes requested):
We could try to move more work into the worker, if it makes sense. Otherwise looking good.

packages/flutter/lib/src/native/java/android_core_worker.dart
buenaflor: we normalize because we want to send it over, cross isolate doesn't accept arbitrary classes unfortunately

packages/flutter/lib/src/native/java/android_core_worker.dart
buenaflor: not possible unfortunately, cross isolate data transfer only works with normalized data

packages/flutter/lib/src/native/java/android_core_worker.dart
buenaflor: generally yes, but I want to make a separate PR for that, I want to create like a reusable pattern that should be used codebase wide (similar to the internalLogger)

packages/flutter/lib/src/native/java/android_core_worker.dart
buenaflor: we can't remove it, _clearBreadcrumbs etc... are top-level functions (static) so they can be resolved at compile time and therefore be easily used in background isolates - so we have to pass them here

packages/flutter/lib/src/native/java/sentry_native_java.dart
buenaflor: we are not re-assigning it right now but we should (in a separate PR) so it still works when closing and re-initializing the SDK

packages/flutter/lib/src/native/java/android_core_worker.dart
buenaflor: I think _isClosed is still needed. It prevents start() and fallback JNI calls after close(), and it lets _start() handle the race where close() happens while _spawn is in flight.

packages/flutter/lib/src/native/java/android_core_worker.dart
buenaflor: I think it still makes sense because the functions can return `FutureOr`

inlining `return () async {` doesnt look so nice

buenaflor:
failures are because of Github..

packages/flutter/lib/src/native/java/sentry_native_java.dart
```diff
@@ -58,98 +58,15 @@ class SentryNativeJava extends SentryNativeChannel {
...
-      if (options.automatedTestMode) {
-        rethrow;
-      }
-    } finally {
-      contextsUtf8JsonBytes?.release();
-    }
-
-    return null;
-  }
+  FutureOr<Map<String, dynamic>?> loadContexts() => _coreWorker?.loadContexts();
```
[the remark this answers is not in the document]
buenaflor: false positive

denrase (approved):

PR #3721, opened 2026-05-27
feat(grpc): Add integration support for GRPC

### LINKED ISSUE #1120: Support `gRPC` tracing/breadcrumbs/Client errors
### Description

[https://pub.dev/packages/grpc](<https://pub.dev/packages/grpc>)<br>We can add an interceptor [https://pub.dev/documentation/grpc/latest/service_api/ClientInterceptor-class.html](<https://pub.dev/documentation/grpc/latest/service_api/ClientInterceptor-class.html>)

lucas-zimerman:
## :scroll: Summary

- Attach structured gRPC error details (`ErrorInfo`, `BadRequest`, `RetryInfo`, `DebugInfo`, `PreconditionFailure`, `ResourceInfo`, `QuotaFailure`) to spans on failed calls
- Remove `captureBody` — no other Sentry SDK gRPC integration captures the request body
- Improve example apps: typed `DummyMessage` class (replaces raw `List<int>` hand-encoding), `WithHeaders` button demonstrating custom gRPC metadata, `SentryFlutter` init in grpc4 example

Brings the Dart gRPC interceptor closer to parity with the Python and JS integrations for request header propagation, while adding richer error detail capture that those SDKs don't have.

- [Ran](https://sentry-sdks.sentry.io/explore/discover/trace/fb29638b87904fbe93cecc5513a33a8c/?dataset=transactions&field=title&field=project&field=user.display&field=timestamp&field=replayId&id=21705&name=&node=span-27bf17b80aca49f0&project=5428562&query=&queryDataset=transaction-like&sort=-timestamp&source=discover&statsPeriod=1h&timestamp=1779850513&topEvents=5&yAxis=count%28%29) both grpc4 and grpc5 example apps against `grpcb.in`
- Verified spans appear in Sentry with `http.request.header.*` data when metadata is set
- Verified error details are attached on failed calls
- Pre-commit hooks (analyze + format) pass on all affected packages

## :pencil: Checklist
- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps

Close https://github.com/getsentry/sentry-dart/issues/1120


packages/grpc/CHANGELOG.md
```diff
@@ -0,0 +1,7 @@
+# Changelog
+
+## 9.20.0
+
+### Features
+
+- Add gRPC client instrumentation with distributed tracing, breadcrumbs, and error capture ([#TODO](https://github.com/getsentry/sentry-dart/pull/TODO))
```
buenaflor: no need for a changelog entry, we use automatic changelog generation (check out the contributing guide, it has some infos)

packages/grpc/test/mocks/mock_hub.dart
```diff
@@ -0,0 +1,72 @@
...
+  @override
+  // ignore: invalid_use_of_internal_member
+  SentryOptions get options => _options;
+
+  void reset() {
+    captureExceptionCalls = [];
+    addBreadcrumbCalls = [];
+    closeCalls = 0;
+    _isEnabled = true;
+  }
```
buenaflor, line 26: I think we already have a MockHub in the `dart` package, maybe it makes sense to move that mockhub from the dart package to the `_sentry_testing` package so we can use it easily across packages

packages/grpc/lib/src/sentry_grpc_interceptor.dart
```diff
@@ -0,0 +1,300 @@
...
+        span.setData('grpc.debug_info.detail', detail.detail);
+      } else if (detail is rpc.PreconditionFailure) {
+        span.setData(
+          'grpc.precondition_failure.violations',
+          detail.violations
+              .map((v) => '${v.type}: ${v.subject} - ${v.description}')
+              .join('; '),
+        );
+      } else if (detail is rpc.ResourceInfo) {
+        span.setData('grpc.resource_info.type', detail.resourceType);
```
buenaflor: Since we add a bunch of new attributes double check if OTEL has any equivalent attribute: https://opentelemetry.io/docs/specs/semconv/rpc/

if not, we should open a PR adding new ones in https://getsentry.github.io/sentry-conventions/

Let's double check also the [python](https://docs.sentry.io/platforms/python/integrations/grpc/) and [go](https://docs.sentry.io/platforms/go/guides/grpc/) instrumentation to see how they structure their attributes and span ops

also let's add them to `packages/dart/constants.dart` in `SemanticAttributesConstants`

buenaflor (commented):
Did a quick PR review for now (didn't look into the details yet)

packages/grpc/test/mocks/mock_hub.dart
lucas-zimerman: Is it ok if I do this task on a follow-up PR?

packages/grpc/CHANGELOG.md
lucas-zimerman: Fixed on https://github.com/getsentry/sentry-dart/pull/3721/commits/311bc7056f29d969b3750b5ae7609a88152ca601

packages/grpc/lib/src/sentry_grpc_interceptor.dart
```diff
@@ -0,0 +1,300 @@
...
+          await _addBreadcrumb(
+            method.path,
+            StatusCode.ok,
+            'OK',
+            stopwatch.elapsed,
+            SentryLevel.info,
+          );
+        }
+      },
+      onError: (Object error) async {
```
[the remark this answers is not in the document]
lucas-zimerman: Fixed on https://github.com/getsentry/sentry-dart/pull/3721/commits/311bc7056f29d969b3750b5ae7609a88152ca601

packages/grpc/test/sentry_grpc_interceptor_test.dart
```diff
@@ -0,0 +1,507 @@
...
+
+          expect(
+            fixture.service.lastReceivedMetadata?.containsKey('baggage'),
+            isTrue,
+          );
+        });
+
+        test('injects sentry-trace header even without an active transaction',
+            () async {
+          final client = fixture.getSut(mockHub: fixture.mockHub);
```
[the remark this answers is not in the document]
lucas-zimerman: Fixed on https://github.com/getsentry/sentry-dart/pull/3721/commits/311bc7056f29d969b3750b5ae7609a88152ca601

packages/grpc/test/mocks/mock_hub.dart
buenaflor: sounds good

packages/dart/lib/src/constants.dart
```diff
@@ -414,6 +414,61 @@ abstract class SemanticAttributesConstants {
...
+
+  /// The fully-qualified logical name of the method (`ServiceName/MethodName`).
+  ///
+  /// See https://opentelemetry.io/docs/specs/semconv/rpc/grpc/
+  static const rpcMethod = 'rpc.method';
+
+  /// The string representation of the gRPC status code (e.g., `'OK'`, `'NOT_FOUND'`).
+  ///
+  /// See https://opentelemetry.io/docs/specs/semconv/rpc/grpc/
+  static const rpcResponseStatusCode = 'rpc.response.status_code';
```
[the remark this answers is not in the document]
lucas-zimerman: `rpc.grpc.status_code` is not there, as odd as it sounds, they adopted to use a general key name from `rpc` instead of adding `grpc` to the key name

packages/dart/lib/src/constants.dart
```diff
@@ -414,6 +414,61 @@ abstract class SemanticAttributesConstants {
   /// The database name.
   // TODO: deprecated, needs to be replaced later by db.namespace
   static const dbName = 'db.name';
+
+  /// The RPC system identifier. For gRPC, always `'grpc'`.
+  ///
+  /// See https://opentelemetry.io/docs/specs/semconv/rpc/grpc/
+  static const rpcSystemName = 'rpc.system.name';
```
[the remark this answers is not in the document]
lucas-zimerman: `rpc.name` doesn't exist

packages/grpc/lib/src/sentry_grpc_interceptor.dart
lucas-zimerman: https://github.com/getsentry/sentry-conventions/issues/97 reference issue about the attributes

packages/grpc/lib/src/version.dart
```diff
@@ -0,0 +1,5 @@
+/// The SDK version reported to Sentry.io in the submitted events.
+const String sdkVersion = '9.20.0';
```
buenaflor: newest version is `9.21.0` now

packages/grpc/test/mocks/mock_hub.dart
buenaflor: pls create an issue as a follow up so we don't forget this (regarding moving this to the testing pkg)

packages/grpc/CHANGELOG.md
```diff
@@ -0,0 +1 @@
+# Changelog
```
buenaflor: this should be a symlink to the root changelog

packages/grpc/lib/src/sentry_grpc_interceptor.dart
```diff
@@ -0,0 +1,364 @@
...
+  /// Set [recordBreadcrumbs] to `false` to disable breadcrumb recording.
+  ///
+  /// Set [captureRequestHeaders] to `false` to suppress attaching outgoing
+  /// gRPC metadata (request headers) to the span. Sensitive headers
+  /// (`authorization`, `cookie`, etc.) are redacted unless
+  /// [SentryOptions.sendDefaultPii] is enabled. Defaults to `true`.
+  SentryGrpcInterceptor({
+    Hub? hub,
+    bool? captureFailedRequests,
+    bool recordBreadcrumbs = true,
```
buenaflor: let's call it `enableBreadcrumbs` instead of `recordBreadcrumbs` mainly so we can later add `enableLogs` and `enableMetrics` as well

packages/grpc/lib/src/sentry_grpc_interceptor.dart
```diff
@@ -0,0 +1,364 @@
...
+    bool captureRequestHeaders = true,
+  })  : _hub = hub ?? HubAdapter(),
+        _captureFailedRequests = captureFailedRequests,
+        _recordBreadcrumbs = recordBreadcrumbs,
+        _captureRequestHeaders = captureRequestHeaders {
+    _spanFactory = _hub.options.spanFactory;
+    if (_hub.options.isTracingEnabled()) {
+      _hub.options.sdk.addIntegration(integrationName);
+    }
```
buenaflor:
```suggestion
    _hub.options.sdk.addIntegration(integrationName);
```

let's add it regardless of if tracing is on or not, this is useful later to generally track adoption of this class, someone might use the integration with tracing off for example.

packages/grpc/lib/src/sentry_grpc_interceptor.dart
```diff
@@ -0,0 +1,364 @@
...
+
+    span?.origin = SentryTraceOrigins.autoGrpcClientInterceptor;
+
+    if (span != null) {
+      _attachRpcAttributes(span, method.path);
+      _attachRequestData(span, method, request, options);
+    }
+
+    final modifiedOptions = _buildModifiedOptions(options, span, method.path);
+    final stopwatch = _recordBreadcrumbs ? (Stopwatch()..start()) : null;
```
buenaflor, line 95: we're creating a new stopwatch instance every request, we should prolly store an instance of it and just start/stop/reset I think

minor issue though

packages/grpc/lib/src/sentry_grpc_interceptor.dart
```diff
@@ -0,0 +1,364 @@
...
+/// ```
+///
+/// To capture failed RPCs (overriding [SentryOptions.captureFailedRequests]):
+///
+/// ```dart
+/// final interceptor = SentryGrpcInterceptor(captureFailedRequests: true);
+/// ```
+class SentryGrpcInterceptor extends ClientInterceptor {
+  /// Integration name registered with the Sentry SDK.
+  static const String integrationName = 'GrpcClientTracing';
```
buenaflor: Maybe we should just call it `GrpcClient`? since it not only does tracing

buenaflor (commented):
nice, will do another review later

packages/grpc/CHANGELOG.md
lucas-zimerman: thank you! wasn't sure how to deal with the new changelog

packages/grpc/example/grpc5/lib/main.dart
```diff
@@ -0,0 +1,209 @@
...
+    tr.status = const SpanStatus.internalError();
+    await Sentry.captureException(e, stackTrace: s);
+    print('FAILED: $e');
+  } finally {
+    await tr.finish();
+  }
+}
+
+// Demonstrates request header capture: 'meat: vegetable' will appear
+// as http.request.header.meat in the span data.
```
buenaflor:
```suggestion
// as rpc.request.metadata.meat in the span data.
```

packages/grpc/example/grpc5/README.md
```diff
@@ -0,0 +1,36 @@
+# grpc5 — sentry_grpc Dart example
+
+Dart CLI example demonstrating `SentryGrpcInterceptor` with grpc 5.x.
+
+## What it shows
+
+- `Sentry.init` with `tracesSampleRate` and `captureFailedRequests`
+- `SentryGrpcInterceptor` attached to a `ClientChannel` targeting `grpcb.in:9001`
+- Request header capture in spans — the `WithHeaders` call passes `meat: vegetable` custom metadata, visible as `http.request.header.meat` in the span data
```
buenaflor:
```suggestion
- Request header capture in spans — the `WithHeaders` call passes `meat: vegetable` custom metadata, visible as `rpc.request.metadata.meat` in the span data
```

buenaflor (commented):
`LICENSE` seems to be missing as well for the gRPC package

packages/grpc/example/grpc5/README.md
lucas-zimerman: Fixed on https://github.com/getsentry/sentry-dart/pull/3721/commits/e9f643a39b94a1daae9b159991725a6b4e683599

packages/grpc/example/grpc5/lib/main.dart
lucas-zimerman: Fixed on https://github.com/getsentry/sentry-dart/pull/3721/commits/e9f643a39b94a1daae9b159991725a6b4e683599

packages/grpc/lib/src/sentry_grpc_interceptor.dart
lucas-zimerman: Fixed on https://github.com/getsentry/sentry-dart/pull/3721/commits/e9f643a39b94a1daae9b159991725a6b4e683599

packages/grpc/lib/src/sentry_grpc_interceptor.dart
lucas-zimerman: Fixed on https://github.com/getsentry/sentry-dart/pull/3721/commits/e9f643a39b94a1daae9b159991725a6b4e683599

packages/grpc/lib/src/version.dart
lucas-zimerman: Fixed on https://github.com/getsentry/sentry-dart/pull/3721/commits/e9f643a39b94a1daae9b159991725a6b4e683599

packages/grpc/lib/src/sentry_grpc_interceptor.dart
lucas-zimerman: Renamed on [GrpcClient](https://github.com/getsentry/sentry-dart/pull/3721/commits/1d928b374762b36bf69c6c0766436e361056235a)

packages/grpc/lib/src/sentry_grpc_interceptor.dart
lucas-zimerman: That's a good idea, although I'm afraid it would break if users fire multiple concurrent gRPC requests, since they'd share the same stopwatch instance and each new call would reset it mid-flight.

packages/grpc/test/mocks/mock_hub.dart
lucas-zimerman: Issue created https://github.com/getsentry/sentry-dart/issues/3749

lucas-zimerman:
I can't see which test failed on [sentry-flutter / macos | macos | beta (pull_request)](https://github.com/getsentry/sentry-dart/actions/runs/27217883555/job/80364516959?pr=3721)
sentry-flutter / macos | macos | beta (pull_request)
but it seems to be on flutter and not on the grpc package.

packages/grpc/lib/src/version.dart
```diff
@@ -0,0 +1,5 @@
+/// The SDK version reported to Sentry.io in the submitted events.
+const String sdkVersion = '9.21.0';
```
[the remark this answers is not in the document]
lucas-zimerman: It is aligned with the main repo https://github.com/getsentry/sentry-dart/blob/main/packages/dart/lib/src/version.dart

buenaflor:
@lucas-zimerman yeah unfortunately we have a test that seems to fail without outputting a test failure, it's pronably some async behaviour.

It's flaky so you can retry it, I'll spend some time soon to figure out what's wrong

packages/grpc/lib/src/sentry_grpc_interceptor.dart
buenaflor: ah nevermind then, let's keep it as it is

packages/grpc/lib/src/sentry_grpc_interceptor.dart
```diff
@@ -0,0 +1,366 @@
...
+              method.path,
+              grpcError,
+              stackTrace,
+            );
+          }
+        },
+      ),
+    );
+
+    return response;
```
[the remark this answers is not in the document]
lucas-zimerman: That is an ok tradeoff, calling sentry.close right after finishing a grpc call is not an expected behavior, still, testing it on a running program, there were no issues with the grpc and parent spans

packages/grpc/lib/src/sentry_grpc_interceptor.dart
```diff
@@ -0,0 +1,366 @@
+// ignore_for_file: invalid_use_of_internal_member, implementation_imports
+
+import 'dart:async';
+
+import 'package:grpc/grpc.dart';
+import 'package:meta/meta.dart';
+import 'package:grpc/src/generated/google/rpc/error_details.pb.dart' as rpc;
```
buenaflor: do you know if this is safe to use since it's a generated file which can change between versions, not sure if there is an alternative though, wdyt? maybe it's fine and we can keep it

packages/grpc/lib/src/sentry_grpc_interceptor.dart
```diff
@@ -0,0 +1,366 @@
...
+      options.metadata.forEach((key, value) {
+        final normalizedKey = key.toLowerCase();
+        if (!sendPii && _sensitiveHeaders.contains(normalizedKey)) return;
+        span.setData(
+          '${SemanticAttributesConstants.rpcRequestMetadataPrefix}$normalizedKey',
+          value,
+        );
+      });
+    }
+  }
```
buenaflor: seems like `method` and `request` parameters are unused here

packages/grpc/test/sentry_grpc_interceptor_test.dart
```diff
@@ -0,0 +1,1012 @@
...
+          await tr.finish();
+
+          expect(
+            fixture.service.lastReceivedMetadata?.containsKey('sentry-trace'),
+            isTrue,
+          );
+        });
+      });
+
+      group('_attachErrorDetails', () {
```
buenaflor: minor:

```suggestion
      group('on error with rich error details', () {
```

wdyt?

packages/grpc/lib/src/sentry_grpc_interceptor.dart
```diff
@@ -0,0 +1,366 @@
...
+  ///
+  /// Set [captureRequestHeaders] to `false` to suppress attaching outgoing
+  /// gRPC metadata (request headers) to the span. Sensitive headers
+  /// (`authorization`, `cookie`, etc.) are redacted unless
+  /// [SentryOptions.sendDefaultPii] is enabled. Defaults to `true`.
+  SentryGrpcInterceptor({
+    Hub? hub,
+    bool? captureFailedRequests,
+    bool enableBreadcrumbs = true,
+    bool captureRequestHeaders = true,
```
buenaflor: can you check if `captureRequestHeaders` exists in other SDKs?

for example in our HTTP integration (`failed_request_client.dart`) we gate all request headers behind `sendDefaultPii` which might also not be correct fwiw

packages/grpc/lib/src/sentry_grpc_interceptor.dart
```diff
@@ -0,0 +1,366 @@
+// ignore_for_file: invalid_use_of_internal_member, implementation_imports
+
+import 'dart:async';
+
+import 'package:grpc/grpc.dart';
```
buenaflor:
```suggestion
import 'package:grpc/service_api.dart';
import 'package:grpc/grpc_or_grpcweb.dart';
```

currently `package:grpc/grpc.dart` only supports native targets (not web), pls double check this

buenaflor (commented):
What is still missing:
- dedicated `.github/workflows/grpc.yml` file that runs the grpc tests - check out `drift.yml`
- entry in`.craft.yml`
- debug `internalLogger`, pls look at how other packages define it e.g `drift` `internal_logger.dart` and add some sensible debug logs here where it makes sense
- we can support web (see my comment): so it makes sense to also have a quick smoke test for that, maybe something like `test/web_compile_test.dart` with               `@TestOn('browser')`, imports the lib, constructs the interceptor and so on and in `.github/workflows/grpc.yml` we additionally run `dart test -p chrome`

packages/grpc/lib/src/sentry_grpc_interceptor.dart
lucas-zimerman: fixed on https://github.com/getsentry/sentry-dart/pull/3721/commits/79fede276a2f6a0cb0403b760a4b26d39231a9a4

packages/grpc/test/sentry_grpc_interceptor_test.dart
lucas-zimerman: deffo better, fixed on https://github.com/getsentry/sentry-dart/pull/3721/commits/79fede276a2f6a0cb0403b760a4b26d39231a9a4

packages/grpc/lib/src/sentry_grpc_interceptor.dart
lucas-zimerman: Fixed on https://github.com/getsentry/sentry-dart/pull/3721/commits/79fede276a2f6a0cb0403b760a4b26d39231a9a4

packages/grpc/lib/src/sentry_grpc_interceptor.dart
lucas-zimerman: Only our SDK implements `captureRequestHeaders`

Checking the other SDKs I got the following data:

- Always capture, redact sensitive (Python, Go, Java server-side) — Headers are always included; sensitive ones (Authorization, Cookie, API keys, etc.) are filtered/redacted. sendDefaultPii=true removes the filter.
- Entirely gated behind sendDefaultPii (Java client-side, .NET client-side) — No headers at all when sendDefaultPii=false; when true, still filters sensitive headers via a deny-list.
- Granular dataCollection API (JavaScript v11) — Fine-grained control with dataCollection.httpHeaders supporting true/false/allow-list/deny-list.

So we have mixed output. It is fair to drop `captureRequestHeaders` would only make sense if we always capture the headers despite what was set on `sendDefaultPii`.

At any case, I will drop `captureRequestHeaders` and opt for only dropping the header based on `sendDefaultPii` to match the HTTP integration of dart.

Would be nice if the SDKs would opt for a unique way for dealing with headers, but that will be a break change for us so lets keep it simple for now and patch it on a major version.

lucas-zimerman: removed on https://github.com/getsentry/sentry-dart/pull/3721/commits/18ee73d7c1d95e75f2c2d3975a910eda79691709

packages/grpc/lib/src/sentry_grpc_interceptor.dart
lucas-zimerman: It's re-exported from `package:grpc/grpc.dart` in `v4` but they dropped that export in `v5` (file still exists at the same path though). So the src/ import is actually the most portable option across both major versions.

The proto types come from the stable `google/rpc/error_details.proto` spec so they shouldn't change in a breaking way, and any change would bump grpc's major version anyway. I think we're fine keeping it.

packages/grpc/lib/src/sentry_grpc_interceptor.dart
```diff
@@ -0,0 +1,360 @@
...
+    if (!_hub.options.sendDefaultPii) return;
+    options.metadata.forEach((key, value) {
+      final normalizedKey = key.toLowerCase();
+      span.setData(
+        '${SemanticAttributesConstants.rpcRequestMetadataPrefix}$normalizedKey',
+        value,
+      );
+    });
+  }
+
```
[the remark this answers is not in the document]
lucas-zimerman: the state has changed.  Will update the description

buenaflor (commented):
overall looks good, regarding the attributes, for now let's just use the otel ones (even if they are generic) and create a separate issue for richer gRPC details.

please check the AI reviews as well.

packages/grpc/lib/src/sentry_grpc_interceptor.dart
```diff
@@ -0,0 +1,366 @@
...
+        onError: (Object error, StackTrace stackTrace) async {
+          final grpcError = error is GrpcError ? error : null;
+          span?.throwable = error;
+          span?.status = _grpcStatusToSpanStatus(grpcError);
+          span?.setData(
+            SemanticAttributesConstants.rpcResponseStatusCode,
+            grpcError?.codeName ?? 'UNKNOWN',
+          );
+          if (span != null && grpcError != null) {
+            _attachErrorDetails(span, grpcError);
```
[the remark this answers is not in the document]
lucas-zimerman: Since _attachErrorDetails was removed this comment does not apply

packages/grpc/lib/src/sentry_grpc_interceptor.dart
```diff
@@ -0,0 +1,299 @@
...
+
+  CallOptions _buildModifiedOptions(
+    CallOptions options,
+    InstrumentationSpan? span,
+    String methodPath,
+  ) {
+    final headers = <String, dynamic>{};
+    if (containsTargetOrMatchesRegExp(
+      _hub.options.tracePropagationTargets,
+      methodPath,
```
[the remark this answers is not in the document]
lucas-zimerman: The` tracePropagationTargets` check runs against the method path because that's the only context available at intercept time. Users with custom hostname-based targets can match against the method path pattern instead as a workaround.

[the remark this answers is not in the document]
lucas-zimerman: Created an issue to track this since it's not blocking the initial implementation https://github.com/getsentry/sentry-dart/issues/3792

packages/grpc/pubspec.yaml
```diff
@@ -0,0 +1,33 @@
+name: sentry_grpc
+description: An integration which adds support for performance tracing and error monitoring for gRPC calls.
+version: 9.20.0
```
buenaflor: needs to be updated to `9.22.0` then we're good to go

buenaflor (commented):
nice! one last thing about the version then we're good

packages/grpc/pubspec.yaml
lucas-zimerman:
```suggestion
version: 9.22.0
```

buenaflor (approved):
nice 👍

PR #3733, opened 2026-06-02
chore(example): Add app hang button and enable Sentry Gradle plugin

buenaflor:
Adds a configurable app-hang trigger to the example app's events screen.

The new "Trigger app hang" button blocks the main isolate for a duration
specified via an accompanying text field (defaulting to 5 seconds), making
it easy to manually verify app-hang reporting in the Flutter SDK.

To support the controller and state, `EventsScreen` is converted from a
`StatelessWidget` to a `StatefulWidget`.

### DIFF
--- a/packages/flutter/example/android/app/build.gradle
+++ b/packages/flutter/example/android/app/build.gradle
@@ -2,9 +2,7 @@ plugins {
     id "com.android.application"
     id "kotlin-android"
     id "dev.flutter.flutter-gradle-plugin"
-//    uncomment this to upload proguard mapping file and debug symbols to Sentry
-//    and also set the token (sentry.properties) file
-   id "io.sentry.android.gradle"
+    id "io.sentry.android.gradle"
 }
 
 def localProperties = new Properties()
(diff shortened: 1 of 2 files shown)


packages/flutter/example/android/app/build.gradle
```diff
@@ -98,8 +98,15 @@ dependencies {
...
+// Disabled by default because symbol upload fails on CI when SENTRY_AUTH_TOKEN
+// is unavailable (e.g. on fork PRs). Uncomment together with the
+// io.sentry.android.gradle plugin above to upload debug symbols to Sentry locally.
 // sentry {
+//     autoInstallation {
+//       enabled = false
+//     }
 //     uploadNativeSymbols = true
-//     includeNativeSources = true
+//     org = "sentry-sdks"
```
denrase, line 108: We have those in the `pubspec.yaml`, why is this additionally needed?

denrase (approved):
Just one question regarding the gradle setup.

packages/flutter/example/android/app/build.gradle
buenaflor: this is for uploading proguard mapping files mainly which is not handled by our plugin

PR #3742, opened 2026-06-05
ci: Fix JUnit result conversion

buenaflor:
Fix JUnit test result conversion after Pub started rejecting globally activated packages with Flutter SDK constraints.

Both Dart and Flutter test composite actions now use the pure-Dart `better_test_reporter` CLI pinned to `0.1.1` to convert `test-vm.jsonl` into `TEST-vm-junit.xml`.

Dart-only jobs keep using the SDK installed by `dart-lang/setup-dart`, so report conversion no longer installs Flutter or mutates `PATH`. The Dart 3.5 matrix still skips conversion because `better_test_reporter` requires Dart 3.7.2 or newer.

Flutter jobs no longer need the temporary `flutter_junitreport` git dependency; they use the same converter after `flutter test --file-reporter json:test-vm.jsonl`.


.github/actions/flutter-test/action.yml
```diff
@@ -32,10 +32,10 @@ runs:
       shell: bash
       working-directory: ${{ inputs.directory }}
 
-    - name: Install flutter_junitreport
+    - name: Install better_test_reporter
       if: matrix.target == 'linux' || matrix.target == 'windows' || matrix.target == 'macos'
```
denrase: We could also check !cancelled, same as Test VM

denrase (approved):

.github/actions/flutter-test/action.yml
buenaflor: done

PR #3744, opened 2026-06-08
feat(flutter): Add replay trace ID sync for Android

### LINKED ISSUE #3729: Populate trace_ids in replay events to enable searching replays by trace ID
## Problem

When searching for replays by trace ID in Sentry's Explore UI, replays from Dart/Flutter apps are not discoverable. The SDK doesn't collect or include `trace_ids` in replay events.

## Solution

When a transaction is captured while replay is recording, the trace ID should be registered with the replay system and included in the next replay segment's `trace_ids` field.

### Reference implementations
- JS: https://github.com/getsentry/sentry-javascript/blob/30eb68fff5077211c30c61ba74625e66ab514870/packages/replay-internal/src/coreHandlers/handleAfterSendEvent.ts#L34-L43
- Java: https://github.com/getsentry/sentry-java/pull/5473

### Key points
- Collect trace IDs when transactions are successfully captured during replay recording
- Limit to max 100 trace IDs per segment (matching JS SDK)
- Clear collected trace IDs after each segment is sent

### Note
Since Flutter delegates replay to native SDKs, this may need coordination with sentry-java (Android) and sentry-cocoa (iOS) implementations, plus a mechanism to pass trace IDs from Dart transactions to the native replay system.

Action taken on behalf of Roman Zavarnitsyn.

buenaflor:
## :scroll: Description
Register Dart transaction and stream segment trace IDs with native Android replay through the new `registerTraceId` JNI binding. This lets native replay segments collect `trace_ids` for Flutter app activity so replays can be found by trace ID.

Flutter delegates replay recording to the native SDK, but Dart-traceId were not forwarded to native replay for trace ID collection. The SDK now dispatches an internal lifecycle event after Dart transactions are captured and lets replay telemetry register transaction and segment trace IDs with the native binding.

Partly resolves https://github.com/getsentry/sentry-dart/issues/3729 - only for Android currently

- `fvm dart test test/sentry_client_sdk_lifecycle_test.dart`
- `fvm flutter test test/integrations/replay_telemetry_integration_test.dart`
- `fvm flutter test test/integrations/native_trace_sync_integration_test.dart`
- `fvm dart analyze lib/src/sdk_lifecycle_hooks.dart lib/src/sentry_client.dart test/sentry_client_sdk_lifecycle_test.dart`
- `fvm flutter analyze lib/src/integrations/replay_telemetry_integration.dart lib/src/native/sentry_native_binding.dart lib/src/native/java/sentry_native_java.dart lib/src/native/c/sentry_native.dart lib/src/native/sentry_native_channel.dart lib/src/web/sentry_web.dart test/integrations/replay_telemetry_integration_test.dart test/mocks.mocks.dart`
- Pre-commit hook ran package-level Dart and Flutter analyzers successfully.

## :pencil: Checklist
- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps
Native Android behavior is covered through the generated JNI binding and existing native SDK implementation; the Dart/Flutter tests cover lifecycle dispatch and adapter forwarding.

### DIFF
--- a/packages/flutter/test/integrations/replay_telemetry_integration_test.dart
+++ b/packages/flutter/test/integrations/replay_telemetry_integration_test.dart
@@ -1,5 +1,7 @@
 // ignore_for_file: invalid_use_of_internal_member
 
+import 'dart:async';
+
 import 'package:flutter_test/flutter_test.dart';
 import 'package:mockito/mockito.dart';
 import 'package:sentry_flutter/sentry_flutter.dart';
@@ -61,6 +63,33 @@ void main() {
             .dispatchCallback(OnGenerateNewTrace(traceId, spanId));
 
         verify(fixture.nativeBinding.registerTraceId(traceId)).called(1);
+      });
+
+      test('waits for generated trace id registration with native replay',
+          () async {
+        fixture.options.replay.sessionSampleRate = 0.5;
+        final traceId = SentryId.newId();
+        final spanId = SpanId.newId();
+        final completer = Completer<void>();
+        var dispatchCompleted = false;
+
+        when(fixture.nativeBinding.registerTraceId(traceId))
+            .thenAnswer((_) => completer.future);
+
+        await fixture.getSut().call(fixture.hub, fixture.options);
+        final dispatch = fixture.options.lifecycleRegistry
+            .dispatchCallback(OnGenerateNewTrace(traceId, spanId));
+        unawaited(Future<void>.value(dispatch).then((_) {
+          dispatchCompleted = true;
+        }));
+        await Future<void>.delayed(Duration.zero);
+
+        expect(dispatchCompleted, false);
+
+        completer.complete();
+        await dispatch;
+
+        expect(dispatchCompleted, true);
       });
 
       test('does not register segment span trace id with native replay',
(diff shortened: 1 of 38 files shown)


packages/flutter/test/integrations/replay_telemetry_integration_test.dart
```diff
@@ -39,12 +41,88 @@ void main() {
...
+        )).called(1);
+      });
+
+      test('registers generated trace id with native replay', () async {
+        fixture.options.replay.sessionSampleRate = 0.5;
+        final traceId = SentryId.newId();
+        final spanId = SpanId.newId();
+
+        await fixture.getSut().call(fixture.hub, fixture.options);
+        clearInteractions(fixture.nativeBinding);
```
denrase, line 61: Why do we need to clear interactions here?

denrase (approved):

packages/flutter/test/integrations/replay_telemetry_integration_test.dart
buenaflor: because when the integration is called we register the trace id, without this we would have two calls because we trigger the callback

PR #3745, opened 2026-06-08
fix(dart): Add missing log byte outcomes

### LINKED ISSUE #3367: Implement Log Outcomes
* SDK must count each dropped log.
* SDK must calculate the size of each dropped log.
* SDK must emit both values as client outcomes.
* precise byte counting is not required; an approximate size is sufficient because:
  * Dropped logs are not billable.
  * sufficient to provide a rough overview to users (count + approximate size)<br>

If the SDK already does it then close this issue

[LOGS-482](https://linear.app/getsentry/issue/LOGS-482/fix-sdk-to-send-correct-outcomes-and-improve-ui-hover-info-to-show)

buenaflor:
## :scroll: Description

Adds `log_byte` client report outcomes for discarded logs while preserving the existing `{ reason, category, quantity }` client report shape. Dropped logs now record paired `log_item` and `log_byte` entries for `beforeSendLog`, rate limit filtering, transport failures, and oversized telemetry buffer drops.

The client reports spec requires SDKs to report dropped log volume as a separate `log_byte` category in addition to dropped log count. This gives users visibility into the approximate amount of log data discarded by the SDK.

Fixes #3367

- Unit tests

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps

None.


packages/dart/lib/src/transport/rate_limiter.dart
```diff
@@ -21,10 +21,15 @@ class RateLimiter {
         dropItems ??= [];
         dropItems.add(item);
 
-        _options.recorder.recordLostEvent(
-          DiscardReason.rateLimitBackoff,
-          DataCategory.fromItemType(item.header.type),
-        );
+        final category = DataCategory.fromItemType(item.header.type);
+        if (category == DataCategory.logItem) {
+          _recordRateLimitedLogItem(item);
```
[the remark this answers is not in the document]
buenaflor: Fixed in 02c08422 by treating log_byte rate limits as applying to log envelope items, with regression coverage in rate_limiter_test.dart.

*— Claude Code*

packages/dart/lib/src/client_reports/discarded_event.dart
```diff
@@ -35,8 +35,12 @@ extension _OutcomeExtension on DiscardReason {
         return 'queue_overflow';
       case DiscardReason.cacheOverflow:
         return 'cache_overflow';
+      case DiscardReason.bufferOverflow:
+        return 'buffer_overflow';
       case DiscardReason.rateLimitBackoff:
         return 'ratelimit_backoff';
+      case DiscardReason.internalSdkError:
+        return 'internal_sdk_error';
```
buenaflor, line 43: official discard reasons: https://develop.sentry.dev/sdk/telemetry/client-reports/#discard-reasons-1

packages/dart/lib/src/client_reports/discard_reason.dart
```diff
@@ -10,6 +10,8 @@ enum DiscardReason {
   networkError,
   queueOverflow,
   cacheOverflow,
+  bufferOverflow,
   rateLimitBackoff,
+  internalSdkError,
   ignored,
 }
```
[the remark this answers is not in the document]
buenaflor: it is a valid definition

packages/dart/lib/src/client_reports/client_report_recorder.dart
```diff
@@ -20,6 +20,20 @@ class ClientReportRecorder {
...
   }
 
+  /// Records a dropped log as a [DataCategory.logItem] count and, when the
+  /// size is known, an additional [DataCategory.logByte] outcome as required
+  /// by the client reports spec for log byte outcomes.
+  ///
+  /// Pass a null [bytes] when the size cannot be determined (e.g. the log
+  /// could not be encoded); the [DataCategory.logByte] outcome is then omitted
+  /// rather than reported as zero.
+  void recordLostLog(final DiscardReason reason, {int count = 1, int? bytes}) {
```
denrase, line 30: Not sure if we should just do this on the call site if count/bytes are optional parameters anyway, or as an alternative, pass the log and do the counting here as an implementation detail (if possible).

packages/dart/lib/src/telemetry/processing/in_memory_buffer.dart
denrase, line 142: Should we also call the callback in this case?

denrase (approved):
Just one additional error path where it might make sense to check if we should also create a cleint report. Also the API with two optional parameters might be something we could tighten up if possible.

packages/dart/lib/src/client_reports/client_report_recorder.dart
buenaflor: > pass the log and do the counting here as an implementation detail

the buffer already reports the size so imo it's easier to have this function be unaware on how it functions

according to spec the logItem and logByte should be paired but in some edge cases (e.g encoding fails) we cannot report any bytes

packages/dart/lib/src/telemetry/processing/in_memory_buffer.dart
buenaflor: no need to, when flushing the transport already records lost events so if we did it here as well we would double count

PR #3746, opened 2026-06-08
Migrate sentry_flutter to built-in Kotlin

ubxty:
Migrates the `sentry_flutter` subpackage of this monorepo (which lives at `packages/flutter/`, with `name: sentry_flutter` in its `pubspec.yaml`) to use the built-in Kotlin support that ships with the Flutter Gradle plugin.

Follows the official Flutter migration guide: https://docs.flutter.dev/release/breaking-changes/migrate-to-built-in-kotlin/for-plugin-authors

## Scope

This PR covers the `sentry_flutter` subpackage only. The other subpackages in this monorepo (`sentry`, `sentry_dio`, `sentry_drift`, `sentry_file`, `sentry_firebase_remote_config`, `sentry_hive`, `sentry_isar`, `sentry_link`, `sentry_logging`, `sentry_sqflite`, `sentry_supabase`) are pure-Dart packages with no Android `build.gradle` and need no changes.

## Changes to `packages/flutter/android/build.gradle`

- Remove `classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"` and `ext.kotlin_version = '1.8.0'` from the `buildscript` block (KGP is now bundled with the Flutter Gradle plugin).
- Remove `apply plugin: 'kotlin-android'`.
- Remove the `kotlinOptions { jvmTarget = JavaVersion.VERSION_1_8 }` block from inside `android { }`.
- Add a top-level `kotlin { compilerOptions { jvmTarget = org.jetbrains.kotlin.gradle.dsl.JvmTarget.JVM_17 } }` block (the new DSL that replaces the deprecated `kotlinOptions`).
- Hardcode the `kotlin-stdlib-jdk8` dependency to `1.8.0` (the previous `$kotlin_version` ext property no longer exists).

## Other changes

- `packages/flutter/pubspec.yaml`: bump `version: 9.21.0` -> `9.22.0`, `sdk: '>=3.5.0 <4.0.0'` -> `^3.12.0`, `flutter: '>=3.24.0'` -> `'>=3.44.0'`. The Flutter bump is required by the new `kotlin.compilerOptions` DSL.
- `packages/flutter/lib/src/version.dart`: bump `sdkVersion` to `9.22.0` to keep it in sync with the pubspec (the `version_test.dart` test enforces this).
- `CHANGELOG.md`: new `9.22.0` entry documenting the migration.

## Validation

- Ran `flutter test` in `packages/flutter` on Flutter 3.44.1 / Dart 3.12.1: **all 996 tests pass** (3 skipped, 0 failed).
- The example app at `packages/flutter/example/` was **not** built or migrated in this PR per the task scope. The example app's own `build.gradle` still uses `id "kotlin-android"` and its `gradle.properties` does not yet have the `android.builtInKotlin=true` / `android.newDsl=true` flags, so it will still print the KGP deprecation warning until a follow-up PR migrates the example app. This PR eliminates the KGP warning from the plugin itself, which is what fixes the warning that appears when `sentry_flutter` is consumed as a dependency.

Fixes the warning: "Your app uses the following plugins that apply Kotlin Gradle Plugin (KGP)".

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,17 @@
 # Changelog
 
+## 9.22.0
+
+### Changes
+
+- Migrate `sentry_flutter` (the Flutter subpackage of the sentry-dart monorepo) to use built-in Kotlin, per the official Flutter migration guide: https://docs.flutter.dev/release/breaking-changes/migrate-to-built-in-kotlin/for-plugin-authors
+  - Remove the `org.jetbrains.kotlin:kotlin-gradle-plugin` classpath and the `ext.kotlin_version` property from `packages/flutter/android/build.gradle`
+  - Remove `apply plugin: 'kotlin-android'` from `packages/flutter/android/build.gradle`
+  - Remove the `kotlinOptions { jvmTarget = ... }` block from `packages/flutter/android/build.gradle`
+  - Add a top-level `kotlin { compilerOptions { jvmTarget = org.jetbrains.kotlin.gradle.dsl.JvmTarget.JVM_17 } }` block to `packages/flutter/android/build.gradle`
+  - Hardcode the `kotlin-stdlib-jdk8` dependency version (the previous `$kotlin_version` ext property is no longer defined)
+- Bump minimum Dart SDK to `^3.12.0` and minimum Flutter version to `>=3.44.0`, as required by the new `kotlin.compilerOptions` DSL.
+
 ## 9.21.0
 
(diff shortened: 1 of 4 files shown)


buenaflor:
Unfortunately we're blocked here by the JNI plugin which is not compatible either yet so the build would fail regardless. We'll have to wait until they update the JNI package.

PR #3760, opened 2026-06-15
enh(flutter): Support int64 values from sentry-native

### LINKED ISSUE #3669: Add support for `int64` in native
### Description

Currently we do not support `int64` values that interact with sentry-native.

This might lead to silent errors or the SDK silently swallowing values without us knowing.

e.g if the native SDK now stores image_size as int64, those values silently become null

buenaflor:
Native `int64`/`uint64` values from sentry-native are no longer silently dropped, and large Dart `int` values are no longer coerced to strings when sent to native.

Previously, reading a native value typed as `int64` or `uint64` (for example `image_size` on a debug image) fell through to the `default` case in `castPrimitive`, which logged a warning and returned `null`. In the write direction, any Dart `int` outside the int32 range was converted to a string instead of an integer. Both were silent type-fidelity losses.

**Read path**

`castPrimitive` now handles `SENTRY_VALUE_TYPE_INT64` and `SENTRY_VALUE_TYPE_UINT64` via `sentry_value_as_int64`/`sentry_value_as_uint64`. Dart's `int` is signed 64-bit, so a `uint64` above 2^63-1 wraps to a negative number; native values that reach here in practice (e.g. `image_size`) are well within the signed range.

**Write path**

Dart ints outside the int32 range now use `sentry_value_new_int64` instead of `toString()`. Note that values above 2^53 may lose precision once serialized as a JSON number downstream — this matches how other SDKs send large integers.

**FFI bindings**

`sentry_value_new_int64`/`uint64` and `sentry_value_as_int64`/`uint64` are added to `ffi-native.yaml` and `binding.dart`. The `sentry_value_u` extension is now named (`SentryValueExtension`) so the new round-trip test can exercise `castPrimitive`.

The FFI test suite runs on Linux and Windows CI (it is gated by `@TestOn('vm && (windows || linux)')` and cannot run on macOS, which uses sentry-cocoa).

Fixes GH-3669

### DIFF
--- a/packages/flutter/lib/src/native/c/sentry_native.dart
+++ b/packages/flutter/lib/src/native/c/sentry_native.dart
@@ -12,6 +12,7 @@ import 'package:sentry/src/utils/iterable_utils.dart';
 
 import '../../../sentry_flutter.dart';
 import '../../replay/replay_config.dart';
+import '../../utils/internal_logger.dart';
 import '../native_app_start.dart';
 import '../sentry_native_binding.dart';
 import '../sentry_native_invoker.dart';
@@ -39,13 +40,13 @@ class SentryNative with SentryNativeSafeInvoker implements SentryNativeBinding {
 
   SentryNative(this.options);
 
-  void _logNotSupported(String operation) => options.log(
-      SentryLevel.debug, 'SentryNative: $operation is not supported');
+  void _logNotSupported(String operation) =>
+      internalLogger.debug('SentryNative: $operation is not supported');
 
   @override
   FutureOr<void> init(Hub hub) {
     if (!options.enableNativeCrashHandling) {
-      options.log(SentryLevel.info, 'SentryNative crash handling is disabled');
+      internalLogger.info('SentryNative crash handling is disabled');
     } else {
       tryCatchSync("init", () {
         final cOptions = createOptions(options);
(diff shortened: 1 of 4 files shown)


packages/flutter/lib/src/native/c/sentry_native.dart
```diff
@@ -363,6 +358,13 @@ extension on binding.sentry_value_u {
         return (SentryNative.native.value_is_true(this) == 1) as T;
       case binding.sentry_value_type_t.SENTRY_VALUE_TYPE_INT32:
         return SentryNative.native.value_as_int32(this) as T;
+      case binding.sentry_value_type_t.SENTRY_VALUE_TYPE_INT64:
+        return SentryNative.native.value_as_int64(this) as T;
+      case binding.sentry_value_type_t.SENTRY_VALUE_TYPE_UINT64:
```
denrase, line 363: We could consider loggin if we exceed signed range.

 ```dart
   case binding.sentry_value_type_t.SENTRY_VALUE_TYPE_UINT64:
     final value = SentryNative.native.value_as_uint64(this);
     if (value < 0) {
       logger(
         SentryLevel.warning,
         'SentryNative: uint64 value exceeds signed int64 range and cannot be represented as Dart int',
       );
       return null;
     }
     return value as T;
 ```

denrase (approved):
Just one suggestion.

packages/flutter/lib/src/native/c/sentry_native.dart
buenaflor: done

PR #3765, opened 2026-06-15
deps(flutter): Bump jni to 1.0.0 and jnigen to 0.16.0 in v10

buenaflor:
Moves the Android JNI bindings from `jni`/`jnigen` 0.14.2 to `jni` 1.0.0 / `jnigen` 0.16.0. This keeps the existing architecture — `jni` remains a direct dependency and still bundles the native lib as an `ffiPlugin`; `jni_flutter` is intentionally **not** added because it only carries the `Jni.androidApplicationContext`/`androidActivity` helpers, which this SDK doesn't use (it gets the context via its own `SentryFlutterPlugin.applicationContext`).

### Why exact pins (not caret)
`jni` is pinned to `1.0.0` and `jnigen` to `0.16.0`. jni 1.0.0 adds a compile-time consistency check between the runtime and the JNIgen-generated code. Because `binding.dart` is committed (not regenerated downstream), a caret range could let a consumer resolve a newer `jni` that rejects our committed binding at *their* build. Note `jni` and `jnigen` are not on the same version line (they're published together), so caret-on-both would not keep them aligned.

### What changed
- **Regeneration moved to a programmatic generator** (`tool/jnigen.dart`, replacing `ffi-jni.yaml`). JNIgen 1.0.0 faithfully mirrors Java APIs where a `@Nullable` getter pairs with a non-null setter, which Dart rejects as `getter_not_subtype_setter_types`. A small visitor excludes four such **unused** setters (`SentryAndroidOptions.beforeScreenshotCaptureCallback`/`beforeViewHierarchyCaptureCallback`, `SentryOptions.gestureTargetLocators`, `SentryEvent.timestamp`), keeping the nullable getters and producing a clean binding.
- **Interop + integration tests migrated to the 1.0.0 API**: `getX()`/`setX()` → getters/setters (with `release$1`/`type$1` where names collide with `release()`/the static `type`), `JList.array`/`JMap.hash` → `JArrayList`/`JHashMap`, `JByteArray.from` → allocate-then-fill, `intValue`/`longValue`/`doubleValue`/`booleanValue(releaseOriginal:)` → `toDartInt`/`toDartDouble`/`toDartBool`, generic type witnesses (`JList.type(...)`) → bare `JList.type`, and Dart-collection access via `.asDart()`.
- **Regeneration scripts** updated to invoke `tool/jnigen.dart`. `safe-regenerate-jni.sh` already implemented the bootstrap fallback (build only the plugin jar, then run JNIgen) that's needed when bumping `jni` to an incompatible version — exactly the path used to regenerate here.

### Verification
- `flutter analyze` (whole package, incl. `lib/`, `example/`, `tool/`) → **No issues found**; `dart format` clean.
- ⚠️ **On-device Android integration tests have not been run yet.** Per the package conventions, JNI is only verifiable through integration tests on a device/emulator, so that run is the required correctness gate before this leaves draft. Kept as a draft for that reason.


packages/flutter/lib/src/native/java/sentry_native_java.dart
```diff
@@ -328,18 +326,23 @@ JList<JObject> dartToJList(List<dynamic> values) {
...
+    jMap.put(jk, jv);
     jk.release();
     jv.release();
   }
   return jMap;
 }
 
+/// Builds a [JByteArray] from Dart [bytes]. JNIgen 1.0.0 dropped the
+/// `JByteArray.from` factory in favour of allocate-then-fill.
+JByteArray _toJByteArray(List<int> bytes) =>
```
denrase: m: We have duplicated this, we could think about extracting into a helper and re-use it.

denrase (approved):
We do update the min version in the sample and pubspec. If we refernce these in any other file, we should update them as well.

```yaml
sdk: '3.9.2'
flutter: '3.35.6'
```

We could also think about making an issue where we update deprecated dart/flutter APIs if anything changed.

packages/flutter/lib/src/native/java/sentry_native_java.dart
buenaflor: done

gabrielgarciagava:
I see this got merged today. I just would like to know if there is any release planned already to have this included.

buenaflor:
Hey, we plan to release a v10 alpha this week

PR #3778, opened 2026-06-22
feat(dart): Add array attributes to telemetry

### LINKED ISSUE #3665: Add support for arrays in `SentryAttribute`
## Summary
`SentryAttribute` currently only supports primitive attribute values.

## Problem
There is no support for array values, which limits attribute expressiveness for SDK users.

## Proposal
Add support for arrays in `SentryAttribute` so users can attach list-like metadata without flattening or stringifying manually.

## Blocked

This is currently blocked by various points in the pipeline. If you send attribute arrays now they dont show up in Sentry

Will be unblocked by https://github.com/getsentry/sentry/pull/113567

buenaflor:
## :scroll: Description

Adds support for homogeneous scalar array attributes on streaming telemetry — usable on spans, logs, and metrics. Four typed factories are added to `SentryAttribute`:

```dart
SentryAttribute.stringArray(List<String> value)
SentryAttribute.intArray(List<int> value)
SentryAttribute.doubleArray(List<double> value)
SentryAttribute.boolArray(List<bool> value)
```

Each serializes to the wire format `{"value": [...], "type": "array"}`. `StreamingInstrumentationSpan.setData` also now maps `List<String|int|double|bool>` values to these factories; heterogeneous, nullable, and untyped lists fall through to the existing ignore path and are not sent.

The factories are intentionally typed (rather than a single inferring `array(List)`): the element type *is* the constraint, so homogeneity is enforced at compile time and `List<dynamic>`/mixed lists can't be constructed. Logs and metrics required no production change — both already accept `Map<String, SentryAttribute>` and serialize attributes via the same `value.toJson()` passthrough as spans, so arrays flow through unchanged.

Sentry's attribute protocol ingests homogeneous arrays of primitive types using the `array` type. This brings the Dart SDK in line so users can attach list-valued dimensions to spans, logs, and metrics. Note: array attributes are ingested but not yet visible in the Sentry UI (server-side limitation). Objects and mixed/nested arrays are not yet supported by the protocol and are intentionally out of scope; the design is additive, so they can be layered on later without breaking changes.

Closes https://github.com/getsentry/sentry-dart/issues/3665

- Unit tests for each factory verifying the `{value, type: 'array'}` serialization, including the empty-list case.
- Serialization tests confirming array attributes survive end-to-end on logs, metrics, and the span capture pipeline (i.e. no sink strips them).
- Dispatch tests for `setData` covering all four list types plus rejection of mixed-element and null-containing lists.
- A debug-formatting test for array attribute output.

The full core SDK suite and analyzer pass green.

## :pencil: Checklist
- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps

Array attributes are ingested but not yet surfaced in the Sentry UI. Support for object and mixed/nested array attribute values can be added additively (a general `array`/`object` factory + matching `setData` branches) once the protocol and UI support them.


packages/dart/lib/src/protocol/sentry_attribute.dart
```diff
@@ -45,6 +45,22 @@ class SentryAttribute {
     return SentryAttribute(value, 'double');
   }
 
+  factory SentryAttribute.stringArray(List<String> value) {
```
denrase, line 48: nit: order coul be the same as the inidvidual types, so string, bool, int, double

denrase (approved):
LGTM 👍

packages/dart/lib/src/protocol/sentry_attribute.dart
buenaflor: Reordered the array factories to match the scalar order (string, bool, int, double) in e667d4037.

PR #3797, opened 2026-06-29
fix: Correct feature flag scope buffer updates

denrase:
## :scroll: Description

Fixes scope feature flag updates so existing flags are refreshed as the newest entry before enforcing the 100-entry limit. This prevents full-buffer updates from dropping unrelated flags or shrinking the buffer, and adds regression coverage for the 100 most recent unique flags behavior.

https://develop.sentry.dev/sdk/foundations/client/integrations/feature-flags/#tracking-feature-flag-evaluations

Added regression test.

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes


buenaflor (commented):
I think we're still missing the active span specific implementation where the first 10 flag evaluations end up as `flag.evaluation.*` attributes on the span if I understand the dev doc correctly

denrase:
@buenaflor Yes, will do this in a separate PR, this is just a fix for the scope level buffer update behaviour.

denrase:
Here's the other PR: https://github.com/getsentry/sentry-dart/pull/3806

buenaflor (approved):

denrase:
@buenaflor ~~Addressed bot feedback if you also wanna take a look.~~ Was unrelated, doing proper copy when cloning scopes now.

packages/dart/lib/src/protocol/sentry_feature_flag.dart
```diff
@@ -33,6 +33,15 @@ class SentryFeatureFlag {
     };
   }
 
+  @internal
+  SentryFeatureFlag copy() {
```
buenaflor, line 38: should we call it clone() to be consistent with others e.g scope clone(), contexts clone() etc..

denrase: We already have clone, which is deprecated. Will elave as copy for now and add a comment to use clone instead when deprecation is removed and we can make it internal.

buenaflor: yeah we should probably make clone internal across the board

PR #3805, opened 2026-06-30
feat(telemetry): Align span attributes with Sentry Conventions

buenaflor:
Closes https://github.com/getsentry/sentry-dart/issues/3818

Aligns the SDK's span/telemetry attribute keys with the [Sentry Conventions](https://getsentry.github.io/sentry-conventions/attributes/) source of truth. Every flagged attribute in `packages/dart/lib/src/constants.dart` was cross-checked against the conventions model (`getsentry/sentry-conventions`).

> [!NOTE]
> The Supabase `db.query.text`/`db.query.summary` work was split into the stacked follow-up #3817. This PR keeps only Supabase's foundational key migration.

### Database span attributes (wire change)

- `db.system` → `db.system.name` across all database integrations (drift, hive, isar, sqflite, supabase).
- `db.name` → `db.namespace` across all database integrations.
- `db.collection` → `db.collection.name` in isar.
- Supabase emits `db.operation.name` and `db.collection.name` instead of `db.operation` and `db.table`. (Supabase still emits the raw filter list under `db.query.text` and the generated SQL under `db.sql.query`; the query-text rework moved to #3817.)

### HTTP span attributes (wire change)

- `url` → `url.full` (the bare `url` key is not a convention).
- `http.response_content_length` → `http.response.body.size`.

### App start (wire change)

- The Flutter app start handler emits `app.vitals.start.type` instead of the custom `app_start_type`, matching the span-v2 handler.

### Internal cleanup (no wire change)

- Removed deprecated, unused attribute constants whose stable replacement was already in use (`app.app_*`, `os.build`, `device.connection_type`) and corrected the `device.locale` / `device.timezone` comments — neither is actually deprecated in the conventions.
- Removed the `SentrySpanData` key bag. Codified keys live in `SemanticAttributesConstants`; keys with no convention home (`db.schema`, `db.url`, `db.sdk`, `db.body`, `file.path`, `file.size`) moved to `ProposedSemanticAttributes`, and `app.vitals.start.value` moved there too.
- Dropped the per-integration `dbSystemKey` / `dbNameKey` / `dbCollectionKey` constants in hive/isar/sqflite in favor of `SemanticAttributesConstants`; the integration-specific `db.system.name` value constants (`sqlite`, `isar`, `flutter_hive`, `postgresql`) stay local and are named with a `dbSystemName` prefix.
- Production code references `SemanticAttributesConstants` for these keys instead of string literals (dart HTTP client, dio, supabase, flutter app start, file). Unit tests assert literal attribute keys (no tautology).

### Compatibility

Relay backfills the old key (soft migration):

- `db.system` → `db.system.name` (`_status: backfill`).
- `http.response_content_length` → `http.response.body.size` (`_status: backfill`).

**Hard cutovers** — no Relay backfill, consumers querying the old key must migrate:

- `db.name` → `db.namespace`.
- `db.collection` (isar) and supabase `db.table` → `db.collection.name`.
- `url` → `url.full`.
- `app_start_type` → `app.vitals.start.type`.

(`db.operation` → `db.operation.name` is a `normalize` deprecation.)


packages/drift/lib/src/constants.dart
```diff
@@ -1 +1,4 @@
 const String integrationName = 'sentryDriftTracing';
+
+/// The `db.system.name` value for SQLite databases.
+const String dbSystemSqlite = 'sqlite';
```
denrase: Shold the prefix reflect the new key format, so `dbSystemNameSqlite`?

packages/drift/test/sentry_drift_test.dart
```diff
@@ -832,11 +832,11 @@ void _verifySpan(
   expect(span?.status, status);
   expect(span?.origin, SentryTraceOrigins.autoDbDriftQueryInterceptor);
   expect(
-    span?.data[SentrySpanData.dbSystemKey],
-    SentrySpanData.dbSystemSqlite,
+    span?.data[SemanticAttributesConstants.dbSystemName],
+    'sqlite',
```
denrase, line 836: Shouldn't we use the new key for this value?

packages/drift/test/spanv2_integration_test.dart
```diff
@@ -66,8 +66,8 @@ void main() {
         equals('db.sql.query'),
       );
       expect(
-        span.attributes[SemanticAttributesConstants.dbSystem]?.value,
-        equals('db.sqlite'),
+        span.attributes[SemanticAttributesConstants.dbSystemName]?.value,
+        equals('sqlite'),
```
denrase, line 70: Also in this file.

packages/flutter/lib/src/integrations/native_app_start_handler_v2.dart
```diff
@@ -110,7 +110,7 @@ class NativeAppStartHandlerV2 {
     };
     appStartSpan.setAttribute(legacyValueKey, durationMs);
     appStartSpan.setAttribute(
-      SemanticAttributesConstants.appVitalsStartValue,
+      ProposedSemanticAttributes.appVitalsStartValue,
```
denrase, line 113: Are these really proposed semantic attributes that will evnetually make it into the convention as-is? If yes, would any of the keys need to change to match the new format? If not, is there a more fitting name, like "NonSemanticAttributesConstants" or "TempSemanticAttributesConstants", etc.

packages/hive/lib/src/sentry_hive_impl.dart
```diff
@@ -21,7 +21,7 @@ class SentryHiveImpl implements SentryHiveInterface {
 
   @internal
   // ignore: public_member_api_docs
-  static const dbSystemKey = 'db.system';
+  static const dbSystemKey = 'db.system.name';
```
denrase: We should also change the const to match the value -> `dbSystemNameKey`

packages/hive/test/sentry_box_base_test.dart
```diff
@@ -23,7 +23,7 @@ void main() {
     expect(span?.status, SpanStatus.ok());
     // ignore: invalid_use_of_internal_member
     expect(span?.origin, SentryTraceOrigins.autoDbHiveBoxBase);
-    expect(span?.data[SentryHiveImpl.dbSystemKey], SentryHiveImpl.dbSystem);
+    expect(span?.data[SentryHiveImpl.dbSystemKey], 'flutter_hive');
```
denrase: Why not keep using the defined constant? It did not change.

packages/isar/lib/src/sentry_isar.dart
```diff
@@ -15,7 +15,7 @@ class SentryIsar implements Isar {
 
   @internal
   // ignore: public_member_api_docs
-  static const dbSystemKey = 'db.system';
+  static const dbSystemKey = 'db.system.name';
```
denrase:
```suggestion
  static const dbSystemNameKey = 'db.system.name';
```

denrase: Also, isn't this a const a duplicate of `SemanticAttributesConstants.dbSystemName`?

packages/sqflite/lib/src/sentry_database.dart
```diff
@@ -45,7 +45,7 @@ class SentryDatabase extends SentryDatabaseExecutor
 
   @internal
   // ignore: public_member_api_docs
-  static const dbSystemKey = 'db.system';
+  static const dbSystemKey = 'db.system.name';
```
denrase:
```suggestion
  static const dbSystemNameKey = 'db.system.name';
```

or remove and use the correct seamntic attibute.

packages/sqflite/test/sentry_batch_test.dart
```diff
@@ -502,7 +502,7 @@ SELECT * FROM Product''';
       await batch.apply();
 
       final span = fixture.tracer.children.last;
-      expect(span.data[SentryDatabase.dbSystemKey], SentryDatabase.dbSystem);
+      expect(span.data[SentryDatabase.dbSystemKey], 'sqlite');
```
denrase: Dito, why remove the constant and repeat `sqlite` multiple times?

packages/supabase/lib/src/constants.dart
```diff
@@ -2,3 +2,6 @@ const String integrationNameTracing = 'sentrySupabaseTracing';
 const String integrationNameErrors = 'sentrySupabaseErrors';
 const String integrationNameBreadcrumbs = 'sentrySupabaseBreadcrumbs';
 const String loggerName = 'sentry_supabase';
+
+/// The `db.system.name` value for PostgreSQL databases.
+const String dbSystemPostgresql = 'postgresql';
```
denrase:
```suggestion
const String dbSystemNamePostgresql = 'postgresql';
```

denrase:
@buenaflor  Read https://github.com/getsentry/sentry-dart/pull/3807 after this PR, ignore the constant comments 👍

packages/drift/test/sentry_drift_test.dart
buenaflor: this is a test assertion improvement - going forward we should not use the (same) constants in tests but rather assert the value directly. this could lead to cases where we accidentally change the constant and tests still pass which are at the end useless tests then (tautological tests)

buenaflor: rule added here: https://github.com/getsentry/sentry-dart/pull/3807/

packages/flutter/lib/src/integrations/native_app_start_handler_v2.dart
buenaflor: they are yes, I'm following the way relay defines these attributes that are not yet in sentry-conventions but will be

packages/drift/test/spanv2_integration_test.dart
buenaflor: replied in other comment

packages/sqflite/test/sentry_batch_test.dart
buenaflor: see my other comment

packages/file/lib/src/sentry_file.dart
```diff
@@ -439,8 +439,8 @@ class SentryFile implements File {
     breadcrumbData['file.async'] = true;
```
buenaflor: we will remove `file.async` in the next PR

packages/supabase/lib/src/sentry_supabase_request.dart
```diff
@@ -188,7 +188,9 @@ class SentrySupabaseRequest {
     final body = this.body;
     switch (operation) {
       case Operation.select:
-        return 'SELECT * FROM "$table"';
+        final whereClause = _buildWhereClause();
+        return 'SELECT ${_buildSelectColumns()} FROM "$table"'
+            '${whereClause.isNotEmpty ? ' WHERE $whereClause' : ''}';
```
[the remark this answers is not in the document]
buenaflor: Good catch — this is a pre-existing flaw (`_getOperatorSql` had no `not` case and fell through to `=`, so it already affected UPDATE/DELETE on v10). The SELECT WHERE completion that exposed it has been moved out of this PR into #3817, and the negation is fixed there in 3dba0a8 — `.not(column, op, value)` now renders as `NOT (column <op> ?)` with a regression test. This PR no longer adds a WHERE clause to SELECT, so the finding no longer applies here.

denrase (approved):

PR #3806, opened 2026-06-30
feat: Add feature flags to hub span

### LINKED ISSUE #3766: Feature flags not tracked on spans (spec gap)
`FeatureFlagsIntegration` only records flag evaluations to the current scope
(error events via `event.contexts.flags`). It does not record flags as span
attributes, which is required by the stable SDK spec.

**Current behavior**

`addFeatureFlag()` in [`feature_flags_integration.dart`](https://github.com/getsentry/sentry-dart/blob/HEAD/packages/dart/lib/src/feature_flags_integration.dart)
writes flags into `SentryFeatureFlags` on the scope only. There is no
equivalent of `_INTERNAL_addFeatureFlagToActiveSpan` found in the JS SDK.

**Gap**

The [feature flags SDK spec](https://github.com/getsentry/sentry-docs/blob/HEAD/develop-docs/sdk/foundations/client/integrations/feature-flags.mdx)
(v1.0.1, stable) requires:

> When tracking feature flag evaluations on spans, we track the first 10
> feature flags evaluated within the span's scope. Evaluations are span
> attributes and follow the existing span attribute schema.

Expected attribute format: `flag.evaluation.<name>` (bool), max 10 per span.

**What needs to change**

- On each `addFeatureFlag` call, also set `flag.evaluation.<flagName>` on the
  active span (if one exists).
- Cap at 10 unique flags per span; update in-place if the flag already exists.
- Non-boolean values remain a no-op (already enforced for scope path).

denrase:
## :scroll: Description

Adds feature flag evaluations to the currently active span through the hub when using Sentry.addFeatureFlag(...), without exposing feature flag APIs on spans directly.

Closes #3766 

Unit tests.

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes


packages/dart/lib/src/telemetry/span/recording_sentry_span_v2.dart
```diff
@@ -1,5 +1,10 @@
 part of 'sentry_span_v2.dart';
 
+const _spanFeatureFlagPrefix = 'flag.evaluation.';
```
buenaflor: pls add this to `SemanticAttributesConstants` instead, feel free to make it a function there

packages/dart/lib/src/telemetry/span/sentry_span_v2.dart
```diff
@@ -65,4 +65,10 @@ sealed class SentrySpanV2 {
 
   /// Removes an attribute on this span with a matching key.
   void removeAttribute(String key);
+
+  /// Adds or updates a feature flag evaluation on this span.
+  void addFeatureFlag(String flag, bool result);
+
+  /// Removes a feature flag evaluation from this span.
+  void removeFeatureFlag(String flag);
```
buenaflor: is that by spec? double check if JS also has this on their public API

buenaflor: I'm not sure if we need this, we can just use the setAttributes directly from within the feature flags integration

denrase: We are also adding it to sentry-cocoa, but we could also do this in a separate PR for the whole API surface.

buenaflor: generally I would opt-out of adding a separate API on the spans (for now) - we can always do it later but right now I'd not add it and only set it through the feature flags integration

denrase: Re-checking with the team.

denrase: Removed this for now, we can add it later.

packages/dart/test/feature_flags_integration_test.dart
```diff
@@ -3,143 +3,387 @@ library;
...
+
+      sut.call(fixture.hub, fixture.options);
+
+      fixture.hub.activeSpan = parent;
+      await sut.addFeatureFlag('parent', true);
+      fixture.hub.activeSpan = child;
+      await sut.addFeatureFlag('child', true);
+
+      expect(parent.attributes, isNot(contains('flag.evaluation.child')));
+      expect(child.attributes, isNot(contains('flag.evaluation.parent')));
```
buenaflor, line 385: can you double check if this behaviour is correct? I am not sure and the devdocs doesn't really specify it (the same for spanv2)

packages/dart/lib/src/feature_flags_integration.dart
```diff
@@ -35,7 +50,81 @@ class FeatureFlagsIntegration extends Integration<SentryOptions> {
...
+    if (activeSpan != null) {
+      _addFeatureFlagToActiveSpanV2(activeSpan, flag, result);
+      return;
+    }
+
+    final activeLegacySpan = hub.getSpan();
+    if (activeLegacySpan != null) {
+      _addFeatureFlagToActiveLegacySpan(activeLegacySpan, flag, result);
+    }
+  }
```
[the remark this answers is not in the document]
buenaflor: I'm not sure if this is an actual issue

buenaflor:
There are some failing tests

buenaflor (approved):

PR #3821, opened 2026-07-01
feat: Report blocked_main_thread on streaming spans

buenaflor:
`ThreadInfoIntegration` only handled v1 spans (`OnSpanStart`/`OnSpanFinish`), so streaming/v2 spans received no `thread.id`/`thread.name` and never got `blocked_main_thread`. The internal `sync` marker set by the `file`/`hive`/`isar` instrumentation was also written onto v2 spans but never consumed there, so it leaked to Sentry.

This brings v2/streaming spans to parity with v1. `call()`/`close()` branch on `options.traceLifecycle`:

- **static** — `OnSpanStart`/`OnSpanFinish` (unchanged).
- **stream** — `OnSpanStartV2` sets `thread.id`/`thread.name`; `OnProcessSpan` derives `blocked_main_thread` from the sync marker when on the main isolate, then strips the marker.

The responsibilities stay clean: instrumentation states a fact it knows (the op is synchronous), and `ThreadInfoIntegration` — which knows the isolate — is the sole setter of `blocked_main_thread`. The previously raw `'sync'` data key is wrapped in a small set of internal extensions so the key lives in one place:

- `InstrumentationSpan.markSynchronous()` — used by `file`/`hive`/`isar`.
- `SentrySpan.isSynchronous` / `clearSynchronous()` (v1) and `SentrySpanV2.isSynchronous` / `clearSynchronous()` (v2) — used by the integration.

No new state is added to the span classes. Behavior is unchanged on v1.

### DIFF
--- a/packages/dart/lib/src/tracing/instrumentation/instrumentation_span.dart
+++ b/packages/dart/lib/src/tracing/instrumentation/instrumentation_span.dart
@@ -3,41 +3,6 @@ import 'package:meta/meta.dart';
 import '../../../sentry.dart';
 import '../../utils/internal_logger.dart';
 
-/// Data key marking a span as wrapping a synchronous operation.
-///
-/// This is an internal signal between auto-instrumentation (which knows an
-/// operation ran synchronously) and the Flutter `ThreadInfoIntegration` (which
-/// knows the isolate). The integration consumes it to derive
-/// `blocked_main_thread` and strips it, so it is not sent to Sentry.
-const _synchronousAttributeKey = 'sync';
-
-/// Marks a span as wrapping a synchronous operation.
-@internal
-extension SynchronousInstrumentationSpan on InstrumentationSpan {
-  /// Flags this span as a synchronous operation for main-thread detection.
-  void markSynchronous() => setData(_synchronousAttributeKey, true);
-}
-
-/// Reads and clears the synchronous marker on a v1 span.
-@internal
-extension SynchronousSentrySpan on SentrySpan {
-  /// Whether this span was flagged as a synchronous operation.
-  bool get isSynchronous => data[_synchronousAttributeKey] == true;
-
-  /// Removes the synchronous marker so it is not sent to Sentry.
-  void clearSynchronous() => removeData(_synchronousAttributeKey);
-}
-
-/// Reads and clears the synchronous marker on a v2 span.
-@internal
-extension SynchronousSentrySpanV2 on SentrySpanV2 {
-  /// Whether this span was flagged as a synchronous operation.
-  bool get isSynchronous => attributes[_synchronousAttributeKey]?.value == true;
-
-  /// Removes the synchronous marker so it is not sent to Sentry.
-  void clearSynchronous() => removeAttribute(_synchronousAttributeKey);
-}
-
 /// Opaque span handle enabling swappable tracing backends.
 @internal
 abstract class InstrumentationSpan {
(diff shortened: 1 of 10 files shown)


packages/dart/lib/src/tracing/instrumentation/instrumentation_span.dart
```diff
@@ -3,6 +3,41 @@ import 'package:meta/meta.dart';
...
+/// Marks a span as wrapping a synchronous operation.
+@internal
+extension SynchronousInstrumentationSpan on InstrumentationSpan {
+  /// Flags this span as a synchronous operation for main-thread detection.
+  void markSynchronous() => setData(_synchronousAttributeKey, true);
+}
+
+/// Reads and clears the synchronous marker on a v1 span.
+@internal
+extension SynchronousSentrySpan on SentrySpan {
```
denrase: This is the `instrumentation_span.dart`. Extensions specific to other spans (SentrySpan, SentrySpanV2) should be moved to their files.

packages/flutter/test/integrations/thread_info_integration_test.dart
```diff
@@ -273,20 +259,208 @@ void main() {
...
+            isFalse);
+      });
+
+      test('does not set blocked_main_thread when sync is false', () async {
+        final hub = Hub(fixture.options);
+        final integration = fixture.getSut();
+        integration.call(hub, fixture.options);
+
+        final span = fixture.startStreamingSpan(hub);
+        span.setAttribute('sync', SentryAttribute.bool(false));
```
denrase, line 433: AFAIK, If we set the `sync` attribute, we also need to verify that it was removed.

denrase (approved):
LGTM, smaller feedback notes.

packages/flutter/lib/src/integrations/thread_info_integration.dart
```diff
@@ -32,63 +32,103 @@ class ThreadInfoIntegration implements Integration<SentryFlutterOptions> {
...
+      span.setData(SemanticAttributesConstants.threadId,
+          isolateName.hashCode.toString());
+      span.setData(SemanticAttributesConstants.threadName, isolateName);
     }
   }
 
   void _processSyncSpanOnFinish(OnSpanFinish event) {
     final span = event.span;
-    if (span is! SentrySpan) {
+    if (span is! SentrySpan || !span.isSynchronous) {
```
denrase: In the case we have `sync: false` by leaving early here, we will not clean up the flag, is this intended?

denrase: This is a regression from before, where we would always remove `sync` flags on static flags. Please verify if this is intendend.

denrase (changes requested):

packages/flutter/lib/src/integrations/thread_info_integration.dart
buenaflor: we should actually remove the sync flag in general, it's merely meant as an internal data object for this thread info integration - there's no product around `sync`

packages/dart/lib/src/tracing/instrumentation/instrumentation_span.dart
buenaflor: I would maybe move this to their own file then, I think grouping them is a bit clearer than spreading them across

buenaflor: I moved them to their own file `synchronous_span_marker.dart`

packages/flutter/test/integrations/thread_info_integration_test.dart
buenaflor: added tests for this

denrase (approved):

PR #3840, opened 2026-07-01
feat(tracing): Simplify span v2 status to ok/error

### LINKED ISSUE #3839: Simplify span status v2 to ok/error and define idle span finish semantics
### Background

The internal `SentrySpanStatusV2` enum currently carries four values: `ok`, `error`, `cancelled`, and `deadlineExceeded`. Spans in the v2 telemetry pipeline should only distinguish success from failure; the extra states don't map cleanly to the streaming span model and diverge from how other SDKs report idle-span outcomes.

### Desired behavior

Reduce `SentrySpanStatusV2` to `ok` and `error`, and convey the finer-grained outcome through semantic attributes instead:

| Scenario | Status | Attribute |
|---|---|---|
| Idle span final timeout | `error` | `sentry.status.message` = `deadline_exceeded` |
| Idle span idle timeout | `ok` | — |
| Still-active children force-finished on timeout | `ok` | — |
| External cancel (navigation route change, user-interaction different-widget tap) | `ok` | `sentry.idle_span_finish_reason` = `cancelled` |

The `sentry.idle_span_finish_reason` attribute mirrors the `browserTracingIntegration` behavior in the JS SDK.

### Notes

`SentrySpanStatusV2` is `@internal` and not exported from any public barrel, so this is not a breaking change for SDK users. The v1 `SpanStatus` (with `cancelled()` / `deadlineExceeded()`) is a separate protocol type and is unaffected.

buenaflor:
Closes #3839

`SentrySpanStatusV2` now only distinguishes `ok` from `error`. The `cancelled` and `deadlineExceeded` values are removed, and the finer-grained outcome of a v2 idle span moves onto semantic attributes set when the span finishes:

| Scenario | Status | Attribute |
|---|---|---|
| Final timeout | `error` | `sentry.status.message` = `deadline_exceeded` |
| Idle timeout | `ok` | — |
| Still-active children force-finished on timeout | `ok` | — |
| External cancel (navigation route change, user-interaction different-widget tap) | `ok` | `sentry.idle_span_finish_reason` = `cancelled` |

`sentry.idle_span_finish_reason` mirrors the JS SDK's `browserTracingIntegration`. Two new `@internal` `SemanticAttributesConstants` back these attributes: `sentryStatusMessage` and `sentryIdleSpanFinishReason`.

`SentrySpanStatusV2` is `@internal` and not exported from any public barrel, so SDK users are unaffected. The v1 `SpanStatus` (`cancelled()` / `deadlineExceeded()`) is a separate protocol type and is left untouched — its use in gRPC and the v1 navigation trackers is unchanged.

### DIFF
--- a/packages/dart/lib/src/telemetry/span/sentry_span_status_v2.dart
+++ b/packages/dart/lib/src/telemetry/span/sentry_span_status_v2.dart
@@ -1,7 +1,5 @@
 enum SentrySpanStatusV2 {
   error('error'),
-  cancelled('cancelled'),
-  deadlineExceeded('deadline_exceeded'),
   ok('ok');
 
   final String value;
--- a/packages/dart/lib/src/tracing/instrumentation/instrumentation_span.dart
+++ b/packages/dart/lib/src/tracing/instrumentation/instrumentation_span.dart
@@ -203,22 +203,12 @@ class StreamingInstrumentationSpan implements InstrumentationSpan {
         return SpanStatus.ok();
       case SentrySpanStatusV2.error:
         return SpanStatus.unknownError();
-      case SentrySpanStatusV2.cancelled:
-        return SpanStatus.cancelled();
-      case SentrySpanStatusV2.deadlineExceeded:
-        return SpanStatus.deadlineExceeded();
     }
   }
 
   SentrySpanStatusV2 _convertToV2Status(SpanStatus status) {
     if (status == SpanStatus.ok()) {
       return SentrySpanStatusV2.ok;
-    }
-    if (status == SpanStatus.cancelled()) {
-      return SentrySpanStatusV2.cancelled;
-    }
-    if (status == SpanStatus.deadlineExceeded()) {
-      return SentrySpanStatusV2.deadlineExceeded;
     }
     return SentrySpanStatusV2.error;
   }
--- a/packages/dart/lib/src/constants.dart
+++ b/packages/dart/lib/src/constants.dart
@@ -104,6 +104,14 @@ abstract class SemanticAttributesConstants {
   /// The operation name of a span.
   static const sentryOp = 'sentry.op';
 
+  /// A human-readable message providing additional context about a span's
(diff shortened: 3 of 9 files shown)


packages/dart/lib/src/telemetry/span/sentry_span_status_v2.dart
```diff
@@ -1,7 +1,5 @@
 enum SentrySpanStatusV2 {
```
denrase, line 1: If what warden says is correct, we should at least mark it now as internal, no?

packages/dart/lib/src/tracing/instrumentation/instrumentation_span.dart
```diff
@@ -203,23 +203,13 @@ class StreamingInstrumentationSpan implements InstrumentationSpan {
...
         return SpanStatus.unknownError();
-      case SentrySpanStatusV2.cancelled:
-        return SpanStatus.cancelled();
-      case SentrySpanStatusV2.deadlineExceeded:
-        return SpanStatus.deadlineExceeded();
     }
   }
 
   SentrySpanStatusV2 _convertToV2Status(SpanStatus status) {
     if (status == SpanStatus.ok()) {
```
denrase, line 210: Are there now cases in `SpanStatus` that would also match to error, like `deadlineExceeded`?

packages/dart/lib/src/telemetry/span/sentry_span_status_v2.dart
buenaflor: imo it's okay, span-streaming is not yet GA (no docs, etc...)

packages/dart/lib/src/tracing/instrumentation/instrumentation_span.dart
buenaflor: yeah, which is fine imo. I dont expect this to be perfect now

denrase (approved):

PR #3842, opened 2026-07-01
ref(logs): Make traceId required on SentryLog

### LINKED ISSUE #3835: [v10] Make `traceId` required on `SentryLog`
## Context
Part of the [Major v10 Task Collection](https://github.com/getsentry/sentry-dart/issues/3487).

## Problem
`SentryLog.traceId` is nullable and defaults to `SentryId.empty()`, so logs can ship without a real trace association. There is an existing `TODO(major-v10)`.

## Expected outcome
- `traceId` becomes non-nullable and a required constructor parameter.
- The `SentryId.empty()` fallback is removed.

## Pointers
- `packages/dart/lib/src/telemetry/log/log.dart:7-27`

**Breaking change:** Compile-time for direct `SentryLog` constructors (mostly internal callers).

buenaflor:
`SentryLog.traceId` is now a required, non-nullable constructor parameter. It was previously nullable and defaulted to `SentryId.empty()`, so a log could be built without a real trace association and only got patched later in the capture pipeline.

With the id guaranteed at construction, the pipeline's compensating "populate traceId from scope if empty" block is removed; the scope attribute merge is unchanged. The only production caller (`DefaultSentryLogger`) already passes `propagationContext.traceId`, so runtime behavior is unaffected.

Part of the v10 cleanup (#3487).

**Breaking change:** compile-time — constructing `SentryLog` without a `traceId` no longer compiles. Callers are mostly internal.

Fixes #3835


packages/dart/lib/src/telemetry/log/log_capture_pipeline.dart
```diff
@@ -22,11 +22,6 @@ class LogCapturePipeline {
 
     try {
       if (scope != null) {
-        // Populate traceId from scope if not already set
-        // TODO(major-v10): this can be removed once we make the traceId required on the log
-        if (log.traceId == null || log.traceId == SentryId.empty()) {
-          log.traceId = scope.propagationContext.traceId;
```
denrase, line 28: Just want to understand, was this dead code before? I do not see any change in how the log.traceId is bein set at init time in this pr.

buenaflor: it was pretty much since we always created a log with a traceId already

denrase (approved):

PR #3567, opened 2026-03-17
feat: Implement strict trace continuation

antonis:
## :scroll: Description

Implement [Strict Trace Continuation](https://develop.sentry.dev/sdk/foundations/trace-propagation/#strict-trace-continuation) in the Dart SDK.

- Add `strictTraceContinuation` (bool) and `orgId` (String?) options on `SentryOptions`
- Extract org_id from DSN host; `effectiveOrgId` prefers explicit over DSN (trims whitespace)
- Propagate `sentry-org_id` in baggage and trace context header
- Validate incoming traces in `SentryTransactionContext.fromSentryTrace` per the spec decision matrix
- Pass options through to the native Android SDK on init

## :bulb: Motivation and Context

Prevents cross-organization trace continuation by validating org IDs in distributed traces.

Part of the [Strict Trace Continuation initiative](https://linear.app/getsentry/initiative/strict-trace-continuation-575e9631fd98).

### Platform coverage

| Platform | Coverage |
|----------|----------|
| **Dart** | Full |
| **Android** | Full — via sentry-android 8.39.1 ([sentry-java#5136](https://github.com/getsentry/sentry-java/pull/5136)) |
| **iOS** | Dart layer only — native support requires Sentry Cocoa v9 ([#7705](https://github.com/getsentry/sentry-cocoa/pull/7705)); Flutter is pinned to `Sentry/HybridSDK 8.x` |

## :green_heart: How did you test it?

46 new unit tests covering DSN parsing, option resolution, decision matrix, baggage/header propagation, and `fromSentryTrace` validation.

## :pencil: Checklist
- [x] I reviewed submitted code
- [x] I added tests to verify changes
- [x] No new PII added or SDK only sends newly added PII if `sendDefaultPii` is enabled
- [ ] I updated the docs if needed
- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps
- Follow-up: enable native iOS coverage once Flutter migrates to Sentry Cocoa v9
- Docs PR: https://github.com/getsentry/sentry-docs/pull/16986

### Changelog Entry

- Prevent cross-organization trace continuation
  - By default, the SDK now extracts the organization ID from the DSN (e.g. `o123.ingest.sentry.io`) and compares it with the `sentry-org_id` value in incoming baggage headers. When the two differ, the SDK starts a fresh trace instead of continuing the foreign one. This guards against accidentally linking traces across organizations.
[...]

PR #3845, opened 2026-07-01
ref: Make clone() internal and remove deprecated protocol clones

### LINKED ISSUE #3824: [v10] Make `clone()` internal
## Context
Part of the [Major v10 Task Collection](https://github.com/getsentry/sentry-dart/issues/3487).

## Problem
`clone()` is public on `Hub`, `Scope`, `NoOpHub`, and many protocol classes, but it is only intended for SDK-internal use. `Contexts.clone()` is already `@Deprecated`.

## Expected outcome
- `clone()` is marked `@internal` (deprecate-then-internal where appropriate) across Hub/Scope/NoOpHub and protocol classes.
- The SDK is audited to confirm no public API path depends on user-callable `clone()`.

## Pointers
- `packages/dart/lib/src/hub.dart:435`
- `packages/dart/lib/src/scope.dart:469`
- `packages/dart/lib/src/noop_hub.dart`
- `packages/dart/lib/src/protocol/contexts.dart:326` and other protocol classes

**Breaking change:** Medium — only affects users calling `clone()` directly (expected to be rare).

buenaflor:
Two related `clone()` cleanups for v10:

- **`Hub`/`Scope`/`NoOpHub`/`HubAdapter`/`Sentry.clone()`** are now `@internal`. They exist only for SDK-internal scope/hub duplication; no user or cross-package code calls them.
- **The protocol `clone()` methods** (`Contexts` and its `device`/`os`/`app`/`browser`/`culture`/`gpu`/`runtime`/`response`/`trace`/`feedback`/`flags` children) are **removed**. They were already `@Deprecated`, and formed a dead subtree — only called from within `Contexts.clone()`, which itself had no remaining callers. The alias-only "clone context" test is dropped.
- **`SentryFeatureFlag(s).copy()`** is renamed to `clone()` (review feedback) — the `copy` name only existed to avoid clashing with the deprecated public `clone()`, which this PR removes.

Part of the v10 cleanup (#3487).

**Breaking change:** deprecated protocol `clone()` is removed (migration: assign directly / use `copyWith` where still present); `Hub`/`Scope.clone()` become `@internal` — low impact, direct calls were rare.

Also carries two v10-branch CI unblockers that pre-date this PR (every v10 PR currently fails on them):

- `thread_info_integration_test.dart` still called `OnProcessSpan(span)` with one argument after #3847 made `hint` required — the flutter test suite didn't compile (analyze + js/linux/macos jobs). The five call sites now pass `Hint()`.
- The Android SDK-metrics binary-size diff (~1409 KiB) outgrew the 1350 KiB budget; `diffMax` is raised to 1450 KiB, same pattern as #3324 / #3567.

Fixes #3824


packages/dart/lib/src/protocol/sentry_feature_flag.dart
```diff
@@ -55,7 +55,4 @@ class SentryFeatureFlag {
       unknown: unknown ?? this.unknown,
     );
   }
-
-  @Deprecated('Will be removed in a future version.')
-  SentryFeatureFlag clone() => copyWith();
```
denrase, line 60: On hub we made clone internal, here we have copy. We should aling naming.

denrase (changes requested):

denrase (approved):

Issue #3767, opened 2026-06-16
Add API to extend app start (Flutter)


buenaflor:
Add an API to the Sentry Flutter SDK to extend the app start measurement past its default end point and instrument the extended region with child spans.

```dart
// Simple: extend the app start window, then finish
SentryFlutter.extendAppStart();
await initializeRemoteConfig();
SentryFlutter.finishExtendedAppStart();
```

```dart
// With child spans: instrument the extended region
SentryFlutter.extendAppStart();
final span = Sentry.getExtendedAppStartSpan();

final configSpan = span.startChild('app.init', description: 'fetch remote config');
await fetchRemoteConfig();
await configSpan.finish();

SentryFlutter.finishExtendedAppStart();
```

**Reference:** getsentry/sentry-cocoa#6886 (Done, `SentryExtendedAppLaunchManager`). **Siblings:** getsentry/sentry-java#5553, getsentry/sentry-react-native#6303 — shared contract, per-platform deltas below.

## API

* `extendAppStart(): void` — creates the extended app start span (op `app.start.extended`, description `"Extended App Start"`).
* `finishExtendedAppStart()` — finishes it (returns void); triggers finalization.
* `getExtendedAppStartSpan(): Span` — returns the span for attaching children; a no-op span when there's no active extension.

> `finishExtendedAppStart()` is equivalent to `.finish()` on the returned span.

## Lifecycle

* `extendAppStart`**:** creates the span (start = call time). No-ops if app start already finished, if none is in progress, or on repeat calls (first wins).
* `getExtendedAppStartSpan`**:** returns the extended span, else a no-op span.
* `finishExtendedAppStart`**:** no-ops if not extended or called twice. Doesn't finalize directly — the parent (auto-generated, `waitForChildren`) finalizes when its last child finishes. If the child never finishes the final deadline will

## Duration & measurement

* Extending can only push the end *later*; finishing before the default end keeps the default (never shorter than a non-extended app start). The extended span is just one participant.

* App start measurement = that final duration (process start → finish extended app start time), set at **finalization**. The normal completion-time measurement must be suppressed in extended mode so the early (non-extended) value doesn't win.

## Timeout

[...]

PR #3853, opened 2026-07-02
feat(flutter): Add standalone app start tracing

### LINKED ISSUE #3634: feat: Send standalone app start transactions [Flutter]
### Problem

Currently, app start data is captured directly within the first ``` ui``.load ``` (screen load) transaction. If no qualifying transaction starts, app start data can be lost entirely.

The goal is to send app start as its own standalone transaction for cleaner separation, and to eventually allow capturing app starts that don't necessarily have a UI (e.g. background services).

### Proposal

Instead of attaching app start spans and measurements to the `ui.load` transaction, the SDK should send a standalone app start transaction as soon as app start data is available. Flutter is a hybrid SDK: it pulls raw app start timing from the native layer (via the native `getAppStartMeasurement()` bridge) and constructs the transaction in the Dart layer. The native `enableStandaloneAppStartTracing` must therefore stay inactive for hybrid consumers, to avoid duplicate transactions.

This transaction carries the same timing and metadata, but is independent of any screen transaction.

### Span spec (Span V2 / EAP)

* **Transaction / span op:** `app.start`
* **Span name / description:** `App Start`
* **Attributes:**
  * `app.vitals.start.value` — app start duration
  * `app.vitals.start.type` — `cold` / `warm`

This is the V2/EAP attribute-based encoding. Relay backfills `app.vitals.start.value` / `app.vitals.start.type` (and `app.vitals.start.screen`) for V1 transactions from the legacy `app_start_cold` / `app_start_warm` measurements and transaction name, so the SDK should emit the V2 attributes directly. Do **not** use the legacy per-type op (`app.start.cold` / `app.start.warm`).

* **Rollout:** Opt-in via `enableStandaloneAppStartTracing` (experimental).

### Reference

Android (getsentry/sentry-java#5046) and Cocoa (getsentry/sentry-cocoa#6883) shipped the standalone-transaction behavior but using the legacy per-type op encoding (`app.start.cold` / `app.start.warm`, `app_start_*` measurements, `app.start_type` context). Flutter should follow the same standalone-transaction approach but adopt the V2 `app.start` + `app.vitals.start.*` spec above.

Related: [https://github.com/getsentry/sentry-dart/issues/3767](<https://github.com/getsentry/sentry-dart/issues/3767>)

buenaflor:
Adds opt-in standalone native app-start tracing for Flutter. With
`options.enableStandaloneAppStartTracing` enabled, native app start is emitted
as a detached `App Start` transaction with op `app.start` and origin
`auto.app.start`; with the flag disabled, native app start remains attached to
the initial `ui.load` transaction.

## Architecture

This keeps each telemetry signal under one owner:

- `TimeToDisplayTracker` / `TimeToDisplayTrackerV2` own the initial `ui.load`
  root, TTID/TTFD, `currentDisplay`, and `reportFullyDisplayed` compatibility.
- `StaticStandaloneAppStartEmitter` / `StreamStandaloneAppStartEmitter` own
  only the detached `App Start` / `app.start` root.
- `NativeAppStartIntegration` owns native source orchestration: first-frame
  callback registration, duplicate processing guard, native fetch/parse, display
  branch invocation, standalone branch invocation, and failure cleanup.
- App-start span writers own formatting only: legacy attached measurement and
  breakdown shape, standalone breakdown shape, app-start type data/attributes,
  and finalization-time app-start vitals encoding.

In standalone mode, the `ui.load` root still tracks TTID/TTFD but no longer
carries the app-start measurement or breakdown spans. The standalone root keeps
the current public shape:

```text
App Start                                 op: app.start
├─ <native breakdown spans>               op: app.start.native
├─ App start to plugin registration       op: app.start.plugin_registration
├─ Before Sentry Init Setup               op: app.start.sentry_setup
└─ First frame render                     op: app.start.first_frame_render
```

The static standalone root deliberately does not use `trimEnd`: the explicit
native app-start end timestamp is authoritative, and trimming could let an
out-of-range native span stretch the transaction duration beyond the app-start
measurement.

## Scope

This PR is native app start only. Generic standalone app start remains future
work and should reuse the same display-tracker entrypoints and standalone
emitter seam when conventions can represent its app-start type honestly.

#3767 extended app start remains follow-up work. The emitter seam is intended to
support that later by letting the standalone finalization policy evolve; a
session/tracker should be added then only if it owns real lifecycle state such
as extension first-wins behavior, deadline handling, or delayed finalization.

Fixes #3634
Refs #3767


buenaflor:
Will be replaced by another impl

PR #3859, opened 2026-07-06
fix(replay): Accept Double timestamps in Android network breadcrumb conversion

aqrc:
## :scroll: Description

Fixes Session Replay being broken on Android since `9.21.0`: replays were either missing entirely or uploaded as short glitchy fragments.

`SentryFlutterReplayBreadcrumbConverter.convertNetworkBreadcrumb` cast `start_timestamp` / `end_timestamp` with an unchecked `as Long`. Since #3708, breadcrumbs are synced to Android as JSON bytes and re-parsed by sentry-java's `Breadcrumb.Deserializer`, whose `JsonObjectDeserializer.nextNumber()` tries `nextInt()` → `nextDouble()` → `nextLong()`. Epoch-millisecond values overflow `Integer`, so they are parsed as `Double` and `nextLong()` is never reached. The cast then threw for every `http` breadcrumb:

```
E/Sentry: Failed to execute task SessionCaptureStrategy.add_frame
E/Sentry: java.lang.ClassCastException: java.lang.Double cannot be cast to java.lang.Long
	at io.sentry.flutter.SentryFlutterReplayBreadcrumbConverter.convertNetworkBreadcrumb(SentryFlutterReplayBreadcrumbConverter.kt:84)
	at io.sentry.flutter.SentryFlutterReplayBreadcrumbConverter.convert(SentryFlutterReplayBreadcrumbConverter.kt:30)
	at io.sentry.android.replay.capture.CaptureStrategy$Companion.buildReplay(CaptureStrategy.kt:200)
	at io.sentry.android.replay.capture.CaptureStrategy$Companion.createSegment(CaptureStrategy.kt:113)
```

Because the exception is thrown inside replay segment creation, any segment containing an `http` breadcrumb was dropped — in apps with regular network traffic this effectively disables Session Replay on Android.

This change accepts any `Number` for the timestamps and returns `null` when `url` or the timestamps are missing or mistyped instead of throwing, matching what the iOS converter (`convertNetwork:` in `SentryFlutterReplayBreadcrumbConverter.m`) already does with its `isKindOfClass:` guards.

Replay worked on `9.20.0` (typed method-channel scope sync preserved `Long`) and broke on `9.21.0`+ for Android. The same numeric-subtype issue was fixed in sentry-react-native's converter in getsentry/sentry-react-native#6288.

- Reproduced the `ClassCastException` against the unmodified converter by invoking `convert()` with an `http` breadcrumb whose `start_timestamp`/`end_timestamp` are `Double` (compiled against `sentry-android 8.47.0`).
- Verified the fixed converter with the same harness:
  - `Double` timestamps → `RRWebSpanEvent` with correct second-precision start/end timestamps
  - `Long` timestamps (pre-9.21 path) → unchanged behavior
  - missing `url` or non-numeric timestamp → returns `null` instead of throwing
- `ktlint` passes.
- Reproduced the problem in the app, then verified that it disappeared with this fix.

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps

The root cause — sentry-java's `JsonObjectDeserializer.nextNumber()` returning `Double` for any value that overflows `Integer` — may affect other consumers of JSON-synced scope data and could be worth fixing upstream in sentry-java as well.

### DIFF
--- a/packages/flutter/android/src/main/kotlin/io/sentry/flutter/SentryFlutterReplayBreadcrumbConverter.kt
+++ b/packages/flutter/android/src/main/kotlin/io/sentry/flutter/SentryFlutterReplayBreadcrumbConverter.kt
@@ -71,25 +71,32 @@ class SentryFlutterReplayBreadcrumbConverter : DefaultReplayBreadcrumbConverter(
   private fun doubleTimestamp(timestamp: Long) = timestamp / MILLIS_PER_SECOND
 
   private fun convertNetworkBreadcrumb(breadcrumb: Breadcrumb): RRWebEvent? {
-    var rrWebEvent = super.convert(breadcrumb)
-    if (rrWebEvent == null &&
-      breadcrumb.data.containsKey("start_timestamp") &&
-      breadcrumb.data.containsKey("end_timestamp")
-    ) {
-      rrWebEvent =
-        RRWebSpanEvent().apply {
-          op = "resource.http"
-          timestamp = breadcrumb.timestamp.time
-          description = breadcrumb.data["url"] as String
-          startTimestamp = doubleTimestamp(breadcrumb.data["start_timestamp"] as Long)
-          endTimestamp = doubleTimestamp(breadcrumb.data["end_timestamp"] as Long)
-          data =
-            breadcrumb.data
-              .filterKeys { key -> supportedNetworkData.containsKey(key) }
-              .mapKeys { (key, _) -> supportedNetworkData[key] }
-        }
+    val rrWebEvent = super.convert(breadcrumb)
+    if (rrWebEvent != null) {
+      return rrWebEvent
     }
-    return rrWebEvent
+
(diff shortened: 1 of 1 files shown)


aqrc:
hey, sorry in advance that this PR is not impeccable, I've been looking into this problem with the help of Claude then asked if he "can fix it and create a PR" theoretically, but he actually went full throttle 😬 The problem is real and the fix should be working too, but I am going to manually review and test everything now

buenaflor:
@aqrc thx! ignore the CI failures, the e2e test is not running due to missing auth permissions since it's a fork

aqrc:
I reviewed the code - looks legit to me, only questionable thing is the comment (it is up to you to keep it or to remove it).

I also imported the fix into a real app and I confirm that Replay works correctly now.

buenaflor (approved):
thx, lgtm 👍

PR #3875, opened 2026-07-10
feat(replay): Add request details  to Session Replay on Android 

lucas-zimerman:
## 📜 Description

Adds opt-in capture of HTTP request/response headers and bodies for Session Replay network breadcrumbs, gated by `networkDetailAllowUrls`/`networkDetailDenyUrls` being non-empty, scoped to Dart-only r8equests.

`NetworkDetailsCapture` (Dart) attaches the captured data to the http breadcrumb. On Android, the replay breadcrumb converter forwards this into the `RRWebSpanEvent` unchanged, and the native `SentryReplayOptions` are synced at init so the SDK reports `networkDetailHasUrls: true` in the replay's options event — the Sentry frontend gates rendering of captured request/response detail on that flag.

iOS and Web support are not included in this PR — the iOS native replay integration doesn't yet forward this data to the replay, so no platform-specific iOS work was added.

## 💡 Motivation and Context

Session Replay currently shows network spans without request/response detail. This lets users opt in to seeing headers and bodies alongside those spans for debugging, starting with Android.

## 💚 How did you test it?

* Added unit tests for `NetworkDetailsCapture` and `BreadcrumbClient`.
* Manually verified via the example app's "Dart: Web request with network details" button on Android, confirming the captured request/response data appears in the http breadcrumb and is forwarded to the replay.

## :pencil: Checklist

- [X] All tests passing
- [X] No breaking changes

## 🔮 Next steps

* Extend native forwarding to iOS once the native replay integration supports it.

## Changelog Entry

Add HTTP request/response header and body capture for Session Replay network breadcrumbs, gated by `networkDetailAllowUrls`/`networkDetailDenyUrls`.


packages/dart/lib/src/http_client/sentry_http_client.dart
```diff
@@ -4,6 +4,7 @@ import '../hub.dart';
 import '../hub_adapter.dart';
 import 'breadcrumb_client.dart';
 import 'failed_request_client.dart';
+import 'network_details_capture.dart';
```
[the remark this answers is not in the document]
lucas-zimerman: Fixed in 83a9680e

packages/flutter/lib/src/native/java/sentry_native_java_init.dart
```diff
@@ -299,6 +299,18 @@ void configureAndroidOptions({
...
+    if (options.enableReplayNetworkDetailsCapturing) {
+      sessionReplay.setNetworkDetailAllowUrls(
+          dartToJStringList(options.networkDetailAllowUrls)..releasedBy(arena));
+      sessionReplay.setNetworkDetailDenyUrls(
+          dartToJStringList(options.networkDetailDenyUrls)..releasedBy(arena));
+      sessionReplay.setNetworkRequestHeaders(
+          dartToJStringList(options.networkRequestHeaders)..releasedBy(arena));
+      sessionReplay.setNetworkResponseHeaders(
+          dartToJStringList(options.networkResponseHeaders)..releasedBy(arena));
+      sessionReplay.setNetworkCaptureBodies(options.networkCaptureBodies);
```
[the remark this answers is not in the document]
lucas-zimerman: Fixed on 83a9680e

packages/dart/lib/src/http_client/network_details_capture.dart
```diff
@@ -0,0 +1,150 @@
...
+      return null;
+    }
+    // Only `Request` exposes its body synchronously without finalizing the
+    // request stream, which `BaseClient` implementations are only allowed to
+    // do once.
+    if (request is! Request) {
+      return null;
+    }
+    if (!_isCapturableContentType(request.headers['content-type'])) {
+      return null;
```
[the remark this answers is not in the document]
lucas-zimerman: false positive

packages/dart/lib/src/http_client/network_details_capture.dart
```diff
@@ -0,0 +1,150 @@
...
+
+    if (!_options.networkCaptureBodies ||
+        !_isCapturableContentType(response.headers['content-type'])) {
+      return (response, data);
+    }
+
+    try {
+      final bytes = await response.stream.toBytes();
+      data['body'] = _truncate(utf8.decode(bytes, allowMalformed: true));
+      return (_copyWithBytes(response, bytes), data);
```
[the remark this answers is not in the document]
lucas-zimerman: Acknowledged, not fixing: `http.Response.fromStream()` already buffers the full body for the common `get()`/`post()` path regardless, so this wouldn't lower real peak memory there. A proper fix needs a live stream-tee, which is more complexity than this opt-in, allowlist-gated debug feature warrants right now.

packages/dart/lib/src/http_client/network_details_capture.dart
```diff
@@ -0,0 +1,171 @@
...
+    }
+  }
+
+  final SentryOptions _options;
+
+  static const _defaultHeaders = ['content-type', 'content-length', 'accept'];
+
+  /// Matches native's `SentryReplayOptions.MAX_NETWORK_BODY_SIZE`.
+  static const maxBodySize = 150 * 1024;
+
```
buenaflor: nit: I know we have it in some classes constructor first but it's pretty much a Flutter specific pattern.

I think I would prefer something like fields → constructors → public API

packages/dart/lib/src/http_client/network_details_capture.dart
```diff
@@ -0,0 +1,171 @@
...
+/// Captures HTTP request/response headers and bodies for [SentryHttpClient]
+/// requests, gated by [SentryOptions.enableReplayNetworkDetailsCapturing] and
+/// [SentryOptions.networkDetailAllowUrls], so they can be shown alongside
+/// network spans in Session Replay.
+@internal
+class NetworkDetailsCapture {
+  NetworkDetailsCapture(this._options) {
+    if (_options.enableReplayNetworkDetailsCapturing) {
+      _options.sdk.addFeature(SentryFeatures.replayNetworkDetailsCapturing);
+    }
```
buenaflor: I dont think we need this flag, only cocoa has it.

other SDKs only check that the `networkDetailAllowUrls` is not empty

packages/dart/lib/src/http_client/network_details_capture.dart
```diff
@@ -0,0 +1,171 @@
...
+    // request stream, which `BaseClient` implementations are only allowed to
+    // do once.
+    if (request is! Request) {
+      return null;
+    }
+    if (!_isCapturableContentType(request.headers['content-type'])) {
+      return null;
+    }
+    try {
+      return _truncate(request.body);
```
buenaflor: we should use `request.bodyBytes` instead with `_truncateBytes`

packages/dart/lib/src/http_client/network_details_capture.dart
```diff
@@ -0,0 +1,171 @@
...
+        normalized.contains('x-www-form-urlencoded');
+  }
+
+  String _truncate(String body) {
+    final bytes = utf8.encode(body);
+    if (bytes.length <= maxBodySize) {
+      return body;
+    }
+    return utf8.decode(bytes.sublist(0, maxBodySize), allowMalformed: true);
+  }
```
buenaflor: we are always fully encoding the entire String body which is unbounded. since we have access to the bytes directly above we should modify this

```dart
  String _truncateBytes(Uint8List bytes) {
    final view = bytes.length <= maxBodySize
        ? bytes
        : Uint8List.sublistView(bytes, 0, maxBodySize);
    return utf8.decode(view, allowMalformed: true);
  }
```

packages/dart/lib/src/http_client/network_details_capture.dart
```diff
@@ -0,0 +1,171 @@
...
+      // The stream may be partially consumed at this point, so there's no
+      // safe way to forward it to the original caller anymore.
+      return (response, data);
+    }
+
+    // The body is already fully read at this point, so decoding failures
+    // (e.g. malformed encoding info) shouldn't prevent forwarding it as-is.
+    final forwarded = _copyWithBytes(response, bytes);
+    try {
+      data['body'] = _truncate(utf8.decode(bytes, allowMalformed: true));
```
buenaflor: we should use `_truncateBytes` (see other comment)

packages/dart/lib/src/http_client/network_details_capture.dart
```diff
@@ -0,0 +1,171 @@
...
+          _filterHeaders(response.headers, _options.networkResponseHeaders),
+    };
+
+    if (!_options.sendDefaultPii ||
+        !_options.networkCaptureBodies ||
+        !_isCapturableContentType(response.headers['content-type'])) {
+      return (response, data);
+    }
+
+    List<int> bytes;
```
buenaflor:
```suggestion
    Uint8List bytes;
```

packages/dart/lib/src/http_client/network_details_capture.dart
```diff
@@ -0,0 +1,171 @@
...
+    final data = <String, dynamic>{
+      'headers':
+          _filterHeaders(response.headers, _options.networkResponseHeaders),
+    };
+
+    if (!_options.sendDefaultPii ||
+        !_options.networkCaptureBodies ||
+        !_isCapturableContentType(response.headers['content-type'])) {
+      return (response, data);
+    }
```
buenaflor: we should also check `response.contentLength`

packages/dart/lib/src/http_client/network_details_capture.dart
buenaflor: should be safe to remove this flag from here, let's remove this

packages/dart/lib/src/http_client/network_details_capture.dart
lucas-zimerman: Fixed on 2f41f60b

packages/dart/lib/src/http_client/network_details_capture.dart
lucas-zimerman: Fixed on 5f60324c

packages/dart/lib/src/http_client/network_details_capture.dart
lucas-zimerman: Fixed on 5f60324c

packages/dart/lib/src/http_client/network_details_capture.dart
lucas-zimerman: Fixed on 5f60324c

packages/dart/lib/src/http_client/network_details_capture.dart
lucas-zimerman: Fixed on 5f60324c

packages/dart/lib/src/http_client/network_details_capture.dart
lucas-zimerman: Fixed on 01004074

packages/dart/lib/src/http_client/network_details_capture.dart
lucas-zimerman: I will keep that in mind for future changes!
Fixed on 0b5f2e65

lucas-zimerman:
All CI failures on this PR are pre-existing on main

packages/dart/lib/src/http_client/network_details_capture.dart
```diff
@@ -0,0 +1,170 @@
...
+    if (!_options.sendDefaultPii ||
+        !_options.networkCaptureBodies ||
+        !_isCapturableContentType(response.headers['content-type']) ||
+        (contentLength != null && contentLength > maxBodySize)) {
+      return (response, data);
+    }
+
+    Uint8List bytes;
+    try {
+      bytes = await response.stream.toBytes();
```
buenaflor: hm we are still reading the full body here. I experimented what kind of code AI will give me to optimize this and it's essentially using `StreamIterator` and `BytesBuilder`. pls have a look there maybe we can do this as a helper so we can re-use this elsewhere too, I think we ran into this problem before as well.

Here is what AI told me (summarized):

```
- StreamIterator — pull-based reading with a stopping condition. This is what lets you break out of the loop at the cap and still hand the unconsumed rest of the stream onward.
- BytesBuilder — efficiently accumulates the chunks you did read without copying on every add (with copy: false it just holds references), then takeBytes() gives you one consolidated Uint8List for the prefix.
- async* generator — yields the buffered prefix, then keeps pulling from the same iterator so the caller gets the full body streamed through.
```

packages/dart/lib/src/utils/stream_utils.dart
```diff
@@ -0,0 +1,52 @@
...
+@internal
+Future<(Uint8List prefix, Stream<List<int>> forwarded)> bufferStreamPrefix(
+  Stream<List<int>> stream, {
+  required int maxBytes,
+}) async {
+  final iterator = StreamIterator(stream);
+  final builder = BytesBuilder(copy: false);
+  List<int>? overflow;
+
+  while (builder.length < maxBytes && await iterator.moveNext()) {
```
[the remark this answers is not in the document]
lucas-zimerman: Session Replay HTTP body capture leaks into every breadcrumb-bearing event, not just replay. Bodies (up to ~150KB each) get attached to the shared http breadcrumb via `BreadcrumbClient`, so once `enableReplayNetworkDetailsCapturing` is on, that payload rides along with every subsequent error/transaction, and native scope sync means Android/iOS crash events pick it up too, not just replay recordings. A real fix needs a replay-only side channel (e.g. via Hint) with new native bridge plumbing, since nothing like that exists today. Impact is bounded (opt-in, `allowlist-gated`) 
What do you think @buenaflor? Should I solve this issue here or on a follow-up PR?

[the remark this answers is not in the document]
buenaflor: please create a separate issue then, we can keep it like it is now, so feel free to revert the changes regarding optimizing `stream.toBytes`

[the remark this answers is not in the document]
lucas-zimerman: Changes reverted and issue created: https://github.com/getsentry/sentry-dart/issues/3900

packages/dart/lib/src/http_client/breadcrumb_client.dart
```diff
@@ -60,17 +68,32 @@ class BreadcrumbClient extends BaseClient {
...
+    final capture = _networkDetailsCapture;
+    final captureNetworkDetails =
+        capture != null && capture.shouldCapture(request.url);
+    if (captureNetworkDetails) {
+      requestDetail = capture.captureRequest(request);
+    }
+
     try {
-      final response = await _client.send(request);
+      var response = await _client.send(request);
```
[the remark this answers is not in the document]
lucas-zimerman: Was fixed on da583d9b

packages/dart/lib/src/sentry_options.dart
```diff
@@ -480,6 +480,54 @@ class SentryOptions {
...
+  /// Currently only forwarded to Session Replay on Android; on other
+  /// platforms the data is captured but not yet surfaced in the replay.
+  ///
+  /// Empty by default, meaning no request is captured. Request/response
+  /// bodies may contain PII — review this list carefully before adding
+  /// entries.
+  ///
+  /// This data is attached to HTTP breadcrumbs, so [recordHttpBreadcrumbs]
+  /// must also be enabled (the default) for this option to have any effect.
+  final List<String> networkDetailAllowUrls = [];
```
buenaflor: shouldn't these options be part of `options.replay.*` namespace

lucas-zimerman: Good point! fixed on ce004caf

packages/dart/lib/src/sentry_options.dart
```diff
@@ -480,6 +480,54 @@ class SentryOptions {
...
+  /// PII.
+  final List<String> networkRequestHeaders = [];
+
+  /// Additional response header names to capture for URLs matching
+  /// [networkDetailAllowUrls].
+  ///
+  /// `Content-Type`, `Content-Length`, and `Accept` are always captured.
+  /// Any other header only takes effect when [sendDefaultPii] is also
+  /// enabled, since headers such as `Set-Cookie` may contain PII.
+  final List<String> networkResponseHeaders = [];
```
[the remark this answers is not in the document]
lucas-zimerman: Fixed on ce004caf

packages/dart/lib/src/http_client/network_details_capture.dart
```diff
@@ -0,0 +1,175 @@
...
```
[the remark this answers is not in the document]
lucas-zimerman: will be fixed on a follow up PR

packages/dart/lib/src/sentry_options.dart
buenaflor: I was actually thinking of this should be in the `replay` namespace in the Flutter package. didn't realize this is the dart package

buenaflor: but I guess the challenge is we need to make it avaialble somehow in the dart package while setting hte options in the flutter package

there are multiple alternatives we can do here to make that work, can you check pls

lucas-zimerman: Good call! went with the interface + nullable hook approach on `SentryOptions`, mirroring the existing `Transport`/`FileSystemTransport` pattern:

- sentry (dart) now only exposes `@internal` abstract class `NetworkDetailsCapture` and a `nullable` `SentryOptions.networkDetailsCapture` hook, no concrete capture logic, no replay options.
- `sentry_flutter` owns `SentryReplayOptions` (with the network fields) and provides the concrete `FlutterNetworkDetailsCapture`, wired up in `_initDefaultValues` alongside where transport is set.

So the dart package stays replay-agnostic, and everything replay-specific lives in Flutter. Pushed in b282d2db

packages/flutter/lib/src/replay/network_details_capture.dart
```diff
@@ -0,0 +1,180 @@
...
+    Uint8List bytes;
+    try {
+      bytes = await response.stream.toBytes();
+    } catch (exception) {
+      internalLogger.warning(
+        () => 'Failed to capture response body for replay: $exception',
+      );
+      // The stream may be partially consumed at this point, so there's no
+      // safe way to forward it to the original caller anymore.
+      return (response, data);
```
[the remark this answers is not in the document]
lucas-zimerman: Agreed, this predates this PR's refactor (same behavior existed in the original dart implementation before the move). Will address in a follow-up PR.

packages/dart/lib/src/http_client/breadcrumb_client.dart
```diff
@@ -60,24 +68,44 @@ class BreadcrumbClient extends BaseClient {
...
       statusCode = response.statusCode;
       reason = response.reasonPhrase;
       responseBodySize = response.contentLength;
 
+      if (captureNetworkDetails) {
+        final result = await capture.captureResponse(response);
+        response = result.$1;
+        responseDetail = result.$2;
+      }
+
```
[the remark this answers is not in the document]
lucas-zimerman: Good catch, will address in a follow-up PR.

lucas-zimerman:
Sample replay on the latest commit: https://sentry-sdks.sentry.io/explore/replays/b14e6b2a28b447f4b3cfe14ce2070c89/?referrer=/explore/discover/homepage/&t_main=network&n_detail_row=1

buenaflor:
@lucas-zimerman pls take a look at the latest AI reviews

packages/flutter/lib/src/replay/network_details_capture.dart
```diff
@@ -0,0 +1,180 @@
...
+    if (!_options.sendDefaultPii ||
+        !_options.replay.networkCaptureBodies ||
+        !_isCapturableContentType(response.headers['content-type']) ||
+        (contentLength != null && contentLength > maxBodySize)) {
+      return (response, data);
+    }
+
+    Uint8List bytes;
+    try {
+      bytes = await response.stream.toBytes();
```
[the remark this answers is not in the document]
lucas-zimerman: Deferring this to #3900 — bounding this buffering was already attempted in 27b7c5b5/cb9bba37/7f5bcc52 and reverted in a3a7899f to keep this PR's scope on the header/body capture feature itself. The follow-up work in #3900 (moving captured data to a replay-only `Hint` side channel with new native bridge plumbing) touches this same `captureResponse` path, so the bounded-read fix belongs there instead of being built twice.

packages/flutter/lib/src/replay/network_details_capture.dart
```diff
@@ -0,0 +1,180 @@
...
+  static const _defaultHeaders = ['content-type', 'content-length', 'accept'];
+
+  /// Matches native's `SentryReplayOptions.MAX_NETWORK_BODY_SIZE`.
+  static const maxBodySize = 150 * 1024;
+
+  FlutterNetworkDetailsCapture(this._options) {
+    if (_options.replay.networkDetailAllowUrls.isNotEmpty) {
+      _options.sdk.addFeature(SentryFeatures.replayNetworkDetailsCapturing);
+    }
+  }
```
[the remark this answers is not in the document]
lucas-zimerman: Already fixed in 4a815644 (before this comment landed) — moved the `addFeature` call from the constructor into `shouldCapture()`, which runs lazily per-request after `optionsConfiguration` has populated `networkDetailAllowUrls`. Covered by the "adds feature flag when allow list is non-empty" / "does not add feature flag when allow list is empty" tests.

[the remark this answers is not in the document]
lucas-zimerman: Confirmed, this was fixed in 4a815644 before this finding was posted. Nothing further needed.

packages/dart/lib/src/http_client/breadcrumb_client.dart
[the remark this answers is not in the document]
lucas-zimerman: Fixed in 2330d71c — stop the stopwatch right after headers arrive (in the try block, plus on the exception path) instead of in `finally`, so `requestDuration` measures time-to-headers consistently whether or not network details are captured. Added a regression test ("duration excludes response body capture time") using a capture fake with an injected delay to confirm the measured duration no longer includes it.

packages/flutter/test/replay/network_details_capture_test.dart
```diff
@@ -0,0 +1,329 @@
...
+        final sut = fixture.getSut();
+
+        final request = Request('POST', Uri.parse('https://example.com'))
+          ..headers['content-type'] = 'text/plain'
+          ..body = 'a' * FlutterNetworkDetailsCapture.maxBodySize;
+
+        final data = sut.captureRequest(request);
+
+        expect((data['body'] as String).length,
+            FlutterNetworkDetailsCapture.maxBodySize);
```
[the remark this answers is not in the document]
lucas-zimerman: Fixed in 4d88b2ce — pinned the expected length as a literal (150 * 1024) instead of FlutterNetworkDetailsCapture.maxBodySize. Also fixed the identical tautology in the sibling captureResponse test ("truncates and forwards body exceeding max size when contentLength is unknown", same pattern, not separately flagged but same bug).

packages/dart/lib/src/http_client/breadcrumb_client.dart
```diff
@@ -60,23 +68,48 @@ class BreadcrumbClient extends BaseClient {
...
       return response;
     } catch (_) {
+      stopwatch.stop();
```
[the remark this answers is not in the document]
lucas-zimerman: Good catch — this was a side effect of my earlier fix in 2330d71c (rethrow on capture failure, closer above). Fixed in 4707b53b: split the inner `_client.send` try/catch from the capture step, so only a transport-level failure sets `requestHadException`. A capture failure after a real response still rethrows (so callers don't get a broken/consumed stream, per the earlier fix), but the breadcrumb level now stays based on the actual status code instead of forcing error. Added a regression test covering this.

packages/dart/lib/src/http_client/breadcrumb_client.dart
```diff
@@ -60,23 +68,56 @@ class BreadcrumbClient extends BaseClient {
...
-      rethrow;
     } finally {
-      stopwatch.stop();
+      // Captured after `_client.send` returns (or throws) rather than
+      // before, so that headers set by clients further down the chain
+      // (e.g. TracingClient's sentry-trace/baggage) are reflected instead
+      // of a pre-dispatch snapshot.
+      if (captureNetworkDetails) {
+        requestDetail = capture.captureRequest(request);
+      }
```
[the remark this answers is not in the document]
lucas-zimerman: Another good catch from the same fix, and same root cause as the error-level one above: `Breadcrumb.http`'s `timestamp` param defaults to "now" if not passed explicitly, and `BreadcrumbClient` wasn't passing it — so it defaulted to whenever the breadcrumb factory was called in the outer `finally`, after body capture. Fixed in f5f2345e: capture `getUtcDateTime()` at the same point the stopwatch stops (right after headers arrive, success or failure) and pass it explicitly as `timestamp:`, so `start_timestamp`/`end_timestamp` anchor to header arrival regardless of how long capture takes afterward. Added a regression test.

packages/dart/lib/src/http_client/breadcrumb_client.dart
```diff
@@ -60,24 +68,44 @@ class BreadcrumbClient extends BaseClient {
...
+        final result = await capture.captureResponse(response);
+        response = result.$1;
+        responseDetail = result.$2;
+      }
+
       return response;
     } catch (_) {
       requestHadException = true;
       rethrow;
     } finally {
```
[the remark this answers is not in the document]
lucas-zimerman: This is a duplicate of the finding below (cursor[bot], same commit) — fixed in 4707b53b. See that reply for details.

packages/flutter/lib/src/replay/network_details_capture.dart
```diff
@@ -0,0 +1,180 @@
...
+      if (_options.sendDefaultPii)
+        ...extra.map((header) => header.toLowerCase()),
+    };
+    final result = <String, String>{};
+    headers.forEach((key, value) {
+      if (allowed.contains(key.toLowerCase())) {
+        result[key] = value;
+      }
+    });
+    return result;
```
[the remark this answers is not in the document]
buenaflor: @lucas-zimerman

buenaflor (commented):
there's one more AI review open

packages/flutter/lib/src/replay/network_details_capture.dart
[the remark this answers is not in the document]
lucas-zimerman: This is a false positive — Dart generics are covariant, so `Map<String, String> is Map<String, dynamic>` evaluates to `true` at runtime (verified directly: `{'a':'b'} is Map<String, dynamic>` → `true`). `_filterHeaders`'s result therefore hits the `Map<String, dynamic>` recurse branch in `normalize()` (packages/flutter/lib/src/native/utils/data_normalizer.dart:6-12), not the `.toString()` fallback, so headers cross the bridge as a structured map rather than being stringified. `data_normalizer_test.dart` corroborates this for nested maps generally. Resolving as not applicable.

packages/flutter/lib/src/replay/network_details_capture.dart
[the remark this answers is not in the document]
lucas-zimerman: Tracked in #3900, which reworks captured replay data onto a replay-only `Hint` side channel — the bounded-read fix belongs there since it touches this same `captureResponse` path. Resolving here to avoid tracking it twice.

packages/flutter/pubspec.yaml
```diff
@@ -29,6 +29,7 @@ dependencies:
   ffi: ^2.0.0
   web: ^1.1.0
   jni: 0.14.2
+  http: '>=0.13.0 <2.0.0'
```
buenaflor, line 32: is including this dependency necessary?

buenaflor: I guess it's added anyway since dart also includes it?

buenaflor: question is if we explicitly need to add it still

lucas-zimerman: I know it looks strange, but `flutter analyze` complains if we don't include it. It's the same version that dart uses.
The issue started when I moved the http logic into flutter.
Since `packages/flutter/lib/src/replay/network_details_capture.dart` has `import 'package:http/http.dart';` flutter analyze will complain about it, despite being on Sentry Dart

Sentry dart under the hood already includes it and sentry flutter depends on it, there will be no break changes to end users. it's more of a CI nitpicking

buenaflor: okay then let's keep it like this for now

packages/flutter/lib/src/sentry_replay_options.dart
```diff
@@ -4,6 +4,51 @@ import 'replay/replay_quality.dart';
...
+  /// with `SentryHttpClient` headers/bodies are captured, to be shown
+  /// alongside network spans in Session Replay.
+  ///
+  /// Empty by default, meaning no request is captured. Request/response
+  /// bodies may contain PII — review this list carefully before adding
+  /// entries.
+  ///
+  /// This data is attached to HTTP breadcrumbs, so `recordHttpBreadcrumbs`
+  /// must also be enabled (the default) for this option to have any effect.
+  final List<String> networkDetailAllowUrls = [];
```
buenaflor, line 19: I think we should also add here that these are only working on Android at the moment

buenaflor (approved):
LGTM, one last comment about the options docs

packages/flutter/lib/src/sentry_replay_options.dart
lucas-zimerman: Good idea!  added a note on https://github.com/getsentry/sentry-dart/pull/3875/commits/ac0af8dd22ed714b06fbf6ffe532da0e4e2dcd12

packages/flutter/lib/src/sentry_flutter.dart
```diff
@@ -37,6 +37,7 @@ import 'native/native_scope_observer.dart';
 import 'native/sentry_native_binding.dart';
 import 'profiling.dart';
 import 'replay/integration.dart';
+import 'replay/network_details_capture.dart';
```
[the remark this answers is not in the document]
buenaflor: @lucas-zimerman I think this is what we already discussed or is this something else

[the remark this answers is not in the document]
lucas-zimerman: Yeah it will be addressed on a follow up PR

PR #3876, opened 2026-07-12
fix(flutter): Prevent StateError when delayed frames list is empty

muhammadkamel:
## Description
This PR fixes a crash in delayed frames tracking when _delayedFrames becomes empty after filtering in removeIrrelevantFrames.

### What changed
- Replaced .first with .firstOrNull when reading _delayedFrames
- Kept _oldestFrameEndTimestamp as null when no delayed frames remain

### Why
Calling .first on an empty list can throw:
StateError (Bad state: No element)
Screenshot 2026-07-12 at 7 22 04 PM


buenaflor (approved):
Thanks. we've had this report before that VSCode catches this error, this doesn't cause any production issues since it only happens with VSCode but we should nonetheless update it with `firstOrNull` 👍

Issue #3823, opened 2026-07-01
[v10] Remove deprecated `copyWith` methods


buenaflor:
## Context
Part of the [Major v10 Task Collection](https://github.com/getsentry/sentry-dart/issues/3487).

## Problem
~31 `copyWith` methods across the protocol classes are annotated `@Deprecated('Assign values directly to the instance.')`. They should not survive into v10.

## Expected outcome
- All deprecated `copyWith` methods are deleted.
- The abstract `copyWith` member is removed from the `SentryEventLike` mixin.
- No internal or test callers of `copyWith` remain.

## Pointers
- `packages/dart/lib/src/protocol/*` (SentryEvent, Breadcrumb, Contexts, SentryUser, Mechanism, …)
- `packages/dart/lib/src/sentry_event_like.dart:5`

**Breaking change:** Yes — public API removal, high blast radius. Migration: assign values directly to the instance.

PR #3877, opened 2026-07-13
ref(dart): Remove deprecated copyWith methods

buenaflor:
## 📜 Description

Removes the deprecated `copyWith` methods from Sentry protocol classes and the abstract member from `SentryEventLike`. This also removes `SentryRuntime.copyWith`, whose deprecation annotation was accidentally omitted.

Mutable protocol objects should be updated directly. Immutable `SentryFeatureFlag`, `SentryProxy`, and `SentryTransactionInfo` values should be replaced by constructing new instances.

## 💡 Motivation and Context

These APIs became obsolete after protocol objects became mutable and should not carry into v10.

Closes getsentry/sentry-dart#3823

## 💚 How did you test it?

* `fvm dart analyze`
* `fvm dart test`
* `fvm dart format --output=none --set-exit-if-changed lib test`
* `fvm flutter analyze test/mocks.mocks.dart`

## :pencil: Checklist

- [X] All tests passing
- [ ] No breaking changes


packages/dart/lib/src/sentry_event_like.dart
```diff
@@ -1,7 +1,2 @@
-import 'protocol.dart';
-
 /// A [SentryEventLike] mixin that is extended by [SentryEvent] and [SentryTransaction]
-mixin SentryEventLike<T> {
-  @Deprecated('Assign values directly to the instance.')
-  T copyWith();
-}
+mixin SentryEventLike<T> {}
```
denrase: If this was the only method exposed, we can remove this mixin as well.

denrase (approved):
Looking good, just one suggestion.

packages/dart/lib/src/sentry_event_like.dart
buenaflor: good catch, didn't see this

buenaflor: good catch 👍

PR #3894, opened 2026-07-15
fix(dart): Read Windows total physical memory via GlobalMemoryStatusEx instead of wmic.exe

cf-rhett:
## :scroll: Description

On Windows, the pure-Dart IO device-context enricher (`PlatformMemory`) read `TotalPhysicalMemory` by shelling out to `wmic.exe` (`wmic ComputerSystem get TotalPhysicalMemory /VALUE`), falling back to `powershell.exe` when WMIC was absent. This replaces both process spawns with a direct kernel call to [`GlobalMemoryStatusEx`](https://learn.microsoft.com/en-us/windows/win32/api/sysinfoapi/nf-sysinfoapi-globalmemorystatusex) from `kernel32.dll` via `dart:ffi`, reading `MEMORYSTATUSEX.ullTotalPhys`.

The returned value keeps the previous semantics (bytes), so `device.memorySize` is unchanged. The `useWindowsWmci` / `useWindowsPowerShell` probing, the WMIC/PowerShell helpers, and the `wmic.exe`/`powershell.exe` existence checks are all removed. Adds `ffi: ^2.0.0` as a dependency — already used by `sentry_flutter`.

WMIC is [deprecated and being removed](https://techcommunity.microsoft.com/blog/windows-itpro-blog/wmi-command-line-wmic-utility-deprecation-next-steps/4039242) from current Windows installs. Two problems with the current approach:

1. **Endpoint-security noise.** Every enriched event triggers a `wmic.exe` (or `powershell.exe`) child process under the host application. EDR/XDR tooling flags these spawns, and users ask why the app is invoking a deprecated Windows utility.
2. **Cost.** Spawning a process to read a single integer is far heavier than a syscall — the caching added in #2760 reduced how often it happens, but the spawn is still there on the first enriched event.

`GlobalMemoryStatusEx` is the documented Win32 API for this value, needs no child process, and works regardless of whether WMIC is installed. The library the current code was copied from (`system_info2` / `onepub-dev/system_info`) still uses WMIC on its `master` branch, so there's no upstream version to lean on here.

Covered by the existing `test/event_processor/enricher/io_platform_memory_test.dart`, which asserts a non-null, positive `TotalPhysicalMemory` on Windows and Linux and `null` elsewhere. `dart analyze` and `dart format` are clean on the change.

## :pencil: Checklist

- [x] I added tests to verify changes (existing test covers the Windows path)
- [x] All tests passing
- [x] No breaking changes

## Changelog Entry

Read Windows total physical memory via the native `GlobalMemoryStatusEx` API instead of spawning the deprecated `wmic.exe` (or PowerShell) process.


packages/dart/lib/src/event_processor/enricher/io_platform_memory.dart
```diff
@@ -1,63 +1,61 @@
...
-    if (size == null) {
+  int? _getWindowsTotalPhysicalMemory() {
+    final statusPointer = calloc<_MemoryStatusEx>();
+    try {
+      statusPointer.ref.dwLength = sizeOf<_MemoryStatusEx>();
+      // kernel32 is always mapped into the process, so resolve the symbol from
+      // the process itself rather than acquiring a fresh module handle.
+      final globalMemoryStatusEx = DynamicLibrary.process().lookupFunction<
+          _GlobalMemoryStatusExNative,
+          _GlobalMemoryStatusExDart>('GlobalMemoryStatusEx');
```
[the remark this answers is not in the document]
cf-rhett: Thanks for the flag — I checked this and it isn't a real bug. `DynamicLibrary.process()` is supported on Windows and does resolve `GlobalMemoryStatusEx`; it does not throw `UnsupportedError`.

**Why `process()` works on Windows**

The API contract for [`DynamicLibrary.process()`](https://api.dart.dev/stable/dart-ffi/DynamicLibrary/DynamicLibrary.process.html) is platform-neutral — it holds *all* globally-loaded symbols, "including the executable itself." On Windows the runtime resolves a symbol by walking every loaded module and calling `GetProcAddress` on each. See the Win32 code path in the SDK (pinned to tag `3.12.2`):

- [`LookupSymbolInProcess` — `EnumProcessModules` + `GetProcAddress`](https://github.com/dart-lang/sdk/blob/3.12.2/runtime/lib/ffi_dynamic_library.cc#L98-L135)
- [`#if defined(DART_HOST_OS_WINDOWS)` dispatch for process lookups](https://github.com/dart-lang/sdk/blob/3.12.2/runtime/lib/ffi_dynamic_library.cc#L137-L141)

`GlobalMemoryStatusEx` is exported by [`Kernel32.dll`](https://learn.microsoft.com/en-us/windows/win32/api/sysinfoapi/nf-sysinfoapi-globalmemorystatusex) (*Requirements → DLL: Kernel32.dll*), which is always mapped into every Win32 process, so it's always in that enumeration and resolves.

**Empirical check on real Windows**

I ran a standalone `dart:ffi` parity check comparing `process()`, `open('kernel32.dll')`, `wmic`, and PowerShell CIM. All four agree to the byte:

```
sizeOf<MemoryStatusEx> = 64 (expect 64)
Native GlobalMemoryStatusEx:
  [process()] ullTotalPhys = 8580235264
  [open('kernel32.dll')] ullTotalPhys = 8580235264

wmic TotalPhysicalMemory = 8580235264
powershell CIM Total     = 8580235264

PASS: native (8580235264) == wmic (8580235264)
```

So `process()` returns a correct, non-null value and the unit test passes on Windows.

Note your suggested fix (`open('kernel32.dll')`) is also correct and returns the identical value (second line above). I chose `process()` deliberately: `kernel32.dll` is guaranteed to already be mapped, so there's no need to acquire a fresh module handle. Happy to switch to `open('kernel32.dll')` if the team finds it clearer.

<details>
<summary>Parity check script (<code>verify-winmem.ps1</code>) run in a Windows guest</summary>

```powershell
# Self-contained Windows physical-memory parity check.
# Writes a pure dart:ffi program (no pub get needed) and runs it, comparing
# native GlobalMemoryStatusEx against wmic and PowerShell CIM.
#
# Prereq: Dart SDK at C:\dart\dart-sdk\bin\dart.exe (adjust $dart if elsewhere).
# Run in the guest:  powershell -NoProfile -ExecutionPolicy Bypass -File verify-winmem.ps1

$ErrorActionPreference = 'Stop'
$dartFile = Join-Path $env:TEMP 'verify-winmem.dart'

$dart = (Get-Command dart -ErrorAction SilentlyContinue).Source
if (-not $dart) {
  Write-Error "dart not found on PATH. Install the SDK first, then re-run."
  exit 1
}
Write-Output "Using dart: $dart"

$src = @'
// Standalone parity check: native GlobalMemoryStatusEx vs wmic vs PowerShell.
import 'dart:ffi';
import 'dart:io';

final class MemoryStatusEx extends Struct {
  @Uint32()
  external int dwLength;
  @Uint32()
  external int dwMemoryLoad;
  @Uint64()
  external int ullTotalPhys;
  @Uint64()
  external int ullAvailPhys;
  @Uint64()
  external int ullTotalPageFile;
  @Uint64()
  external int ullAvailPageFile;
  @Uint64()
  external int ullTotalVirtual;
  @Uint64()
  external int ullAvailVirtual;
  @Uint64()
  external int ullAvailExtendedVirtual;
}

typedef _GmseNative = Int32 Function(Pointer<MemoryStatusEx>);
typedef _GmseDart = int Function(Pointer<MemoryStatusEx>);
typedef _AllocNative = Pointer<Void> Function(Uint32, IntPtr);
typedef _AllocDart = Pointer<Void> Function(int, int);
typedef _FreeNative = Pointer<Void> Function(Pointer<Void>);
typedef _FreeDart = Pointer<Void> Function(Pointer<Void>);

int? nativeTotalPhys(DynamicLibrary lib) {
  final gmse = lib.lookupFunction<_GmseNative, _GmseDart>('GlobalMemoryStatusEx');
  final alloc = lib.lookupFunction<_AllocNative, _AllocDart>('GlobalAlloc');
  final free = lib.lookupFunction<_FreeNative, _FreeDart>('GlobalFree');
  final ptr = alloc(0x0040, sizeOf<MemoryStatusEx>()).cast<MemoryStatusEx>();
  if (ptr.address == 0) return null;
  try {
    ptr.ref.dwLength = sizeOf<MemoryStatusEx>();
    if (gmse(ptr) == 0) return null;
    return ptr.ref.ullTotalPhys;
  } finally {
    free(ptr.cast());
  }
}

int? tryLib(String label, DynamicLibrary Function() open) {
  try {
    final v = nativeTotalPhys(open());
    print('  [$label] ullTotalPhys = $v');
    return v;
  } catch (e) {
    print('  [$label] FAILED: $e');
    return null;
  }
}

int? wmicValue() {
  try {
    final r = Process.runSync('wmic', [
      'ComputerSystem',
      'get',
      'TotalPhysicalMemory',
      '/VALUE',
    ]);
    final m = RegExp(r'TotalPhysicalMemory=(\d+)').firstMatch(r.stdout as String);
    return m == null ? null : int.parse(m.group(1)!);
  } catch (e) {
    print('  wmic FAILED: $e');
    return null;
  }
}

int? psValue() {
  try {
    final r = Process.runSync('powershell', [
      '-NoProfile',
      '-Command',
      '(Get-CimInstance Win32_ComputerSystem).TotalPhysicalMemory',
    ]);
    return int.tryParse((r.stdout as String).trim());
  } catch (e) {
    print('  powershell FAILED: $e');
    return null;
  }
}

void main() {
  print('sizeOf<MemoryStatusEx> = ${sizeOf<MemoryStatusEx>()} (expect 64)');
  print('Native GlobalMemoryStatusEx:');
  final proc = tryLib('process()', DynamicLibrary.process);
  final k32 = tryLib("open('kernel32.dll')", () => DynamicLibrary.open('kernel32.dll'));
  final wmic = wmicValue();
  final ps = psValue();
  print('');
  print('wmic TotalPhysicalMemory = $wmic');
  print('powershell CIM Total     = $ps');
  final native = proc ?? k32;
  final ok = native != null && wmic != null && native == wmic;
  print('');
  print(
    ok
        ? 'PASS: native ($native) == wmic ($wmic)'
        : 'FAIL: process()=$proc kernel32=$k32 wmic=$wmic powershell=$ps',
  );
}
'@

Set-Content -Path $dartFile -Value $src -Encoding UTF8
Write-Output "Wrote $dartFile"
Write-Output "Running..."
& $dart $dartFile
```
</details>

packages/dart/lib/src/event_processor/enricher/io_platform_memory.dart
```diff
@@ -1,63 +1,61 @@
...
+
+      if (globalMemoryStatusEx(statusPointer) == 0) {
+        return null;
+      }
+      return statusPointer.ref.ullTotalPhys;
+    } catch (e) {
+      options.log(
```
lucas-zimerman: We should use `internalLogger` here since `options.log`

cf-rhett: Switched to `internalLogger.warning`. I'd used `options.log` to match `_exec` just below it, which is still on `options.log` in main. Want that one converted here too, or should I keep this PR to the Windows path?

lucas-zimerman: No need, let's keep the PR exclusively to the changes for Windows

lucas-zimerman (approved):
Looks good to me, CI passed with the changes in regard to windows, there are some failures on other platforms but probably unrelated to this PR.

@buenaflor Anything that you would like to point out? I checked locally, and the PR is working as intended, if there is nothing else to change from your part, LGTM!

buenaflor:
LGTM if you tested it as well. CI failures are because of missing auth for E2E tests, but that's fine

buenaflor:
There seems to be Analyze errors

PR #3896, opened 2026-07-16
feat(flutter): Add standalone app start tracing

### LINKED ISSUE #3634: feat: Send standalone app start transactions [Flutter]

buenaflor:
## 📜 Description

Opt-in standalone app-start tracing for Android/iOS. When `enableStandaloneAppStartTracing` is enabled, Flutter emits an independent `App Start` root (`app.start` / `auto.app.start`) from process start to first frame, instead of nesting app-start breakdown under the synthetic initial`ui.load`.

The flag is experimental and defaults to `false`. With it off, the existing `ui.load` app-start path is unchanged. With it on, static and stream lifecycles share the same hierarchy and vitals semantics while using each lifecycle's native payload shape.

Shared timing/parsing lives in `AppStartData`; static and stream each have a dedicated trace builder, owned by `StandaloneAppStartLifecycle`. Idle roots gain a hard final timeout so deadline events can still flush without a duration vital.

## 💡 Motivation and Context

Today native app start is attached to a synthetic initial `ui.load`, which mixes startup and screen-display telemetry and prevents an independent lifecycle/sampling decision for app start.

Fixes getsentry/sentry-dart#3634

## 💚 How did you test it?

E2E:

* [Static](<https://sentry-sdks.sentry.io/explore/traces/trace/c6f3cd0a88664a4888166eecef91c0bb/?node=span-c00189e324fc40fd&project=5428562&query=span.op%3Aapp.start&source=traces&statsPeriod=90d&targetId=c00189e324fc40fd&timestamp=1784137863>):
* [Streamed](<https://sentry-sdks.sentry.io/explore/traces/trace/b4c11717dd364400a6a795c927b398a8/?node=span-f65c51ebaf754fc8&project=5428562&source=traces&statsPeriod=1h&targetId=47ac9319bcc449b7&timestamp=1784539483>)

## :pencil: Checklist

- [ ] All tests passing
- [X] No breaking changes

## 🔮 Next steps

* Strip the temporary example debug transport / dart-define harness before<br>merge (local verification only)
* Confirm docs-site follow-up once the experimental flag is ready to advertise

Made with [Cursor](<https://cursor.com>)


packages/flutter/lib/src/app_start/standalone/standalone_app_start_handler.dart
```diff
@@ -0,0 +1,170 @@
...
+    await _trace?.close();
+    switch (_options.traceLifecycle) {
+      case SentryTraceLifecycle.static:
+        _options.timeToDisplayTracker.clear();
+      case SentryTraceLifecycle.stream:
+        _options.timeToDisplayTrackerV2.cancelCurrentRoute();
+    }
+    _trace = null;
+    _startScreenName = null;
+  }
```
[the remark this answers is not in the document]
buenaflor: Won't fix for now. close() before first frame is rare outside re-init/tests; the callback self-removes on the first non-empty timings batch. Separately hardened close() to tolerate closed/no-op hub options (the CI cast crash).

*— Claude Code*

packages/flutter/lib/src/navigation/time_to_display_tracker.dart
```diff
@@ -117,9 +153,13 @@ class TimeToDisplayTracker {
   }
 
   void clear() {
+    final preparedInitialDisplay = _preparedInitialDisplay;
+    _preparedInitialDisplay = null;
+    if (preparedInitialDisplay != null) {
```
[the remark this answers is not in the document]
buenaflor: Fixed in 470f47c0d4 by removing abandon() entirely. clear() only drops the prepared reference; idle auto-finish + the childless-idle drop own teardown (and finish() clears scope.span).

*— Claude Code*

packages/flutter/lib/src/app_start/standalone/streaming_app_start_trace.dart
```diff
@@ -0,0 +1,187 @@
...
+
+      final firstFrameBarrier = hub.startInactiveSpan(
+        appStartFirstFrameRenderDescription,
+        parentSpan: createdRoot,
+        startTimestamp: data.sentrySetupTimestamp,
+        attributes: _childAttributes(
+          data,
+          SentrySpanOperations.appStartFirstFrameRender,
+        ),
+      );
```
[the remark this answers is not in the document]
buenaflor: Won't fix. With no children attached, the idle timer ends the root after appStartIdleTimeout (3s). Matches JS idle-span behavior for ignored/no-op children; edge-casey via ignoreSpans.

*— Claude Code*

packages/dart/lib/src/sentry_tracer.dart
```diff
@@ -94,25 +111,64 @@ class SentryTracer extends ISentrySpan {
...
+    Hint? hint,
+  }) {
+    if (_captured) return Future.value();
+    final inFlight = _finalizeFuture;
+    if (inFlight != null) return inFlight;
+
+    _finishRequested = false;
+    final future = _finalize(endTimestamp: endTimestamp, hint: hint);
+    _finalizeFuture = future;
+    return future;
```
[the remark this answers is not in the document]
buenaflor: Not actionable for this PR. The cached failed future can matter if `onFinish` throws before the root finishes and the same transaction is explicitly finished again, but standalone app start uses a synchronous `onFinish` that catches its own errors. Capture failures happen after the root has finished and were not retriable before this change either. Avoiding additional retry state here.

packages/dart/lib/src/sentry_tracer.dart
```diff
@@ -150,26 +213,107 @@ class SentryTracer extends ISentrySpan {
...
+  }
+
+  Future<void> _finishAtDeadline() {
+    if (_captured) return Future.value();
+
+    _deadlineFinalization = true;
+    _requestedStatus = SpanStatus.deadlineExceeded();
+    final inFlight = _finalizeFuture;
+    if (inFlight != null) return inFlight;
+
```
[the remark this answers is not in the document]
buenaflor: This race is not reachable in the standalone app-start path. The only production caller of `tryScheduleFinalTimeout` supplies a synchronous `onFinish` callback that catches its own errors, so the deadline timer cannot fire during an awaited `onFinish`. Other async `onFinish` call sites do not schedule this timeout. No change needed.

packages/dart/lib/src/telemetry/span/idle_recording_sentry_span_v2.dart
```diff
@@ -142,20 +144,25 @@ final class IdleRecordingSentrySpanV2 extends RecordingSentrySpanV2 {
...
         SemanticAttributesConstants.sentryStatusMessage,
         SentryAttribute.string(SentrySpanStatusMessages.deadlineExceeded),
       );
     }
 
-    _finishActiveDescendants(idleEndTimestamp);
+    _finishActiveDescendants(
+      idleEndTimestamp,
+      deadlineExceeded: deadlineExceeded,
+    );
```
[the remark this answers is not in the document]
buenaflor: Fixed in 8b933d2176. Final-timeout roots now retain the deadline timestamp, and descendants that start after the pinned deadline are ended at the current time with `deadlineExceeded`. Added deterministic regression coverage for both cases.

packages/flutter/lib/src/app_start/standalone/streaming_app_start_trace.dart
```diff
@@ -0,0 +1,211 @@
...
+    }
+  }
+
+  @override
+  void close() {
+    if (_closed || _completed) return;
+    _closed = true;
+    _root.end();
+  }
```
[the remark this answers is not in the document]
buenaflor: Closing as non-actionable. `close()` intentionally ends the root and starts the pre-capture processing pipeline. `_processSpan` is the one-shot enrichment hook for that ended root and removes itself before the span enters the telemetry processor. Guarding it with `_closed` or unregistering it in `close()` would drop required type and screen metadata, including the behavior covered by the close regression test.

packages/flutter/lib/src/app_start/ui_load_attached/native_app_start_integration.dart
```diff
@@ -29,6 +29,8 @@ class NativeAppStartIntegration extends Integration<SentryFlutterOptions> {
 
   @override
   void call(Hub hub, SentryFlutterOptions options) async {
+    if (options.enableStandaloneAppStartTracing) return;
```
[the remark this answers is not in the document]
buenaflor: Closing as non-actionable. Standalone app start is intentionally limited to Android and iOS. Although the legacy integration is registered on macOS, the native plugin returns no app-start timing there and both legacy handlers emit nothing for null timing, so enabling this unsupported option does not suppress an existing macOS app-start signal.

packages/flutter/lib/src/navigation/time_to_display_tracker.dart
```diff
@@ -117,9 +154,11 @@ class TimeToDisplayTracker {
...
 
   void clear() {
+    // Drop the prepared root reference only. Idle auto-finish (and the
+    // childless-idle drop in SentryTracer) owns teardown without capture.
+    _preparedInitialDisplay = null;
     transactionId = null;
     _pendingTTFDEndTimestamp = null;
-
     _ttidTracker.clear();
     if (options.enableTimeToFullDisplayTracing) {
```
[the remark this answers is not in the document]
buenaflor: Closing as non-actionable. During Sentry.close(), the global hub is replaced with NoOpHub before integrations are closed. The prepared transaction remains associated only with the discarded old hub until its idle timeout; new spans and a subsequent SDK initialization use a different hub, so they cannot attach to it. Immediate scope cleanup could be separate hardening, but the reported production mis-parenting window is not reachable.

packages/flutter/lib/src/app_start/standalone/standalone_app_start_handler.dart
```diff
@@ -0,0 +1,241 @@
...
+
+      final setupTimestamp = SentryFlutter.sentrySetupStartTime;
+      if (nativeAppStart != null && setupTimestamp != null) {
+        data = AppStartData.tryParse(
+          nativeAppStart,
+          sentrySetupTimestamp: setupTimestamp,
+          // Eager parse: only trust timestamps through setup. First frame
+          // is the measurement end later, not part of this ceiling.
+          validUntil: setupTimestamp,
+        );
```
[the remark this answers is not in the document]
buenaflor: Fixed in ad20cf38ab. Eager standalone parsing now validates native phases against the completed snapshot time rather than SDK setup start, retaining lifecycle spans returned after setup began. Added a regression test covering an Activity.onCreate phase that completes between setup start and snapshot.

docs/standalone-app-start-spec.md
```diff
@@ -0,0 +1,77 @@
+# Standalone App Start
```
denrase: Do we ship the specs in the PRs?

docs/standalone-app-start-spec.md
```diff
@@ -0,0 +1,77 @@
...
+
+## Acceptance criteria
+
+- `enableStandaloneAppStartTracing` is experimental, defaults to `false`, and
+  does not independently enable tracing.
+- When disabled, existing app-start attachment to `ui.load` remains unchanged.
+- When enabled, exactly one app-start representation is emitted.
+- The root is named `App Start`, uses `app.start`, and has origin
+  `auto.app.start`.
+- Its natural duration runs from process start to the first frame drawn.
```
denrase: How does this work with use-cases where we do not have app UI?

docs/standalone-app-start-spec.md
```diff
@@ -0,0 +1,77 @@
...
+- the duration measurement/value is omitted; and
+- type and required screen metadata are retained.
+
+If no first frame arrives, this diagnostic deadline event may still be
+captured without a duration vital. This is not headless app-start support.
+
+## Non-goals
+
+- Extended app-start APIs or measurement windows.
+- Headless or background app-start measurement.
```
denrase: Ok, ignore my previous comment then.

packages/dart/lib/src/hub.dart
```diff
@@ -900,16 +900,24 @@ class Hub {
...
+  /// the span is detached — it never becomes the default parent, does not
+  /// block a bound idle span, and the caller owns ending it (the idle and
+  /// final timeouts remain as backstops).
   @internal
   SentrySpanV2 startIdleSpan(
     String name, {
     Duration idleTimeout = const Duration(seconds: 3),
     Duration finalTimeout = const Duration(seconds: 30),
     bool trimIdleSpanEndTimestamp = true,
+    bool setAsActive = true,
```
denrase: Other params have 'idle' int hame, we could also name this accordingly, so t is clear they belong together conceptually.

packages/dart/lib/src/sentry_tracer.dart
```diff
@@ -94,25 +111,64 @@ class SentryTracer extends ISentrySpan {
...
   @override
-  Future<void> finish(
-      {SpanStatus? status, DateTime? endTimestamp, Hint? hint}) async {
-    final commonEndTimestamp = endTimestamp ?? _hub.options.clock();
+  Future<void> finish({
+    SpanStatus? status,
+    DateTime? endTimestamp,
+    Hint? hint,
+  }) {
+    if (_captured) return Future.value();
```
denrase: Are captured and finished two differnt things, or could we derive caprtured from properties we already have?

packages/dart/lib/src/sentry_tracer.dart
```diff
@@ -150,26 +213,107 @@ class SentryTracer extends ISentrySpan {
...
+    if (deadline == null) return;
+
+    final deadlineStatus = SpanStatus.deadlineExceeded();
+    _rootSpan.status = deadlineStatus;
+    final children = List<SentrySpan>.of(_children);
+    for (final child in children) {
+      if (child.finished) continue;
+      child.status = deadlineStatus;
+      if (child.startTimestamp.isAfter(deadline)) {
+        _children.remove(child);
```
denrase: Should we update the spec to include removal of late started children (after deadline)?

packages/dart/lib/src/sentry_tracer.dart
```diff
@@ -23,13 +23,30 @@ class SentryTracer extends ISentrySpan {
...
-  var _finishStatus = SentryTracerFinishStatus.notFinishing();
+
+  /// [finish] was called but [waitForChildren] is still blocking capture.
+  bool _finishRequested = false;
+
+  /// In-flight capture path; concurrent callers join this future.
+  Future<void>? _finalizeFuture;
+
+  /// Successfully captured (or root already finished before finalize).
+  bool _captured = false;
```
denrase: We now have multiple properties and futures deciding the finishing status, which makes it really difficult to reason about, I do not grasp how this works together. It adds substantial state-machine and async complexity to the core SentryTracer for an app-start-specific deadline. Can we simplify the design before merging?

 - _finishRequested
 - _finalizeFuture
 - _captured
 - _requestedStatus
 - _deadlineFinalization
 - _finalDeadlineTimestamp
 - two timers

packages/flutter/lib/src/app_start/standalone/standalone_app_start_lifecycle.dart
```diff
@@ -0,0 +1,242 @@
...
+
+        // Freeze the launch screen during first frame before enrichment;
+        // the user may navigate away before the app-start span finishes.
+        _startScreenName ??= SentryNavigatorObserver.currentRouteName;
+
+        final options = _flutterOptions;
+        if (options != null) {
+          switch (options.traceLifecycle) {
+            case SentryTraceLifecycle.static:
+              await options.timeToDisplayTracker.recordInitialDisplay(
```
denrase: recordInitialDisplay internally also awaits ttfd, which could timeout app start behaviour, when we only care about ttid, no?

denrase (changes requested):

docs/standalone-app-start-spec.md
buenaflor: we don't support it in Flutter (we also dont support now with the normal app start)

docs/standalone-app-start-spec.md
buenaflor: no, it was accidental, I'll remove it before merging

packages/dart/lib/src/sentry_tracer.dart
buenaflor: in some other SDKs it lives in the tracer directly but I think we can move it to `StaticAppStartTrace`  since it's only used there right now

Moving it to `StaticAppStartTrace`

packages/flutter/lib/src/app_start/standalone/standalone_app_start_lifecycle.dart
buenaflor: swapped the order so it runs last and won't timeout app start behaviour. 

we care about both TTID and TTFD here , we wanna do the tracking simultaenously

packages/dart/lib/src/sentry_tracer.dart
buenaflor: moved the implementation so it's simpler now

packages/dart/lib/src/sentry_tracer.dart
buenaflor: I moved the impl to the static app start trace which is now much simpler

the SDK does this automatically anyway when finishing a transaction

```
      // remove span where its endTimestamp is before startTimestamp
      _children.removeWhere(
          (span) => !_hasSpanSuitableTimestamps(span, commonEndTimestamp));
```

packages/dart/lib/src/hub.dart
buenaflor: updated to `bindToHub` which has a better representation imo

packages/flutter/lib/src/app_start/standalone/streaming_app_start_trace.dart
```diff
@@ -0,0 +1,212 @@
...
+        trace._processSpan,
+      );
+      return trace;
+    } catch (error, stackTrace) {
+      internalLogger.error(
+        'Failed to create streaming standalone app start',
+        error: error,
+        stackTrace: stackTrace,
+      );
+      return root == null ? null : _abort(root, children: createdChildren);
```
denrase: Streaming spans cannot iterate their children?

packages/flutter/lib/src/app_start/standalone/streaming_app_start_trace.dart
```diff
@@ -0,0 +1,212 @@
...
+  /// Children are ended here rather than left to the root's own teardown,
+  /// which force-ends descendants with timestamps derived for a trace that is
+  /// being reported — not one being abandoned.
+  static StreamingAppStartTrace? _abort(
+    IdleRecordingSentrySpanV2 root, {
+    Iterable<RecordingSentrySpanV2> children = const [],
+    String? reason,
+  }) {
+    if (reason != null) {
+      internalLogger.info('Skipping streaming standalone app start: $reason');
```
denrase, line 137: Reason is only used for logging?

packages/flutter/lib/src/navigation/time_to_display_tracker.dart
```diff
@@ -30,10 +34,45 @@ class TimeToDisplayTracker {
...
 
   // The timestamp where report TTFD was called before the transaction was started.
   DateTime? get pendingTTFDEndTimestamp => _pendingTTFDEndTimestamp;
 
+  /// Reserves the initial standalone `ui.load` identity for the app start.
+  ///
+  /// Only the identity: [recordInitialDisplay] starts the transaction. Starting
+  /// it here would arm its idle timer before a first frame exists, and a start
+  /// slower than that timer drops the childless root along with TTID and TTFD.
+  void prepareInitialDisplay(DateTime startTimestamp) {
```
denrase, line 47: Would it make sense to return context instead of holding it, so we avoid additional state we have to manage?

packages/flutter/lib/src/navigation/time_to_display_tracker.dart
```diff
@@ -30,10 +34,45 @@ class TimeToDisplayTracker {
...
+    final context = initialDisplayTransactionContext();
+    transactionId = context.spanId;
+    _preparedContext = context;
+    _preparedStartTimestamp = startTimestamp;
+  }
+
+  /// Starts the reserved initial standalone display root and records TTID/TTFD.
+  Future<void> recordInitialDisplay(DateTime endTimestamp) async {
```
denrase, line 69: ... and just pass it here as parameter.

denrase (approved):
Just something a review pass uncovered, could we verify that a named initial route in static mode doesn’t clear the prepared ui.load and create it under a new trace instead of sharing the app-start trace ID?

packages/flutter/lib/src/app_start/standalone/streaming_app_start_trace.dart
buenaflor: yup, to add more context

packages/flutter/lib/src/app_start/standalone/streaming_app_start_trace.dart
buenaflor: Updated it, we should be good to go by just ending the root idle span without tracking the children (the idle span already tracks them)

packages/flutter/lib/src/navigation/time_to_display_tracker.dart
buenaflor: we can't since the tracker isn't the only consumer, I've consolidated them into a single nullable `_PreparedInitialDisplay` so it's a bit clearer

PR #3918, opened 2026-07-22
feat(flutter): Extend standalone app start

### LINKED ISSUE #3767: Add API to extend app start (Flutter)

buenaflor:
## 📜 Description

Experimental `SentryFlutter` API for extending the standalone App Start past the first frame, so startup work that finishes later is part of the reported duration.

```dart
appRunner: () async {
  SentryFlutter.extendAppStart();
  runApp(const MyApp());

  try {
    await completeStartupWork();
  } finally {
    await SentryFlutter.finishExtendedAppStart();
  }
},
```

`extendAppStart()` opens one `app.start.extended` child of the `App Start` root and holds the root open. `finishExtendedAppStart()` closes it and releases the app start to report, measuring to the extension end when it outlasts the first frame. While an extension is still open the duration is withheld, rather than reporting a window that ends mid-extension.

To nest work under the extension, take the span itself: `getExtendedAppStartSpan()` for the V1 `ISentrySpan` on the static lifecycle, `getExtendedAppStartSpanV2()` for the V2 `SentrySpanV2` on the streaming one — two getters because the protocols don't merge into anything worth exposing publicly. Each returns `null` on the other lifecycle and once the extension has ended. Finishing the span completes the extension on its own.

Requires `enableStandaloneAppStartTracing` on iOS or Android. All entry points are guarded at the boundary so an SDK bug can't throw into `appRunner`, and refusals are logged rather than returned. Operation and method names are verbatim from sentry-java and sentry-cocoa.

> < />!NOTE< />  <br>Stacked on getsentry/sentry-dart#3896 — that is the base branch and should be read first.

## 💡 Motivation and Context

Standalone App Start measures process start to first frame, so work finishing after it — remote config, auth restore, flag hydration — falls outside the window and understates startup. This lets the app say where startup really ends.

Closes getsentry/sentry-dart#3767

## 💚 How did you test it?

Unit coverage on both lifecycles: hierarchy and status, nested descendants, the deadline interaction, finishing the span directly instead of via the API, refused extensions and refused finishes, vitals endpoint resolution, and unpublish-on-complete. The example app is wired up end to end.

## :pencil: Checklist

- [X] All tests passing
- [X] No breaking changes

## 🔮 Next steps

* Attach E2E trace links for both lifecycles, matching getsentry/sentry-dart#3896
* Merge getsentry/sentry-dart#3896 first — this targets that branch
* docs-site follow-up: add the Flutter section alongside the existing Java / Cocoa / React Native ones

Made with [Cursor](<https://cursor.com>)

### DIFF
--- a/packages/flutter/example/lib/main.dart
+++ b/packages/flutter/example/lib/main.dart
@@ -14,18 +14,63 @@ import 'app_config.dart' as config;
 import 'home_screen.dart';
 import 'theme_provider.dart';
 
+Future<void>? _startupConfiguration;
+
 Future<void> main() async {
   await setupSentry(
-    () => runApp(
-      SentryWidget(
-        child: DefaultAssetBundle(
-          bundle: SentryAssetBundle(),
-          child: const MyApp(),
+    () async {
+      SentryFlutter.extendAppStart();
+      final startupConfiguration = _loadStartupConfiguration();
+      _startupConfiguration = startupConfiguration;
+      runApp(
+        SentryWidget(
+          child: DefaultAssetBundle(
+            bundle: SentryAssetBundle(),
+            child: const MyApp(),
+          ),
         ),
-      ),
-    ),
+      );
+      try {
+        await startupConfiguration;
+      } finally {
+        await SentryFlutter.finishExtendedAppStart();
+      }
+    },
     config.exampleDsn,
   );
+}
+
+Future<void> _loadStartupConfiguration() async {
+  final staticParent = SentryFlutter.getExtendedAppStartSpan();
+  if (staticParent != null) {
+    final child = staticParent.startChild(
+      'app.init',
+      description: 'Load startup configuration',
+    );
+    try {
+      await Future.delayed(const Duration(seconds: 1));
+    } finally {
+      await child.finish();
+    }
+    return;
+  }
+
+  final streamParent = SentryFlutter.getExtendedAppStartSpanV2();
+  if (streamParent != null) {
+    final child = Sentry.startInactiveSpan(
+      'Load startup configuration',
+      parentSpan: streamParent,
+      attributes: {
+        SemanticAttributesConstants.sentryOp:
+            SentryAttribute.string('app.init'),
+      },
+    );
+    try {
+      await Future.delayed(const Duration(seconds: 1));
+    } finally {
+      child.end();
+    }
+  }
 }
 
 Future<void> setupSentryWithCustomInit(
@@ -134,9 +179,10 @@ class _MyAppState extends State<MyApp> {
   void doWork() async {
     final rootDisplay = SentryFlutter.currentDisplay();
 
-    await Sentry.startSpan('Custom span that runs during app start', (_) async {
-      await Future.delayed(const Duration(seconds: 1));
-    });
+    final startupConfiguration = _startupConfiguration;
+    if (startupConfiguration != null) {
+      await startupConfiguration;
+    }
 
     rootDisplay?.reportFullyDisplayed();
   }
--- a/packages/flutter/example/lib/main.dart
+++ b/packages/flutter/example/lib/main.dart
@@ -14,14 +14,9 @@ import 'app_config.dart' as config;
 import 'home_screen.dart';
 import 'theme_provider.dart';
 
-Future<void>? _startupConfiguration;
-
 Future<void> main() async {
   await setupSentry(
     () async {
-      SentryFlutter.extendAppStart();
-      final startupConfiguration = _loadStartupConfiguration();
-      _startupConfiguration = startupConfiguration;
       runApp(
         SentryWidget(
           child: DefaultAssetBundle(
@@ -30,11 +25,6 @@ Future<void> main() async {
           ),
         ),
       );
-      try {
-        await startupConfiguration;
-      } finally {
-        await SentryFlutter.finishExtendedAppStart();
-      }
     },
     config.exampleDsn,
   );
@@ -177,11 +167,15 @@ class _MyAppState extends State<MyApp> {
   }
 
   void doWork() async {
+    // Extend before the first frame renders, so the App Start measures up to
+    // `finishExtendedAppStart` instead of the first frame.
+    SentryFlutter.extendAppStart();
     final rootDisplay = SentryFlutter.currentDisplay();
 
-    final startupConfiguration = _startupConfiguration;
-    if (startupConfiguration != null) {
-      await startupConfiguration;
+    try {
+      await _loadStartupConfiguration();
+    } finally {
+      await SentryFlutter.finishExtendedAppStart();
     }
 
     rootDisplay?.reportFullyDisplayed();
--- a/packages/flutter/lib/src/app_start/standalone/app_start_vitals.dart
+++ b/packages/flutter/lib/src/app_start/standalone/app_start_vitals.dart
@@ -51,24 +51,21 @@ final class AppStartVitals {
     );
   }
 
-  /// The endpoint the app start is measured to, or `null` when the first frame
-  /// never arrived.
+  /// The later of the first frame and the extension, or `null` when the first
+  /// frame never arrived.
   ///
-  /// An extension that settled after the first frame moves the endpoint out to
-  /// it. One that contributes nothing — never started, still running, or
-  /// force-ended by the root's deadline — leaves the app start measured to the
-  /// first frame rather than withholding it, matching how sentry-java and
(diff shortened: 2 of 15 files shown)


buenaflor:
@cursor review

buenaflor:
@sentry review

buenaflor:
@cursor review

buenaflor:
@sentry review

buenaflor:
@cursor review

buenaflor:
@sentry review

buenaflor:
@cursor review

packages/flutter/example/lib/main.dart
```diff
@@ -14,20 +14,65 @@ import 'app_config.dart' as config;
...
+            child: const MyApp(),
+          ),
         ),
-      ),
-    ),
+      );
+      try {
+        await startupConfiguration;
+      } finally {
+        await SentryFlutter.finishExtendedAppStart();
```
denrase: Couldn't we have called this in the `doWork` method and move everything we do in `_loadStartupConfiguration` there? Would be a simpler approach for a readable sample, no?

packages/flutter/lib/src/app_start/standalone/app_start_vitals.dart
```diff
@@ -32,20 +32,43 @@ final class AppStartVitals {
...
+    required DateTime? extensionEndTimestamp,
+  }) {
+    final measurementEnd =
+        _resolveMeasurementEnd(endTimestamp, extensionEndTimestamp);
+    return AppStartVitals._(
+      type: timing.type,
+      screen: screen,
+      duration: measurementEnd == null
+          ? null
+          : timing.reportableDurationUntil(measurementEnd),
```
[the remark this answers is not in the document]
denrase: @buenaflor Worth checking if this is a real issue and write a regression test for it.

packages/flutter/lib/src/app_start/standalone/app_start_vitals.dart
```diff
@@ -32,20 +32,43 @@ final class AppStartVitals {
...
+  /// The later of the first frame and the extension, or `null` when the first
+  /// frame never arrived.
+  ///
+  /// An extension contributing nothing — never started, still running, or
+  /// force-ended by the root's deadline — therefore leaves the app start
+  /// measured to the first frame rather than withholding it, matching how
+  /// sentry-java and sentry-cocoa clamp a timed-out time-to-full-display back
+  /// to time-to-initial-display.
+  static DateTime? _resolveMeasurementEnd(
+    DateTime? endTimestamp,
```
denrase: We could consider renaming `endTimestamp` if it is always first frame. Ignore if others (deadlines) are also used for this parameter.

packages/flutter/lib/src/app_start/standalone/standalone_app_start_handler.dart
```diff
@@ -25,9 +25,11 @@ class StandaloneAppStartHandler {
   AppStartTrace? _trace;
```
denrase: Now that hold the trace on options, can we get rid of this `_trace` ivar or do they have different lifecycles?

packages/flutter/lib/src/app_start/standalone/static_app_start_trace.dart
```diff
@@ -197,19 +250,24 @@ final class StaticAppStartTrace implements AppStartTrace {
...
-    if (_state.isTerminal) return;
+    if (_isFinalizingOrTerminal) return;
+    _finalizing = true;
     _clearFinalTimeout();
 
     final status = SpanStatus.deadlineExceeded();
     _root.status = status;
 
+    await _extensionLifecycle.waitForPendingFinish();
+    if (_state.isTerminal) return;
```
denrase, line 272: We should reset `_finalizing` at all method return points, no?

packages/flutter/lib/src/app_start/standalone/static_app_start_trace.dart
```diff
@@ -259,3 +319,168 @@ final class StaticAppStartTrace implements AppStartTrace {
...
+    if (finishFuture != null) await finishFuture;
+  }
+
+  Future<void> close() async {
+    if (_closed) return;
+    _closed = true;
+
+    final finishFuture = _finishFuture;
+    if (finishFuture != null) {
+      await finishFuture;
```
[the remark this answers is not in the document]
denrase: @buenaflor I'd say this is the whole point, no?

denrase (approved):

packages/flutter/lib/src/app_start/standalone/static_app_start_trace.dart
buenaflor: no, it's set once (one-way) when the final deadline starts draining descendants

packages/flutter/lib/src/app_start/standalone/standalone_app_start_handler.dart
buenaflor: removed it 👍

packages/flutter/lib/src/app_start/standalone/app_start_vitals.dart
buenaflor: yup, it's always teh first frame, renamed it

PR #3924, opened 2026-07-27
perf(flutter): Speed up Android scope sync and replay capture

### LINKED ISSUE #3926: Reduce copying and serialization on the Android JNI path
### Description

The Android JNI path does redundant work on every scope write and every replay frame:

- `AndroidCoreWorker` normalizes each breadcrumb, user, and context twice — once before handing it to the worker isolate, then again inside the `JByteArray` helper.
- `encodeUtf8Json` builds an intermediate UTF-16 `String` via `json.encode` only to re-encode it to UTF-8, instead of the core SDK's single-pass `utf8JsonEncoder`. The read paths likewise decode to a `String` before parsing JSON.
- Replay screenshots are copied through the isolate port on every frame instead of being transferred by ownership.

Separately, non-finite doubles are silently fatal on this path: `json.encode` throws on `Infinity` / `NaN`, and the surrounding `catch` drops the **entire** payload — so a single such value loses the whole context. `normalize` treats every `num` as primitive, so nothing filters it out beforehand.

buenaflor:
## 📜 Description

Three related changes to the Android native path, one per commit and reviewable in that order. Everything touched is `@internal`; there is no public API change.

**Normalize once instead of twice.** `AndroidCoreWorker` normalized every breadcrumb, user, and context twice — once before handing it to the worker isolate, then again inside the `JByteArray` helper. Normalization now happens once at the call site, matching the shape `setContexts` already used.

**Encode in a single pass.** `encodeUtf8Json` now uses the core SDK's `utf8JsonEncoder`, the same `JsonUtf8Encoder` already backing envelopes, logs, spans, and metrics. It writes objects straight to UTF-8 bytes instead of building an intermediate `String` and copying it, and the read paths fuse `Utf8Decoder` with `JsonDecoder` so JSON is parsed from the bytes directly. The return type became `List<int>` because that is what `JsonUtf8Encoder` actually promises — `JByteArray.from` takes an `Iterable<int>`, so no caller cared.

**Transfer replay screenshots by ownership.** Screenshot bytes were copied through the isolate port on every frame; they are now sent as `TransferableTypedData`. Passing the `ByteData` straight in also drops a `buffer.asUint8List()` view that ignored `offsetInBytes` and `lengthInBytes`.

A dead `jsonToJByteArray` in `sentry_native_java.dart` is removed as well. It was only reachable from an integration test and duplicated the worker's conversion under what is now the opposite contract.

### One behavior change

Non-finite doubles now encode as `"Infinity"` / `"NaN"` instead of throwing. Previously `json.encode` threw and the surrounding `catch` dropped the **entire** payload, so a single `double.infinity` in a context silently lost the whole context. `normalize` treats every `num` as primitive, so this was reachable in practice rather than theoretical. The new behavior matches the core SDK's `jsonSerializationFallback`.

## 💡 Motivation and Context

Fixes getsentry/sentry-dart#3926

Breadcrumbs, user updates, and context writes cross the Dart–JNI boundary constantly in a running app, and each one was paying for a redundant tree traversal plus a full UTF-16 `String` that existed only to be re-encoded and thrown away. Replay adds a multi-megabyte copy per frame on top of that.

Measured medians over 7 alternating rounds:

| Path | Payload | Before | After | Speedup |
| -- | -- | -- | -- | -- |
| Scope sync (normalize + encode) | 1.2 KB | 23.6 µs | 13.3 µs | 1.77x |
|  | 23 KB | 410 µs | 239 µs | 1.72x |
|  | 238 KB | 4.43 ms | 2.37 ms | 1.87x |
| Replay hand-off (host → worker) | 1 MB | 228 µs | 86 µs | 2.64x |
|  | 4 MB | 623 µs | 384 µs | 1.62x |
|  | 10 MB | 1188 µs | 902 µs | 1.32x |

Worth calibrating on: these are JIT numbers from `dart run` on macOS arm64, not AOT on a device. The ratios should hold since both arms are dominated by the same VM-level copy and serializer work, but the absolutes will not.

## 💚 How did you test it?

Two tests in `utf8_json_test.dart` cover what this package actually owns: the non-finite double fallback, and that the encoder returns typed data so `JByteArray.from` stays on its bulk-copy path rather than silently degrading to an element-wise copy.

A wider net of seven tests was used while swapping the encoder, written against the old implementation first to prove the new one is byte-identical. It stayed green through the swap and was then trimmed, since the rest were asserting `dart:convert`'s JSON formatting rather than any contract of ours.

Full `packages/flutter` suite passes (912 passed, 3 skipped), with `flutter analyze` and `dart format` clean on both `packages/flutter` and `packages/flutter/example` under Dart 3.12.0.

## :pencil: Checklist

- [X] All tests passing
- [X] No breaking changes

## 🔮 Next steps

* `feat/bump-jni-v1` carries its own copy of this code in a separate `jni_byte_array.dart` under jni 1.0.0. It needs the same treatment and will conflict with this branch.


packages/flutter/lib/src/native/java/sentry_native_java.dart
```diff
@@ -328,10 +328,6 @@ JObject dartToJObject(Object? value) => switch (value) {
       _ => value.toString().toJString()
     };
 
-@visibleForTesting
-JByteArray jsonToJByteArray(Object? value) =>
-    JByteArray.from(encodeUtf8Json(normalize(value)));
```
buenaflor, line 333: this is dead code

packages/flutter/lib/src/native/java/android_core_worker.dart
```diff
@@ -187,28 +187,26 @@ class AndroidCoreWorker {
...
     }
 
-    return _addBreadcrumbFromWorker(client, breadcrumb);
+    return _addBreadcrumbFromWorker(client, normalizedBreadcrumb);
   }
 
   Future<void> _addBreadcrumbFromWorker(
     Worker client,
-    Breadcrumb breadcrumb,
+    Map<String, dynamic> breadcrumb,
```
denrase: l: maybe rename the params with `normalized` prefix as well.

packages/flutter/lib/src/native/java/android_core_worker.dart
```diff
@@ -251,24 +249,23 @@ class AndroidCoreWorker {
...
     }
 
-    return _setUserFromWorker(client, user);
+    return _setUserFromWorker(client, normalizedUser);
   }
 
-  Future<void> _setUserFromWorker(Worker client, SentryUser? user) async {
+  Future<void> _setUserFromWorker(
+    Worker client,
+    Map<String, dynamic>? user,
```
denrase: l: dito

denrase (approved):

PR #3934, opened 2026-07-29
fix(http): Align HTTP client error type and value

### LINKED ISSUE #2914: [sentry_dio] Unhelpful and verbose issue captured following failed HTTP status
### Platform

Dart

### Obfuscation

Enabled

### Debug Info

Enabled

### Doctor

\[✓\] Flutter (Channel stable, 3.29.3, on macOS 15.4.1 24E263 darwin-arm64, locale en-GB) \[4.2s\]<br>    • Flutter version 3.29.3 on channel stable at /Users/Scott/Developer/fvm/versions/stable<br>    • Upstream repository [https://github.com/flutter/flutter.git](<https://github.com/flutter/flutter.git>)<br>    • Framework revision ea121f8859 (4 weeks ago), 2025-04-11 19:10:07 +0000<br>    • Engine revision cf56914b32<br>    • Dart version 3.7.2<br>    • DevTools version 2.42.3

\[✓\] Android toolchain - develop for Android devices (Android SDK version 35.0.0) \[8.7s\]<br>    • Android SDK at /Users/Scott/Library/Android/sdk<br>    • Platform [JAVA-450](https://linear.app/getsentry/issue/JAVA-450/android-report-sdk-size-as-a-comment-in-prs), build-tools 35.0.0<br>    • ANDROID_HOME = /Users/Scott/Library/Android/sdk<br>    • Java binary at: /Users/Scott/Library/Java/JavaVirtualMachines/jdk-17.0.2.jdk/Contents/Home/bin/java<br>      This JDK is specified in your Flutter configuration.<br>      To change the current JDK, run: `flutter config --jdk-dir="path/to/jdk"`.<br>    • Java version OpenJDK Runtime Environment (build 17.0.2+8-86)<br>    • All Android licenses accepted.

\[✓\] Xcode - develop for iOS and macOS (Xcode 16.3) \[1,702ms\]<br>    • Xcode at /Applications/Xcode.app/Contents/Developer<br>    • Build 16E140<br>    • CocoaPods version 1.16.2

\[✓\] Chrome - develop for the web \[33ms\]<br>    • Chrome at /Applications/Google Chrome.app/Contents/MacOS/Google Chrome

\[✓\] Android Studio (version 2024.3) \[32ms\]<br>    • Android Studio at /Users/Scott/Applications/Android Studio.app/Contents<br>    • Flutter plugin can be installed from:<br>      🔨 [https://plugins.jetbrains.com/plugin/9212-flutter](<https://plugins.jetbrains.com/plugin/9212-flutter>)<br>    • Dart plugin can be installed from:<br>      🔨 [https://plugins.jetbrains.com/plugin/6351-dart](<https://plugins.jetbrains.com/plugin/6351-dart>)<br>    • Java version OpenJDK Runtime Environment (build 21.0.6+-13355223-b631.42)

\[✓\] IntelliJ IDEA Ultimate Edition (version 2025.1) \[31ms\]<br>    • IntelliJ at /Users/Scott/Applications/IntelliJ IDEA Ultimate.app<br>    • Flutter plugin can be installed from:<br>      🔨 [https://plugins.jetbrains.com/plugin/9212-flutter](<https://plugins.jetbrains.com/plugin/9212-flutter>)<br>    • Dart plugin can be installed from:<br>      🔨 [https://plugins.jetbrains.com/plugin/6351-dart](<https://plugins.jetbrains.com/plugin/6351-dart>)

\[✓\] Connected device (4 available) \[7.1s\]<br>    • iPhone 16 Plus (mobile)         • 9420837B-BDC8-460F-9FE2-1B060BEC20FB • ios            •<br>      com.apple.CoreSimulator.SimRuntime.iOS-18-4 (simulator)<br>    • macOS (desktop)                 • macos                                • darwin-arm64   • macOS 15.4.1 24E263 darwin-arm64<br>    • Mac Designed for iPad (desktop) • mac-designed-for-ipad                • darwin         • macOS 15.4.1 24E263 darwin-arm64<br>    • Chrome (web)                    • chrome                               • web-javascript • Google Chrome 135.0.7049.115

\[✓\] Network resources \[339ms\]<br>    • All expected network resources are available.

• No issues found!

### Version

8.14.1

### Steps to Reproduce

Initialise Sentry Dio integration and then make an API call that returns a failure status code:

```dart
dio.addSentry(
  captureFailedRequests: false,
  failedRequestStatusCodes: [
    SentryStatusCode.range(500, 599),
    SentryStatusCode(403),
  ],
);
```

### Expected Result

I'm seeing a lot of issues in Sentry with descriptions like the following:

> **Qx**<br>DioException \[bad response\]: This exception was thrown because the response has a status code of 502 and RequestOptions.validateStatus was configured to throw for this status code.<br>The status code of 502 has the following meaning: "Server error - the server failed to fulfil an apparently valid request"<br>Read more about status codes at [https://developer.mozilla.org/en-US/docs/Web/HTTP/Status](<https://developer.mozilla.org/en-US/docs/Web/HTTP/Status>)<br>In order to resolve this exception you typically have either to verify and fix your request code or you have to fix the server code.

Putting aside why we might be getting an HTTP 502 response, it seems to me that this message is quite verbose and that a more useful message might be something like (for example):

> DioException \[bad response\]: Status 502 response calling POST /booking/check

I also don't understand why the issue title is (for example) **Qx**. It seems like obfuscation, but we have uploaded our symbol files and can see accurate call stack details.

### Actual Result

A more useful issue title and body.

None

buenaflor:
## :scroll: Description

HTTP client errors from `package:http`, Dio and Supabase now produce the same short, obfuscation-stable exception the other Sentry SDKs do.

### Issue titles

Dio dumped its own `toString()` into the value — the status code's meaning, an MDN link and generic remediation advice — and its type didn't survive obfuscation:

```diff
- Qx: DioException [bad response]: This exception was thrown because the response has a
- status code of 502 and RequestOptions.validateStatus was configured to throw for this
- status code. […MDN link and remediation advice…]
+ DioException: HTTP Client Error with status code: 502
```

`package:http` and Supabase were already concise but carried a stray `Exception: ` prefix, and their types didn't survive obfuscation either:

```diff
- Qy: Exception: HTTP Client Error with status code: 404
+ SentryHttpClientError: HTTP Client Error with status code: 404
```

Connection-level failures — timeouts, DNS errors, bad certificates — produced no event at all in Dio, because there's no status code to match against. They're now captured, with the specific cause chained rather than in the title, so grouping stays per failure kind instead of splintering per hostname:

```diff
- (no event sent)
+ DioException: HTTP Client Error: connection error
+   └─ SocketException: Failed host lookup: 'example.invalid'
```

### Event payload

| | Before | After |
|---|---|---|
| `contexts.response` (Dio) | absent — the response reached the `beforeSend` hint and stopped there | status code, headers, cookies, body size |
| Response headers | sent unmasked under `sendDefaultPii` | `Authorization`, `Cookie` and friends filtered out |
| `mechanism.handled` | unset | `true` |
| Requests to the DSN | captured like any other | skipped, on an exact host match |
| `failedRequestTargets` (Dio) | matched the `baseUrl`-relative path, so a host target never matched | matched against the resolved URL |

> [!IMPORTANT]
> Issue titles and grouping change for anyone already capturing failed requests, and the newly captured connection failures add event volume. That's why this targets `v10-branch`.

Two deliberate divergences from sentry-javascript: it never captures connection-level failures at all, where Dio now does, matching our own `FailedRequestClient`. And it sets `handled: false` where we set `true` — `handled == false` ends the native session as crashed, and `captureFailedRequests` defaults to on here while js's integration is opt-in, so copying that value would report a crashed session for every captured 5xx.

Closes #2914. 

Unit tests in each affected package. Manually verified on an Android emulator with three new example buttons covering a 404, a connection timeout and a DNS failure.

## :pencil: Checklist

- [x] All tests passing
- [ ] No breaking changes

## :crystal_ball: Next steps

- Cherry-pick the header filtering and the target-matching fix to `main` — neither affects grouping.
- Document that Dio now reports connection-level failures.


packages/dart/lib/src/utils/tracing_utils.dart
```diff
@@ -134,6 +134,25 @@ bool containsTargetOrMatchesRegExp(
...
+    final host = options.parsedDsn.uri?.host;
+    if (host == null || host.isEmpty) {
+      return false;
+    }
+    return url.contains(host);
+  } catch (_) {
+    // The DSN may be unset or unparseable.
+    return false;
+  }
+}
```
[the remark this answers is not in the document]
buenaflor: Good catch, fixed in 8cd78bec81.

`isSentryRequestUrl` now parses the URL and compares hosts for equality rather than testing `contains`, so a DSN on `sentry.example.com` no longer suppresses capture for `not-sentry.example.com`. A relative URL yields an empty host and is kept, since it cannot address the DSN.

sentry-javascript has the same substring flaw (`url.includes(dsn.host)`), but the failure mode here is silent, so being stricter than the reference implementation is worth one `Uri.tryParse` per captured failure. Regression tests added in both `packages/dart` and `packages/dio`.

packages/dio/lib/src/dio_event_processor.dart
```diff
@@ -21,26 +34,58 @@ class DioEventProcessor implements EventProcessor {
...
-    hint.response ??= _responseFrom(dioError);
+    final response = _responseFrom(dioError);
+    hint.response ??= response;
 
     // Don't override just parts of the original request.
     // Keep the original one or if there's none create one.
     event.request = event.request ?? _requestFrom(dioError);
+    // Also on the event, not just the hint: a hint is beforeSend-only and never
+    // serialized, so otherwise the status code, headers and body never arrive.
+    event.contexts.response = event.contexts.response ?? response;
```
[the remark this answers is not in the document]
buenaflor: Correct on every point, fixed in 05120c4aff.

`data` was already built by `_responseFrom` but only ever reached `hint.response`, and hints are never serialized — so assigning that same object to `event.contexts.response` silently put the response body on the wire. With `sendDefaultPii` on that is up to `Hint.maxResponseBodySize` (150 KB) per failed request, which is more than the option asks for and diverges from both `FailedRequestClient` and sentry-javascript.

`_responseFrom` now takes a `withData` flag: the event gets status code, headers, cookies and body size, while the hint keeps the body so `beforeSend` can still read it. Test asserts `contexts.response.data` is null while `hint.response.data` is not.

packages/dio/lib/src/dio_event_processor.dart
```diff
@@ -4,6 +4,16 @@ import 'package:dio/dio.dart';
 import 'package:sentry/sentry.dart';
 import 'dart:convert';
 
+const _failureDescriptions = <DioExceptionType, String>{
```
denrase, line 12: Is it feasible that dio would change exception types or introduce additional ones? We could add alink to the source files in the latter case.

packages/dio/lib/src/dio_event_processor.dart
```diff
@@ -21,26 +31,54 @@ class DioEventProcessor implements EventProcessor {
...
-    for (final exception in event.exceptions ?? []) {
+    for (final exception in event.exceptions ?? <SentryException>[]) {
       final throwable = exception.throwable;
       if (throwable is DioError) {
-        dioError = throwable;
-        break;
+        if (_shouldReplaceValue(throwable)) {
+          exception.value = _valueFrom(throwable);
+        }
+        dioError ??= throwable;
```
denrase, line 45: The behaviour before was that the last `DioError` in the list would be chosen, now it's the first one. Is this intended?

packages/dio/lib/src/failed_request_interceptor.dart
```diff
@@ -29,15 +29,32 @@ class FailedRequestInterceptor extends Interceptor {
...
+      handler.next(err);
+      return;
+    }
+
+    final statusCode = err.response?.statusCode;
+    // A connection-level failure — timeout, DNS error, bad certificate — has no
+    // status code to match against, so there is nothing to filter on. A
+    // cancellation is a deliberate user action, not a failure.
+    final isFailure = statusCode == null
+        ? err.type != DioExceptionType.cancel
```
denrase, line 37: > A cancellation is a deliberate user action, not a failure.

Do we have this on good authority that the ONLY reason for this is deliberate user action.

denrase (approved):

packages/dio/lib/src/failed_request_interceptor.dart
buenaflor: Updated the comment to be more clear, it's only cancelled via `CancelToken`, so it is caller-initiated instead of a transport failure

packages/dio/lib/src/dio_event_processor.dart
buenaflor: Yes, so it's a const map with a `?? type.name` fallback rather than a switch, a new dio type degrades to its enum name instead of breaking the build. Added a link to the enum.

packages/dio/lib/src/dio_event_processor.dart
buenaflor: It's still the first one. The ??= does what the break used to: it only assigns while dioError is null, so the first match wins either way.

The break had to go so the loop keeps walking and normalizes value on every DioException, not just the first

PR #3935, opened 2026-07-30
fix(flutter): Await in-flight replay capture on scheduler stop

### LINKED ISSUE #3923: Session Replay: in-flight capture outlives Vulkan surface teardown → native SIGABRT in ~TextureVK (Android/Impeller)
### Platform

Flutter Mobile Android

### SDK versions

- `sentry_flutter` / `sentry` / `sentry_dio`: **9.24.0**
- Flutter: **3.41.7 stable**, release build (Impeller / Vulkan backend)
- Device: Samsung Galaxy A15 5G (SM-A156E), MediaTek MT6835 / Mali-G57, Android 15

### Configuration

```dart
options.replay.sessionSampleRate = 0.03;
options.replay.onErrorSampleRate = 1.0;
// app wrapped in SentryWidget(child: app)
```

### What happened

A production user hit a native `SIGABRT` **3.000 s after `MainActivity` was destroyed**. The crashing stack is an Impeller Vulkan texture teardown, reached from a `picture.cc` completion lambda being destroyed during a message-loop task flush:

```
android::Looper::pollOnce
fml::MessageLoopAndroid::OnEventFired          (message_loop_android.cc:91)
fml::MessageLoopImpl::RunExpiredTasksNow       (message_loop_impl.cc:139)
fml::MessageLoopImpl::FlushTasks               (message_loop_impl.cc:122)
std::__fl::function<T>::operator=              (function.h:968)
std::__fl::__function::__func<T>::destroy_deallocate (function.h:313)
const::lambda::~lambda                         (picture.cc:244)
sk_sp<T>::~sk_sp                               (SkRefCnt.h:256)
impeller::DlImageImpeller::~DlImageImpeller    (dl_image_impeller.cc:84)
impeller::TextureVK::~TextureVK                (texture_vk.cc:25)
impeller::TextureSourceVK::~TextureSourceVK    (texture_source_vk.cc:11)
impeller::Attachment::~Attachment              (formats.h:660)
...
abort
```

Lifecycle breadcrumbs immediately before it:

```
16:09:26.018  MainActivity paused
16:09:26.592  LOW_MEMORY (TRIM_MEMORY level 40)
16:09:26.601  MainActivity stopped
16:09:26.614  MainActivity destroyed
16:09:29.614  SIGABRT
```

### Why I think this is the replay recorder

I can't attach a reproduction, so this is attribution by elimination — but the elimination is fairly tight:

1. Our app contains exactly **one** `RenderRepaintBoundary.toImage()` call site, on a share-a-post screen. The affected user was a signed-out guest whose entire six-minute session was an onboarding flow; breadcrumbs show they never authenticated and never opened that screen. That path provably did not run.
2. Nothing else in our code calls `toImage` / `toImageSync` / `Picture.toImage`.
3. That leaves the replay recorder as the only in-process producer, and it matches the stack exactly — `lib/src/screenshot/recorder.dart`:
   - L84 `final futureImage = renderObject.toImage(pixelRatio: capture.pixelRatio);`
   - L223 `finalImage = await picture.toImage(width, height);`

   `Picture.toImage` is `Picture::RasterizeToImage`, and the completion lambda in `picture.cc` is what owns the `sk_sp<DlImage>` whose destruction runs `~DlImageImpeller` → `~TextureVK` in the stack above.

### Suspected race

`ScheduledScreenshotRecorder.pause()` / `stop()` (`lib/src/replay/scheduled_recorder.dart` L99–L120) set `_status` and stop the `Scheduler`, but they don't cancel or await a capture that is **already in flight**.

A capture that started just before the lifecycle transition still runs to completion on its own async chain: `await futureImage` → `await image.toByteData()` → `canvas.drawImage` → `picture.toImage(width, height)` → `Screenshot(finalImage)`. Only then does `_onImageCaptured` observe `_status != running` and drop the frame.

The Dart-side `Screenshot.dispose()` does run in the `finally`, but releasing the image only *posts* the GPU-side texture release to the engine's message loop. By the time that task is flushed, the Activity and its Vulkan surface are gone, and destroying `TextureVK` against a dead context aborts inside the Mali driver. The `TRIM_MEMORY 40` immediately beforehand plausibly accelerates the surface teardown and widens the window.

### Suggested direction

On `pause()` / `stop()`, either await in-flight captures before returning (so the textures are released while the context is still alive), or refuse to *start* a capture once the app is no longer `resumed`. Dropping the finished frame at the end is enough for correctness of the replay itself, but not for the GPU resource it already allocated.

### Caveats

Single event, no reproduction, and the native frames come from a release build. I'm reporting it because the code path lines up precisely and the failure mode (deferred GPU resource release racing surface teardown) looks structural rather than device-specific, even if this particular driver is what turns it into an abort.

lucas-zimerman:
## 📜 Description

`Scheduler.stop()` only awaited the delay timer between scheduled runs, not the in-flight capture itself. `ScheduledScreenshotRecorder.pause()`/`stop()` could therefore return while a screenshot capture was still rendering, disposing its image, and releasing its GPU texture on the engine's message loop.

`Scheduler.stop()` now also awaits the currently in-flight capture (`_runningCallback`) before returning.

## 💡 Motivation and Context

Fixes getsentry/sentry-dart#3923. A production user hit a native `SIGABRT` \~3s after `MainActivity` was destroyed, in `~TextureVK` during a message-loop task flush on Android/Impeller (Vulkan backend). The crash was attributed to the session replay recorder: a screenshot capture that was still in flight when the app was backgrounded/destroyed could complete and post its GPU texture release *after* the Activity's Vulkan surface had already been torn down, aborting inside the driver.

`pause()`/`stop()` are invoked from native lifecycle callbacks and are awaited there, but the underlying `Scheduler.stop()` wasn't actually waiting for the capture to finish — only for a delay timer that's a loose proxy for it. This closes that gap so GPU resources are released while the surface is still alive.

## 💚 How did you test it?

* Added a regression test (`test/replay/scheduler_test.dart`) that reproduces the race: it starts a capture gated on an uncompleted `Completer`, calls `stop()`, and asserts `stop()` does not resolve until the capture completes. This test fails against the old implementation and passes with the fix.
* Ran the full `test/replay/` suite plus `flutter analyze` — all green.
* Sanity-tested the example app (replay enabled at 100% sample rate) on an Android emulator: cycled backgrounding, a tight background→foreground race, and a hard `force-stop` while a replay segment was actively encoding. No crashes, ANRs, or hangs; `pause()` returned promptly in all cases.
* Note: the emulator's Impeller backend resolves to OpenGLES rather than Vulkan (engine-level device detection avoids Vulkan on virtual GPUs), so the exact native `TextureVK` path from the report can't be reproduced on an emulator — this fix was verified functionally rather than via a native crash repro. Atm I don't have an Android device with vulkan support for testing this.

## :pencil: Checklist

- [X] All tests passing
- [X] No breaking changes

## 🔮 Next steps

None planned; this is a scoped fix to `Scheduler.stop()`.


buenaflor (commented):
LGTM, but before I approve I wanna test it on my physical Android device

buenaflor:
Sry for the delay, I'll properly test it tomorrow

lucas-zimerman:
> Sry for the delay, I'll properly test it tomorrow

No rush, I ordered an android device to help with tests and it should arrive on day 20

packages/flutter/lib/src/replay/scheduler.dart
```diff
@@ -36,6 +36,11 @@ class Scheduler {
     if (scheduled != null) {
       await scheduled;
     }
+    // The scheduled future above only tracks the delay timer between runs,
+    // not the callback itself, so we also need to await any run that's
+    // currently in flight to make sure its GPU resources are released
+    // before the caller (e.g. the app lifecycle teardown) proceeds.
+    await _runningCallback;
```
buenaflor: What's missing is a timeout

```dart
await _runningCallback?.timeout(_stopTimeout, onTimeout: () {});
_runningCallback = null;
```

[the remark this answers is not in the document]
lucas-zimerman: duplicated comment

lucas-zimerman: fair point, applied the timeout + tests

buenaflor (commented):
Pls check if the AI reviews are legitimate

lucas-zimerman:
> Pls check if the AI reviews are legitimate

They were valid, fixed them on https://github.com/getsentry/sentry-dart/pull/3935/commits/7ea65be0e9118ac85471dcde38c019ab3b64f349

buenaflor:
I am not 100% sure to be honest if this is actually mitigating / fixing the core issue. Check my comment in the issue, I think this is a Flutter bug

lucas-zimerman:
@buenaflor  It may be unrelated to that flutter bug, but still, would you rather me to block this PR/issue until a fix on Flutter is released?

buenaflor:
@lucas-zimerman okay I think this PR should land but I'm unsure if 2s timeout is too aggressive

buenaflor:
Do you think it also makes sense to check if we can capture at all? e.g

from this:
```dart
_scheduler = Scheduler(
  frameDuration,
  (_) => capture(_onImageCaptured),
  _addPostFrameCallback,
);
```

to this:

```dart
// in scheduled_recorder.dart
_scheduler = Scheduler(
  frameDuration,
  (_) async {
    if (!_canStartCapture) {
      internalLogger.debug(() =>
          '$logName: skipping capture, app is not resumed.'));
      return;
    }
    await capture(_onImageCaptured);
  },
  _addPostFrameCallback,
);

/// The engine's surface can be torn down as soon as the app leaves
/// [AppLifecycleState.resumed]. A capture started after that point can
/// outlive the surface backing its GPU texture, and releasing the texture
/// against a dead Vulkan context aborts in the driver (#3923).
///
/// Captures already in flight when the transition happens are handled by
/// [Scheduler.stop]; this only prevents new ones.
bool get _canStartCapture {
  final state = options.bindingUtils.instance?.lifecycleState;
  // null until the first lifecycle message arrives, i.e. during startup,
  // where the app is foreground by definition.
  return state == null || state == AppLifecycleState.resumed;
}
```

packages/flutter/lib/src/replay/scheduled_recorder.dart
```diff
@@ -81,7 +95,14 @@ class ScheduledScreenshotRecorder extends ReplayScreenshotRecorder {
 
     _scheduler = Scheduler(
       frameDuration,
-      (_) => capture(_onImageCaptured),
+      (_) async {
+        if (!_canStartCapture) {
+          options.log(SentryLevel.debug,
+              '$logName: skipping capture, app is not resumed.');
```
buenaflor: let's use the `internalLogger` instead

buenaflor (changes requested):

packages/flutter/lib/src/replay/scheduled_recorder.dart
lucas-zimerman: Fixed on https://github.com/getsentry/sentry-dart/pull/3935/commits/ae9711c6adc4de58dc9cbc0d85ad2e1d95c0c096

packages/flutter/lib/src/replay/scheduler.dart
```diff
@@ -31,11 +41,21 @@ class Scheduler {
...
     }
+    // The scheduled future above only tracks the delay timer between runs,
+    // not the callback itself, so we also need to await any run that's
+    // currently in flight to make sure its GPU resources are released
+    // before the caller (e.g. the app lifecycle teardown) proceeds. If it
+    // times out, drop the reference too: a stalled callback's future never
+    // completes, and leaving it in place would make _runAfterNextFrame()
+    // wait on it forever on the next start(), permanently stalling capture.
+    await _runningCallback?.timeout(_stopTimeout, onTimeout: () {});
+    _runningCallback = null;
```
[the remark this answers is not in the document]
lucas-zimerman: sounds good! fixed on https://github.com/getsentry/sentry-dart/pull/3935/commits/7d0e7edc8ed35b23464c24c496be66e58ba9f39f

buenaflor (approved):

PR #3945, opened 2026-08-04
chore: Add melos commands for example app and e2e tests

### LINKED ISSUE #3660: Add more convenient melos commands
### Description

E.g building the Flutter example, running e2e tests and so on

hkarmoush:
## :scroll: Description

Adds two melos scripts:

- `melos run example:flutter -- <platform>` — builds and runs the Flutter example app (delegates to the existing `packages/flutter/example/run.sh`, so there's still one source of truth for the build logic).
- `melos run e2e:dart` — runs the Dart e2e suite against sentry.io (mirrors `.github/workflows/e2e_dart.yml`, requires `SENTRY_AUTH_TOKEN_E2E` locally).

Also documents the existing and new scripts in a "Common commands" table in `CONTRIBUTING.md`, and fixes a small bug in `run.sh` where calling it with no arguments crashed on an unbound variable (`set -u`) instead of printing the usage message.

Closes #3660

Building the example app and running the e2e suite both require remembering raw shell commands that live in a few different places (`run.sh`, CI workflow files). This wraps them behind `melos run`, which is discoverable via `melos run --list`.

I kept this scoped to the two examples the issue calls out rather than guessing at everything "and so on" might cover.

- `melos run example:flutter -- ""` to confirm the script forwards args correctly and the run.sh usage guard fires as expected.
- `melos run e2e:dart` to confirm it resolves dependencies and fails fast with a clear error when `SENTRY_AUTH_TOKEN_E2E` isn't set (same behavior as CI without the secret).
- `bash -n packages/flutter/example/run.sh` to sanity check the shell syntax.

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps

If there's appetite for it, a follow-up could tighten `test:dart`/`test:flutter`/`analyze:*` to match the flags CI actually uses (randomized test ordering, web/wasm runs, `--fatal-infos`), since those currently drift from what CI checks.

### DIFF
--- a/melos.yaml
+++ b/melos.yaml
@@ -58,7 +58,8 @@ scripts:
     run: cd packages/flutter/example && ./run.sh
     description: |
       Build & run the Flutter example in release mode with obfuscation and symbol upload.
-      Usage: melos run example:flutter -- <ios|android|web|macos|windows|linux>
+      Usage: melos run example:flutter -- <ios|android|web|macos|linux>
+      Not supported on Windows: melos runs scripts through cmd.exe there, which can't execute run.sh.
 
   e2e:dart:
     run: cd e2e_test && dart pub get && dart run --define=SENTRY_ENVIRONMENT=e2e
--- a/melos.yaml
+++ b/melos.yaml
@@ -54,6 +54,20 @@ scripts:
       flutter: true
       dirExists: test
 
+  example:flutter:
+    run: cd packages/flutter/example && ./run.sh
+    description: |
+      Build & run the Flutter example in release mode with obfuscation and symbol upload.
+      Usage: melos run example:flutter -- <ios|android|web|macos|windows|linux>
+
+  e2e:dart:
+    run: cd e2e_test && dart pub get && dart run --define=SENTRY_ENVIRONMENT=e2e
+    description: |
+      Run the Dart e2e test against sentry.io. Requires SENTRY_AUTH_TOKEN_E2E.
+      Keep in sync with .github/workflows/e2e_dart.yml.
+    env:
+      SENTRY_DIST: '1'
+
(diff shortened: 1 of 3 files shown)


melos.yaml
```diff
@@ -54,6 +54,21 @@ scripts:
...
+      Usage: melos run example:flutter -- <ios|android|web|macos|linux>
+      Not supported on Windows: melos runs scripts through cmd.exe there, which can't execute run.sh.
+
+  e2e:dart:
+    run: cd e2e_test && dart pub get && dart run --define=SENTRY_ENVIRONMENT=e2e
+    description: |
+      Run the Dart e2e test against sentry.io. Requires SENTRY_AUTH_TOKEN_E2E.
+      Keep in sync with .github/workflows/e2e_dart.yml.
+    env:
+      SENTRY_DIST: '1'
```
buenaflor: we can remove this, we'll remove this E2E test at some point

hkarmoush: Good call, dropped the `e2e:dart` melos entry — no point adding a convenience wrapper for something that's going away.

melos.yaml
```diff
@@ -54,6 +54,13 @@ scripts:
       flutter: true
       dirExists: test
 
+  example:flutter:
+    run: cd packages/flutter/example && ./run.sh
```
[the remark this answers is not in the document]
hkarmoush: Verified empirically: `melos run example:flutter -- <platform>` does forward the arg through to `run.sh` correctly (tested with a dummy platform name that hits the "isn't supported" branch — $1 was populated as expected). melos's `run:` scripts do pass `--` args through to the underlying shell command. No fix needed here.

PR #3947, opened 2026-08-05
fix(replay): Stop network detail capture leaking into non-replay events

### LINKED ISSUE #3900: Replay HTTP body capture leaks into non-replay breadcrumbs and native crash events
### Problem

When `enableReplayNetworkDetailsCapturing` is on, captured HTTP request/response
bodies (up to `maxBodySize`, ~150KB each) are attached to the shared `http`
breadcrumb produced by `BreadcrumbClient`. That breadcrumb isn't replay-specific
— it rides along with every subsequent error and transaction event, not just
replay recordings.

Because Flutter syncs breadcrumbs into the native scope, Android/iOS crash
events captured natively also pick up these bodies, even though the feature
is meant to enrich Session Replay only.

### Impact

Bounded: the feature is opt-in (`enableReplayNetworkDetailsCapturing`) and
further gated by `networkDetailAllowUrls`, so nothing leaks unless a user
explicitly enables and allowlists it. But once enabled, the leak is broader
than intended.

### Fix

Needs a replay-only side channel for this data (e.g. via `Hint`) instead of
the shared breadcrumb, which requires new native bridge plumbing on
Android/iOS that doesn't exist today.

### Related

Follow-up from #3875, which adds request/response header+body capture for
replay.

hkarmoush:
## :scroll: Description

Session Replay's opt-in HTTP body/header capture attached its data to the shared HTTP breadcrumb, which then rode along into every subsequent Dart-captured error/transaction event and, on Android, into native crash events too — not just replay recordings.

Request/response detail is now delivered via a `Hint` instead of `breadcrumb.data`, so it's never stored on the `Scope`. On Android it reaches the replay breadcrumb converter through a new side channel (`ReplayNetworkDetailCache`, correlated by an opaque `replay_request_id`) that never touches the native `Scope`, closing the crash-event leak without touching the shipped Android replay rendering. iOS/FFI/Web never see the detail at all now, closing a zero-benefit leak there too (iOS doesn't render it in replay).

Fixes #3900.

- Dart: updated/added unit tests for `BreadcrumbClient`, `Scope`, `NativeScopeObserver`, and `MockScopeObserver` covering the new `Hint`-based path.
- Android/Kotlin: **not compiled or tested in this environment** (no Android SDK/JDK/cmake available), including a hand-authored addition to the auto-generated JNI bindings in `lib/src/native/java/binding.dart`. The changes mirror existing generated/native patterns exactly but need verification via `scripts/generate-jni-bindings.sh` regeneration and a real device/emulator run before this is ready to ship — opening as a draft until that's done.

## :pencil: Checklist
- [ ] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps

- Regenerate/verify JNI bindings with `scripts/generate-jni-bindings.sh` and confirm on a real Android device/emulator.
- Convert to ready for review once native side is verified.

### DIFF
--- a/packages/dart/lib/src/http_client/breadcrumb_client.dart
+++ b/packages/dart/lib/src/http_client/breadcrumb_client.dart
@@ -4,6 +4,8 @@ import 'package:meta/meta.dart';
 import '../protocol.dart';
 import '../hub.dart';
 import '../hub_adapter.dart';
+import '../hint.dart';
+import '../type_check_hint.dart';
 import '../utils/breadcrumb_log_level.dart';
 import '../utils/url_details.dart';
 import '../utils/http_sanitizer.dart';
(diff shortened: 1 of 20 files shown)


buenaflor:
Let's discuss an approach in the issue first please. AFAIK this requires changes through the native SDKs as well and cannot be entirely fixed through Flutter

PR #3950, opened 2026-08-06
fix(dart): Fix inApp detection for web stacktraces

### LINKED ISSUE #3510: Fix Flutter Web stacktraces not being correctly set as inApp
### Description

stacktraces on Flutter Web should be applied with default stacktrace rules so they are correctly marked as inApp/not inApp 

e.g frames coming from `.pub-cache` should generally be not inapp

hkarmoush:
## :scroll: Description

On Flutter Web, stack frames never carry `dart:`/`package:` URIs the way VM stack frames do — the browser reports plain `https://…` (or `dart-sdk/lib/...`, `packages/<pkg>/...`) URLs instead. `SentryStackTraceFactory._isInApp` only knew how to classify `dart:`/`package:` schemes, so it always fell through to `considerInAppFramesByDefault` (`true`) on web. In practice, that meant **every** frame on Flutter Web — your own code, the Flutter framework, and every pub dependency — was marked `in_app: true`.

This PR:

- Normalizes DDC (`flutter run -d chrome`) frame URIs served under `packages/<pkg>/...` and `dart-sdk/lib/...` back into `package:pkg/...` / `dart:...` shapes before classification runs. This makes the existing `dart:`/`package:flutter` checks work on web for the first time, and also makes `inAppIncludes`/`inAppExcludes` usable on web (previously silently a no-op, since they key off `Frame.package`, which requires a `package:` scheme). As a side effect, `abs_path` for these frames now keeps the full path instead of being truncated to just the file's basename, and `module` gets populated when `includeModuleInStackTrace` is on.
- Adds a small denylist for known Flutter/engine web bootstrap assets (`flutter.js`, `canvaskit.js`, `dart_sdk.js`, etc.) so those are never marked in-app either.
- Fixes a related, separate bug found while investigating: `package:stack_trace`'s V8 parser assumes the first line of a stack trace string is an exception message and unconditionally skips it. dart2wasm's `StackTrace.toString()` has no such header line, so the parser was silently dropping the first (and often only) real frame on `flutter build web --wasm`. Fixed by detecting that shape and prepending a synthetic header line before parsing.

**Known limitation, called out explicitly rather than silently punted:** in release/profile web builds (dart2js/dart2wasm), the entire app plus every pub dependency compile into a single bundle file, so every frame shares the exact same URL — there is no per-frame signal left for a client-side fix to separate app code from dependency code in that mode. The denylist above is the honest ceiling of what's achievable there. A more complete fix for release-mode `.pub-cache` frames (the example in the original issue) likely also needs a change to Sentry's server-side stack trace enhancer rules, since the existing `.pub-cache` rule is gated to `family:native` and never fires for web's `family:javascript` frames — tracked separately, out of scope for this SDK-side PR.

Fixes #3510. Flutter Web users had no working in-app/not-in-app grouping signal at all — every frame, including framework and third-party code, showed up as in-app, which pollutes Sentry's default grouping and suspect-frame selection for web issues.

Added test coverage in `packages/dart/test/stack_trace_test.dart`:
- DDC `packages/<pkg>/...` and `dart-sdk/lib/...` frame normalization, including that `inAppIncludes`/`inAppExcludes` now apply, `abs_path` keeps the full path, and `module` gets populated.
- The release-mode denylist, plus a pinned regression test confirming a shared-bundle app frame still correctly falls back to `considerInAppFramesByDefault` (documents the deliberate limitation rather than hiding it).
- The dart2wasm first-frame-loss fix, plus a control test confirming a normal V8 trace with a real message header is unaffected.

All 28 tests in `stack_trace_test.dart` pass, along with downstream consumers (`sentry_client_test.dart`, `sentry_exception_factory_test.dart`, `load_dart_debug_images_integration_test.dart`).

**Note:** a full `dart test` run in this environment surfaces ~39 pre-existing failures in unrelated files (`sentry_tracer_test.dart`, `sentry_transaction_test.dart`, etc.), all from the same root cause (`MockHub.configureScope` unimplemented in `packages/_sentry_testing`). Verified via `git stash` that these fail identically without this PR's changes — pre-existing environment issue, not a regression from this change.

## :pencil: Checklist
- [x] All tests passing (see note above re: pre-existing unrelated failures)
- [x] No breaking changes to the public API

## :crystal_ball: Next steps

- File a follow-up against `getsentry/sentry` to extend the `.pub-cache` / `packages/flutter/**` stack trace enhancer rules to `family:javascript` (currently `family:native`-only), which is what actually resolves the `.pub-cache` example from the original issue for release-mode web events.

## Changelog Entry

Fixes Flutter Web stack frames not being correctly marked as `in_app`/not-`in_app`. **Note for reviewers:** this is a behavior change for existing Flutter Web users — frames that were previously (incorrectly) always `in_app: true` will now often be correctly marked `in_app: false` in debug-mode-style traces, which can shift Sentry's default grouping/culprit-frame selection for already-triaged issues. Worth flagging in release notes rather than treating as a silent patch.

### DIFF
--- a/packages/dart/lib/src/sentry_stack_trace_factory.dart
+++ b/packages/dart/lib/src/sentry_stack_trace_factory.dart
@@ -61,6 +61,24 @@ class SentryStackTraceFactory {
     );
   }
 
+  // `package:stack_trace`'s V8 trace parser unconditionally treats the
+  // first line as an exception-message header and skips it. dart2wasm's
+  // `StackTrace.toString()` - unlike a real V8 JS error - never includes
+  // that header, so the parser silently drops the first (often only) real
+  // frame. Detect that shape and prepend a synthetic header line to absorb
+  // the skip instead.
+  static final _v8FrameLineStart = RegExp(r'^\s{0,4}at ');
+
+  String _ensureV8MessageHeader(String stackTrace) {
+    final firstNewline = stackTrace.indexOf('\n');
+    final firstLine =
+        firstNewline == -1 ? stackTrace : stackTrace.substring(0, firstNewline);
+    if (_v8FrameLineStart.hasMatch(firstLine)) {
+      return '\n$stackTrace';
+    }
+    return stackTrace;
+  }
+
   _StackInfo _parseStackTrace(dynamic stackTrace) {
     if (stackTrace is Chain) {
       return _StackInfo(stackTrace.traces);
@@ -87,8 +105,9 @@ class SentryStackTraceFactory {
       //     #01 abs 000000723d637527 _kDartIsolateSnapshotInstructions+0x1e5527
 
       final startOffset = _frameRegex.firstMatch(stackTrace)?.start ?? 0;
-      final chain = Chain.parse(
-          startOffset == 0 ? stackTrace : stackTrace.substring(startOffset));
+      final trimmed =
+          startOffset == 0 ? stackTrace : stackTrace.substring(startOffset);
+      final chain = Chain.parse(_ensureV8MessageHeader(trimmed));
       final info = _StackInfo(chain.traces);
       info.buildId = buildIdRegex.firstMatch(stackTrace)?.group(1);
       info.baseAddr = _baseAddrRegex.firstMatch(stackTrace)?.group(1);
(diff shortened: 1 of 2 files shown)


buenaflor (commented):
I haven't taken a look in detail but in a minified build we have to report all of the frames as inApp anyway, and serverside rules mark known frames as not inApp

hkarmoush:
@buenaflor Good callout, and to confirm — this PR doesn't change that. In release/profile web builds (dart2js/dart2wasm) every frame still falls back to `considerInAppFramesByDefault` (`true`) here: https://github.com/getsentry/sentry-dart/blob/100d916283139d5ec0901e31c4f92612dd13908a/packages/dart/lib/src/sentry_stack_trace_factory.dart#L247-L282 — there's no per-frame signal left once app + deps are compiled into a single bundle, so this PR can't (and doesn't try to) separate them client-side. The only thing added for that mode is a small denylist for known Flutter/engine bootstrap assets (`flutter.js`, `canvaskit.js`, etc.) so those specifically don't get marked in-app.

One nuance worth flagging on the "serverside rules mark known frames as not inApp" part: I checked, and today that doesn't actually fire for web. The existing `.pub-cache` / `packages/flutter/**` enhancer rule on the server is gated to `family:native`, so it never matches web's `family:javascript` frames — meaning minified web frames currently stay `in_app: true` end-to-end, PR or no PR. That's the gap called out in the PR description's "Next steps" as a separate follow-up against `getsentry/sentry`. Happy to file that issue now if useful context for review.

This PR's actual effect is scoped to debug-mode web (`flutter run -d chrome`), where DDC serves `packages/<pkg>/...` / `dart-sdk/lib/...` paths per-frame — those now normalize back into `package:`/`dart:` shapes so the existing classification logic (and `inAppIncludes`/`inAppExcludes`) can work on web for the first time. Test coverage for both the debug-mode fix and the release-mode fallback/denylist is in `packages/dart/test/stack_trace_test.dart`.

buenaflor:
> The existing .pub-cache / packages/flutter/** enhancer rule on the server is gated to family:native, so it never matches web's family:javascript frames

that's what this issue is about and we need to update Sentry server side rules, I updated the issue description.

I'll double check if the PR is worth doing for debug mode

buenaflor:
afaict newer Flutter versions should already have proper debug web paths and wasm is a separate issue and we dont officially support it yet. 

if you see any specific bug, please open a separate issue and let's discuss it there

Issue #3738, opened 2026-06-03
[v10] Migrate sentry_flutter Android to Built-in Kotlin (AGP 9+)


chems17:
I am using `sentry_flutter` **9.21.0** in a Flutter app with **AGP 9.0** and `android.builtInKotlin=false` (temporary Flutter workaround).

The Android implementation still applies `kotlin-android` in its Gradle build file, so Flutter lists `sentry_flutter` among plugins that apply the Kotlin Gradle Plugin (KGP):

```
WARNING: Your app uses the following plugins that apply Kotlin Gradle Plugin (KGP): ... sentry_flutter ...
Future versions of Flutter will fail to build if your app uses plugins that apply KGP.
```

Starting with AGP 9.0, built-in Kotlin is the default; applying KGP explicitly will eventually break builds when Flutter removes its temporary KGP compatibility layer (and when `android.builtInKotlin=false` is no longer supported in AGP 10).

**Request:** Please migrate the Android module(s) for `sentry_flutter` to [built-in Kotlin](https://docs.flutter.dev/release/breaking-changes/migrate-to-built-in-kotlin/for-plugin-authors).

**References:**
- Flutter tracking: https://github.com/flutter/flutter/issues/181383
- Plugin author guide: https://docs.flutter.dev/release/breaking-changes/migrate-to-built-in-kotlin/for-plugin-authors
- Related ecosystem work: https://github.com/flutter/flutter/issues/187261

Thank you!

buenaflor:
Thanks for the issue, we'll have a look

cc @denrase 

since we support min Flutter `3.24.0` we have to follow this https://docs.flutter.dev/release/breaking-changes/migrate-to-built-in-kotlin/for-plugin-authors#supporting-flutter-versions-earlier-than-3-44

buenaflor:
Unfortunately this is blocked by `jni`, since we depend on their package we'll have to wait until they also update their plugin.

rekire:
Just for the record the related issue is here: https://github.com/dart-lang/native/issues/3415

CharlieMacnamara:
Hi team, I'm another user of sentry_flutter who's affected by this. My app is currently pinned to AGP 9.x with `android.builtInKotlin=false` because of this dependency.

Glad to see this is targeted for v10 — is there an ETA for when v10 might release? In the meantime, happy to help test if needed.

buenaflor:
[...]

PR #3961, opened 2026-08-12
build(flutter): Migrate Android plugin to built-in Kotlin

buenaflor:
## :scroll: Description

Follows the Flutter [built-in Kotlin migration guide for plugin authors](https://docs.flutter.dev/release/breaking-changes/migrate-to-built-in-kotlin/for-plugin-authors).

`packages/flutter` is the only package in the monorepo with Android native code. Its `android/build.gradle` now:

- no longer applies `kotlin-android` itself,
- no longer carries its own KGP `1.8.0` `buildscript` classpath,
- replaces `kotlinOptions` with a top-level `kotlin { compilerOptions { jvmTarget = JvmTarget.JVM_1_8 } }`,
- drops the explicit `kotlin-stdlib-jdk8` dependency, which was pinned to the now-removed `kotlin_version` and is added automatically by KGP.

The JVM target stays at 1.8 to match `compileOptions` in the same module — AGP fails the build if the Java and Kotlin targets diverge within a module, so raising it is a separate decision.

No `pubspec.yaml` change was needed: `sentry_flutter` already requires Flutter `>=3.44.0` and Dart `>=3.12.0`, which is exactly the floor the guide asks for. That is why this takes the guide's simple path rather than the backwards-compatible variant with the `agpMajor < 9` check.

Drive-by: `packages/flutter/microbenchmarks/android/settings.gradle` pinned KGP `1.8.22`, below the 2.0.0 minimum that Flutter 3.44 hard-errors on, so it is bumped to `2.0.21`. Note that project also pins AGP `8.1.0` while Flutter 3.44 errors below AGP `8.6.0`, so its Android build stays broken until AGP is bumped too — pre-existing, and CI only runs Dart analysis there.

Flutter is moving Kotlin compilation into AGP. Flutter 3.44 already prints an error-level warning naming `sentry_flutter` as a plugin that applies KGP, and future Flutter versions will fail the build outright for apps depending on such plugins. Migrating now keeps the SDK usable as apps move to AGP 9 and `android.builtInKotlin=true`.

Closes https://github.com/getsentry/sentry-dart/issues/3738

Verified against the Flutter 3.44.0 SDK source that this leaves the plugin buildable rather than assuming the guide holds:

- `FlutterPluginUtils.libPluginRegexGroovy` still matches our `apply plugin: 'com.android.library'` line while the KGP regexes no longer match anything, so Flutter classifies the project as a library without KGP and calls `pluginManager.apply("kotlin-android")` on our behalf.
- That happens inside `addFlutterTasks`, which runs eagerly while the app project's `plugins` block is evaluated, so KGP is applied before the plugin project's build script is evaluated and the `kotlin { }` extension exists by the time we reach it.

No local Android build was run; CI is the first build signal. `kotlin-version-compatibility.yml` is the job to watch — it builds the example APK against KGP 2.0.21 and 2.3.0, and after this change those versions compile the plugin's Kotlin too, instead of the 1.8.0 it was previously pinned to. That is the main risk in this diff: our Kotlin sources get compiled under a 2.x language version for the first time.

## :pencil: Checklist
- [ ] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps

The example app, `min_version_test`, and `microbenchmarks` are apps and still apply KGP, which the guide permits. Migrating them needs AGP 9 + Flutter 3.47 + `android.builtInKotlin=true`, and `min_version_test` cannot move at all while our floor is Flutter 3.44.

## Changelog Entry

Migrate the Android plugin to Flutter's built-in Kotlin. `sentry_flutter` no longer applies the Kotlin Gradle Plugin itself, which silences Flutter 3.44's warning about plugins applying KGP and unblocks apps moving to AGP 9.

### DIFF
--- a/packages/flutter/android/build.gradle
+++ b/packages/flutter/android/build.gradle
@@ -1,5 +1,4 @@
 buildscript {
-    ext.kotlin_version = '1.8.0'
     repositories {
         google()
         mavenCentral()
@@ -7,7 +6,6 @@ buildscript {
 
     dependencies {
         classpath 'com.android.tools.build:gradle:7.4.2'
-        classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
     }
 }
 
@@ -19,7 +17,6 @@ rootProject.allprojects {
 }
 
 apply plugin: 'com.android.library'
-apply plugin: 'kotlin-android'
 
 android {
     compileSdkVersion 36
@@ -56,15 +53,17 @@ android {
         sourceCompatibility = JavaVersion.VERSION_1_8
         targetCompatibility = JavaVersion.VERSION_1_8
     }
-    kotlinOptions {
-        jvmTarget = JavaVersion.VERSION_1_8
+}
+
+kotlin {
+    compilerOptions {
+        jvmTarget = org.jetbrains.kotlin.gradle.dsl.JvmTarget.JVM_1_8
     }
 }
 
 dependencies {
     api 'io.sentry:sentry-android:8.51.0'
     debugImplementation 'io.sentry:sentry-spotlight:8.51.0'
-    implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk8:$kotlin_version"
 
     // Required -- JUnit 4 framework
     testImplementation "junit:junit:4.13.2"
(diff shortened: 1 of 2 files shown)


denrase (approved):
LGTM 👍

PR #3970, opened 2026-08-14
fix(flutter): Support AGP 9's built-in Kotlin

TheoGermain:
AGP 9 makes the `org.jetbrains.kotlin.android` plugin redundant and rejects it outright at build time:

```
Failed to apply plugin 'org.jetbrains.kotlin.android'
The 'org.jetbrains.kotlin.android' plugin is no longer required for Kotlin support since AGP 9.0.
```

Guards the plugin application behind an AGP-major-version check and moves `jvmTarget` to the `compilerOptions` DSL, which works with both the standalone plugin (AGP < 9) and AGP 9's built-in Kotlin support ([reference](https://kotl.in/gradle/agp-built-in-kotlin)).

Tested against a Flutter 3.47 / AGP 9.3.0 app — builds cleanly with this change, fails without it.

### DIFF
--- a/packages/flutter/android/build.gradle
+++ b/packages/flutter/android/build.gradle
@@ -64,10 +64,20 @@ android {
         sourceCompatibility = JavaVersion.VERSION_1_8
         targetCompatibility = JavaVersion.VERSION_1_8
     }
+}
+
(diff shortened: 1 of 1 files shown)


buenaflor (approved):
I made some adjustments, but LGTM now (CI failing is due to missing permissions in e2e test)

PR #3971, opened 2026-08-17
fix(dart): Record client report outcome for dropped spans

### LINKED ISSUE #3751: Record client report outcomes for spans dropped by the telemetry buffer
## Description

#3745 wired the new `onDrop` callback of `InMemoryTelemetryBuffer` for the **log** buffer so that oversized or unencodable logs record client report outcomes (`buffer_overflow` / `internal_sdk_error`).

The **span** buffer (`GroupedInMemoryTelemetryBuffer` created in `InMemoryTelemetryProcessorIntegration._createSpanBuffer`) does not pass `onDrop`, so spans dropped by the buffer (item exceeds `maxBufferSizeBytes`, or the encoder throws) are silently lost without any discarded-event outcome.

## Expected

Spans dropped by the buffer record a client report outcome with the `span` data category:
- `buffer_overflow` when the encoded item exceeds the buffer limit
- `internal_sdk_error` when encoding fails

## Notes

- The `onDrop` infrastructure already exists in `in_memory_buffer.dart`; this is mostly wiring in `processor_integration.dart` analogous to the log buffer.
- See https://develop.sentry.dev/sdk/telemetry/client-reports/ for discard reasons.

Follow-up to #3745.

lucas-zimerman:
## :scroll: Description
Wires the `onDrop` callback of the span buffer (`GroupedInMemoryTelemetryBuffer` created in `InMemoryTelemetryProcessorIntegration._createSpanBuffer`) so that spans dropped by the buffer record a client report outcome, mirroring what #3745 already did for the log buffer.

- `buffer_overflow` when the encoded item exceeds the buffer limit
- `internal_sdk_error` when encoding fails

Both are recorded against the `span` data category via `recorder.recordLostEvent`. Spans have no `span_byte` category (unlike logs/metrics), so no byte-size outcome is recorded.

Previously, spans dropped by the buffer were silently lost without any discarded-event outcome, unlike logs and metrics which already had this wired up.

Fixes #3751

Unit tests. Added a red test (`oversized span records buffer overflow client report`) confirming the missing wiring before the fix, now passing.

I did not add an equivalent "unencodable span" integration test: `RecordingSentrySpanV2` is a `base class` with only a private constructor, so it can't be subclassed from test code to force `toJson()` to fail (unlike `SentryLog`/`SentryMetric`), and the encoder's JSON fallback stringifies otherwise-unencodable values rather than throwing. This matches the existing precedent in the same test file, where the log buffer only has an oversized-case test as well. The `encodeFailed` switch arm is implemented and structurally identical to the already-tested log/metric arms, and the underlying buffer-level `encodeFailed` → `onDrop` mechanism is generically covered in `buffer_test.dart`.

## :pencil: Checklist
- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps
None.

### DIFF
--- a/packages/dart/lib/src/telemetry/processing/processor_integration.dart
+++ b/packages/dart/lib/src/telemetry/processing/processor_integration.dart
@@ -2,6 +2,7 @@ import 'package:meta/meta.dart';
 
 import '../../../sentry.dart';
 import '../../client_reports/discard_reason.dart';
+import '../../transport/data_category.dart';
 import '../../utils/internal_logger.dart';
 import 'in_memory_buffer.dart';
 import 'processor.dart';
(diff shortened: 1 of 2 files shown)


buenaflor (commented):
~we're missing a unit test for the encode error case~ NVM, read the PR description

buenaflor (approved):

PR #3972, opened 2026-08-18
fix(flutter): Narrow Android R8 keep rule from io.sentry.** to bound classes

### LINKED ISSUE #3933: Broad R8 keep rule for io.sentry prevents optimization of the entire Sentry package
### Platform

Flutter Mobile Android

### Obfuscation

Enabled

### Debug Info

Enabled

### Doctor

Doctor summary (to see all details, run flutter doctor -v):
[✓] Flutter (Channel stable, 3.44.8, on macOS 26.5.2 25F84 darwin-arm64, locale en-US)
[✓] Android toolchain - develop for Android devices (Android SDK version 36.0.0)
[✓] Xcode - develop for iOS and macOS (Xcode 26.6)
[✓] Network resources

• No issues found!

### Other Error Monitoring Solution

Yes

### Other Error Monitoring Solution Name

Firebase crashlytics

### Version

9.25.0

### Steps to Reproduce

1. Create or use a Flutter application with `sentry_flutter` enabled.

2. Enable R8 shrinking, optimization, and obfuscation for the Android release build.

3. Using Android Gradle Plugin 9.3+, run the R8 Configuration Analyzer:

```bash
./gradlew :app:analyzeReleaseR8Config
```

4. Open the generated R8 Configuration Analyzer report.

5. Find the following keep rule:

```proguard
-keep class io.sentry.** { *; }
```

6. Observe that this rule prevents R8 from shrinking, optimizing, and obfuscating all affected classes and members under `io.sentry.**`.

In our application, the analyzer reports approximately:

- 1,170 classes kept (2.1%)
- 4,159 fields kept (2.8%)
- 9,745 methods kept (3.2%)
- 3.0% total impact

### Expected Result

The Sentry consumer R8/ProGuard rules should preserve only the classes and members that actually require it, for example those accessed through JNI or reflection.

Classes and members that do not require this protection should remain eligible for R8 shrinking, optimization, and obfuscation.

### Actual Result

The following keep rule prevents R8 from shrinking, optimizing, and obfuscating the affected code across the entire `io.sentry.**` package:

```proguard
-keep class io.sentry.** { *; }
```

In our application, the AGP 9.3 R8 Configuration Analyzer reports that this single rule has approximately a 3.0% total impact and keeps 1,170 classes, 4,159 fields, and 9,745 methods.

The analyzer marks shrinking, optimization, and obfuscation as blocked by this rule.

None

lucas-zimerman:
## 📜 Description

Replaces the blanket `-keep class io.sentry.** { *; }` rule in `packages/flutter/android/proguard-rules.pro` with explicit keep rules for exactly the classes that are genuinely reachable via JNI from `package:jni`/`jnigen` bindings (`lib/src/native/java/binding.dart`), plus their nested classes.

The class list is derived 1:1 from `ffi-jni.yaml`'s `classes:` section (the source of truth for what jnigen actually binds) — 28 `io.sentry.*` classes. I verified the two lists match exactly (diffed programmatically, no manual transcription drift).

Native JNI/reflection needs of the underlying Java SDK (`io.sentry:sentry-android`) are already covered by the consumer ProGuard rules bundled in the `sentry-android-core` and `sentry-android-ndk` AARs — I pulled these directly from `getsentry/sentry-java@8.53.0` (the version this repo currently depends on) to confirm they're scoped narrowly (specific classes, `native <methods>` members only) and get auto-merged by AGP, so the Flutter plugin doesn't need to duplicate them.

## 💡 Motivation and Context

The R8 Configuration Analyzer (AGP 9.3+) reports the blanket rule blocks shrinking/optimization/obfuscation of \~1,170 classes, 4,159 fields, and 9,745 methods (\~3% of a typical app) under `io.sentry.**`, since it protects the entire Sentry Java SDK rather than just what's actually accessed reflectively.

Fixes getsentry/sentry-dart#3933

## 💚 How did you test it?

* Built a minified release APK for the example app (`flutter build apk --release`, `minifyEnabled true`) — built clean, no R8 errors. Installing and driving the UI on an emulator produced no crash either way.
* That "no crash" check turned out to be a weak signal on its own: `NativeSdkIntegration.call()` wraps native init in try/catch by design (so a Sentry init failure never crashes the host app), the resulting log is gated behind `options.debug` (`false` in the example app's `--release` build), and `dart:developer.log()` doesn't route anywhere visible without a VM service attached in release mode. So a real break in the keep rules would *not* have shown up as a crash or a log line with that setup alone.
* To get an actual positive/negative signal, I built a deliberate negative control: temporarily removed the `SentryAndroid` keep rule (the literal JNI entry point everything else depends on — `native.SentryAndroid.init$2(...)` is called directly from Dart on every SDK init) and added a raw `print()` probe around the try/catch. Rebuilding and running produced a clean `ClassNotFoundException: Didn't find class "io/sentry/android/core/SentryAndroid"` — confirming the test setup can actually detect a real break.
* Cross-checked statically with `strings` on the built `classes.dex`: the class name was genuinely absent from the APK in the broken build (not a build-caching artifact — verified via a fresh `mapping.txt` timestamp), and present again once the keep rule was restored.
* Restored the correct (committed) rules, rebuilt, and confirmed both that `Lio/sentry/android/core/SentryAndroid;` survives unrenamed in the dex and that the probe printed a clean success at runtime.
* Reverted both pieces of temporary test instrumentation (the probe and a `debug = true` override) before pushing — none of that is in this PR's diff.
* `flutter analyze --fatal-warnings` on `packages/flutter` — clean.

This validates the single most load-bearing class (the JNI entry point) with a real pass/fail signal, not exhaustive coverage of all 28 classes. Given the blast radius of getting a proguard keep rule wrong (silent runtime breakage only visible in obfuscated release builds, and easy to write a test that looks like it's checking this but isn't), it'd be worth a second look with the R8 Configuration Analyzer from the original issue, and ideally spot-checking a couple of the other bound classes (e.g. one from `protocol.*` and one from `android.replay.*`) the same way before merging.

## :pencil: Checklist

- [X] All tests passing
- [X] No breaking changes

## 🔮 Next steps

Worth confirming with AGP 9.3's R8 Configuration Analyzer (per the issue's repro steps) that this meaningfully reduces the kept-class footprint, and keeping `ffi-jni.yaml` and this rule list in sync if the JNI binding surface changes in the future.


packages/flutter/android/proguard-rules.pro
```diff
@@ -1,5 +1,67 @@
...
+-keep class io.sentry.rrweb.RRWebOptionsEvent$* { *; }
+-keep class io.sentry.rrweb.RRWebEvent { *; }
+-keep class io.sentry.rrweb.RRWebEvent$* { *; }
+-keep class io.sentry.SentryTraceHeader { *; }
+-keep class io.sentry.SentryTraceHeader$* { *; }
+
+# NDK-layer native method signatures and other JNI/reflection needs of the underlying Java SDK
+# are already covered by the consumer proguard rules bundled in the sentry-android-core and
+# sentry-android-ndk AARs -- no need to duplicate them here.
+
```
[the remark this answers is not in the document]
lucas-zimerman: I agree, looking at it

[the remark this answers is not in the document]
lucas-zimerman: @buenaflor `ffi-jni.yaml` is manually made, so I think since we have AI on our side we actually don't need a script for syncing it.
On the other hand, it would be nice to have GH CI check if they are "out of sync". If I were to do that on Sentry React Native, I'd make a simple add-on script for danger so that it warns when the file is out of sync and that's.
This repo no longer has Danger, if it did, I'd use the extra danger file feature here. https://github.com/getsentry/github-workflows/tree/main/danger#usage

Do you have any opinions on how to approach this here?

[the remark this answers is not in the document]
buenaflor: @lucas-zimerman yeah we definitely want a script that can sync them or alert us at least and it should run when update android dependency CI is run, ideally also auto-update the proguard rule file the same way it auto runs jnigen to update the JNI bindings.

Take a look at `update-android.sh` and how it invokes `generate-jni-bindings.sh`

maybe we can even make the proguard rule update a step in `generate-jni-bindings.sh`

I dont think danger is the best fit here.

packages/flutter/android/proguard-rules.pro
```diff
@@ -1,5 +1,67 @@
...
+-keep class io.sentry.protocol.SdkVersion { *; }
+-keep class io.sentry.protocol.SdkVersion$* { *; }
+-keep class io.sentry.protocol.SentryPackage { *; }
+-keep class io.sentry.protocol.SentryPackage$* { *; }
+-keep class io.sentry.rrweb.RRWebOptionsEvent { *; }
+-keep class io.sentry.rrweb.RRWebOptionsEvent$* { *; }
+-keep class io.sentry.rrweb.RRWebEvent { *; }
+-keep class io.sentry.rrweb.RRWebEvent$* { *; }
+-keep class io.sentry.SentryTraceHeader { *; }
+-keep class io.sentry.SentryTraceHeader$* { *; }
```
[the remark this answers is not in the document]
lucas-zimerman: fixed on https://github.com/getsentry/sentry-dart/pull/3972/commits/4b15a6af6635b7547f25b389cbc8c064b10b114e

packages/flutter/android/proguard-rules.pro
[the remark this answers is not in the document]
lucas-zimerman: I see

I added some calls into `update-android.sh` to sync the proguard files with ffi.
Also added some CI check for when `proguard` and `ffi` are out of sync.

Would be nice to have a script to automatically update the `ffi-jni.yaml`, but, to not increase the scope of this PR, I'd rather make a follow-up PR to address it, what do you think @buenaflor?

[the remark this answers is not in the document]
buenaflor: @lucas-zimerman `ffi-jni.yaml` shouldnt be automatically updated - these are the Java classes we bind that are handpicked. unless I misunderstand your use case

packages/flutter/scripts/update-android.sh
```diff
@@ -38,6 +38,9 @@ set-version)
 
     # Regenerate Dart JNI bindings so they stay in sync with the updated Android SDK version
     ../scripts/generate-jni-bindings.sh "$new_version"
+
+    # Regenerate proguard rules so they stay in sync with the JNI class list
+    ../scripts/generate-sentry-java-proguard.sh
```
buenaflor: Let's run `generate-sentry-java-proguard.sh` from `generate-jni-bindings.sh` so work (not only Android version bumps) stays in sync.

e.g if I work on a feature that needs a new JNI binding then I can run this locally as well.

buenaflor (changes requested):

buenaflor (commented):
What I would suggest is also doing `flutter build apk --release` then grep the outputted `mapping.txt` so every yaml io.sentry.* class still maps to itself.

e.g `io.sentry.android.core.SentryAndroid -> io.sentry.android.core.SentryAndroid:`

afaik we already have a step that builds `flutter build apk --debug` but I cant remember why we set it to debug and not release

packages/flutter/scripts/update-android.sh
lucas-zimerman: Moved on https://github.com/getsentry/sentry-dart/pull/3972/commits/1d5f00f381a397f6318bcb611bc573863c8719be

packages/flutter/scripts/generate-sentry-java-proguard.sh
```diff
@@ -0,0 +1,77 @@
...
+
+tmp_file=$(mktemp)
+trap 'rm -f "$tmp_file"' EXIT
+
+awk -v gen="$generated" -v start="$start_marker" -v end="$end_marker" '
+    $0 ~ start { print; print gen; in_block=1; next }
+    $0 ~ end { in_block=0; print; next }
+    in_block { next }
+    { print }
+' "$proguard_file" >"$tmp_file"
```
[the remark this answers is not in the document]
lucas-zimerman: Reproduced locally, fixed on [9dd31b5](https://github.com/getsentry/sentry-dart/pull/3972/commits/9dd31b524778e4d220f7a8787f794b1a9210a84f)

lucas-zimerman:
Fixed the AI issues and also addressed your issue in regard to the mapping check.

buenaflor (approved):
LGTM

.github/workflows/flutter.yml
```diff
@@ -215,6 +215,48 @@ jobs:
...
+      - name: Fail if proguard rules are out of sync with ffi-jni.yaml
+        run: |
+          diff_output=$(git diff -- android/proguard-rules.pro)
+          if [[ -n "$diff_output" ]]; then
+            echo "$diff_output"
+            echo "::error::android/proguard-rules.pro is out of sync with ffi-jni.yaml. Run packages/flutter/scripts/generate-sentry-java-proguard.sh and commit the result."
+            exit 1
+          fi
+
+  proguard-mapping-verify:
```
buenaflor: should we make them one job instead of two parallel?

e.g
- Sync check (fail fast if the committed file doesn’t match yaml)
- Then mapping verify (only if step 1 passed)

buenaflor: also let's add a note somewhere that all of this is needed until proguard rules are automatically supported by jni: https://github.com/dart-lang/native/issues/681

lucas-zimerman: Done. Still think it's worth to keep the ci check as a double check the mapping matches, wdyt?

Opened this issue to track the removal from the introduced change https://github.com/getsentry/sentry-dart/issues/3978

buenaflor: > Still think it's worth to keep the ci check as a double check the mapping matches, wdyt

wdym?

lucas-zimerman: Sorry, wasn't clear, meant `verify-proguard-mapping.sh` specifically, not the job structure (that's settled). Once `dart-lang/native#681` lands, only the sync/generation script becomes redundant; mapping-verify still checks R8's real output, not just text sync, so it should stay.

buenaflor: ah yeah, we should keep it

PR #3976, opened 2026-08-19
ci(testflight): Cancel overlapping TestFlight jobs

buenaflor:
## :scroll: Description

Overlapping TestFlight uploads on the same branch are cancelled instead of racing App Store Connect for the same build number.

The workflow now uses a concurrency group shaped like the other CI workflows (`workflow` + branch), but falls back to `github.ref` instead of `github.run_id`. On `push` to `main`, `head_ref` is empty and `run_id` is unique, so the usual fallback would not cancel anything. A newer `main` push now cancels the stale job. Fastlane still bumps with `latest_testflight_build_number + 1`.

The job fails intermittently with:

> Validation failed (409) Redundant Binary Upload. You've already uploaded a build with build number 'N'.

That happened when two `main` pushes overlapped: both read the same latest TestFlight number and both tried to upload `N+1`. Recent examples: [1115](https://github.com/getsentry/sentry-dart/actions/runs/32237730245), [1113](https://github.com/getsentry/sentry-dart/actions/runs/32142414962), [1111](https://github.com/getsentry/sentry-dart/actions/runs/32141380248).

This is the smaller fix. If 409s continue (for example Apple not indexing a just-uploaded build before the next run queries it), we can add unique CI build numbers later.

Diagnosed from the failed job logs (archive succeeded; `altool` 409 on upload). After merge, watch whether close `main` pushes still 409.

## :pencil: Checklist
- [ ] All tests passing
- [x] No breaking changes

### DIFF
--- a/.github/workflows/testflight.yml
+++ b/.github/workflows/testflight.yml
@@ -40,9 +40,11 @@ jobs:
           APP_STORE_CONNECT_KEY_ID: ${{ secrets.APP_STORE_CONNECT_KEY_ID }}
           APP_STORE_CONNECT_ISSUER_ID: ${{ secrets.APP_STORE_CONNECT_ISSUER_ID }}
           APP_STORE_CONNECT_KEY: ${{ secrets.APP_STORE_CONNECT_KEY }}
-          # Unique per run and re-run. Used as CFBundleVersion so overlapping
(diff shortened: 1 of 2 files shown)


denrase (approved):
COuldn't we sync the jobs instead?

PR #3980, opened 2026-08-24
fix(replay): Route network detail through sentry-java's own hint

hkarmoush:
## :scroll: Description

Reworks the Session Replay network-detail leak fix per @buenaflor's review on #3900.

- Dart side keeps captured HTTP request/response detail off the shared `http` breadcrumb, passing it via a `Hint` instead (unchanged from the earlier attempt).
- Android side no longer uses a bespoke JNI side channel (`ReplayNetworkDetailCache`, keyed by a generated `replay_request_id`) to get that detail into a replay recording. Instead it's delivered through sentry-java's own `SENTRY_REPLAY_NETWORK_DETAILS` hint - the same mechanism `SentryOkHttpInterceptor` already uses - so `DefaultReplayBreadcrumbConverter`'s own (already-tested) conversion logic builds the replay event, rather than a hand-rolled reimplementation.
- `addBreadcrumb` now carries the network detail directly instead of a separate `captureReplayNetworkDetail` call, so there's one atomic native call instead of two - removing the race Bugbot flagged on the previous side-channel approach (#3947).

Closes the leak in #3900: captured HTTP bodies/headers (opt-in via `enableReplayNetworkDetailsCapturing`) were riding along on the shared `http` breadcrumb and ending up on unrelated events and native crash events, not just replay recordings.

Supersedes #3947, which used a bespoke JNI cache that Bugbot flagged a race in and @buenaflor asked to be reworked to route through sentry-java's existing mechanism instead: https://github.com/getsentry/sentry-dart/issues/3900#issuecomment-3427266847

- Unit tests updated at the `NativeScopeObserver` / `SentryNativeJava.addBreadcrumb` seams (single-call contract, request/response detail merging).
- `dart analyze` clean on both `sentry` and `sentry_flutter`; full test suites pass.
- The new Android JNI binding entry (`addBreadcrumbFromJsonBytes` overload) is hand-authored, mirroring the existing pattern for `[B` descriptors in `binding.dart`, since regenerating it needs a full Android build toolchain not available in this environment - flagging for verification against a real `jnigen` run before merge.

## :pencil: Checklist
- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps

Would like a maintainer (or CI with a real Android toolchain) to verify the hand-authored JNI binding entry in `binding.dart` against an actual `jnigen` run before merging.


packages/dart/lib/src/scope.dart
```diff
@@ -250,10 +250,17 @@ class Scope {
...
     if (addedBreadcrumb != null) {
-      await _callScopeObservers((scopeObserver) async =>
-          await scopeObserver.addBreadcrumb(addedBreadcrumb));
+      await _callScopeObservers((scopeObserver) async {
+        if (scopeObserver is HintAwareScopeObserver) {
+          await (scopeObserver as HintAwareScopeObserver)
+              .addBreadcrumbWithHint(addedBreadcrumb, resolvedHint);
+        } else {
+          await scopeObserver.addBreadcrumb(addedBreadcrumb);
+        }
```
buenaflor:
```suggestion
        if (scopeObserver case final HintAwareScopeObserver observer) {
          await observer.addBreadcrumbWithHint(addedBreadcrumb, resolvedHint);
        } else {
          await scopeObserver.addBreadcrumb(addedBreadcrumb);
        }
```

buenaflor: making it a bit more nicer with newer Dart features

packages/flutter/lib/src/native/java/binding.dart
```diff
@@ -5615,6 +5615,50 @@ class SentryFlutterPlugin extends jni$_.JObject {
...
+  // this file requires `scripts/generate-jni-bindings.sh`, which needs a
+  // full Android build toolchain not available in this environment. Re-run
+  // that script (or otherwise verify against a real jnigen run) before
+  // merging.
+  static final _id_addBreadcrumbFromJsonBytes$1 = _class.staticMethodId(
+    r'addBreadcrumbFromJsonBytes',
+    r'([B[B)V',
+  );
+
+  static final _addBreadcrumbFromJsonBytes$1 =
```
buenaflor: let's not hand-roll this. please re-generate this with e.g Flutter 3.44

packages/flutter/android/src/main/kotlin/io/sentry/flutter/SentryFlutterPlugin.kt
```diff
@@ -188,6 +194,89 @@ class SentryFlutterPlugin :
...
+        val statusCode = (breadcrumb.data["status_code"] as? Number)?.toInt() ?: 0
+        networkData.setResponseDetails(statusCode, toReplayNetworkRequestOrResponse(it))
+      }
+
+      breadcrumb.data["start_timestamp"]?.let {
+        breadcrumb.setData(SpanDataConvention.HTTP_START_TIMESTAMP, it)
+      }
+      breadcrumb.data["end_timestamp"]?.let {
+        breadcrumb.setData(SpanDataConvention.HTTP_END_TIMESTAMP, it)
+      }
```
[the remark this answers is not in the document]
hkarmoush: Checked this against sentry-java's own `SentryOkHttpInterceptor.sendBreadcrumb` (the code this hint routing mirrors) — it does the identical thing: unconditionally calls `breadcrumb.setData(SpanDataConvention.HTTP_START_TIMESTAMP, ...)` / `HTTP_END_TIMESTAMP` directly on the breadcrumb before `scopes.addBreadcrumb(breadcrumb, hint)`, with the comment "needs this as unix timestamp for rrweb". So every HTTP breadcrumb already carries these fields on all events in native Android, replay or not — this isn't a leak introduced here, it's the same behavior sentry-java ships today. Leaving as-is.

PR #3983, opened 2026-08-25
feat(dart): Remove need for enableLogs and enableMetrics v10

buenaflor:
## :scroll: Description

Removes `enableLogs` and `enableMetrics` from `SentryOptions`. Logs and metrics are always captured at the Dart layer (still no auto-instrumentation), and manual `Sentry.logger` / `Sentry.metrics` calls stay always-on.

`LoggingIntegration` is the only automatic log source, and adding it is now the whole opt-in: it forwards records at or above `minSentryLogLevel` as structured logs, with no second flag. Raise that threshold, or set it to `Level.OFF`, to keep breadcrumbs and events without logs. This matches sentry-java, where Logback, Log4j2, JUL and Timber pair a level threshold with the global flag and have no per-integration boolean.

Native init now enables both explicitly: Android `logs`/`metrics`, Cocoa `enableLogs`/`enableMetrics`. sentry-native has no equivalent options.

v10 port of #3964, adapted for Cocoa 9.21 top-level options and the JNI property accessors. The 9.x counterpart is #3982, which keeps the options but stops them gating the manual APIs.

Logs and metrics should be on by default at the SDK options layer. Native SDKs still default logs off, so Flutter has to turn them on at init.

The logs policy also rules out a real double opt-in — users should not have to both flip a flag and add an integration. With the global option gone, the integration itself is the opt-in.

- `packages/dart` — options, pipelines, setup integrations, enricher, client lifecycle
- `packages/flutter` — load contexts, replay telemetry, widgets binding observer
- `packages/logging` — logging integration, including `minSentryLogLevel: Level.OFF` forwarding nothing
- Cocoa unit test: `update(options:with: [:])` leaves `enableLogs` and `enableMetrics` true
- `dart analyze` clean on touched files

## :pencil: Checklist

- [x] All tests passing
- [ ] No breaking changes

## :crystal_ball: Next steps

Docs need updating to drop `options.enableLogs` / `options.enableMetrics` from setup snippets, and to note that adding the logging integration forwards logs on its own.

## Changelog Entry

Remove the `enableLogs` and `enableMetrics` options. Logs and metrics are now always enabled and can no longer be disabled through `SentryOptions`. Remove any `options.enableLogs` / `options.enableMetrics` assignments from your SDK init. If you use the [logging integration](https://docs.sentry.io/platforms/dart/integrations/logging/), it now forwards log records as Sentry logs as soon as it is added — raise `minSentryLogLevel`, or set it to `Level.OFF`, to opt out.


denrase (approved):

buenaflor:
@adinauer just tagged you here in case you want to take a quick look since you also did this for java 👍

PR #3987, opened 2026-08-26
fix(flutter): Close workers on isolate exit

### LINKED ISSUE #3960: Android: initializing SentryFlutter in a rendering isolate retains ~80MB per Activity create/destroy (reproduces with an empty DSN)
### Platform

Android

### Obfuscation

Disabled (also reproduces with `isMinifyEnabled = false`)

### Debug Info

Not enabled (no `--split-debug-info`)

### Doctor

```
[✓] Flutter (Channel stable, 3.41.3, on Ubuntu 24.04.4 LTS 6.17.0-20-generic, locale en_US.UTF-8)
[✓] Android toolchain - develop for Android devices (Android SDK version 36.1.0)
[✓] Connected device (1 available)
```

Device: Pixel 7, Android 17 (SDK 37). Impeller/Vulkan (Flutter default; no `EnableImpeller` flag set).

### Other Error Monitoring Solution

No

### Version

`sentry_flutter` 9.14.0 — also reproduced identically on 9.26.0.

### Steps to Reproduce

Our app runs two Flutter engines: a UI engine and a headless background engine (a foreground service keeps the process alive so BLE work continues with no UI). `SentryFlutter.init` was called in the UI isolate.

1. Cold start the app, let it settle.
2. Background it, then remove the task so the Activity — and its engine — is destroyed. The process survives (foreground service).
3. Measure with `adb shell dumpsys meminfo <pkg>` in that headless state.
4. Repeat.

Each Activity create/destroy cycle permanently retains **~17.6 MB of GPU memory (`GL mtrack`)** and **~65 MB of native memory** (`Native Heap` + `Unknown`). Over three cycles the process went from 220 MB to 465 MB and never came back down.

The retention is per Activity cycle; it is *not* time-based. With no UI sessions, five minutes of BLE and socket activity in the background isolate left `GL mtrack` bit-for-bit identical.

### Expected Result

Memory allocated during a UI session is released when that engine is destroyed, so the process does not grow with each app open.

### Actual Result

It is retained for the life of the process. Because the app holds a foreground service, Android classifies the process as foreground and refuses to deliver the trim callbacks that would prompt Flutter to purge caches:

```
IllegalArgumentException: Unable to set a background trim level on a foreground process
```

The foreground-level trims that *are* permitted (`RUNNING_MODERATE` / `RUNNING_LOW` / `RUNNING_CRITICAL`) freed zero bytes.

**What isolates it to `SentryFlutter.init` itself.** Each of the following still leaked at the full rate (+17.6 MB GPU per cycle):

| configuration | leaks? |
|---|---|
| `enableAutoPerformanceTracing = false` | yes |
| `enableWatchdogTerminationTracking = false` + `enableAppHangTracking = false` | yes |
| `enableFramesTracking = false` | yes |
| `tracesSampler → 0.0` (no transactions at all) | yes |
| `sampleRate = 0.0` (no events at all) | yes |
| `autoInitializeNativeSdk = false` **and** manifest `io.sentry.auto-init=false` (no native SDK whatsoever) | yes |
| `runApp` outside the `appRunner` zone | yes |
| **`options.dsn = ''`** — SDK inert | **yes** |
| upgrade 9.14.0 → 9.26.0 | yes |
| **`SentryFlutter.init` not called at all** | **no — 0 kB retained, PSS flat** |

So an inert SDK with no DSN, no native layer, no events and no transactions still retains the memory. That points at something installed unconditionally during `init` rather than at any feature.

Two things we ruled out that might otherwise look like the cause:

- **Impeller.** Disabling it zeroed the GPU line but PSS still grew ~65 MB/cycle, so Impeller was only the allocator the GPU share flowed through.
- **Build mode.** Profile builds appeared not to leak, but only because our `init` is gated on `dart.vm.product`. Forcing Sentry on in a profile build reproduces it in full.

We also controlled for the extra `await` before `runApp`: no Sentry plus an equivalent `await Future.delayed(1s)` retains 0 kB.

**Workaround.** Moving `SentryFlutter.init` into the headless background isolate — which never renders and is created once — eliminates it completely: 0 kB retained per cycle, PSS flat, and the process baseline drops ~70 MB. The UI isolate now installs `FlutterError.onError`, `PlatformDispatcher.onError` and a `runZonedGuarded` itself and forwards errors over a port. That suggests the retention is tied to `init` running in an isolate that renders.

Happy to run patches or targeted builds against this setup — we have a repeatable measurement harness on the device.

We can test and measure patches; we haven't identified the retaining allocation itself, so we can't propose a fix yet.

lucas-zimerman:
Fixes #3960: Android's background JNI worker could outlive its owning Flutter engine and accumulate native memory across Activity destruction/recreation cycles, even with an empty DSN and `autoInitializeNativeSdk = false`.

The worker now registers shutdown on its owning isolate's exit. View detachment does not close the native SDK or worker, so cached engines can continue reporting after reattachment. Explicit `Sentry.close()` starts worker shutdown before awaiting replay cleanup, and repeated close calls share the same completion future. Native SDK ownership guards preserve SDKs initialized by the host application and avoid loading desktop native libraries solely for cleanup.

Regression tests cover detach/reattach without closing the native binding, explicit worker cleanup after detach, synchronous shutdown initiation, and worker termination after owner-isolate exit. Android integration coverage verifies that the real native SDK remains enabled and accepts a native event after simulated Flutter lifecycle transitions.

Validation of the follow-up fixes: 1,117 Flutter unit tests passed (3 skipped), focused analysis passed, and the Android integration regression passed on an API 36 emulator. Using the original [reproduction harness](https://github.com/getsentry/sentry-dart/commit/6677b4bf6d1873623960de75e4249d0491c9e366) with an empty DSN and native initialization disabled, three actual Activity/engine destruction and relaunch cycles retained the same Android process: the Dart VM reported zero isolates after each destruction and exactly one `SentryAndroidCoreWorker` after each relaunch. The harness is not included in the shipped changes. These checks do not simulate owner exit during worker startup.

## Changelog Entry
Fix an Android native/GPU memory leak caused by the background JNI worker outliving its Flutter engine, while preserving native reporting across cached-engine view detachment and reattachment.


packages/flutter/lib/src/native/sentry_native_channel.dart
```diff
@@ -28,6 +28,13 @@ class SentryNativeChannel
   SentryNativeChannel(this.options)
       : channel = SentrySafeMethodChannel(options);
 
+  // Tracks whether the native SDK's own init actually ran, so close() -
+  // which may now run even when autoInitializeNativeSdk is false, since the
+  // native binding may start background resources unconditionally (see
+  // #3960) - doesn't tear down a native SDK the app initialized itself.
+  @protected
+  bool nativeSdkInitialized = false;
```
buenaflor: should this be something like `nativeSdkAutoInitialized`?

packages/flutter/lib/src/native/c/sentry_native.dart
```diff
@@ -40,6 +40,12 @@ class SentryNative with SentryNativeSafeInvoker implements SentryNativeBinding {
 
   SentryNative(this.options);
 
+  // Tracks whether native.init() actually ran, so close() - which may now be
+  // called even when crash handling was never enabled, see #3960 - doesn't
+  // touch the `native` binding (and so trigger its lazy DynamicLibrary.open)
+  // for the first time on a binding that was never initialized.
+  bool _crashHandlingInitialized = false;
```
buenaflor: maybe we can name this the same as the other? `nativeSdkAutoInitialized`

buenaflor (commented):
overall LGTM

just couple question about the naming, maybe we can also reduce the amount of comments

packages/flutter/lib/src/native/c/sentry_native.dart
lucas-zimerman: Works for me, renamed on https://github.com/getsentry/sentry-dart/pull/3987/commits/c33d07e6a0e28eeda1b2e458bd0f02b28fafacf6

packages/flutter/lib/src/native/sentry_native_channel.dart
lucas-zimerman: Yeah makes sense, renamed on https://github.com/getsentry/sentry-dart/pull/3987/commits/c33d07e6a0e28eeda1b2e458bd0f02b28fafacf6

packages/flutter/lib/src/integrations/native_sdk_integration.dart
```diff
@@ -43,20 +55,73 @@ class NativeSdkIntegration implements Integration<SentryFlutterOptions> {
...
+
+    try {
+      await _native.close();
+    } catch (exception, stackTrace) {
+      _options?.log(
+        SentryLevel.fatal,
+        'nativeSdkIntegration failed to be closed',
+        exception: exception,
+        stackTrace: stackTrace,
+      );
```
buenaflor: let's use the new `internalLogger` instead

packages/flutter/lib/src/integrations/native_sdk_integration.dart
```diff
@@ -43,20 +55,73 @@ class NativeSdkIntegration implements Integration<SentryFlutterOptions> {
...
+      // Not awaited - didChangeAppLifecycleState is synchronous, and
+      // SentryNativeBinding.close() implementations are expected to do
+      // their critical shutdown work (if any) synchronously, before their
+      // first `await`, so it runs within this call stack rather than after
+      // a microtask hop that may never come. See #3960.
+      // Routed through _closeNative() so errors from its asynchronous tail
+      // are still logged instead of becoming unhandled Future errors.
+      unawaited(_integration._closeNative());
+    }
+  }
```
buenaflor: I think this cleanup is too broad. `detached` means the engine has no attached host view, but the engine and Dart isolate can continue running. A new host view can later attach to that same engine, and this change would leave its native resources closed without restarting them.

Could we tie worker cleanup to the owning isolate’s lifetime instead? One option to investigate is [addOnExitListener](https://api.dart.dev/dart-isolate/Isolate/addOnExitListener.html), registered in the owning isolate for each worker:

```dart
Isolate.current.addOnExitListener(
  workerCommandPort,
  response: '_shutdown_',
);
```

This would request worker shutdown when the owner exits, while preserving workers across view detachment and reattachment. We should verify that it works during Android engine destruction and handles the owner exiting during worker startup.

buenaflor (changes requested):

packages/flutter/lib/src/integrations/native_sdk_integration.dart
buenaflor: and check other uses of this as well pls

lucas-zimerman: I patched the log calls that were altered on this PR here https://github.com/getsentry/sentry-dart/pull/3987/commits/4ad32a0d42b19c8c272fc26ff1ccc4f679f1a2b8

Is the use of `internalLogger` only because options could be null here? If not, do you want me to open a follow-up PR later to broaden this fix to other unchanged code?
A quick check pointed the following files using the options log.

- debug_print_integration.dart:52
- flutter_error_integration.dart:30,54,97
- frames_tracking_integration.dart:29,34,41,48,79,100,110
- load_release_integration.dart:35,43
- on_error_integration.dart:50
- screenshot_integration.dart:15
- thread_info_integration.dart:28
- web_sdk_integration.dart:43,61
- web_session_integration.dart:35,51,64,72
- widgets_binding_integration.dart:17

packages/flutter/lib/src/integrations/native_sdk_integration.dart
lucas-zimerman: moved worker shutdown off AppLifecycleState.detached, now using Isolate.current.addOnExitListener tied to the owning isolate's own exit - so a cached engine reattaching later keeps its worker alive.

added a VM test that kills the owner isolate abruptly and confirms the worker still gets the shutdown signal.

also validated on device (moto g34, 8 relaunch cycles, real engine teardown not just Isolate.kill): EGL/GL mtrack flat, native heap just GC noise, ~200KB/cycle drift matching the same baseline the original fix landed on. no leak back.

pushed, ready for another look.

packages/flutter/lib/src/integrations/native_sdk_integration.dart
buenaflor: its fine, let's keep it to this PRs change

options.log is the old deprecated way, we should be using internalLogger now

buenaflor:
@lucas-zimerman I'll have a look at the rest, I'd like to merge this today

buenaflor (approved):

PR #4008, opened 2026-09-07
feat(flutter): Unsampled errors affect release health

### LINKED ISSUE #3786: [v10] Sampled-out unhandled errors don't mark session as unhandled
When `sampleRate` is below 1.0, unhandled Dart/Flutter errors that are sampled out never reach the transport layer. On mobile, session crashes are signaled by forwarding an envelope to the native SDK with a `containsUnhandledException` flag, but sampled-out events skip this path entirely. The native session remains healthy, inflating crash-free session and crash-free user metrics for Dart-origin crashes. Native crashes are unaffected since they're captured entirely by the native SDK and bypass Dart's sampling logic.

* **Root cause (medium confidence)**: Sampling is applied before the session-update path; sampled-out events exit early and never trigger the native session envelope forward
* **Scope**: Flutter/Dart unhandled errors only; native crashes are unaffected
* **Related**: getsentry/sentry-dart#3300 improves session handling for Flutter unhandled errors more broadly (new `unhandled` session status), which will also improve this scenario  <br>  <br>Blocked by Native SDK APIs to control sessions

See develop docs: https://develop.sentry.dev/sdk/foundations/client/#event-pipeline

buenaflor:
## :scroll: Description

Add an internal sampled-out event lifecycle hook and use it to keep native session state accurate when an unhandled Dart or Flutter event is dropped by sampling.

The native session integration:
- receives only final events accepted by processors and `beforeSend`
- updates Android or Cocoa with `unhandled: true` when `mechanism.handled == false`
- awaits Android worker persistence before capture returns
- unregisters on close and contains native failures
- remains disabled when automatic session tracking is disabled

Stacked on [#4007](https://github.com/getsentry/sentry-dart/pull/4007), which adopts the non-terminating envelope APIs. That PR is itself stacked on [#4006](https://github.com/getsentry/sentry-dart/pull/4006).

Fixes #3786

A sampled-out unhandled error has no envelope, so the native SDK cannot infer its session side effects from capture. Without a separate update, a gracefully ending session is incorrectly reported as exited instead of unhandled.

- Core `SentryClient` sampling suite: 10 tests passed
- Native session integration, channel, and Android worker suite: 104 tests passed
- Platform registration suite: 22 tests passed
- Android release APK build passed
- Dart and Flutter analysis passed with fatal warnings
- Dart 3.13 formatting verified
- Confirmed the Cocoa 9.27.0 private API signatures against the tagged source

The local macOS example build was blocked before compilation by Flutter's experimental Swift Package Manager override identity mismatch (`flutter` vs `sentry_flutter`); CI provides the platform build lane.

## :pencil: Checklist

- [x] All focused tests passing
- [x] No breaking changes

### DIFF
--- a/packages/dart/test/sentry_client_test.dart
+++ b/packages/dart/test/sentry_client_test.dart
@@ -1176,6 +1176,20 @@ void main() {
       expect(fixture.transport.called(1), true);
     });
 
+    test('does not dispatch sampled out event when captured', () async {
+      var callbacks = 0;
+      fixture.options.lifecycleRegistry.registerCallback<OnEventSampledOut>((
+        event,
+      ) {
+        callbacks++;
+      });
+      final client = fixture.getSut(sampleRate: 1.0);
+
+      await client.captureEvent(fakeEvent);
+
+      expect(callbacks, 0);
+    });
+
     test('do not capture event, sample rate is 0% disabled', () async {
       final client = fixture.getSut(sampleRate: 0.0);
       await client.captureEvent(fakeEvent);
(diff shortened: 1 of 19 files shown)


packages/dart/test/sentry_client_test.dart
```diff
@@ -1214,6 +1228,81 @@ void main() {
...
+      );
+
+      await client.captureEvent(fakeEvent, hint: hint);
+
+      expect(sampledOut?.event, same(finalEvent));
+      expect(sampledOut?.hint, same(hint));
+    });
+
+    test(
+      'does not dispatch sampled out event when beforeSend drops it',
```
denrase, line 1253: Wouldn't we also skew the results if we have a sampled out even due to beforeSend?

packages/dart/test/sentry_client_test.dart
```diff
@@ -1214,6 +1228,81 @@ void main() {
...
+        );
+
+        await client.captureEvent(fakeEvent);
+
+        expect(callbacks, 0);
+      },
+    );
+
+    test(
+      'does not dispatch sampled out event when a processor drops it',
```
denrase, line 1273: Wouldn't we also skew the results if we have a sampled out even due to event processor?

packages/dart/test/sentry_client_test.dart
```diff
@@ -1214,6 +1228,81 @@ void main() {
...
+          eventProcessor: DropAllEventProcessor(),
+        );
+
+        await client.captureEvent(fakeEvent);
+
+        expect(callbacks, 0);
+      },
+    );
+
+    test('does not dispatch sampled out event when ignored', () async {
```
denrase, line 1292: Wouldn't we also skew the results if we have a sampled out even due to ignore?

packages/flutter/lib/src/integrations/native_session_integration.dart
```diff
@@ -0,0 +1,60 @@
...
+    _options = options;
+    options.lifecycleRegistry.registerCallback<OnEventSampledOut>(
+      _updateSessionForSampledOutEvent,
+    );
+    options.sdk.addIntegration(integrationName);
+  }
+
+  @override
+  void close() {
+    _options?.lifecycleRegistry.removeCallback<OnEventSampledOut>(
```
denrase, line 35: Should also nil out options reference. To be extra correct, shouldn't we remove the integration string from options sdk as well? I guess the latter is also true for other integrations.

denrase (commented):
We should check if we need to do this if we drop for ANY reason, not just sampling.

buenaflor:
> We should check if we need to do this if we drop for ANY reason, not just sampling.

We only wanna do this for sampling.

if an event is dropped due to `ignoreErrors`, `eventProcessor` or `beforeSend`, it's filtered on purpose. however when events are not filtered and reach the sampling stage it means this is an error that we consider as "useful" and want to still reflect in the session even though it might be sampled out.

For details: https://develop.sentry.dev/sdk/foundations/client/#event-pipeline

packages/dart/test/sentry_client_test.dart
buenaflor: no cause beforeSend is a user added filter, if the user explicitly filters out an event they wouldn't wanna consider it in a session

packages/dart/test/sentry_client_test.dart
buenaflor: same reason as above with `beforeSend`

packages/dart/test/sentry_client_test.dart
buenaflor: same comment as above

packages/flutter/lib/src/integrations/native_session_integration.dart
buenaflor: yeah for pure correctness we should, I'll update it

denrase (approved):

packages/flutter/lib/src/integrations/native_session_integration.dart
buenaflor: updated to nil out options reference, but the integration string is fine to keep since it's tied to the live options. the integrations are only closed when the whole sdk is closed so it's fine

PR #4010, opened 2026-09-07
feat(flutter): Add manual replay APIs

### LINKED ISSUE #2558: [v10] Replay: add start/stop - pause/resume APIs
### Description

Apps should be able to start and stop replays manually, ignoring session & error sampling rates

buenaflor:
## :scroll: Description

Adds `SentryFlutter.replay` with `start`, `startBuffering`, `pause`, `resume`, `stop`, and `flush`, routed through JNI on Android and the method channel on iOS. The replay recorder now stays available when automatic sampling is off, and native and Dart callbacks are cleaned up on close.

Two changes worth a closer look in review:

**`ReplayTelemetryIntegration` no longer checks the sample rates.** A manually started replay records while both rates are `0`, and the old gates would have dropped its `replay_id` from logs, spans, and metrics. Attributes are now derived from whether a replay is actually recording, which is safe because:

- The session gate was redundant — native writes the replay ID to the scope *only* in session mode, so the scope already answers the question. It also suppressed a legitimate case: a buffered replay flushed by an error keeps recording in session mode.
- The buffering gate guarded a state only real buffering produces — "the binding has an ID but the scope doesn't". iOS can't produce that split at all (the screenshot provider hardcodes `replayIsBuffering = NO`), leaving Android's `replayStarted(id, isBuffering: true)` as its only source.

**iOS now clears the replay ID on stop.** iOS has no `replayStopped` callback, so after `stop()` a finished replay's ID stayed on `scope.replayId` and kept riding along on events and the trace context. `SentryNativeCocoa.stopReplay()` now clears both the binding cache and the scope. Pre-existing and still open: a native-initiated stop (session max duration) has no callback, and it can't be covered by a test because FFI/JNI can't be mocked.

Applications need explicit Session Replay control for opt-in flows and sensitive screens, independent of the automatic sample rates.

Fixes #2558

- Full `sentry_flutter` unit suite, plus new coverage for manual starts with sampling disabled, the "no replay running" cases, and `stopReplay()` clearing both the binding cache and the scope (verified to fail before the fix).
- Replay integration tests on an iOS simulator; Android integration test target builds.
- `flutter analyze` and formatting with the Flutter-bundled Dart that CI uses.

## :pencil: Checklist

- [x] All tests passing
- [x] No breaking changes

## :crystal_ball: Next steps

None.

### DIFF
--- a/packages/flutter/darwin/sentry_flutter/Sources/sentry_flutter/SentryFlutterPlugin.swift
+++ b/packages/flutter/darwin/sentry_flutter/Sources/sentry_flutter/SentryFlutterPlugin.swift
@@ -155,6 +155,42 @@ public class SentryFlutterPlugin: NSObject, FlutterPlugin {
             result(nil)
 #endif
 
+        case "startReplay":
+#if canImport(UIKit) && !SENTRY_NO_UIKIT && (os(iOS) || os(tvOS))
+            SentrySDK.internal.replay.start()
+#endif
+            result("")
+
+        case "startReplayBuffering":
+#if canImport(UIKit) && !SENTRY_NO_UIKIT && (os(iOS) || os(tvOS))
+            SentrySDK.internal.replay.startBuffering()
+#endif
+            result("")
+
+        case "pauseReplay":
+#if canImport(UIKit) && !SENTRY_NO_UIKIT && (os(iOS) || os(tvOS))
+            SentrySDK.internal.replay.pause()
+#endif
+            result("")
+
+        case "resumeReplay":
+#if canImport(UIKit) && !SENTRY_NO_UIKIT && (os(iOS) || os(tvOS))
+            SentrySDK.internal.replay.resume()
+#endif
+            result("")
+
+        case "stopReplay":
+#if canImport(UIKit) && !SENTRY_NO_UIKIT && (os(iOS) || os(tvOS))
+            SentrySDK.internal.replay.stop()
+#endif
+            result("")
+
+        case "flushReplay":
+#if canImport(UIKit) && !SENTRY_NO_UIKIT && (os(iOS) || os(tvOS))
+            SentrySDK.internal.replay.flush()
+#endif
+            result("")
+
         case "setTrace":
             let arguments = call.arguments as? [String: Any?]
             let traceId = arguments?["traceId"] as? String
--- a/packages/flutter/darwin/sentry_flutter/Sources/sentry_flutter/SentryFlutterPlugin.swift
+++ b/packages/flutter/darwin/sentry_flutter/Sources/sentry_flutter/SentryFlutterPlugin.swift
@@ -155,41 +155,8 @@ public class SentryFlutterPlugin: NSObject, FlutterPlugin {
             result(nil)
 #endif
 
-        case "startReplay":
-#if canImport(UIKit) && !SENTRY_NO_UIKIT && (os(iOS) || os(tvOS))
-            SentrySDK.internal.replay.start()
-#endif
-            result("")
-
-        case "startReplayBuffering":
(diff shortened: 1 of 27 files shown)


buenaflor:
@cursor review

packages/flutter/lib/src/native/cocoa/sentry_native_cocoa.dart
```diff
@@ -64,6 +64,18 @@ class SentryNativeCocoa extends SentryNativeChannel {
...
+  Future<void> stopReplay() async {
+    await super.stopReplay();
+    // iOS reports replay IDs only through the screenshot provider, which goes
+    // quiet once recording stops. Android has `replayStopped` for this.
+    _replayId = null;
+    _hub?.configureScope((s) {
+      // ignore: invalid_use_of_internal_member
+      s.replayId = null;
+    });
+  }
```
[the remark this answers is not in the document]
buenaflor: Good catch, fixed in b34b37bb3 — the clearing now happens in a `finally`, so the error still propagates to the caller but Dart state stays consistent. Added a test that stubs the channel to fail and asserts both the cached and scope replay IDs are cleared.

One note on reachability for anyone reading this later: the Swift side cannot actually fail here. All six replay methods funnel through `controlReplay`, which calls a non-throwing Cocoa method and unconditionally ends in `result("")`, never a `FlutterError`. So the only reachable throw is `MissingPluginException` — where no native replay exists and clearing is the right thing to do anyway. Worth doing regardless.

buenaflor:
bugbot run

packages/flutter/darwin/sentry_flutter/Sources/sentry_flutter/SentryFlutterPlugin.swift
```diff
@@ -166,6 +169,29 @@ public class SentryFlutterPlugin: NSObject, FlutterPlugin {
...
+            replay.resume()
+        case "stopReplay":
+            replay.stop()
+        case "flushReplay":
+            replay.flush()
+        default:
+            break
+        }
+#endif
+        result("")
```
lucas-zimerman: Q : Why not follow the same return pattern on `"captureReplay"` on line `155` when the same ifdef condition is not met?

packages/flutter/darwin/sentry_flutter/Sources/sentry_flutter/SentryFlutterPlugin.swift
```diff
@@ -155,6 +155,9 @@ public class SentryFlutterPlugin: NSObject, FlutterPlugin {
             result(nil)
 #endif
 
+        case "startReplay", "startReplayBuffering", "pauseReplay", "resumeReplay", "stopReplay", "flushReplay":
```
lucas-zimerman: Q: What do you think of merging the `captureReplay` logic with the following ones, also renaming `startReplayBuffering` to `startBufferingReplay`.

This way we could use the following `case let method where method.hasSuffix("Replay"):`

And have a single source that skips for `#if canImport(UIKit) && !SENTRY_NO_UIKIT && (os(iOS) || os(tvOS))` on both `captureReplay` and the other replay cases.

lucas-zimerman (approved):
Left two non-blocking comments, LGTM!

packages/flutter/darwin/sentry_flutter/Sources/sentry_flutter/SentryFlutterPlugin.swift
buenaflor: I'd keep the current separation, matching everything ending in "Replay" could silently accept an unknown method

But I merged captureReplay into the controlReplay function

Issue #3838, opened 2026-07-01
[v10] Revamp standalone app-start spans


buenaflor:
## Context and scope

Part of the [Major v10 Task Collection](https://github.com/getsentry/sentry-dart/issues/3487). Implementation: #4018.

Make slow startup work attributable to the phase that caused it. The old `First frame render` interval includes Sentry initialization and application/framework startup work, so its name does not describe what it measures. Engine build/vsync timestamps can also omit earlier framework work during warm-up frames.

This issue applies only to the standalone `App Start` trace (`app.start`), for both static and streaming tracing. The legacy app-start spans attached to a screen-load (`ui.load`) transaction and explicit app-start extension behavior remain unchanged.

The scope below reflects the implementation in #4018. It supersedes the earlier proposal for a nested First Frame Render tree with Vsync Overhead, Raster Handoff, and conditional Post-Init Startup spans. The PR is pending; this is not a release announcement.

## Target span model

All phases are direct children of the standalone `App Start` root:

| Span | Operation | Measured interval |
| --- | --- | --- |
| Pre-Init Startup | `app.start.pre_init` | Native startup timestamp → Sentry initialization start |
| Sentry Initialization | `app.start.sentry_init` | SDK initialization start → completion, excluding app-runner execution |
| Root Widget Attachment | `app.start.root_widget_attachment` | Initial `attachToBuildOwner` entry → successful return |
| Frame Build | `app.start.frame_build` | `handleBeginFrame` entry → successful `drawFrame` return |
| Frame Rasterization | `app.start.frame_raster` | First frame's engine-reported raster start → raster finish |

Native detail spans retain their platform-provided descriptions and `app.start.native` operation. They may overlap Pre-Init Startup; child durations are not additive. Cold/warm identity remains in `app.vitals.start.type`.

Pre-Init Startup replaces the split at native plugin attachment, whose timing depends on plugin ordering. The plugin timestamp remains useful for validating native timing, but does not divide the reported phases. No Dart `main()` entry timestamp or new entrypoint API is introduced.

## Framework and raster observations

- Measure attachment and framework work through Sentry's binding hooks. Frame Build includes frame callbacks, intervening microtasks, widget updates, layout, paint, compositing, and finalization. It excludes post-frame callbacks and does not use `FrameTiming.buildDuration` as its boundary.
- Begin recording builds only after successfully observing initial root attachment. Retain at most ten complete Frame Build intervals, plus the initial attachment, while awaiting the first raster report.
[...]

PR #4018, opened 2026-09-15
feat(flutter): Revamp app-start frame spans

buenaflor:
## :scroll: Description

Split standalone app start into Pre-Init Startup, Sentry Initialization, Root Widget Attachment, Frame Build, and Frame Rasterization.

Static and streaming traces share the same measurements. Capture warm-up and deferred build work, with raster-only detail when binding hooks are unavailable. Includes configurable example workloads and migration notes for changed span names.

### App-start span tree: before and after

The standalone root stays the same. The old broad first-frame interval is replaced by separate initialization, attachment, framework-build, and raster intervals. Native detail remains alongside them.

**Before this PR**

```text
App Start [app.start]
├── App start to plugin registration [app.start.plugin_registration]
│   └─ native startup → plugin attachment
├── Before Sentry Init Setup [app.start.sentry_setup]
│   └─ plugin attachment → Sentry init start
├── First frame render [app.start.first_frame_render]
│   └─ Sentry init start → first raster finish
└── Native detail spans [app.start.native]
```

**After this PR**

```text
App Start [app.start]
├── Pre-Init Startup [app.start.pre_init]
│   └─ native startup → Sentry init start
├── Sentry Initialization [app.start.sentry_init]
│   └─ Sentry init start → init completion, before app-runner execution
├── Root Widget Attachment [app.start.root_widget_attachment]
│   └─ initial attachToBuildOwner entry → successful return
├── Frame Build [app.start.frame_build] (up to 10)
│   └─ handleBeginFrame entry → successful drawFrame return
├── Frame Rasterization [app.start.frame_raster]
│   └─ first frame's engine raster start → raster finish
└── Native detail spans [app.start.native]
```

The indented arrows describe intervals, not additional spans. All named phases are direct children of `App Start`. Multiple warm-up/deferred builds can precede one raster span; there is no enclosing First Frame Render span or one-to-one build/raster pairing. Native detail may overlap Pre-Init Startup, so child durations are not additive.

Attachment/build spans require Sentry's binding hooks; an existing standard binding can still provide raster timing when initialization precedes rendering. Explicit app-start extensions are unchanged and omitted from these trees. The legacy app-start spans attached to `ui.load` are also unchanged.

Make slow startup work attributable to the phase that caused it.

Refs #3838.

- 230 focused tests passed after merging the latest `v10-branch`; formatting and analysis checks passed.
- Tested startup configurations on a physical Android phone and an iOS simulator, including inspection of envelopes before native handoff.
- Confirmed example traces reached Sentry and separated attachment, build, and raster workloads.

## :pencil: Checklist

- [ ] All tests passing
- [ ] No breaking changes

## :crystal_ball: Next steps

Run full CI and review the changed span names and attributes before release.


packages/flutter/lib/src/app_start/standalone/static_app_start_trace.dart
```diff
@@ -87,39 +87,35 @@ final class StaticAppStartTrace implements AppStartTrace {
...
-        description: appStartFirstFrameRenderDescription,
+      final sentryInitSpan = root.startChild(
+        SentrySpanOperations.appStartSentryInit,
+        description: 'Sentry Initialization',
         startTimestamp: timing.sentrySetupTimestamp,
       )..origin = SentryTraceOrigins.autoAppStart;
-      if (firstFrameRenderSpan.samplingDecision?.sampled != true) {
-        return _abort(root, 'first-frame span is not sampled');
+      if (sentryInitSpan.samplingDecision?.sampled != true) {
+        return _abort(root, 'sentry-init span is not sampled');
```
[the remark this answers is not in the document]
buenaflor: This does not occur in the static path with ordinary supported configuration: ignoreSpans filtering is applied during V2 span creation, and the static initialization span is the first child created. maxSpans must be positive and does not count the root, so even maxSpans = 1 permits this child. The separate streaming filtering finding is valid and is being fixed. Resolving this static-path finding as non-actionable.

buenaflor:
Reviewed the suppressed Copilot findings against the implementation and Flutter source:

- The attachment-workload finding is a false positive. Initial attachment synchronously mounts the root and its children, reaching `StatefulElement._firstBuild` and `initState()` before `attachToBuildOwner` returns. The workload is attributed to root attachment as intended.
- The requested First Frame Render/Vsync Overhead/Raster Handoff tree and conditional Post-Init Startup span describe a superseded proposal. Issue #3838 and the PR's before/after trees now document the agreed five-phase model.
- The removed operation constants belong to the explicitly `@internal` surface. This is an intentional v10 telemetry migration, documented in the migration guide; deprecated aliases are not needed for these internal constants.

These suppressed findings have no individual review threads to resolve, so this records their disposition. The static filtering thread has been replied to and resolved separately. The plausible pre-existing deadline race and minor test-group convention comment remain separate from this fix.

packages/flutter/lib/src/app_start/standalone/standalone_app_start_handler.dart
```diff
@@ -144,43 +183,83 @@ class StandaloneAppStartHandler {
...
+  Future<void> _recordStartupWhenReady(SentryFlutterOptions options) async {
+    if (_closed || !_nativeTimingResolved || !_rasterTimingResolved) {
+      return;
+    }
+    // Consume before awaiting display tracking so this result is handled once.
+    final rasterInterval = _pendingRasterInterval;
+    _pendingRasterInterval = null;
+    if (rasterInterval == null) {
+      _disposeRecorder();
+      return;
```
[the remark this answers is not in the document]
buenaflor: Fixed in 4ef0de57a. Both trace implementations now track first-frame observation separately from the optional raster endpoint. Invalid timing releases the idle hold only after initialization also completes, without emitting a startup measurement or substituting a later frame. Regression tests cover both lifecycles and both initialization/observation orderings.

packages/flutter/lib/src/app_start/standalone/standalone_app_start_handler.dart
```diff
@@ -144,43 +183,83 @@ class StandaloneAppStartHandler {
...
+          ? recorder.takeIntervals(
+              processStart: processStart,
+              rasterFinish: rasterInterval.endTimestamp,
+            )
+          : const <AppStartRecordedInterval>[];
+      _disposeRecorder();
+      options.standaloneAppStartTrace?.recordFirstFrame(
+        rasterInterval,
+        frameworkIntervals: frameworkIntervals,
+      );
```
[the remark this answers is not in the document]
buenaflor: Fixed in 4ef0de57a. The shared handler rejects raster intervals whose start precedes native process start before either lifecycle records spans or display timing. Rejected intervals follow the completed-observation failure path, releasing the idle hold without fabricating a measurement. Regression tests cover intervals wholly before process start and intervals crossing it, for static and streaming traces.

packages/flutter/lib/src/app_start/standalone/streaming_app_start_trace.dart
```diff
@@ -77,39 +80,34 @@ final class StreamingAppStartTrace implements AppStartTrace {
...
         attributes: _childAttributes(
           timing,
-          SentrySpanOperations.appStartFirstFrameRender,
+          SentrySpanOperations.appStartSentryInit,
         ),
       );
-      if (firstFrameRenderSpan is! RecordingSentrySpanV2) {
-        return _abort(root, reason: 'first-frame span is not recording');
+      if (sentryInitSpan is! RecordingSentrySpanV2) {
+        return _abort(root, reason: 'sentry-init span is not recording');
```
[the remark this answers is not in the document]
buenaflor: Fixed in 4ef0de57a. The streaming trace now accepts a no-op initialization child while keeping the root paused until initialization and first-frame observation both complete. Updated the previous abort test to verify the root and measurement survive filtering, including when all startup phases are ignored and in both callback orderings. All 224 focused Flutter tests pass; analysis and documentation generation pass.

packages/flutter/lib/src/app_start/standalone/standalone_app_start_handler.dart
```diff
@@ -144,43 +183,86 @@ class StandaloneAppStartHandler {
...
+    }
+    // Consume before awaiting display tracking so this result is handled once.
+    final rasterInterval = _pendingRasterInterval;
+    _pendingRasterInterval = null;
+    final processStart = _processStartTimestamp;
+    if (rasterInterval == null ||
+        (processStart != null &&
+            rasterInterval.startTimestamp.isBefore(processStart))) {
+      _disposeRecorder();
+      options.standaloneAppStartTrace?.recordFirstFrame(null);
```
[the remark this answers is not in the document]
buenaflor: Fixed in de01c51d0. Display endpoint validation is now independent of raster interval validation, so a raster that crosses native process start still records valid display timing. Missing or invalid endpoints cancel only the pending app-start preparation, preserving any newer navigation. Added regression coverage for both trace lifecycles. All 370 startup/navigation tests pass; Flutter analysis is clean.

packages/dart/lib/src/sentry_tracer.dart
```diff
@@ -433,7 +440,33 @@ class SentryTracer extends ISentrySpan {
...
 
+  /// Suspends automatic idle completion and clears a pending finish request.
+  void pauseIdleTimeout() {
+    if (finished) return;
+    _idleTimeoutPaused = true;
+    _autoFinishAfterTimer?.cancel();
+    _finishStatus = SentryTracerFinishStatus.notFinishing();
+  }
+
+  /// Restarts automatic idle completion.
```
lucas-zimerman: Suggestion: Restart implies the time is reset, which is not what this  function is doing.
```suggestion
  /// Resume automatic idle completion.
```

packages/flutter/MIGRATION.md
```diff
@@ -0,0 +1,71 @@
...
+
+The automatic startup endpoint is the first raster-finish timestamp, not callback
+arrival. Post-frame synchronous work may delay callback delivery without extending
+the measurement. Raster completion does not mean physical display presentation or
+full application responsiveness. Explicit app-start extensions retain their
+existing behavior.
+
+## Dashboards and alerts
+
+The old `First frame render` span measured from Sentry initialization through
```
lucas-zimerman: Q: Isnt `post init`, `frame vsync` and `frame raster` handoff only part of v10? 
If so there isn't a need for them to be on the migration, what do you think?

packages/flutter/test/sentry_flutter_test.dart
```diff
@@ -66,6 +66,46 @@ void main() {
...
+        ..methodChannel = native.channel
+        ..clock = () => timestamp;
+      await SentryFlutter.init(
+        (o) {
+          o.dsn = fakeDsn;
+          o.standaloneAppStartTrace = trace;
+        },
+        options: options,
+        appRunner: () {
+          expect(trace.initEnd, timestamp);
```
lucas-zimerman: won't this test quietly fail if runner never runs?

packages/flutter/lib/src/app_start/standalone/standalone_app_start_handler.dart
```diff
@@ -47,6 +59,30 @@ class StandaloneAppStartHandler {
     _started = true;
     _options = options;
 
+    final binding = options.bindingUtils.instance;
+    // A pending raster callback is not proof that rendering has not started.
+    // Requiring an unattached root rejects the submission-to-callback window.
+    if (binding == null ||
```
lucas-zimerman, line 65: won't this skip `_prepareTimeToDisplay` too? if `rootElement` != null or `firstFrameRasterized` is already true when `start()` runs, we return here before line 135 ever executes, so the initial route loses `TTID`/`TTFD`, not just the app-start trace.

The old code called `_prepareTimeToDisplay` unconditionally for exactly this reason, per the comment above that call. Should we keep it outside this early return?

packages/flutter/MIGRATION.md
buenaflor: ah forgot, I will remove this `MIGRATION.md` file, it's not supposed to be part of this PR

packages/dart/lib/src/sentry_tracer.dart
buenaflor: Updated

packages/flutter/test/sentry_flutter_test.dart
buenaflor: Fixed it

packages/flutter/lib/src/app_start/standalone/standalone_app_start_handler.dart
buenaflor: I updated the impl to allow an attached root if the first frame is still deferred, omitting work we missed.

Without a reliable first-frame endpoint, we could measure a later frame as startup. 

Imo overall it's an improvement since we get a more useful startup breakdown and avoid reporting a later frame as the first.

lucas-zimerman (approved):
LGTM!

